PR #2, opened 2015-06-17
Irange tests

tcanabrava:
The int nodes are missing some tests: the  Timeout on the int/buffer and the int/constrain nodes. this patch adds them.


tcanabrava:
Closing this one because I just send a v2

PR #3, opened 2015-06-17
File tests

tcanabrava:
Add a new node to validate text-blobs and use it to test the file/reader node.

### DIFF
--- a/src/modules/flow/test/blob-validator.c
+++ b/src/modules/flow/test/blob-validator.c
@@ -30,33 +30,72 @@
  * OF THIS SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
  */
 
-#include <float.h>
-#include "sol-util.h"
-#include "sol-log-internal.h"
+#include <errno.h>
+#include <string.h>
 
+#include "sol-flow.h"
+#include "sol-log-internal.h"
+#include "sol-mainloop.h"
+#include "sol-util.h"
+
+#include "blob-validator.h"
 #include "test-gen.h"
 
-#include "test-module.h"
-#include "result.h"
-#include "boolean-generator.h"
-#include "boolean-validator.h"
-#include "float-generator.h"
-#include "float-validator.h"
-#include "int-validator.h"
-#include "int-generator.h"
-#include "blob-validator.h"
-
-SOL_LOG_INTERNAL_DECLARE(test_log_domain, "flow-test");
-
-void
-test_init_log_domain(void)
+int
+blob_validator_open(
+    struct sol_flow_node *node,
+    void *data,
+    const struct sol_flow_node_options *options)
 {
-    static bool done = false;
+    struct blob_validator_data *mdata = data;
+    const struct sol_flow_node_type_test_blob_validator_options *opts =
+        (const struct sol_flow_node_type_test_blob_validator_options *)options;
 
-    if (done)
-        return;
-    sol_log_domain_init_level(&test_log_domain);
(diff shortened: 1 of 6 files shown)


src/modules/flow/test/blob-validator.c
```diff
@@ -0,0 +1,101 @@
...
+
+    mdata->done = false;
+
+    if (opts->expected == NULL || *opts->expected == '\0') {
+        SOL_ERR("Option 'expected' is either NULL or empty.");
+        return -EINVAL;
+    }
+    mdata->expected.mem = strdup(opts->expected);
+    mdata->expected.size = strlen(opts->expected);
+    SOL_NULL_CHECK(mdata->expected.mem, -errno);
```
lpereira: I'm not sure if `strlen()` will alter `errno` (doesn't make much sense, but the C library isn't the most consistent thing in the world). In doubt, move the NULL check to happen right after the call to `strdup`.

src/modules/flow/test/blob-validator.c
```diff
@@ -0,0 +1,101 @@
...
+    bool match;
+
+    if (mdata->done) {
+        sol_flow_send_error_packet(node, ECANCELED,
+            "Input stream already deviated from expected data, ignoring packets.");
+        return 0;
+    }
+
+    sol_flow_packet_get_blob(packet, &val);
+    match = (mdata->expected.size == val->size) ?
```
lpereira: Minor, but this can be simplified to `match = (mdata->expected.size == val->size) && memcmp(...) == 0`.

src/modules/flow/test/blob-validator.h
```diff
@@ -0,0 +1,43 @@
...
+ * (INCLUDING NEGLIGENCE OR OTHERWISE) ARISING IN ANY WAY OUT OF THE USE
+ * OF THIS SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
+ */
+
+#include "test-module.h"
+#include "sol-blob.h"
+
+struct blob_validator_data {
+    bool done;
+    struct sol_blob expected;
```
lpereira: Move the `struct` item so it's the first element in the struct. This won't create a hole due to alignment (`bool` is one byte, and there's going to be at least 3 bytes wasted here.)

src/test-fbp/test-blob-data.txt
```diff
@@ -0,0 +1 @@
+test validator data
```
lpereira: It's one test, of course, but since blobs can accept any kind of binary data, it would be nice to test that as well. That's not going to work with the code as is (`strlen` and `strdup` are used, for instance), but it's something that should be tested.

src/modules/flow/test/blob-validator.c
tcanabrava: done.

src/modules/flow/test/blob-validator.h
tcanabrava: forgot about that - done.

src/test-fbp/test-blob-data.txt
tcanabrava: it is. there was a discussion about a binary mode and ascii mode for file/reader and writer, I'll update the test when this is released.

src/modules/flow/test/blob-validator.c
tcanabrava: done

tcanabrava:
Closing this pull request since I'v created a new one that fixed the issues on this one.

PR #6, opened 2015-06-17
Add support for array ports

ibriano:
This patch set adds support to generate node types using array ports,
instead of having several ports that will use the same set of functions.

The parser is changed to understand the notation PORT[idx] and the
necessary work in the building blocks is done to turn said notation
into the port numbers that the base framework undestands.

Exporting ports from an FBP file can be done by exporting individual
members of an array, using the corresponding index, or if no index is
given, the entire array is exported.

A few node types have been converted to use these feature to show case
it.

### DIFF
--- a/data/schemas/node-type.schema
+++ b/data/schemas/node-type.schema
@@ -81,7 +81,7 @@
           "properties": {
             "name": {
               "type": "string",
-              "pattern": "^[A-Z][A-Z0-9_]*$"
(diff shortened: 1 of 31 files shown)


lpereira:
:+1:

bdilly:
Series seems good

ibriano:
Merged

cmarcelo:
+2 (late in the game)

PR #7, opened 2015-06-17
File writer tests

tcanabrava:
Adds a test for the file/writer node. The test tries to read and replicate a file on disk. as soon as the DONE port from the file/writer sends '29' ( the number of bytes on the file) it will be closed as passed.

### DIFF
--- a/src/modules/flow/file/file.json
+++ b/src/modules/flow/file/file.json
@@ -89,6 +89,12 @@
           },
           {
             "data_type": "int",
+            "default": {
+              "max": "INT32_MAX",
+              "min": "INT32_MIN",
+              "step": 1,
+              "val": 644
+            },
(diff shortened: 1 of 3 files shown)


src/modules/flow/file/file.json
```diff
@@ -89,6 +89,12 @@
           },
           {
             "data_type": "int",
+            "default": {
+              "max": "INT32_MAX",
+              "min": "INT32_MIN",
+              "step": 1,
+              "val": 644
```
lpereira: Is this value correct? `644` makes sense if it's in octal, but this is in decimal. Use `420` instead.

tcanabrava: done

tcanabrava:
Updated pull request, closing this one.

PR #8, opened 2015-06-17
Pin mux v2

mbelluzzo:
Applied review comments done by @glima and @cmarcelo. It should be good to go now!

Waiting for ack ;-)


mbelluzzo:
Found I bug, will push fixed version soon

PR #10, opened 2015-06-17
sol-flow-node-type-gen: Properly deal with meta data

anselmolsm:
Parse meta data correctly when many node types are defined in the same
json file, so the generated files contain such information.


barbieri:
-1

But you can drop this, I'm fixing it in my PR. see what I did, but as its optional we can't access directly and the version should be a string in the schema and jsons

anselmolsm:
OK, I'll check that - for general reference, it is PR #11

PR #11, opened 2015-06-17
Node type gen improvements

barbieri:
This patchset will change the generation of our JSON node type definitions (src/modules/flow/*.json, src/lib/flow/builtins.json) so they are generated from generator specifications (src/modules/flow/*/*.json) in Python, so it's platform independent and we can get these even for other OS.

Even the src/lib/flow/builtins.json is generated in Python using a script that loads all the definitions and merging into a single file with the module name being "builtin".


barbieri:
rebased and merged.

PR #12, opened 2015-06-18
Pin mux v2

mbelluzzo:
Applied review comments done by @glima and @cmarcelo. It should be good to go now!

What is new::
Now we have the controller concept to handle multiple controller chipsets cases (mainly in pwm and aio).

Waiting for ack ;-)


mbelluzzo:
merged

PR #14, opened 2015-06-18
Provide simulation nodes

lgywata:
Now, if a flow has exported ports, runner tries to provide
simulation input/output nodes. To use this feature one should
invoke sol-fbp-runner with option '-s'.


cmarcelo:
Looking good. One more iteration to go :-)

cmarcelo:
-1  (which means this needs a new iteration)

PR #21, opened 2015-06-18
File reader tests v2

tcanabrava:
Fixed all issues commented on file_tests ( and renamed. this is actually a reader test ).

The blob validation is currently only for strings because we don't have a sane way to compare blobs yet.


lpereira:
LGTM.

---

Review status: 0 of 7 files reviewed at latest revision, all discussions resolved, all commit checks successful.

---



Comments from the [review on Reviewable.io](https://reviewable.io:443/reviews/solettaproject/soletta/21)

tcanabrava:
Pushed and closed.

PR #23, opened 2015-06-18
flow: fix some error messages while using sol-fbp-runner

ricardotk:
Only fbp errors should be print while using the runner.

The correct way to do this is to expose this kind of errors
and let the caller decide whether to print it or not.

But since it will generate changes on builder API we'll add this
quick fix for now and add a TODO entry for this case.


barbieri:
+1, but hurry with the TODO otherwise users will have less clue of the errors, not more ;-)

ricardotk:
ACK, merged

PR #24, opened 2015-06-18
fbp-generator: handle options string values between quotation marks

ricardotk:
String options like 'prefix="foo"' were being handled wrongly.


lpereira:
Looks sane, :+1:

ricardotk:
merged, thanks!

PR #25, opened 2015-06-18
"Pin mux operational"

mbelluzzo:
Add initialization of pin mux at start-up and triggers the multiplexer setup when one of the supported protocols are used.

Missing aio support because we dont have C api for it **yet**. Its next in my TODO.

### DIFF
--- a/src/lib/common/sol-mainloop.c
+++ b/src/lib/common/sol-mainloop.c
@@ -53,6 +53,21 @@ sol_log_shutdown(void)
 }
 #endif
 
+#ifdef HAVE_PIN_MUX
+extern int sol_pin_mux_init(void);
+extern void sol_pin_mux_shutdown(void);
+#else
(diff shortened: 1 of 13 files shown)


src/lib/common/sol-mainloop.c
```diff
@@ -53,6 +53,21 @@ sol_log_shutdown(void)
...
+sol_pin_mux_init(void)
+{
+    return 0;
+}
+static inline void
+sol_pin_mux_shutdown(void)
+{
+}
+#endif
+
```
lpereira: Why is this in sol-mainloop.c?

mbelluzzo: So we load the pinmux at startup

glima: I think it's a fair place to be, too.

PR #26, opened 2015-06-18
fbp-generator: properly handle rgb types

bdilly:
It was missing support to this type

### DIFF
--- a/src/bin/sol-fbp-generator/type-store.c
+++ b/src/bin/sol-fbp-generator/type-store.c
@@ -67,6 +67,12 @@ CONST_SLICE(VAL_SLICE, "val");
 CONST_SLICE(MIN_SLICE, "min");
 CONST_SLICE(MAX_SLICE, "max");
 CONST_SLICE(STEP_SLICE, "step");
+CONST_SLICE(RED_SLICE, "red");
+CONST_SLICE(RED_MAX_SLICE, "red_max");
+CONST_SLICE(GREEN_SLICE, "green");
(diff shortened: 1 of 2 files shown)


src/bin/sol-fbp-generator/type-store.c
```diff
@@ -384,6 +400,73 @@ parse_range_default_value(struct sol_json_scanner *s, struct option_description
...
+
+    rgb = &o->default_value.rgb;
+    memset(rgb, 0, sizeof(struct option_rgb_value));
+
+    if (!sol_json_scanner_next(s, &tmp))
+        return false;
+    if (sol_json_token_get_type(&tmp) != SOL_JSON_TYPE_OBJECT_START)
+        return false;
+
+    for (;;) {
```
lpereira: There are macros to help you here. See `SOL_JSON_SCANNER_OBJECT_LOOP`. It's used in the OIC stuff if you want a few examples.

src/bin/sol-fbp-generator/type-store.c
```diff
@@ -384,6 +400,73 @@ parse_range_default_value(struct sol_json_scanner *s, struct option_description
 }
 
 static bool
+parse_rgb_default_value(struct sol_json_scanner *s, struct option_description *o)
```
ricardotk: minor: this function is pretty similar to 'parse_range_default_value()', you could try to find a way to reuse this similar code.

bdilly:
v2 sent

PR #27, opened 2015-06-18
flow: add a way to fill builder type description

ricardotk:
Set type description to use, input/output ports and options
are automatically set.

### DIFF
--- a/src/lib/flow/sol-flow-builder.c
+++ b/src/lib/flow/sol-flow-builder.c
@@ -30,6 +30,7 @@
  * OF THIS SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
  */
 
+#include <ctype.h>
 #include <stdlib.h>
 
 #include "sol-flow-builder.h"
@@ -116,6 +117,9 @@ sol_flow_builder_del(struct sol_flow_builder *builder)
     }
     sol_ptr_vector_clear(&builder->ports_out_desc);
 
+    free((char *)builder->type_desc.symbol);
+    free((char *)builder->type_desc.options_symbol);
+
     sol_vector_clear(&builder->exported_in);
     sol_vector_clear(&builder->exported_out);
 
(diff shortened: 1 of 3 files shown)


src/lib/flow/sol-flow-builder.c
```diff
@@ -151,6 +155,55 @@ sol_flow_builder_set_resolver(struct sol_flow_builder *builder,
...
+        SOL_WRN("Couldn't set type description options_symbol for '%s' builder", name);
+}
+
+SOL_API void
+sol_flow_builder_set_type_description(struct sol_flow_builder *builder, const char *name, const char *category,
+    const char *description, const char *author, const char *url, const char *license, const char *version)
+{
+    SOL_NULL_CHECK(builder);
+
+    if (memchr(name, ' ', strlen(name))) {
```
lpereira: Just use `strchr()`.

src/lib/flow/sol-flow-builder.c
```diff
@@ -116,6 +117,9 @@ sol_flow_builder_del(struct sol_flow_builder *builder)
     }
     sol_ptr_vector_clear(&builder->ports_out_desc);
 
+    free((char *)builder->type_desc.symbol);
+    free((char *)builder->type_desc.options_symbol);
```
barbieri: isn't it better to use a sol_arena to store those and cleanup the whole thing at the end?

src/lib/flow/sol-flow-builder.h
```diff
@@ -60,6 +60,10 @@ int sol_flow_builder_del(struct sol_flow_builder *builder);
  * fallback to using the default resolver. */
 void sol_flow_builder_set_resolver(struct sol_flow_builder *builder, const struct sol_flow_resolver *resolver);
 
+/* Set type description to use. Input/output ports,
+ * and options descriptions are automatically set. */
+void sol_flow_builder_set_type_description(struct sol_flow_builder *builder, const char *name, const char *category, const char *description, const char *author, const char *url, const char *license, const char *version);
```
barbieri: document we keep references to these strings during the lifetime of the builder, so the user knows he must keep the strings alive or it will segv.

src/lib/flow/sol-flow-builder.c
cmarcelo: Agreed. @ricardotk take the chance and use arena in other strdups that will need it. (Make the "builder use arena" patch come first, and then your pathc uses it)

cmarcelo:
-1

src/lib/flow/sol-flow-builder.c
ricardotk: ACK

src/lib/flow/sol-flow-builder.h
ricardotk: ACK

ricardotk:
I'll send a v2, thanks for the comments

PR #28, opened 2015-06-18
oicgen.py enhancements

lpereira:
These two patches reduces the amount of generated code for OIC node types. One is essentially cosmetic (removing too much whitespace), the other makes use of `sol_str_table` (slightly reducing code size), and the other only tries to decode a value once (won't be a huge performance boost, but should help a little bit).


barbieri:
+1

PR #29, opened 2015-06-18
Irange tests v2

tcanabrava:
Fixed a crash on the fbp-runner and adds a few tests for the constrain.

### DIFF
--- a/src/modules/flow/int/int.c
+++ b/src/modules/flow/int/int.c
@@ -451,12 +451,17 @@ irange_constrain_open(struct sol_flow_node *node, void *data, const struct sol_f
     return 0;
 }
 
-static void
+static int
 irange_constrain(struct sol_irange *value)
 {
+    if(value->step == 0) {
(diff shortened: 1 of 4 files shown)


src/modules/flow/int/int.c
```diff
@@ -451,12 +451,17 @@ irange_constrain_open(struct sol_flow_node *node, void *data, const struct sol_f
     return 0;
 }
 
-static void
+static int
 irange_constrain(struct sol_irange *value)
 {
+    if(value->step == 0) {
```
lpereira: Please use the `check-style` tool prior to submitting a patch.

Also -- instead of making this an error, why not just ignore `value->step` when it's zero? The value will be constrained nonetheless:

    if (step)
        val -= (val - min) % step;
    if (val < min) val = min;
    if (val > max) val = max;

tcanabrava: Done, sending v3.

tcanabrava:
closing because a v3 was created

PR #31, opened 2015-06-19

ulissesf:
Move to common to better organize thread handling and integration with
the mainloops. Each worker thread implementation needs to use the
corresponding mainloop library to deal with threads otherwise we may
run into all sorts of unpredictable behaviors.

The POSIX threads implementation now blocks the worker threads from
being interrupted to service signals we want to handle in our main
thread.

It compiles just fine using --without-pthread but there's still need
to add a worker thread implementation using only idlers.

### DIFF
--- a/src/lib/common/sol-worker-thread-impl-glib.c
+++ b/src/lib/common/sol-worker-thread-impl-glib.c
@@ -30,25 +30,14 @@
  * OF THIS SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
  */
 
-/* TODO: threads it is common to have blocking calls and for those
- * soletta should offer a way to help.
- *
- * I think that a sol_worker_add(setup, iterate, flush, data) would
- * work, in systems with threads it would start a thread and call
- * those functions from there, on non-threaded systems it would use
- * idlers to run each iteration.
- */
-#include <pthread.h>
+#include <stdio.h>
+#include <stdlib.h>
+#include <glib.h>
 
-#define SOL_LOG_DOMAIN &_log_domain
-#include "sol-log-internal.h"
 #include "sol-mainloop.h"
-#include "sol-util.h"
-#include "sol-worker-thread.h"
+#include "sol-worker-thread-impl.h"
 
-SOL_LOG_INTERNAL_DECLARE_STATIC(_log_domain, "worker-thread-pthread");
-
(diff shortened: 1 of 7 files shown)


src/lib/common/sol-worker-thread-impl-glib.c
```diff
@@ -0,0 +1,207 @@
...
+sol_worker_thread_impl_new(bool (*setup)(void *data),
+    void (*cleanup)(void *data),
+    bool (*iterate)(void *data),
+    void (*cancel)(void *data),
+    void (*finished)(void *data),
+    void (*feedback)(void *data),
+    const void *data)
+{
+    struct sol_worker_thread_glib *thread;
+    char name[16];
```
lpereira: Is 16 characters long enough to hold "%p"? You need at least 19 bytes for a 64-bit hex value + "0x" + NUL character.

lpereira:
LGTM with comment.

src/lib/common/sol-worker-thread-impl-glib.c
ulissesf: May not be enough, I think. Documentation says the length is 16 for naming threads. I could put a counter of some sort but was lazy and tried to put %p knowing that it might be truncated. I can put a counter but this is g_thread's naming and for workers it doesn't really matter, I think. Moreover, we already print the same %p in other places without using this naming by glib.

src/lib/common/sol-worker-thread.c
```diff
@@ -0,0 +1,66 @@
...
+
+#include <stdlib.h>
+
+#include "sol-worker-thread.h"
+#include "sol-worker-thread-impl.h"
+
+SOL_LOG_INTERNAL_DECLARE(_sol_worker_thread_log_domain, "worker-thread");
+
+SOL_API struct sol_worker_thread *
+sol_worker_thread_new(bool (*setup)(void *data),
```
cmarcelo: Minor suggestion: take a struct instead of a row of functions. It must be versioned since it will be exposed to the API. It is less error prone from the caller side, easier to add/change stuff from implementation side.

ulissesf: Agreed. I'll add it to my TODO (:-D) after integrating it. Let's also make more users, maybe? Or at least think of more usage and we may need to change something more.

ulissesf:
Version 2 in PR #42.

PR #32, opened 2015-06-19
Generator rgb v2

bdilly:
v1 changelog:
  * use json_loop macro on generator
  * create some functions to be reused by default and rgb parsers
  * new commit fixing rgb option parser

### DIFF
--- a/src/bin/sol-fbp-generator/type-store.c
+++ b/src/bin/sol-fbp-generator/type-store.c
@@ -67,6 +67,12 @@ CONST_SLICE(VAL_SLICE, "val");
 CONST_SLICE(MIN_SLICE, "min");
 CONST_SLICE(MAX_SLICE, "max");
 CONST_SLICE(STEP_SLICE, "step");
+CONST_SLICE(RED_SLICE, "red");
+CONST_SLICE(RED_MAX_SLICE, "red_max");
+CONST_SLICE(GREEN_SLICE, "green");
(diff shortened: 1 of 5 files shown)


src/bin/sol-fbp-generator/type-store.c
```diff
@@ -321,11 +337,37 @@ read_default_value(struct decoder *d, struct option_description *o)
 }
 
 static bool
+check_types(struct sol_json_token *key,  struct sol_json_token *tmp, struct sol_json_token *value)
+{
+    if (sol_json_token_get_type(key) != SOL_JSON_TYPE_STRING)
```
lpereira: No need to check this; the macro does that for you.

src/bin/sol-fbp-generator/type-store.c
```diff
@@ -321,11 +337,37 @@ read_default_value(struct decoder *d, struct option_description *o)
...
+        return false;
+    if (sol_json_token_get_type(tmp) != SOL_JSON_TYPE_PAIR_SEP)
+        return false;
+    if (sol_json_token_get_type(value) != SOL_JSON_TYPE_NUMBER)
```
lpereira: The value type, however, you still need to check.

ibriano:
You need to add the Signed-off-by to every commit.

bdilly:
Nice, changes done. Merged.

PR #39, opened 2015-06-19
Provide simulation nodes

lgywata:
Now, if a flow has exported ports, runner tries to provide
simulation input/output nodes. To use this feature one should
invoke sol-fbp-runner with option '-s'.


anselmolsm:
I guess this is v2 for #14 so we can close the old version, right?

lgywata:
Yes, I've closed #14. Thanks.

PR #40, opened 2015-06-19
modules/converter: Add converter from empty to rgb and from rgb to empty

otaviobp:
Converters created based on the idea of int-to-empty and empty-to-int
converters.
Empty to rgb converter can be used, for example to change the color of
an rgb led when empty packets are received. Rgb to int converter can be
used, for example, to filter rgb input by rgb color.

Changelog from v2:
 - Use red_max, green_max and blue_max to use the color proportion in order to compare rgb.

### DIFF
--- a/src/modules/flow/converter/converter.c
+++ b/src/modules/flow/converter/converter.c
@@ -551,6 +551,22 @@ empty_to_boolean_open(struct sol_flow_node *node, void *data, const struct sol_f
 }
 
 static int
+empty_to_rgb_open(struct sol_flow_node *node, void *data, const struct sol_flow_node_options *options)
+{
+    struct sol_rgb *mdata = data;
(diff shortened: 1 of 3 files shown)


otaviobp:
New version with Signed-off-by

anselmolsm:
@otaviobp, check the output of ./tools/check-style.sh ;)

otaviobp:
Fixed the reported errors and pushed it again

otaviobp:
But what about the converter/int-to-empty? It is following the same idea, but using a range

anselmolsm:
LGTM

otaviobp:
As it was not appropriated, I removed the converter/rgb-to-empty and pushed it again

anselmolsm:
LGTM. Just Remember to remove the mention about "rgb to empty" from the commit message ;)

PR #42, opened 2015-06-19

ulissesf:
worker-threads-v2

Difference from v1:

### DIFF
--- a/src/lib/common/sol-worker-thread-impl-glib.c
+++ b/src/lib/common/sol-worker-thread-impl-glib.c
@@ -30,25 +30,14 @@
  * OF THIS SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
  */
 
-/* TODO: threads it is common to have blocking calls and for those
- * soletta should offer a way to help.
- *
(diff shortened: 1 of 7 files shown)


src/lib/common/sol-worker-thread-impl-glib.c
```diff
@@ -0,0 +1,208 @@
...
+    thread->setup = setup;
+    thread->cleanup = cleanup;
+    thread->iterate = iterate;
+    thread->cancel = cancel;
+    thread->finished = finished;
+    thread->feedback = feedback;
+
+    g_mutex_init(&thread->lock);
+
+    snprintf(name, 16, "thr-%u", thr_cnt++);
```
lpereira: Shouldn't you assert if you're running on the main thread here?

ulissesf: There's no reason this operation could not be done from another thread, however this counter might complicate things. Since I don't really care about name maybe I'll just put something fixed in there for now. :-) Otherwise I'll probably need an atomic inc here just to be safe and clear. Not sure it's worth it. It depends on what we want to offer here long term, and I think we need more users to define it better and evolve it before 1.0.

ulissesf:
Check new PR #75.

PR #43, opened 2015-06-19
Builder set type desc v2

ricardotk:
Set type description to use.

Since this type is not a generated one, symbol and options_symbol
will not be used by the fbp-generator, therefore, they are hardcoded
strings that indicate they're "builder types".

This is useful for debugging, libsoletta.so-gdb.py uses this
information.

v1 changelog:
    * start using sol_arena on builder
    * check if the type is already "closed" before changing its description
    * do not override api_version on type_desc struct
    * change symbol and options_symbol handling

### DIFF
--- a/src/lib/flow/sol-flow-builder.c
+++ b/src/lib/flow/sol-flow-builder.c
@@ -32,6 +32,7 @@
 
 #include <stdlib.h>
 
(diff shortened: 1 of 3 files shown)


ricardotk:
sending a v3, thanks for the comments

PR #48, opened 2015-06-19
kconfig

dorileo:
v2:
   + get rid of "rename test programs" patches;
   + get rid of "suite test simplification" patches;
   + change the resource lookup approach;
   + split up the Kconfig and Makefiles;
   + put the build system files together in tools/bs/;
   + added README.kconfig;
   + build in a mirrored sysroot tree;
   + added patch removing autotools;

v3:
  + make sure we [re]generate .h files if .config is newer;
  + make sure we [re]compile libs and modules if .config is newer;
  + make sure we [re]compile libs and modules if Makefiles are newer;
  + include tools/kconfig/* on its own patch and make the world a better place to live on;
  + introduced distclean target;
  + new patch: util/resolver-conffile: guess install root dir;
  + fixed issue with make check failure;
  + fixed issue with module's dest dir composition;
  + added check-valgrind target;
  + added check-fbp target;
  + added check-fbp-valgrind target - involved slightly changing the run script;
  + moved the autotools removal to "kconfig usage" patch - as suggested by Caio;
  + README.kconfig - fixed typo, update instruction command, an orthography;
  + autoconf.py:
    + added python module checks - now we fail if jsonschema is not found;
    - changed pkg's dictionary indentation as suggested by Caio;
    - fixed unlink (unlink also output file on header_deps tests) - as suggested by Glima;
    - renamed the script to dependency-resolver.py to avoid further misleading;
  + renamed tools/bs/ to tools/build/ as suggested by Caio;
  + reset default value to flow modules as suggested by Anselmo;
  + added converage target;
  + adapted the code to use the new description generator;
  + added pin-mux support to kconfig;
  + new patch: "pin-mux: guess install root dir";
  + added USE_{DEPENDENCY} support so we use its cflags only when required;
  + dropped patch: "sol-platform-impl-linux-micro.c: include micro builtins only if platform is linux-micro";
  + dropped PLATFORM_LINUX, now we have just PLATFORM_LINUX_MICRO;


dorileo:
v4 comming soon...

PR #51, opened 2015-06-22
Grove sensors

edersondisouza:
It depends on pull req #25 (I've included its commits in this pull req).


dorileo:
seems these murillo's patches don't belong here, doesn't it?

edersondisouza:
If one wants to test, those patches will be needed. That's why I've said that it depends on pull #25.

dorileo:
all right...

edersondisouza:
V2 on #52

PR #52, opened 2015-06-22
Grove sensors v2

edersondisouza:
v2:
 - More meaningful name for struct member
 - No more relative includes.

It depends on pull req #25 (I've included its commits in this pull req).


glima:
Since your new module is named grove, I'd move lcd-grove's contents in there too, also.

edersondisouza:
v3 on #64

PR #55, opened 2015-06-22
I2c fix v2

glima:
Again, this is code supporting new sensor nodes for the following HW
* ADXL345 accelerometer and
* L3G4200D gyroscope

Differences since v1:
 * issues found on i2c code touches fixed
 * actual nodes now present on PR
 * new packet type to support the nodes added, along with converter nodes and tests
 * some cosmetic commits

The accelerometer started to give me funny readings today, I'm still searching why (actually it's just the final part of the fifo reading that comes b0rked -- I tried to slow down the sampling rate but that did not help).

This is the sample flow testing what I wrote, I'll commit something with more utility once this gets an ack:

`timer(timer:interval=1000) OUT -> TICK gyro(gyroscope/l3g4200d:i2c_bus=7)
timer OUT -> TICK accel(accelerometer/adxl345:i2c_bus=7)
gyro OUT -> IN _(console:prefix="gyro (deg/s): ")
accel OUT -> IN _(console:prefix="accel (g): ")`

[Review on Reviewable](https://reviewable.io/reviews/solettaproject/soletta/55)


glima:
But shortening the interval of readings to 10ms, the trash readings go away. Lucas said the i2c driver, at the version we got there, is buggy for sure (they sent patches upstream). I think I'm going to put a comparison while passing through the buffered readings that if more than a delta from one vector to other is detected, we break.

Also, the units for the accelerometer should be m/s^2, I'm fixing that.

glima:
@lucasdemarchi would you check at least the beginning of the series again, please?

glima:
Merging the more trivial commits and sending a new version of the others.

PR #56, opened 2015-06-23
flow: Fix export port

ceolin:
Taking in consideration the last index of the previous port to set the
base port index.


ibriano:
+2

PR #58, opened 2015-06-23
Add support to export options in FBP files.

ibriano:
Make FBPs configurable by allowing them to export options from their
sub-nodes.

An options_description will be created for the type, copying the
option_description_member from the corresponding option of the node,
which means that required options from a sub-node will turn into
required options for the parent. Defaults are taken from the default
from the sub-node type, or the current value for that node's instance
if it has options.

### DIFF
--- a/src/lib/flow/sol-flow-builder.c
+++ b/src/lib/flow/sol-flow-builder.c
@@ -35,6 +35,7 @@
 #include "sol-flow-builder.h"
 #include "sol-flow-internal.h"
 #include "sol-flow-resolver.h"
+#include "sol-str-table.h"
 
 struct sol_flow_builder {
     struct sol_vector nodes;
@@ -51,6 +52,9 @@ struct sol_flow_builder {
     struct sol_ptr_vector ports_in_desc;
     struct sol_ptr_vector ports_out_desc;
 
+    struct sol_vector options_description;
+    size_t options_size;
+
     struct sol_flow_node_type_description type_desc;
 };
 
@@ -58,9 +62,24 @@ struct sol_flow_builder_node_spec {
     struct sol_flow_static_node_spec spec;
     char *name;
 
+    struct sol_vector exported_options;
+
     /* Whether builder owns the options for this node. */
     bool owns_opts;
 };
+
+struct sol_flow_builder_node_exported_option {
+    uint16_t parent_offset, child_offset;
+    uint16_t size;
+    bool is_string;
+};
+
+struct sol_flow_builder_options {
+    struct sol_flow_node_options base;
+    struct sol_flow_builder *builder;
+};
+
+#define SOL_FLOW_BUILDER_API_VERSION 1
 
 static void
 sol_flow_builder_init(struct sol_flow_builder *builder)
@@ -71,6 +90,7 @@ sol_flow_builder_init(struct sol_flow_builder *builder)
     sol_vector_init(&builder->exported_out, sizeof(struct sol_flow_static_port_spec));
     sol_ptr_vector_init(&builder->ports_in_desc);
     sol_ptr_vector_init(&builder->ports_out_desc);
+    sol_vector_init(&builder->options_description, sizeof(struct sol_flow_node_options_member_description));
 
     sol_flow_builder_set_resolver(builder, NULL);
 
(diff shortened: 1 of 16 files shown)


ceolin:
i've tested and this series looks good, although this is piece of code that i'm not familiar with.

+1

src/lib/flow/sol-flow-builder.c
```diff
@@ -544,6 +616,15 @@ get_type_description(struct sol_flow_builder *builder)
         builder->type_desc.ports_out = builder->ports_out_desc.base.data;
     }
 
+    if (builder->options_description.len > 0) {
+        struct sol_flow_node_options_member_description *sentinel;
+        sentinel = sol_vector_append(&builder->options_description);
+        SOL_NULL_CHECK(sentinel, NULL);
+        memset(sentinel, 0, sizeof(*sentinel));
+        builder->type_desc.options = get_options_description(builder);
+        SOL_NULL_CHECK(builder->type_desc.options, NULL);
```
lpereira: `sentinel` leaks here if the NULL-check fails. Is this OK?

ibriano: Yes, if this fails, everything about this builder instance will fail and the whole vector will be freed.

src/lib/flow/sol-flow-builder.c
```diff
@@ -581,6 +662,107 @@ get_exported_ports(
...
+        SOL_FLOW_NODE_OPTIONS_SUB_API_CHECK(copy_from, SOL_FLOW_BUILDER_API_VERSION, NULL);
+    }
+
+    opts = calloc(1, builder->options_size);
+    SOL_NULL_CHECK(opts, NULL);
+
+    opts->base.api_version = SOL_FLOW_NODE_OPTIONS_API_VERSION;
+    opts->base.sub_api = SOL_FLOW_BUILDER_API_VERSION;
+
+    for (member = type->description->options->members; member->name; member++) {
```
lpereira: Holy indirection, Batman!

src/lib/flow/sol-flow-builder.c
```diff
@@ -581,6 +662,107 @@ get_exported_ports(
...
+        dst = (char *)opts + member->offset;
+        if (copy_from)
+            src = (const char **)((char *)copy_from + member->offset);
+        else
+            src = (const char **)&member->defvalue.ptr;
+
+        if (is_string) {
+            char **s = (char **)dst;
+            free(*s);
+            *s = *src ? strdup(*src) : NULL;
```
lpereira: If `is_string` and `strdup()` fails, will the called know how to handle `*s` being NULL?

src/lib/flow/sol-flow-builder.c
```diff
@@ -955,3 +1143,97 @@ sol_flow_builder_export_out_port(struct sol_flow_builder *builder, const char *n
...
+        builder->options_size = sizeof(struct sol_flow_builder_options);
+
+    member_alignment = get_member_alignment(opt);
+    padding = builder->options_size % member_alignment;
+    exported_opt->offset = builder->options_size + padding;
+
+    builder->options_size += exported_opt->size + padding;
+    r = node_spec_add_options_reference(builder, node, exported_opt, opt);
+    if (r < 0) {
+        SOL_ERR("Failed to export option '%s' from node '%s'", option_name, node_name);
```
lpereira: `exported_opt` leaks.

src/lib/flow/sol-flow-builder.c
```diff
@@ -955,3 +1143,97 @@ sol_flow_builder_export_out_port(struct sol_flow_builder *builder, const char *n
 
     return 0;
 }
+
+static size_t
+get_member_alignment(const struct sol_flow_node_options_member_description *member)
+{
+    struct sol_str_slice t;
+    static const struct sol_str_table alignments[] = {
+        SOL_STR_TABLE_ITEM("boolean", __alignof__(member->defvalue.b)),
```
lpereira: `__alignof__` is a GCC extension. Is it supported by all compilers we're also supporting?

ibriano: It'd better be or nothing of this pull request is worth considering.

cmarcelo:
Looking good. Two most important changes I would ask: to not include the builder in options and pipe it through the data type, and take out the foosball change -- explanation is there in the commit.

There's a test for declare that creates an adder that always adds 22. Could you add a variant of adder that takes the value to add from an option and feed it to a constant inside? This would complement the test you already added (that just checks for the members being there and from the same type).

Great work!

ibriano:
New version in #150

PR #61, opened 2015-06-23
minnow-calamari: Make lever sample work again

anselmolsm:
Both calamari-led and calamari-lever were using the same
lever config, however max=1000 was valid only for
calamari-led.


ibriano:
ok

PR #63, opened 2015-06-24
Subprocess v2

ceolin:
- v2
 - Fix code style in macro
 - Split function now works like python's split (except that maxsplit == 0 means get all tokens)
 - Just setup watches on subprocess if start == true (open())
 - Removed duplicated code on linux-micro (fork_run)
 - Removed unused variable/argument in stdin

This series introduce the process module. There are 4 nodes: stderr, stdout, stdin and subprocess.

Basically it allows us to spawn a subprocess and monitor its life cycle and outupt/input.
Each node is more detailed on commits.


ibriano:
Discussed in person about the str_split function, but other than that it looks good.

PR #64, opened 2015-06-24
Grove sensors v3

edersondisouza:

v3:
  Exposing more parameters on nodes, like potential divider resistor resistance;
  Correctly using input range mask on formulae;
  Using an exponential approximation (obtained from Grove Starter Kit booklet table, using least squares) for light sensor Lux measurements;
  Grove fbp examples updated to use the new nodes;
  lcd-grove moved to grove module.

v2:
  More meaningful name for struct member
  No more relative includes.


edersondisouza:
v5 on #76

PR #66, opened 2015-06-24
Process error

lgywata:
Today all *process* callbacks return an integer value that is simply disregarded by the framework, even in case of failure. This series uses the ERROR channel to report failures and is the first step to make all process callbacks return void in the future. The remaining modules will be changed in future PRs.

### DIFF
--- a/src/modules/flow/app/app.c
+++ b/src/modules/flow/app/app.c
@@ -91,13 +91,19 @@ argv_process(struct sol_flow_node *node, void *data, uint16_t port, uint16_t con
     int r;
 
     r = sol_flow_packet_get_irange_value(packet, &in_value);
-    SOL_INT_CHECK(r, < 0, r);
+    SOL_INT_CHECK_GOTO(r, < 0, error);
 
     r = check_index(node, in_value);
-    SOL_INT_CHECK(r, < 0, r);
+    SOL_INT_CHECK_GOTO(r, < 0, error);
 
-    return sol_flow_send_string_packet(node,
+    sol_flow_send_string_packet(node,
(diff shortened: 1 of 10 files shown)


barbieri:
You're mixing error scopes with this description.

Process() return is the error of the method itself, like not being able to create a packet. Like a system error, core could take action such as delaying the dispatch to free memory pressure. The programmer cannot do anything to help, but the core can try. 

Things like sending ERROR on that port is different, like when an expected error in that context, such as integer overflows in calculations. Then programmer can do something about it.

dorileo:
Remember your patches in the tree will not have this pull request message, please include all the relevant information in the commit message... Imagine one reading your patch in the future, will it tell your intention with each patch in this series?

lgywata:
Yes, the PR comment is mixing things. However, I think the commits individually don't. Besides, the idea to make process() functions return void in the future was taken from a conversation with @cmarcelo. We may use this PR to discuss this matter further. Thanks.

src/modules/flow/app/app.c
```diff
@@ -114,9 +120,14 @@ quit_with_code_process(struct sol_flow_node *node, void *data, uint16_t port, ui
...
-    SOL_INT_CHECK(r, < 0, r);
+    SOL_INT_CHECK_GOTO(r, < 0, error);
 
     sol_quit_with_code(in_value);
+
+    return 0;
+
+error:
+    sol_flow_send_error_packet(node, -r, sol_util_strerrora(-r));
     return 0;
```
lpereira: I'd quit the mainloop here in these particular cases with `EXIT_FAILURE` code.

lgywata:
Ack. Thanks.

PR #71, opened 2015-06-24
sol-flow-node-type-gen.py: Remove internal nodes from description json

anselmolsm:
Nodes in category 'internal' should not be in the description files,
as they are implementation details and should not be listed as part
of the API.


ibriano:
OK

ricardotk:
+1

anselmolsm:
Applied, tks.

PR #72, opened 2015-06-24
I2c fix v3

glima:
Again, this is code supporting new sensor nodes for the following HW

ADXL345 accelerometer and
L3G4200D gyroscope

Differences since v2:

Funny readings issue resolved by a workaround.
Simpler commits that were acked already merged.

We're still to find a better name for vector-3f, since there were complaints.
I'm thinking of going with triple-f, any other idea?

This is the sample flow testing what I wrote, I'll commit something with more utility once this gets an ack:

`timer(timer:interval=1000) OUT -> TICK gyro(gyroscope/l3g4200d:i2c_bus=7) timer OUT -> TICK accel(accelerometer/adxl345:i2c_bus=7) gyro OUT -> IN _(console:prefix="gyro (deg/s): ") accel OUT -> IN _(console:prefix="accel (g): ")`


barbieri:
Instead of data type, name it after purpose: DirectionVector or DirectionalVector

glima:
I can live with that (modulo the camelcase). Any other name suggestions?

glima:
Ok, timeout. Working towards the name change now.

PR #73, opened 2015-06-24

dorileo:

New feature I would like to introduce whenever I get a chance is the option --apply to run the script and squash the fixes on each commit.


mbelluzzo:
:thumbsup:

dorileo:
new version coming....

PR #74, opened 2015-06-24
sol-fbp-generator: Fix return in startup function

anselmolsm:
The generated function 'void startup(void)' had a return NULL.


ibriano:
:+1:

anselmolsm:
Applied with the change suggested by @ceolin . Thanks.

PR #75, opened 2015-06-25
[v3] worker-threads: move to common and organize different implementations

ulissesf:
v3:
   - use atomic inc to maintain the static counter consistent among threads

v2:


ulissesf:
Check new PR #102.

PR #81, opened 2015-06-25
Fix out-of-source running of autogen.sh

thiagomacieira:
Fix for issue 79


anselmolsm:
Yep, autogen.sh should not be called out of source. Besides that, #67 is coming ;)

lucasdemarchi:
autotools is dead for soletta, closing this.

PR #82, opened 2015-06-25
fbp-generator: fix range and rgb default value parsing in type store

ricardotk:
Due to a bug in JSON parsing, '(null)' was being placed instead.


anselmolsm:
:+1:

PR #84, opened 2015-06-25
log: embed log domain inits on node generation scripts

glima:
This way freeing the user to have to worry with log domains for their
nodes. If a node is built-in, the log domain is the global one. If not,
a domain following the node's name-space is automatically created from
the .json declaration, with the declaration on its *-gen.h. The
difference now is that this file has to be the 1st Soletta header
getting an #include in a node. The *-gen.c file will have, at the type
init function, a call to actually init (once) the log domain, so we
don't have to do that at node opening time anymore.

Exceptions had to be made for modules that shared a domain via an
external log domain variable.

### DIFF
--- a/data/oic/oicgen.py
+++ b/data/oic/oicgen.py
@@ -904,9 +904,7 @@ def master_c_as_string(generated):
 #include <stdlib.h>
 #include <sys/socket.h>
 
-#define SOL_LOG_DOMAIN &_log_domain
-#include "sol-log-internal.h"
-static SOL_LOG_INTERNAL_DECLARE(_log_domain, "oic");
+#include "oic-gen.h"
 
 #include "sol-coap.h"
 #include "sol-json.h"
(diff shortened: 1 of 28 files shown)


dorileo:
The patch itself looks good to me. I've tested it and seems to be ok as well, running the check-fbp would be enough to tests the changes? is there any test case (corner case)? if not +2.

barbieri:
Doesn't make sense to change the log domain based on being built in or not. Always create a lot domain and use it.

dorileo:
Makes sense...

glima:
@barbieri, I did that because the built-in module's log messages would use the global log domain. So we ditch that behavior?

barbieri:
Yes, that's something we shouldn't keep. It was more like being lazy than a feature

glima:
Okay, fixing that and merging.

PR #87, opened 2015-06-25
Fixes

ceolin:
- Fix the build on kernel older than 3.16
- Remove unecessary check
- Fix a wrong variable type

More details can be found in which commit's message

### DIFF
--- a/configure.ac
+++ b/configure.ac
@@ -25,6 +25,16 @@ AM_CONDITIONAL([ENABLE_STATIC], [test "${enable_static}" = "yes"])
 AC_PROG_MKDIR_P
 PKG_PROG_PKG_CONFIG
 
+AC_CHECK_DECLS([IFLA_INET6_ADDR_GEN_MODE],
+[], [], [[
+#include <netinet/in.h>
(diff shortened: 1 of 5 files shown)


thiagomacieira:
Please fix the warnings:

```
src/lib/common/sol-missing.h:80:6: warning: "HAVE_DECL_IFLA_INET6_ADDR_GEN_MODE" is not defined [-Wundef]
```

ceolin:
Ohhh, i saw the error in that commit. Sorry for the noise, i'll remove it.

ceolin:
barbieri, about the if !

Unlike the other `AC_CHECK_*S' macros, when a symbol is not declared, HAVE_DECL_symbol is defined to `0' instead of leaving HAVE_DECL_symbol undeclared. When you are sure that the check was performed, use HAVE_DECL_symbol just like any other result of Autoconf:

#if !HAVE_DECL_SYMBOL
extern char *symbol;
#endif

ceolin:
closing ... fixes on 
https://github.com/solettaproject/soletta/pull/94

PR #88, opened 2015-06-25

dorileo:
v2:
   + added colorized output;
   + removed a left over import statement;


.gitignore
```diff
@@ -44,3 +44,4 @@
 /sol_flow_modules.mk
 /sol_linux_micro_modules.mk
 /sol_pin_mux_modules.mk
+*.uncrustify
```
barbieri: Missing newline?

tools/check-style.py
```diff
@@ -0,0 +1,132 @@
...
+        return output
+    except subprocess.CalledProcessError as e:
+        try:
+            print("%s%s" % (msg_prefix, e.output))
+        except BrokenPipeError as e:
+            return None
+        return None
+
+def check_diff():
+    diff = which("diff")
```
barbieri: Hum, why not use difflib directly?
https://docs.python.org/3.4/library/difflib.html

.gitignore
dorileo: ok

tools/check-style.py
dorileo: seems a good thing...

dorileo:
new version coming....

PR #90, opened 2015-06-25
Let rerunning of autogen.sh "remember" the old configure arguments

thiagomacieira:


lucasdemarchi:
Autotols is dead for soletta. Closing this.

PR #91, opened 2015-06-25
Fix warning about returning a value in a void function

thiagomacieira:
```
gyroscope.c:124:9: warning: ‘return’ with a value, in function returning void
```


glima:
Merged.

PR #92, opened 2015-06-25
log: embed log domain inits on node generation scripts

glima:
V2:
  * check out the changes in sol-flow-node-types-h-gen.py, are they ok?
  * we now generate log domains for EVERY MODULE

--

This way freeing the user to have to worry with log domains for their
nodes. If a node is built-in, the log domain is the global one. If not,
a domain following the node's name-space is automatically created from
the .json declaration, with the declaration on its *-gen.h. The
difference now is that this file has to be the 1st Soletta header
getting an #include in a node. The *-gen.c file will have, at the type
init function, a call to actually init (once) the log domain, so we
don't have to do that at node opening time anymore.

Exceptions had to be made for modules that shared a domain via an
external log domain variable.

### DIFF
--- a/data/oic/oicgen.py
+++ b/data/oic/oicgen.py
@@ -904,9 +904,7 @@ def master_c_as_string(generated):
 #include <stdlib.h>
 #include <sys/socket.h>
 
-#define SOL_LOG_DOMAIN &_log_domain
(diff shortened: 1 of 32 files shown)


glima:
Alright, thanks, now merging.

PR #93, opened 2015-06-25

dorileo:
v2:
  + added colorized output;
  + removed a left over import statement;

v3:
  + use difflib instead of triggering diff;
  + added new line to .gitignore;


lucasdemarchi:
No license?

dorileo:
merged with license...

PR #95, opened 2015-06-25
Fix exiting of forked process by way of _exit()

thiagomacieira:
Only one of the two sides of a fork() can exit via regular
exit(). That's because it will try to flush any open FILE* and run
atexit() functions, among other things (e.g., running C++
destructors). If both sides do it, weird things happen, which are very,
VERY hard to debug.

Since the parent is not exiting with _exit() in any of those cases, we
conclude the child process must.

### DIFF
--- a/src/lib/common/sol-platform-linux-common.c
+++ b/src/lib/common/sol-platform-linux-common.c
@@ -111,7 +111,7 @@ sol_platform_linux_fork_run(void (*on_fork)(void *data), void (*on_child_exit)(v
 
         errno = 0;
         on_fork((void *)data);
-        exit(EXIT_SUCCESS);
+        _exit(EXIT_SUCCESS);
     } else if (pid < 0) {
         close(pfds[0]);
         close(pfds[1]);
--- a/src/modules/linux-micro/rc-d/rc-d.c
+++ b/src/modules/linux-micro/rc-d/rc-d.c
@@ -82,7 +82,7 @@ find_exec(const char *service, const char *arg)
     }
 
     SOL_WRN("service not found: %s", service);
-    exit(EXIT_FAILURE);
+    _exit(EXIT_FAILURE);
 }
 
 static void
--- a/src/test/test-mainloop-linux.c
+++ b/src/test/test-mainloop-linux.c
@@ -118,7 +118,7 @@ main(int argc, char *argv[])
 
         usleep(100);
         close(fds[1]);
-        exit(EXIT_SUCCESS);
+        _exit(EXIT_SUCCESS);
     }
 
     /* test if we handle graceful termination with SIGTERM */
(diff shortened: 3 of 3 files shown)


barbieri:
Hum, What if user calls exit() from its on_fork?

I guess it's better to assume children will use exit() and main process will not. At least we usually control it if using our macros.

Alternatively we can declare a function that we say on_fork() should call to exit and it would do _exit()

Last but not least, is the usage of abort() of any impact?

thiagomacieira:
abort() is fine. execXX() is fine.

The problem with exit() is the other process continuing after that has happened. You cannot use stdout, for example. So I don't see a way out.

The user should not exit() from the on_fork handler. This is basic fork API: if the user is going to do forking, they should know not to exit() and instead use _exit.

ceolin:
There are other places where the child process is using exit():

src/modules/linux-micro/bluetooth/bluetooth.c
src/modules/linux-micro/dbus/dbus.c
src/modules/linux-micro/console/console.c

even in src/lib/common/sol-platform-linux-common.c

         while (read(pfds[0], &msg, 1) < 0) {
             if (errno == EINTR)
                 continue;
             else {
                 SOL_WRN("failed to read from pipe: %s", sol_util_strerrora(errno));
                 close(pfds[0]);
                 exit(EXIT_FAILURE);
             }
         }

Also, please add a doc in sol_platform_linux_fork_run() explaining to use _exit() instead of exit(). With these fixes +1 for you patch :)


@barbieri i'm afraid we can not let the child use exit() and the parent _exit() because there are some places that it is using atexit().

barbieri:
I'd suggest adding a function or a macro and refer using it instead of exit() or _exit(), such as sol_platform_linux_fork_run_exit(). The idea is to match and keep them documented in a single group.

What's your take on that? @lucasdemarchi @thiagomacieira

lpereira:
I'd prefer if we began using `posix` instead of `linux` for these things unless they're specific to `linux`. This will help porting efforts for other POSIX systems in the future.

thiagomacieira:
I don't see why a macro or wrapper is necessary. fork() is a POSIX function and so is _exit().

If you introduce a wrapper for fork(), I'll do the same for _exit(). Otherwise, I won't.

barbieri:
moving some stuff to posix could be done, particularly this linux-common file, but that's another patch.

As I said to thiago in another issue/pr, since we don't support other systems such as BSD, there is no need to pretend we do. Let's stick with linux right now and once someone steps to actually support another OS then we can do those renames. In this PoV it's better to have files named after "Linux" and allow other OS to be supported in their own files, eventually creating common files (or posix) once patterns are confirmed to work.

barbieri:
that's the case, we don't call fork(), we call sol_platform_linux_fork_run() and similar functions such as sol_platform_linux_fork_run_stop(), sol_platform_linux_fork_run_get_pid(). The purpose is that we keep a list of pids we forked to be checked with waitpid() further on our main loop.

See sol-platform-linux-common.c

barbieri:
Took some time to review the fork-run implementation and document it, see PR #403 .

barbieri:
PR #403 rebased and merged, should fix all of the issues presented in this PR. May I close it?

thiagomacieira:
Yes, go ahead and close it.

PR #96, opened 2015-06-25
Fix thread-safety of pipe2, accept4 and socket

thiagomacieira:


barbieri:
I dislike the new calls if there is a better standard version. In that case just add to sol-missing.h an implementation that tries to do that. Such as accept4() that uses fcntl(). It was also the opinion of Lucas and others. 

Also, what are the use of doing another call if the former fails. Like trying accept4() and if it did not work try accept()+fcntl(). To me there is not much advantage in doing it. 

@lucasdemarchi

thiagomacieira:
I don't see a way without the SOCK_CLOEXEC, SOCK_NONBLOCK and O_CLOEXEC macros. If they aren't defined, you can't properly call accept4 & family. We'd have three situations:
* no macros - the fallback does the right ting
* with macros, user used SOCK_CLOEXEC - the right thing
* with macros, but the user forgot - file descriptor leak

So, no, I think the best solution is to have a function that enforces FD_CLOEXEC. That's what this patch does.

ceolin:
Anyway, i also dislike new calls when we already have well-known functions.

I completely agree with the  usage of pipe2()

thiagomacieira:
Guys, please read the comment for the first of the two commits. The "well-known functions" are not enough, as proven by the attempt currently present at providing a fallback accept4. The fallback either does not compile or does not work as intended.

ceolin:
I see your points, but this brings another question, are we intending to support other unixes systems ?
There are a lot thing that is linux specific and the other ports does its own implementation.

If you are going to support other unixes i agree with you it's better to share the code instead of have millions implementations.

thiagomacieira:
Let me rephrase it back: are you going to **ban** other systems? Because I see a RIOT port inside the code, so why should we support RIOT but not a BSD? Or, for that matter, glibc 2.9 or older, which are the first ones to include the pipe2/dup3/accept4 family of functions? Or kernels older than 2.6.28, which are the first to implement those syscalls?

I'm offering a solution that abstracts this problem away, DoesTheRightThing™ by always setting FD_CLOEXEC, fixes the fallback problem and is cross-platform. The cross-platformness is just one of many benefits.

barbieri:
I'd not ban, but not willing to pretend we support systems we don't. IOW until we do support those, then do not add such things.

RIOT or Contiki does not even use POSIX like networking. They are all about specific handles and functions for that. No select(), connect()... Instead you set callbacks and create handles such as udp_new()

lucasdemarchi:
> I don't see a way without the SOCK_CLOEXEC, SOCK_NONBLOCK and O_CLOEXEC macros. If they aren't defined, you can't properly call accept4 & family. We'd have three situations:

The way it could be done would be to define the function and inside the same function call libc. It adds the complexity though, either via linker magics or dlopen.

The resistance in accepting such a patch is that you are actually doing 2 different things here. 1) Trying to fix the build for those poor environments without SOCK_CLOEXEC and SOCK_NONBLOCK and  2) making a wrapper to force  the use o CLOEXEC. (1) IMO is the major issue here:  you are forcing the more evolved ones to use the wrappers; For (2) it should be named safe_socket/safe_accept or something like that and not socket_create.

(1) should only be done when we know we support (or need to support) an OS that doesn't have that feature. That's the case of RIOT.

thiagomacieira:
Actually, no, the important fix is to get FD_CLOEXEC applied consistently in all file descriptors, preferably atomically. In order to avoid forgetting SOCK_CLOEXEC/O_CLOEXEC, it's best to have a function that does it for you. That's what this is doing.

In addition to that, it's calling the one that does it atomically on Linux, if those are defined. Note that there are multiple versions of libc that Soletta might run on (glibc, uClibc, MUSL, Bionic), so even if we're talking about Linux-only support, the fallback needs to work. For example, Bionic added support for pipe2 and accept4 only in 2014, so it's very possible someone will try to run Soletta on a board with an older (pre-Lollipop) Android BSP.

Do you want me to rename the functions? That's easy to do.

Are there other arguments for not having this?

lucasdemarchi:
> Do you want me to rename the functions? That's easy to do.

yes

> Are there other arguments for not having this?

I dislike the additional call because if whoever added the method remembered to use the function he might as well had remembered to use the flag ;-).  Anyway,  let's put the change in and see if it's bad in the long run? @barbieri, @bdilly, @cmarcelo, opinions?

thiagomacieira:
> I dislike the additional call because if whoever added the method remembered to use the function he might as well had remembered to use the flag ;-). Anyway, let's put the change in and see if it's bad in the long run?

That's assuming it's a simple one-call. As you can see from the bodies of the functions in questions, it's hardly that in the fallback case.

Anyway, I need to update now that Kbuild is in and there's no more AC_CHECK_FUNCS.

src/shared/sol-util.h
```diff
@@ -159,3 +168,60 @@ align_power2(unsigned int u)
  *       elements found are returned.
  */
 struct sol_vector sol_util_str_split(const struct sol_str_slice slice, const char *delim, size_t maxsplit);
+
+/* Sets the F_SETFL flags and sets FD_CLOEXEC in fd.
+ * If either operation fails, closes the fd and returns -1.
+*/
+static inline int fd_set_flags_cloexec(int fd, int flags)
```
lpereira: Maybe there should be a `sol_util_posix_fd_set_flags()` function instead? There are many other places where `fcntl(F_GETFD)` / `fcntl(F_SETFD)` are used, each with their own copy-pasted error handling.

ceolin:
as this pr is outdated and so far we did not face any problem with the current implementation i'm closing it. If necessary it'll be reconsidered.

thiagomacieira:
I'm updating some of the patches. Since I didn't get an answer on whether Soletta plans on supporting non-Linux systems in its POSIX layer, I'll just use the Linux-specific calls, which have found their way to some other OSes as POSIX extensions too.

thiagomacieira:
Commits updated, please reopen.

PR #98, opened 2015-06-25
sol-aio: Add C API to handle Analog I/O

mbelluzzo:
This also remove the code from aio node that would be Linux specific
(which was used to create sol-aio-linux).

WARNING: Riot code needs to be tested, I'm working on it. But is here for general review.

### DIFF
--- a/src/modules/flow/aio/aio.c
+++ b/src/modules/flow/aio/aio.c
@@ -32,6 +32,7 @@
 
 #include "aio-gen.h"
 
+#include "sol-aio.h"
 #include "sol-flow-internal.h"
 #include "sol-flow.h"
 #include "sol-mainloop.h"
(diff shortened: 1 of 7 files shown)


src/shared/sol-aio.h
```diff
@@ -0,0 +1,81 @@
...
+ */
+void sol_aio_close(struct sol_aio *aio);
+
+/**
+ * Read the value of AIO 'pin' on 'device'.
+ *
+ * @param aio A valid AIO handler for the desired 'device'/'pin' pair.
+ * @return The value read. -1 in case of error.
+ */
+int sol_aio_get_value(const struct sol_aio *aio);
```
barbieri: Int32_t to be specific on the return size.

mbelluzzo: All APIs are returning int if I remember well (at mobile right now),
including the system ones. Why convert to int32_t only here?

src/modules/flow/aio/aio.c
```diff
@@ -80,8 +62,8 @@ aio_close(struct sol_flow_node *node, void *data)
 
     SOL_NULL_CHECK(mdata);
 
-    if (mdata->fp)
-        fclose(mdata->fp);
+    if (mdata->aio)
+            sol_aio_close(mdata->aio);
```
lpereira: Weird indenting here.

mbelluzzo: Really weird

src/shared/sol-aio.h
ibriano: Because Arduino and 16-bit ints, if you read from the device with a 16-bit mask you have no way of knowing if you got a valid value or -1 for error.

mbelluzzo: already changed to int32_t

mbelluzzo:
inc v2

PR #101, opened 2015-06-26
float: fpclassify wrapper node

edersondisouza:
If one needs to classify a float (that can be NaN, Infinite, Zero,
Subnormal or Normal), this node comes handy.
It receives a float as input and outputs it to the due port.
Tests to this node were added too.

### DIFF
--- a/src/modules/flow/float/float.c
+++ b/src/modules/flow/float/float.c
@@ -962,4 +962,36 @@ wave_generator_sinusoidal_open(struct sol_flow_node *node,
     return 0;
 }
 
+// =============================================================================
+// FPCLASSIFY
+// =============================================================================
+
+static int
+classify_process(struct sol_flow_node *node, void *data, uint16_t port, uint16_t conn_id, const struct sol_flow_packet *packet)
+{
+    double value;
+    int r;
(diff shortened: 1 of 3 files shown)


glima:
It does not hurt to have this, IMO. +1

src/modules/flow/float/float.c
```diff
@@ -962,4 +962,36 @@ wave_generator_sinusoidal_open(struct sol_flow_node *node,
...
+classify_process(struct sol_flow_node *node, void *data, uint16_t port, uint16_t conn_id, const struct sol_flow_packet *packet)
+{
+    double value;
+    int r;
+
+    r = sol_flow_packet_get_drange_value(packet, &value);
+    SOL_INT_CHECK(r, < 0, r);
+
+    switch (fpclassify(value)) {
+        case FP_NAN:
```
lpereira: Minor coding style issue: please don't indent `case` labels. When in doubt, use the `check-style.sh` script.

mbelluzzo:
+2 with indent fix

edersondisouza:
Integrated with indent fix.

PR #104, opened 2015-06-26
Fix array port export

lgywata:
When trying to export an array port, sub indexes for that port
were all being set to zero, raising a "exported in/out ports
not sorted" warning and failing the flow creation.


lgywata:
This is V2 for PR #100

ibriano:
+2

lgywata:
Applied.

PR #105, opened 2015-06-26
minnow-calamari: Rewrite README, remove misleading files

anselmolsm:
[Based on QA team's feedback]

The idea of providing calamari-3.1*.conf files were not effective.
This patch adds a new file, sol-flow-new.conf,  with the GPIO pin
numbers for Linux Kernel>=3.17. The README was updated to
explain how to use sol-flow-new.conf.


ibriano:
Fine by me.

barbieri:
Ok, but we should remember that and try to allow kernel name and version in our software and as replacements in conffile loading/inclusion.

Like ${PLATFORM} and it uses the code from Murilo. ${KERNEL_NAME} to get linux, riot... And ${KERNEL_VERSION}. Problem is how to cope with things like "if newer than version"

anselmolsm:
@barbieri Agreed.

V2 submitted: #110

PR #107, opened 2015-06-26
flow:converter: Add string to blob

ceolin:
A converter node that receives a string packet in its IN port
and send a blob packet in the OUT port.


ceolin:
merged with the fixes. Thanks !

PR #109, opened 2015-06-26
Small fixes

edersondisouza:


edersondisouza:
Merged - but with a change suggested by @ibriano on 'shared: Avoid signed extension on multiplication'

PR #111, opened 2015-06-26
Container nodes: Avoid issues when options are missing

anselmolsm:
Container nodes should copy the free_options function in order
to avoid issues when options are omitted in conf files.


ibriano:
Ok

anselmolsm:
Applied, tks

PR #112, opened 2015-06-26
If no priv data exists for a node, don't pass a valid data ptr to it.

glima:
This way, on

* int (*open)(struct sol_flow_node *node, void *data, const struct  sol_flow_node_options *options)
* void (*close)(struct sol_flow_node *node, void *data)

The node won't get a non-NULL but invalid pointer to act on.


ibriano:
Ack

ibriano:
Aaaand now I see it's missing in close()

glima:
Thanks a lot. Fixing and merging.

Issue #97, opened 2015-06-25
sol-missing.h defines EBADR to silly value


thiagomacieira:
sol-missing.h has:
```c
#define EBADR       53  /* Invalid request descriptor */
```
This macro is used outside of Linux-specific files. 

That value is silly, since it matches ECONNABORTED on the BSDs (including OS X).

ceolin:
all EBADR references were removed. This issue no longer exists, just closing it.

PR #113, opened 2015-06-26
Fixes

ceolin:
Fixes for problems spotted by coverity and a fixes for the issue https://github.com/solettaproject/soletta/issues/97


ibriano:
Fine by me

ceolin:
Ok, i'll create a new pullrequest with the sugestions (plus another fix) and i'll let this commit (EBADR) away from the pullreq. This error code is been used in many places that is not platform-specific and for most of them its usage makes sense.

ceolin:
closed and the fixes are in https://github.com/solettaproject/soletta/pull/122

PR #114, opened 2015-06-26
kbuild: fix issue with ubuntu 14.04

dorileo:
The ubuntu 14.04 linker will not work if the libs are not the last arguments.


lucasdemarchi:
Applied

PR #116, opened 2015-06-27
random kbuild fixes

dorileo:
This series fixes issues raised by @ibriano  and @lucasdemarchi about dependency selection related platform, glib, and so on.


lucasdemarchi:
Applied. Comments are unrelated to this particular pull request and should be solved on top.

PR #120, opened 2015-06-29
Tools: Update scripts for the kbuild era

anselmolsm:
calc-lib-size still need more attention, though.


anselmolsm:
Applied. tks.

PR #121, opened 2015-06-29
build: fix issue with systemd-journal dependent code

dorileo:
Since the libs are split the build system is not using the systemd-journal
flags. This patch introduces SYSTEM_JOURNAL dependency to the resolver
and use it wherever it's needed.

### DIFF
--- a/Kconfig
+++ b/Kconfig
@@ -30,6 +30,10 @@ config USE_SYSTEMD
 	bool
 	default n
 
+config USE_SYSTEMD_JOURNAL
+	bool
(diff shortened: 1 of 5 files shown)


dorileo:
please don't apply, new version coming ....

dorileo:
@lucasdemarchi please open an issue describing the way you think it should be, but either way, it ends up being similar to what ./configure does, this is the first step before you build the project, if you changed your environment it's just a matter of running make reconf and it's all done...

PR #122, opened 2015-06-29
Fixes v2

ceolin:
-v2
 - Removed commit about EBADR
 - Added fix for  a null pointer dereference
 - Fix a copy/paster in the fcntl patch

### DIFF
--- a/src/modules/flow/process/output.c
+++ b/src/modules/flow/process/output.c
@@ -152,7 +152,8 @@ watch_start(struct output_data *output)
     flags = fcntl(output->fd, F_GETFL);
     SOL_INT_CHECK(flags, < 0, -errno);
     flags |= O_NONBLOCK;
-    fcntl(output->fd, F_SETFL, flags);
+    if (fcntl(output->fd, F_SETFL, flags) < 0)
+        return -errno;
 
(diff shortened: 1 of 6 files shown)


mbelluzzo:
So, github doesn't likes to mark comments in commits as comments in pull requests :-/

src/modules/flow/process/output.c
```diff
@@ -152,7 +152,8 @@ watch_start(struct output_data *output)
     flags = fcntl(output->fd, F_GETFL);
     SOL_INT_CHECK(flags, < 0, -errno);
     flags |= O_NONBLOCK;
-    fcntl(output->fd, F_SETFL, flags);
+    if (fcntl(output->fd, F_SETFL, flags) < 0)
+        return -errno;
```
lpereira: I'd rather have a function to set a flag on a file descriptor than replicate this code all over. Maybe `int sol_util_posix_fd_set_flag(int fd, int flag)`?

ceolin:
cool, I'm going to apply the commits that have received +2 the oher one will require the re-factoring of sol-util so it'll come in another pullreq.

Thanks folks.

Issue #118, opened 2015-06-27
kbuild: missing auto-include of sol_config.h


lucasdemarchi:

dorileo:
pull request https://github.com/solettaproject/soletta/pull/131 brings it back.

dorileo:
fixes in...

PR #123, opened 2015-06-29
build: fix issue with systemd-journal dependent code

dorileo:
v2:
  + don't need to search for libsystemd-journal since the flags are the same as the systemd itself;

sol-log-impl-linux.c should get the SYSTEMD_CFLAGS as extra-cflags
in order to link agains systemd-journal.

### DIFF
--- a/src/lib/common/Kconfig
+++ b/src/lib/common/Kconfig
@@ -53,6 +53,7 @@ endchoice
 
 config LOG
 	bool "Log"
+	select USE_SYSTEMD
 	default y
 
 choice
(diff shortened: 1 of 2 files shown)


bdilly:
commit seems ok, but when building it with default configs, I'm seeing the following warning:
./src/lib/common/sol-log-impl-linux.c:658:5: warning: implicit declaration of function ‘sd_journal_send_with_location’ [-Wimplicit-function-declaration]

ibriano:
I guess that's related to #118, so a define is missing and sd-journal.h is not getting included there.

dorileo:
all right, I'll send a v2 with this -include change

dorileo:
new version coming...

PR #125, opened 2015-06-29
Fix signature of sd-bus functions now that it's officially supported

ibriano:
Starting with systemd 221, sd-bus is officially supported and the
signature of the functions changed from the ones we were using.


lucasdemarchi:
Ack

PR #127, opened 2015-06-29
log: Fix extern log domain variable names so they don't clash

glima:
when we built all modules as built-in.


mbelluzzo:
+2

ricardotk:
ACK.
minor: what do you think about add this kind of prefix in all the other places where we use log domain variables?

glima:
@ricardotk The common case is to have these variables declared as *static*, so we're safe.

ricardotk:
Ok, good

glima:
Merged.

PR #129, opened 2015-06-29
kconfig: sort platform options and select dummy as default

bdilly:
All the other options are sorted alphabetically.
Also using dummy as default instead of micro-linux
seems more straighforward for general users.


dorileo:
LGTM, +2

anselmolsm:
:+1:

bdilly:
merged, thanks

PR #130, opened 2015-06-29
build: fix mod-so rule dependency

dorileo:
The mod-so rules are depending on an nonexistent object since it's
using the wrong variable, it's depending on LIB_SO but the proper
variable is SOL_LIB_SO. This may be causing dependency problems
on modules being built before libsoletta.so.


dorileo:
applied...

PR #132, opened 2015-06-29
kconfig: use SOL_PLATFORM_LINUX to check if it's linux

bdilly:
Dorileo, please check if it's the expected behaviour.

Instead of using SYSTEMD || MICROLINUX, or even SOCKET_LINUX

Using such approaches were leaving dummy without many stuff,
including tools.


bdilly:
merged, thanks

PR #133, opened 2015-06-29
build: Fix installing with the wrong prefix

vcgomes:
When DESTDIR is specified and it doesn't exist, cp -R will not create
it, which will cause the first part of the prefix to be omitted.

For example, prefix=/usr/x86_64-pc-linux-gnu and DESTDIR=/tmp/soletta
will cause it to be installed on /tmp/soletta/x86_64-pc-linux-gnu, if
/tmp/soletta doesn't exist.


vcgomes:
I will make another PR shortly, with another path related fix. Closing this.

PR #134, opened 2015-06-29
samples: add example of password queues

bdilly:
--

It's based on a discussion with Ederson a few weeks ago.
It wasn't doable until REPLACE_PACKET flag was finally
removed.

--

This is an example that simulates two password machines:
one for regular and other for preferencial passwords.
And a cashier, that when press the button call the next
password, preferencials first. It won't call anybody
if such queues are empty already.


bdilly:
merged using tickets.
thanks

PR #135, opened 2015-06-29

ulissesf:
v5:
   - rebased and ported to Kbuild

v4:
   - rebased

v3:
   - use atomic inc to maintain the static counter consistent among threads

v2:


ulissesf:
Merged v5. Check out the previous PRs for comments.

PR #137, opened 2015-06-29
build: make sure all the compilation uses the same -std flag

dorileo:
The language standard is being set only for lib cflags, it should
be applied to common_cflags and be used to all compilations.


anselmolsm:
:+1:

dorileo:
applied

PR #140, opened 2015-06-29
test: add test for converter/string-to-blob

bdilly:
Ceolin, please review this pr

--

It was missing on commit 28137a334a


lpereira:
Blobs can contain any kind of arbitrary data, including a NUL terminator (which might not even be a terminator byte). This PR looks sane, as every node handling blob types need to take its length into consideration, and not if it has a NUL-byte or not.

:+1:

bdilly:
v2 -> https://github.com/solettaproject/soletta/pull/145

PR #141, opened 2015-06-30
[v1] pthread: fix building with or without pthread support

ulissesf:
When some dependency checks were removed the pthread one went from
PTHREAD_H to PTHREAD and thus breaking the build.

However, the bigger problem is that we were always compiling with
pthread support because the dependency solver always defines
HAVE_PTHREAD_H even when we mark the config option PTHREAD as 'n'.
Previously autotools defined HAVE_PTHREAD_H when we had pthread.h
_and_ wanted to compile with pthreads support. Now that's not the
case anymore with Kbuild so we should test for PTHREAD in the code
and not HAVE_PTHREAD_H.

Note that the ldflags still need to be properly sorted out even though
everything compiles and works now.

### DIFF
--- a/data/scripts/dependency-resolver.py
+++ b/data/scripts/dependency-resolver.py
@@ -47,7 +47,7 @@ pkg_config_deps = {
 
 header_deps = {
     "KDBUS": "<systemd/sd-bus.h>",
-    "PTHREAD": "<pthread.h>",
+    "PTHREAD_H": "<pthread.h>",
     "RIOTOS": "<riotos/cpu.h>",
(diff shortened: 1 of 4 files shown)


dorileo:
ok, besides my confusion between the PTHREAD and the HAVE_PTHREAD_H configs usage here the patch looks good. Also, I'm sending a patchset on top of this patch so I don't need to [re]fix things related to pthread already present in this one.

dorileo:
this patch was applied via pull req https://github.com/solettaproject/soletta/pull/149

PR #142, opened 2015-06-30
Build dep resolver and ldflags

dorileo:

@ulissesf could you check where we should set the -extra-ldflags for pthread since you've touched these codes?

### DIFF
--- a/data/jsons/dependencies.json
+++ b/data/jsons/dependencies.json
@@ -1,0 +1,123 @@
+{
+    "dependencies": [
+	{
+	    "dependency": "check",
+	    "type": "pkg-config",
+	    "pkgname": "check"
+	},
+	{
+	    "dependency": "glib",
+	    "type": "pkg-config",
+	    "pkgname": "glib-2.0"
+	},
+	{
+	    "dependency": "gtk",
+	    "type": "pkg-config",
+	    "pkgname": "gtk+-3.0"
+	},
+	{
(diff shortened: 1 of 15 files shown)


dorileo:
look, this patchset brings a patch from @ulissesf which is already here https://github.com/solettaproject/soletta/pull/141

dorileo:
I'm checking why this is failing on ubuntu...

dorileo:
ok, now it builds on ubuntu :)

data/jsons/dependencies.json
```diff
@@ -0,0 +1,123 @@
...
+	},
+	{
+	    "dependency": "systemd",
+	    "type": "pkg-config",
+	    "pkgname": "libsystemd"
+	},
+	{
+	    "dependency": "udev",
+	    "type": "pkg-config",
+	    "pkgname": "libudev"
```
lpereira: Shouldn't the versions of pkg-config things be checked as well?

dorileo: do you mean with --modversion ?

dorileo: Do you think it's worth it? maybe for some cases, right?

dorileo:
just sent new version, https://github.com/solettaproject/soletta/pull/143
closing this one
@lpereira this v2 I've just sent doesn't implement any version check.

PR #143, opened 2015-06-30
Build dep resolver v2

dorileo:
v2:
    + properly set the pthread_h_ldflags;
    + added new patch to set the flow-resolver ldflags;

### DIFF
--- a/data/jsons/dependencies.json
+++ b/data/jsons/dependencies.json
@@ -1,0 +1,123 @@
+{
+    "dependencies": [
+	{
+	    "dependency": "check",
+	    "type": "pkg-config",
+	    "pkgname": "check"
+	},
+	{
+	    "dependency": "glib",
+	    "type": "pkg-config",
+	    "pkgname": "glib-2.0"
+	},
+	{
+	    "dependency": "gtk",
+	    "type": "pkg-config",
+	    "pkgname": "gtk+-3.0"
+	},
+	{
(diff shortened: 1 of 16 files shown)


data/jsons/dependencies.json
```diff
@@ -0,0 +1,123 @@
+{
+    "dependencies": [
+	{
+	    "dependency": "check",
+	    "type": "pkg-config",
+	    "pkgname": "check"
```
lpereira: Having `atleast-version`, `max-version`, and `exact-version` keys for `pkg-config` types is important, as we often use features available in latest or close-to-latest versions of libraries.

dorileo: right, I was wondering if we *alread* have a user for that, talking to @glima we realized systemd would a first user... I'm writing that now. thanks.

glima: Yep. Besides the version checking think, r+ for the series. I'm working on top of it for a Galileo build and it's fine. Let's merge this ASAP, because I already have patches on top of it.

ulissesf:
I'd say we merge this and @dorileo works on the version checking. The build is broken for mainloop posix and other cases right now.

dorileo:
this is done, I'm sending a new version...

dorileo:
new version sent
https://github.com/solettaproject/soletta/pull/149

PR #145, opened 2015-06-30
v2 changelog

bdilly:
* instead of changing converter, add option to blob-validator
 * fix error packet initialization
 * add tests for error packet and hub family


bdilly:
merged, thanks

PR #146, opened 2015-06-30
Change APIs signatures

lgywata:
Sometimes these APIs can be useful when we have no flow
(sol_flow_node) but we do have a sol_flow_node_type.


lgywata:
applied. thanks.

PR #148, opened 2015-06-30
fbp-generator: add support for DECLARE statement

ricardotk:
Now the generator handles the declaration and usage of new types in the FBP itself.

e.g.
    DECLARE=MyType:fbp:my-type.fbp
    my_node(MyType) OUT -> ...

In the generated C code now, we first build the node spec with a placeholder and after
we change this to the proper value, since we need to create this new types in runtime.

### DIFF
--- a/src/bin/sol-fbp-generator/main.c
+++ b/src/bin/sol-fbp-generator/main.c
@@ -32,6 +32,7 @@
 
 #include <ctype.h>
 #include <errno.h>
+#include <libgen.h>
 #include <stdio.h>
 #include <unistd.h>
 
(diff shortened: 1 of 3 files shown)


src/bin/sol-fbp-generator/main.c
```diff
@@ -412,11 +424,14 @@ generate_exports(const struct sol_fbp_graph *g)
...
-generate(struct sol_fbp_graph *g, struct type_description **descs)
+generate(struct sol_vector *fbp_data_vector)
 {
+    const size_t symbol_placeholder_prefix_size = strlen(SYMBOL_PLACEHOLDER_PREFIX);
+    const size_t symbol_placeholder_suffix_size = strlen(SYMBOL_PLACEHOLDER_SUFFIX);
+    struct fbp_data *data;
     struct sol_fbp_meta *m;
     struct sol_fbp_node *n;
-    uint16_t i, j;
+    uint16_t i, j, k;
```
lpereira: When you have three nested loops, it's often a good idea to split the function.

ibriano:
Is the problem of fbp A declaring a type using fbp B, which declares a type using fbp A been addressed here?
I algo agree that this code would benefit a lot from splitting some things into more functions. Maybe go through all the declarations, ensuring you have resolved all the types needed for the program to run, then create one function for each type that will hold its specs, exported ports, options, etc. Create the type from there and return that for the parent. It's a recursive solution, but I think it will make it more clear to see what's happening, when and where.

barbieri:
I agree with Sachiel that the type-store should have a scope of the loader. If a FBP is loaded, then all DECLARE that appears inside it should be constrained to that FBP and not leak into siblings, parent or children.

ricardotk:
Ok, I'll send a v2, thanks!

PR #149, opened 2015-06-30
Build dep resolver v3

dorileo:
v2:
+ properly set the pthread_h_ldflags;
+ added new patch to set the flow-resolver ldflags;

v3:
  + added atleast-version, max-version, and exact-version keys for pkg-config;
  + set atleast-version to 221 for systemd;


dorileo:
applied...

PR #150, opened 2015-06-30
Make FBPs configurable by allowing them to export options from their sub-nodes.

ibriano:
An options_description will be created for the type, copying the
option_description_member from the corresponding option of the node,
with default values being taken either from the default in the sub-node's
type, or from the options used to instantiate that specific instance of the
node.

Changes since v1:
 * Fixed 'leaks' for correctness, even though they would be freed when the
   vectors they come from are freed.
 * Return error when strdup() fails while creating options.
 * Killed foosball example changes, added them to the declare test instead.
 * All options related methods receive the type, so we don't need to pass it
   along in the options anymore.

### DIFF
--- a/src/shared/sol-fbp-parser.c
+++ b/src/shared/sol-fbp-parser.c
@@ -371,6 +371,57 @@ parse_declare_stmt(struct sol_fbp_parser *p)
 }
 
 static bool
+parse_option_stmt(struct sol_fbp_parser *p)
+{
+    struct sol_str_slice opt_name, node_name, node_opt, empty = SOL_STR_SLICE_EMPTY;
+    struct sol_fbp_node *err_node;
+    struct sol_fbp_position pos;
+    int node_idx, err;
+
+    assert(next_token(p) == SOL_FBP_TOKEN_OPTION_KEYWORD);
+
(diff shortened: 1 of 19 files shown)


src/shared/sol-fbp-parser.c
```diff
@@ -371,6 +371,57 @@ parse_declare_stmt(struct sol_fbp_parser *p)
...
+        return handle_node_error(p, &node_name, &pos, node_idx, err_node);
+
+    err = sol_fbp_graph_option(p->graph, node_idx, opt_name, node_opt, pos);
+    if (err == -EEXIST) {
+        p->error_pos = pos;
+        return set_parse_error(p, "Option '%.*s' already declared", SOL_STR_SLICE_PRINT(opt_name));
+    } else if (err == -EINVAL) {
+        p->error_pos = pos;
+        return set_parse_error(p, "Option '%.*s' with invalid values", SOL_STR_SLICE_PRINT(opt_name));
+    }
```
lpereira: `} else if (err < 0) {`?

ibriano: Agh, done.

cmarcelo:
+2

PR #152, opened 2015-06-30
Add array ports to the generator

ibriano:
And related fixes needed to make that work.


src/bin/sol-fbp-generator/type-store.c
```diff
@@ -157,6 +159,30 @@ get_string(struct sol_json_token *token)
     return strndup(token->start + 1, token->end - token->start - 2);
 }
 
+static int
+get_int(struct sol_json_token *token, int *value)
+{
+    int v;
+    char *str, *endptr = NULL;
+
+    str = strndupa(token->start, token->end - token->start);
```
lpereira: There's no need to copy the string: call `strtol()`, and check if `endptr` points to `token->end` (if not, there was a parsing error).

ibriano:
Comments fixed locally, if more things arise I'll send a v2

src/bin/sol-fbp-generator/type-store.c
barbieri: This is not right if JSON is malformed it may cause out of bounds access. strtol() relies on trailing \0

ricardotk:
+1

ibriano:
Merged.

PR #153, opened 2015-06-30
Fixes for building external modules

vcgomes:
The first one was found when installing Soletta in a non-existant directory.

The second one was found when using the node-type code generator from a external module.

There are still problems when building external modules, that will be fixed in later pull requests.


dorileo:
+2

vcgomes:
Applied.

PR #155, opened 2015-06-30
type-store: add field checks while reading types from JSON

ricardotk:
Wrong JSON entries were causing segmentation fault in generator.

Pointed out by coverity.


lpereira:
If the code can handle early returns without leaking memory or breaking the world, :+1:

PR #156, opened 2015-06-30
Check and test the return value of func-call

tcanabrava:
There's always a possibility to have an failed func call, so
let's test it before using the value.


tcanabrava:
Updating V2

PR #157, opened 2015-06-30
sol-util: Make get_progname fail when execfn is NULL

anselmolsm:


ricardotk:
+1

anselmolsm:
Applied, tks.

PR #158, opened 2015-06-30
type-store: add field checks while reading types from JSON v2

ricardotk:
Differences from v1:
    - added checks for ports

Wrong JSON entries were causing segmentation fault in generator.

Pointed out by coverity.


ricardotk:
merged, thanks

PR #159, opened 2015-06-30
I2c small fixes

anselmolsm:
The first commit fixes a KW complaint.
The second patch is to follow the ioctl doc.


src/shared/sol-i2c-linux.c
```diff
@@ -99,6 +99,7 @@ sol_i2c_open_raw(uint8_t bus, enum sol_i2c_speed speed)
     return i2c;
 
 open_error:
+    close(i2c->dev);
```
lpereira: If `open()` fails, this will attempt to close an undefined file descriptor.

anselmolsm: Ops, true story. I'll fix that. tks.

anselmolsm:
v2: Fix issue pointed out by @lpereira .

anselmolsm:
Applied, tks.

PR #160, opened 2015-06-30
Check and test the return value of func-call

tcanabrava:
There's always a possibility to have an failed func call, so
let's test it before using the value.

### DIFF
--- a/src/test/test-json.c
+++ b/src/test/test-json.c
@@ -157,7 +157,10 @@ test_json(void)
         output = scan_tests[i].output;
 
         for (j = 0; j < scan_tests[i].expected_elements; j++) {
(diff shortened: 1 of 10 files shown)


tcanabrava:
applyed and closed.

src/test/test-json.c
```diff
@@ -157,7 +157,10 @@ test_json(void)
         output = scan_tests[i].output;
 
         for (j = 0; j < scan_tests[i].expected_elements; j++) {
-            sol_json_scanner_next(&scanner, &input);
+            if(!sol_json_scanner_next(&scanner, &input)) {
```
lpereira: Space missing between `if` and `(`. Please use the `check-style.sh` tool.

PR #161, opened 2015-06-30
build: fix rpath cleanup on install and dependency chain

dorileo:
This patch fixes the rpath cleanup step on install, it was composing
the rule name with the dest resource path both for the rule itself and
the dependency, since we don't have a rule to make a resource (in this
case to compile the bins out) the install target would never run due
dependency chain error.


brunobottazzini:
LGTM. 

The rpath problem on install phase has gone when applying this patch.

dorileo:
applied, thanks.

PR #163, opened 2015-06-30
Build fixes

mbelluzzo:


glima:
No major flaws, it seems (just minor comments).

mbelluzzo:
Merged

src/bin/sol-fbp-generator/Kconfig
```diff
@@ -1,4 +1,4 @@
 config FBP_GENERATOR
 	bool "fbp generator"
-	depends on SOL_PLATFORM_LINUX && FLOW && RESOLVER_CONFFILE && INSPECTOR
+	depends on SOL_PLATFORM_LINUX && FLOW && NODE_DESCRIPTION && INSPECTOR
```
barbieri: I guess here we need a resolver, not only the descriptions otherwise .conf won't work. Same for runner below.

src/bin/sol-fbp-runner/Kconfig
```diff
@@ -1,4 +1,4 @@
 config FBP_RUNNER
 	bool "fbp runner"
-	depends on SOL_PLATFORM_LINUX && FLOW && RESOLVER_CONFFILE && INSPECTOR
+	depends on SOL_PLATFORM_LINUX && FLOW && NODE_DESCRIPTION && INSPECTOR
```
barbieri: This one also needs conffile resolver.

src/bin/sol-fbp-to-dot/Kconfig
```diff
@@ -1,4 +1,4 @@
 config FBP_TO_DOT
 	bool "fbp-to-dot converter"
-	depends on SOL_PLATFORM_LINUX && RESOLVER_CONFFILE && INSPECTOR
+	depends on SOL_PLATFORM_LINUX && NODE_DESCRIPTION && INSPECTOR
```
barbieri: This one also needs conffile resolver.

src/lib/flow/Kconfig
```diff
@@ -4,8 +4,60 @@ config FLOW
...
 config RESOLVER_CONFFILE
 	bool "Resolver conffile"
+	depends on FLOW
 	default y
 
-config FLOW_RESOLVER
-	bool "Flow resolver"
+config NODE_DESCRIPTION
+	bool "Node description support"
+	depends on RESOLVER_CONFFILE
```
barbieri: This is not right, one can have NODE_DESCRIPTION without a conffile resolver (i.e.: one can provide one externally in his binary)

OTOH, RESOLVER_CONFFILE needs NODE_DESCRIPTION.

src/lib/flow/Kconfig
```diff
@@ -4,8 +4,60 @@ config FLOW
...
 	default y
+	help
+            Add node description support to enable runtime introspection.
```
barbieri: Be more descriptive here, Kbuild makes long descriptions nice to read. I'd say:

Keep meta information such as author, version, ports and options names in the binaries. This allows runtime to resolve a port index from its name or to parse options from text as it will know about the options, their names, types and requirements. These are essential for tool such as sol-fbp-runner or sol-fbp-generator.

The meta information will slightly increase the binary size, thus may be disabled for systems with constrained memory requirements. In this case only the sol-flow.h (low level C api) can be used as sol-flow-builder.h and FBP parsers are all dependent on this option.

tools/build/Makefile.rules
```diff
@@ -167,7 +174,7 @@ endef
 $(foreach curr,$(tests),$(eval $(call make-test,$(curr))))
 
 find-gen-hdrs = $(foreach gen,$($(2)-gens),$($(gen)-hdr))
-external-module-flags = -DSOL_FLOW_NODE_TYPE_MODULE_EXTERNAL=1 -DSOL_PLATFORM_LINUX_MICRO_MODULE_EXTERNAL=1
+external-module-flags = -DSOL_FLOW_NODE_TYPE_MODULE_EXTERNAL=1 -DSOL_PLATFORM_LINUX_MICRO_MODULE_EXTERNAL=1 -DPIN_MUX_MODULE=1
```
barbieri: we should use -DSOL_PIN_MUX_MODULE_EXTERNAL=1 here to be consistent with other defines.

PR #164, opened 2015-06-30
Kbuild

glima:
With these 2 commits, we're finally able to cross-compile to the Galileo, with
```make CFLAGS="$CFLAGS --sysroot=/extra/galileo-sdk-ii/sysroots/i586-poky-linux-uclibc -m32 -march=i586" TOOLCHAIN_PREFIX="i586-poky-linux-uclibc-"```

### DIFF
--- a/src/lib/common/sol-missing.h
+++ b/src/lib/common/sol-missing.h
@@ -98,3 +98,7 @@ err:
 #ifndef I2C_RDRW_IOCTL_MAX_MSGS
 #define I2C_RDRW_IOCTL_MAX_MSGS 42
 #endif
+
+#if !HAVE_SYS_AUXV_H
+#define AT_EXECFN  31   /* filename of program */
+#endif // HAVE_SYS_AUXV_H
--- a/src/shared/sol-util.c
+++ b/src/shared/sol-util.c
@@ -35,7 +35,9 @@
 #include <limits.h>
 #include <stdio.h>
 #include <stdlib.h>
+#ifdef HAVE_SYS_AUXV_H
 #include <sys/auxv.h>
+#endif
 #include <sys/stat.h>
 #include <sys/types.h>
 #include <unistd.h>
@@ -43,6 +45,7 @@
 #include "sol_config.h"
 #include "sol-util.h"
 #include "sol-str-slice.h"
+#include "sol-missing.h"
 
 #define CHUNK_SIZE 4096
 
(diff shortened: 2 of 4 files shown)

### CODE REVIEW

File: src/shared/sol-util.c [hunk 1]
```diff
@@ -312,14 +315,37 @@ sol_util_str_split(const struct sol_str_slice slice, const char *delim, size_t m
...
 {
-    char cwd[PATH_MAX] = { NULL }, *execfn;
+    char cwd[PATH_MAX] = { NULL }, *execfn = NULL;
 
     if (getcwd(cwd, sizeof(cwd)) == NULL)
-        return -1;
+        return -errno;
 
+#ifdef HAVE_SYS_AUXV_H
     execfn = (char *)getauxval(AT_EXECFN);
```
lpereira: This is defined in `sol-missing.h` but this code won't be built because `HAVE_SYS_AUXV_H` isn't defined.

src/shared/sol-util.c
```diff
@@ -312,14 +315,37 @@ sol_util_str_split(const struct sol_str_slice slice, const char *delim, size_t m
...
+
     if (!execfn)
-        return -1;
+        return -errno;
+#else
+    unsigned char data[2 * 8];
+    FILE *f = fopen("/proc/self/auxv", "r");
```
lpereira: Can't `readlink("/proc/self/exe")` be used here?

src/lib/common/sol-missing.h
```diff
@@ -98,3 +98,7 @@ accept4(int sockfd, struct sockaddr *addr, socklen_t *len, int flags)
 #ifndef I2C_RDRW_IOCTL_MAX_MSGS
 #define I2C_RDRW_IOCTL_MAX_MSGS 42
 #endif
+
+#if !HAVE_SYS_AUXV_H
+#define AT_EXECFN  31   /* filename of program */
```
lpereira: If `sys/auxv.h` isn't available, then it's likely that `getauxval()` isn't as well. Not sure if this is correct.

src/shared/sol-util.c
glima: The sol-missing entry is for when the header does *not* exist, hence it will be there in this case.

src/shared/sol-util.c
glima: I can check, thanks.

src/lib/common/sol-missing.h
glima: Then the other path will run, as I said.

src/shared/sol-util.c
barbieri: I guess getauxval() is not defined only on !Linux, right? If that's the case create a sol-util-linux.[ch] and define it there, do not use it in RIOT and other systems where getting the program name would not have a meaning.

Actually, the only user is sol_util_get_rootdir() that has no meaning outside of linux as Contiki, RIOT and others won't have rootdir() or a filesystem at all.

lpereira: If the header does not exist, then `getauxval()` won't exist, and this value won't be needed.

glima: And old Linuxes. I thought of embedding an additional check (&& is_linux) there, but @dorileo commented that the using these functions does not even make sense on riot et al (single, static, binary).

barbieri: manpage says it was introduced in glibc 2.19 released on Feb 2014. I'm okay to not support it unless we have demand.

glima: It's uclibc on the Galileo image, moreover.

ceolin: Why won't you check for the symbol instead only check for the header (same thing is done for accept4), and if it's not defined create it on sol-missing ? 

I don't see where it is different from the strndupa and accept4 cases ?

glima: Fair, @ceolin, doing that (and testing @lpereira's approach too).

glima: @ceolin Thinking better, getauxval() returns ptrs to static memory allocated somewhere else, and I can't do that on the clone (nor have a free() afterwards version). I'll stick with the alternate behavior in place.

PR #166, opened 2015-06-30
build: make sure we include all objects for static lib

dorileo:
For static lib we should include all the objects not only the builtins
but also what comes from src/shared/.


zehortigoza:
Tested on Contiki +2

dorileo:
thanks, applied.

Issue #119, opened 2015-06-27
kbuild: kdbus in kconfig is misleading


lucasdemarchi:
See https://github.com/dorileo/soletta/commit/d556d826bfbb02fb7efbba21585a3888e512db41#commitcomment-11894119

glima:
Is this already solved?

PR #167, opened 2015-07-01
Build selective hardware

dorileo:

This patch also brings two other non related patches, 1 to fix the issue https://github.com/solettaproject/soletta/issues/119 and one fixes a build dependency for test-flow.

### DIFF
--- a/Kconfig
+++ b/Kconfig
@@ -22,6 +22,11 @@ endmenu
 menu "Core library"
 source "src/lib/common/Kconfig"
 source "src/lib/flow/Kconfig"
+
+config NETWORK
+        bool "Network support"
+	default y
+
 source "src/lib/coap/Kconfig"
 source "src/shared/Kconfig"
 endmenu
--- a/src/lib/common/Kconfig
+++ b/src/lib/common/Kconfig
@@ -98,8 +98,13 @@ config HARDWARE_SPI
 	help
           Globaly enables spi hardware support.
 	default y
-endmenu
 
+config HARDWARE_UART
+	bool "uart"
+	help
+          Globaly enables uart hardware support.
+	default y
+endmenu
 
 config SOCKET_LINUX
 	bool "Linux sockets"
--- a/src/lib/common/Kconfig
+++ b/src/lib/common/Kconfig
@@ -110,6 +110,12 @@ config HARDWARE_I2C
 	help
           Globaly enables i2c hardware support.
 	default y
+
+config HARDWARE_GPIO
+	bool "gpio"
+	help
+          Globaly enables gpio hardware support.
+	default y
 endmenu
 
 config SOCKET_LINUX
--- a/src/lib/common/Kconfig
+++ b/src/lib/common/Kconfig
@@ -86,6 +86,15 @@ config SOL_BUS
 	depends on SOL_PLATFORM_LINUX && HAVE_SYSTEMD && HAVE_SD_BUS
 	default n
 
+menu "Hardware support"
+config HARDWARE_PWM
+	bool "pwm"
+	help
+          Globaly enables pwm hardware support.
+	default y
+endmenu
+
+
 config SOCKET_LINUX
 	bool "Linux sockets"
 	depends on SOL_PLATFORM_LINUX
--- a/src/lib/common/Kconfig
+++ b/src/lib/common/Kconfig
@@ -27,7 +27,7 @@ config PLATFORM_SYSTEMD
 	bool "systemd"
 	depends on HAVE_SYSTEMD
 	select SOL_PLATFORM_LINUX
-	select KDBUS
+	select SOL_BUS
 endchoice
 
 choice
@@ -81,9 +81,9 @@ config DEFAULT_LOG_LEVEL_DEBUG
 	bool "debug"
 endchoice
 
-config KDBUS
-	bool "Kdbus"
-	depends on SOL_PLATFORM_LINUX && HAVE_SYSTEMD && HAVE_KDBUS
+config SOL_BUS
+	bool "Sol-bus"
+	depends on SOL_PLATFORM_LINUX && HAVE_SYSTEMD && HAVE_SD_BUS
 	default n
 
 config SOCKET_LINUX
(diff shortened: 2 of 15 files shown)


src/lib/common/Kconfig
```diff
@@ -81,11 +81,43 @@ config DEFAULT_LOG_LEVEL_DEBUG
...
+config SOL_BUS
+	bool "Sol-bus"
+	depends on SOL_PLATFORM_LINUX && HAVE_SYSTEMD && HAVE_SD_BUS
 	default n
 
+menu "Hardware support"
+config HARDWARE_PWM
+	bool "pwm"
+	help
+          Globaly enables pwm hardware support.
```
lpereira: s/Globaly/Globally/
s/pwm/PWM/
s/spi/SPI/
s/uart/UART/
s/i2c/I2C/
s/gpio/GPIO/

mbelluzzo: s#Globaly enables#Enable/Disable# or just Enable. I dont think that Globally is needed.

src/lib/common/sol-interrupt_scheduler_riot.c
```diff
@@ -40,8 +40,10 @@ static kernel_pid_t pid;
 
 enum interrupt_type {
     GPIO,
```
mbelluzzo: What if HARDWARE_GPIO isn't enabled as well?

src/lib/common/Kconfig
```diff
@@ -81,11 +81,43 @@ config DEFAULT_LOG_LEVEL_DEBUG
...
+          Globaly enables i2c hardware support.
+	default y
+
+config HARDWARE_GPIO
+	bool "gpio"
+	help
+          Globaly enables gpio hardware support.
+	default y
+endmenu
+
```
mbelluzzo: I prefer to keep a standard and s/HARDWARE_*/HAVE_*/

Kconfig
```diff
@@ -22,6 +22,11 @@ endmenu
 menu "Core library"
 source "src/lib/common/Kconfig"
 source "src/lib/flow/Kconfig"
+
+config NETWORK
```
mbelluzzo: s/NETWORK/HAVE_NETWORK/

src/shared/Makefile
```diff
@@ -19,11 +29,55 @@ obj-libshared-m-extra-ldflags += $(GLIB_LDFLAGS)
...
+obj-libshared-m += sol-i2c-riot.o
+endif
+
+ifeq (y,$(HARDWARE_GPIO))
+obj-libshared-m += sol-gpio-riot.o
+endif
+
+ifeq (y,$(NETWORK))
+obj-libshared-m += sol-network-riot.o
+endif
```
mbelluzzo: Does the following works?
```Makefile
ifeq (y, $PLATFORM_RIOTOS))
obj-libshared-$(HAVE_*) += sol-*-riot.o
...
obj-libshared-$(HAVE_*) += sol-*-riot.o
endif
```

mbelluzzo:
Also, its missing a dependence on HAVE_GPIO in Pin Multiplexer. The whole Pin Multiplexer support is dependent on GPIO.

src/lib/common/sol-interrupt_scheduler_riot.c
dorileo: if gpio is ripped out the gpio scheduler will never get called ever - well uart will never be called as well - but in the case of uart it depends on uart_t which will not be present in that case (if the platform doesn't support or implement it) that's the why I'm ifdef'ing this code here, actually this enum could not even be ifdef'ed but for "consistency sake" I'm also guarding it... if you think we should keep consistency between gpio and uart we could ifdef this GPIO enum as well, but it's not required.

src/lib/common/Kconfig
dorileo: we're using HAVE_ for a different purpose and I wouldn't like to mix them all. HAVE_* always indicates what your host system provides and these HAVE_* variables are defined by the dependency-resolver, it doesn't tell you what soletta has but what the host system provides us.

About the kconfig docs. Where did you see this suggestion? this is totally up to the users to define the patterns they want to follow, if you see the different kconfig users you'll identify multiple patterns. If you're talking about kconfig-language.txt or their friends they're indicating/suggesting common uses for linux kernel itself, since we're not the linux kernel we can use whatever we want. Remember, the building context/environment for linux kernel is totally different to ours, they don't link against shared libraries for example, and they know very little about the host system.

Kconfig
dorileo: ditto...

src/shared/Makefile
dorileo: not at all...

dorileo:
all right for pin mux... I must investigate why the build is not breaking in a scenario without gpio but with pin mux enabled.

src/lib/common/Kconfig
dorileo: all right...

src/lib/common/Kconfig
dorileo: just a complimentary comment, I agree to maybe not stick to HARDWARE_* variables and come up with a better prefix, but I'll not want to mix the uses of HAVE_* variables.

dorileo: @mbelluzzo  again, I saw you introduced the HAVE_PIN_MUX configuration, I didn't see it before it got applied, but I would send a patch renaming this config to avoid future mixing (like I mentioned before) and the pattern propagation.

mbelluzzo: I think you can send the fix with the modifications you will have to do
when check for GPIO on pin mux.

mbelluzzo: "HAVE_* always indicates what your host system provides and these HAVE_* variables are defined by the dependency-resolver,"

And the main purpose of being able to turn on/off those options instead based on the platform/host (hardware + os) providing us such features? We may lack automatic dependency  check for those now, but may be possible in the future and even if not, it looks to me that HAVE_ meet its purpose here.

dorileo:
v2 coming...

PR #169, opened 2015-07-01
Split sol-util to each platform

zehortigoza:
Riot don't have a filesystem also Contiki don't have, so spliting
sol-util to fix build.

### DIFF
--- a/src/shared/sol-util-riot.c
+++ b/src/shared/sol-util-riot.c
@@ -1,0 +1,88 @@
+/*
+ * This file is part of the Soletta Project
+ *
+ * Copyright (C) 2015 Intel Corporation. All rights reserved.
+ *
+ * Redistribution and use in source and binary forms, with or without
+ * modification, are permitted provided that the following conditions
+ * are met:
+ *
+ *   * Redistributions of source code must retain the above copyright
+ *     notice, this list of conditions and the following disclaimer.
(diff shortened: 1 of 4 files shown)


dorileo:
tested and looks good to me.
+1

ceolin:
:+1:  I've done the same thing here while testing the RIOT's build

src/shared/sol-util-riot.c
```diff
@@ -0,0 +1,88 @@
...
+sol_util_timespec_get_current(void)
+{
+    struct timespec t;
+
+    clock_gettime(CLOCK_MONOTONIC, &t);
+    return t;
+}
+
+int
+sol_util_write_file(const char *path, const char *fmt, ...)
```
barbieri: I see no point in creating these.

Change the sol-util.h to include the sol-util-common.h (to match sol-util-common.c) and conditionally include the util for the platform using the defines in sol-common-buildopts.h

Then we can catch people using unsupported code in their call site.

zehortigoza: Yes, for this patch alone the sol_util_timespec_get_current(void) should stay on sol-util-common.c, I left on each platform because Contiki don't have clock_gettime() so there is a different implementation of sol_util_timespec_get_current().

PR #170, opened 2015-07-01
Share common code between Riot and Posix implementation of mainloop

zehortigoza:
Part of the code was being duplicated between those implementations
of mainloop.

### DIFF
--- a/src/lib/common/sol-mainloop-impl-common.c
+++ b/src/lib/common/sol-mainloop-impl-common.c
@@ -1,0 +1,255 @@
+/*
+ * This file is part of the Soletta Project
+ *
+ * Copyright (C) 2015 Intel Corporation. All rights reserved.
+ *
+ * Redistribution and use in source and binary forms, with or without
+ * modification, are permitted provided that the following conditions
+ * are met:
+ *
+ *   * Redistributions of source code must retain the above copyright
+ *     notice, this list of conditions and the following disclaimer.
+ *   * Redistributions in binary form must reproduce the above copyright
+ *     notice, this list of conditions and the following disclaimer in
+ *     the documentation and/or other materials provided with the
+ *     distribution.
+ *   * Neither the name of Intel Corporation nor the names of its
+ *     contributors may be used to endorse or promote products derived
+ *     from this software without specific prior written permission.
+ *
+ * THIS SOFTWARE IS PROVIDED BY THE COPYRIGHT HOLDERS AND CONTRIBUTORS
+ * "AS IS" AND ANY EXPRESS OR IMPLIED WARRANTIES, INCLUDING, BUT NOT
+ * LIMITED TO, THE IMPLIED WARRANTIES OF MERCHANTABILITY AND FITNESS FOR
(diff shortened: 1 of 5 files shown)


src/lib/common/sol-mainloop-impl-common.c
```diff
@@ -0,0 +1,255 @@
...
+int
+sol_mainloop_impl_timeout_compare(const void *data1, const void *data2)
+{
+    const struct sol_timeout_common *a = data1, *b = data2;
+
+    return sol_util_timespec_compare(&a->expire, &b->expire);
+}
+
+void
+sol_mainloop_impl_common_shutdown(void)
```
lpereira: I believe `_common_` infix is all that's needed for these functions: they're not platform-specific functions, so there's no need to use the `_impl_` prefix.

src/lib/common/sol-mainloop-impl-riot.c
```diff
@@ -238,16 +87,16 @@ timex_set_until_next_timeout(timex_t *timex)
 void
 sol_mainloop_impl_run(void)
```
lpereira: This should be the `sol_mainloop_run()` function, which should call `sol_mainloop_impl_iter()` (or other similarly-named function), left to be implemented by the platform-specific code. This way, there's no need to expose the `_common_` functions.

src/lib/common/sol-mainloop-impl-riot.c
```diff
@@ -259,78 +108,31 @@ sol_mainloop_impl_run(void)
...
-    now = sol_util_timespec_get_current();
-    sol_util_timespec_sum(&now, &timeout->timeout, &timeout->expire);
-    ret = sol_ptr_vector_insert_sorted(&timeout_vector, timeout, timeout_compare);
-    SOL_INT_CHECK_GOTO(ret, != 0, clean);
-    return timeout;
-
-clean:
-    free(timeout);
-    return NULL;
+    return sol_mainloop_impl_common_timeout_add(timeout_ms, cb, data);
```
lpereira: Is the function call here really needed? If all platforms are going to use the same functions for these, do we still need `_impl_` versions of these functions?

src/lib/common/sol-mainloop-impl-common.c
mbelluzzo: +1 to that

src/lib/common/sol-mainloop-impl-riot.c
ulissesf: That's a good question. Right now POSIX just adds locking, and even that we could generalize to share more code. We won't support child watch on RIOT, I think, right?

src/lib/common/sol-mainloop-impl-riot.c
ulissesf: @lpereira, he's just following the abstraction already in place. We can however improve it. We can drop this and I take a look at pushing (maybe) just the threading part as implementation specific on each OS? This way we could share more code among POSIX and small OSes mainloops.

lpereira: I think it might be a good idea to try that and see if it works, yes.

zehortigoza:
Closing, we are going forward with https://github.com/solettaproject/soletta/pull/191

PR #173, opened 2015-07-01
Change installed headers

bdilly:


glima:
I like the approach, and it helps fix an issue with double inclusions of headers later.

vcgomes:
:+1: , I tested this and it goes a long way solving the problem of building external modules.

bdilly:
thanks for reviewing it, guys

PR #168, opened 2015-07-01
build: bring back the cheat-sheet target

dorileo:
We missed the cheat-sheet target, this patch brings it back.

Signed-off-by: Leandro Dorileo <leandro.maciel.dorileo@intel.com>

PR #174, opened 2015-07-01
Build bring back cheat sheet v2

dorileo:
v2:
  + introduced a new path to conditionally install builtins.json (install only if we have external modules), this fixes the issue pointed out by @anselmolsm here: https://github.com/solettaproject/soletta/pull/168#issuecomment-117692764

We missed the cheat-sheet target, this patch brings it back.

Signed-off-by: Leandro Dorileo leandro.maciel.dorileo@intel.com


glima:
Ack. I'll work on top of this to generate doxygen too.

anselmolsm:
LGTM :+1:

dorileo:
applied, thanks.

PR #175, opened 2015-07-01
log: Fix exposing internal header for external users of node gen script

glima:
Also fix double inclusion of some headers that is exposed with this.


vcgomes:
:+1: Just tested and it works.

glima:
Cool, merging.

PR #176, opened 2015-07-01
build: -gen.c should be generated inside build_stagedir

dorileo:
As far as we understand (and suggested by Ivan) we should be generating
the -gen.c files inside build_stagedir not $(top_srcdir)/src/.../


dorileo:
some missing parts... v2 coming.

PR #177, opened 2015-07-01
build: -gen.c should be generated inside build_stagedir

dorileo:
v2:
   + properly adjust cflags to out of src generated code location - since it's #include'ed

As far as we understand (and suggested by Ivan) we should be generating
the -gen.c files inside build_stagedir not $(top_srcdir)/src/.../


dorileo:
applied, thanks.

PR #178, opened 2015-07-01
build: fix a type on soletta.pc generation

dorileo:
Since we were not providing the correct for ${prefix} substitution
the soletta.pc was being generated without the proper value for
prefix key.


otaviobp:
+1 I tested this patch and it is working fine for me

dorileo:
applied fixing the typo in the commit msg...

PR #182, opened 2015-07-01
Build fix module less config

dorileo:
This series fixes issues on build soletta without MODULES support.


glima:
Both make sense.

dorileo:
applied...

PR #183, opened 2015-07-01
Build selective hardware v2

dorileo:
v2:
  + fixed indentation;
  + changed the config variables from HARDWARE_ to USE_;
  + pinmux now depends on gpio;
  + changed labels and help according to @lpereira and @mbelluzzo suggestions;
  + added a patch to rename HAVE_PIN_MUX to USE_PIN_MUX so we don't spread the mixed and wrong use of HAVE_ variables;

This patch also brings two other non related patches, 1 to fix the issue #119 and one fixes a build dependency for test-flow.

### DIFF
--- a/src/lib/common/Kconfig
+++ b/src/lib/common/Kconfig
@@ -93,6 +93,15 @@ config SOL_BUS
 	depends on SOL_PLATFORM_LINUX && HAVE_SYSTEMD && HAVE_SD_BUS
 	default n
 
+menu "Hardware Options"
+config USE_PWM
+	bool "PWM Support"
+	help
+          Enables PWM support.
+	default y
+endmenu
+
+
 config SOCKET_LINUX
 	bool "Linux sockets"
 	depends on SOL_PLATFORM_LINUX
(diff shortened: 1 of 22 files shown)


src/lib/common/Kconfig
```diff
@@ -88,18 +88,51 @@ config DEFAULT_LOG_LEVEL_DEBUG
...
+config SOL_BUS
+	bool "Sol-bus"
+	depends on SOL_PLATFORM_LINUX && HAVE_SYSTEMD && HAVE_SD_BUS
 	default n
 
+menu "Hardware Options"
+config USE_PWM
+	bool "PWM Support"
+	help
+          Enables PWM support.
```
lpereira: The help text isn't helping much. They're pretty similar to these comments, in fact:

    int i = 0; /* Declare an integer variable named "i", and initialize it to "0" */
    i++; /* Increment integer variable i */

Maybe expand the acronym in the help text ("Enable Pulse-Width Modulation support", "Enables Universal Asynchronous Receiver/Transmitter support", "Enable General Purpose I/O support", etc), give an example or two (e.g. for PWM, "Generally used to control the power supplied to devices connected to an I/O port, such as motors or LEDs."), and optionally a Wikipedia link (e.g. https://en.wikipedia.org/wiki/Pulse-width_modulation).

dorileo: well, actually I wasn't take too much care on this help message for now, but I can change it to a more descriptive text following you suggestion. Maybe include at least pat of this commit message as well.

Thanks...

dorileo:
v3 coming

PR #185, opened 2015-07-02
Expose C API for parsers, datatypes and hardware

bdilly:


bdilly:
merged with changes proposed by Caio

PR #186, opened 2015-07-02
Fix invalid read reported by valgrind

ibriano:
When we copy the defaults from an actual options struct, we need to
ensure we only read the real size of the type.


lpereira:
:+1:

cmarcelo:
+2

PR #187, opened 2015-07-02

dorileo:

### DIFF
--- a/data/jsons/dependencies.json
+++ b/data/jsons/dependencies.json
@@ -126,12 +126,14 @@
 	{
 	    "dependency": "chrpath",
 	    "type": "exec",
-	    "exec": "chrpath"
+	    "exec": "chrpath",
+            "required": "true"
 	},
 	{
 	    "dependency": "jsonschema",
 	    "type": "python",
-	    "pkgname": "jsonschema"
+	    "pkgname": "jsonschema",
+            "required": "true"
 	}
     ]
 }
--- a/data/scripts/dependency-resolver.py
+++ b/data/scripts/dependency-resolver.py
@@ -42,6 +42,7 @@ cfg_cflags = {}
 cfg_ldflags = {}
 cfg_kconfig = {}
 makefile_vars = {}
+nof_deps = {}
 
 def run_command(cmd):
     try:
@@ -51,7 +52,6 @@ def run_command(cmd):
     except subprocess.CalledProcessError as e:
         return e.output, False
 
-
 def handle_pkgconfig_check(args, conf):
     dep = conf["dependency"].upper()
     pkg = conf["pkgname"]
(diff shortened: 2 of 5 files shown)


data/jsons/dependencies.json
```diff
@@ -126,12 +126,14 @@
 	{
 	    "dependency": "chrpath",
 	    "type": "exec",
-	    "exec": "chrpath"
+	    "exec": "chrpath",
+            "required": "true"
```
barbieri: indentation here and for the next "required".

data/scripts/dependency-resolver.py
```diff
@@ -59,6 +59,7 @@ def handle_pkgconfig_check(args, conf):
     max_ver = conf.get("max-version")
     exact_ver = conf.get("exact-version")
     ver_match = True
+    required = True if conf.get("required") else False
```
barbieri: ugly as Perl.

Python version is: required = conf.get("required", False)

Also, make sure you're using the JSON boolean (true, false) not a string!

data/jsons/dependencies.json
barbieri: also, it shouldn't have quotes in "true", just a boolean as supported by JSON.

it would be worth to write a JSON Schema for your dependencies.json, use our common load+check block as done by node-type-gen.py so we can avoid mistakes in there.

data/scripts/dependency-resolver.py
```diff
@@ -144,7 +156,13 @@ def handle_python_check(args, conf):
...
     cmd = "%s %s" % (sys.executable, f.name)
     output, status = run_command(cmd)
-    makefile_vars["HAVE_PYTHON_%s" % dep] = "y" if status else "n"
+
+    if status:
+        makefile_vars["HAVE_PYTHON_%s" % dep] = "y"
+    else:
+        makefile_vars["HAVE_PYTHON_%s" % dep] = "n"
+        if required:
+            nof_deps[conf.get("pkgname")] = "python module"
```
barbieri: if pkgname is not defined, this will set nof_deps[None] = "python module", which doesn't make sense.

In such case it seems pkgname is mandatory, then use: nof_deps[conf["pkgname"]].

If you want to be safe against broken dependencies.json, write a schema and use our common function to load+verify the JSON against a schema.

data/scripts/dependency-resolver.py
```diff
@@ -160,11 +178,23 @@ def var_str(items):
         output += "%s ?= %s\n" % (k, v)
     return output
 
+def notfound_deps(items):
+    if not len(items):
```
barbieri: Python is: "if not items:", an empty tuple or list evaluates to False.

data/jsons/dependencies.json
dorileo: would I need jsonschema module to use the json schema? if so it doesn't make sense since this scripts is run before anything and to check even for jsonschema presence.

data/scripts/dependency-resolver.py
dorileo: all right...

dorileo:
v2 coming.

PR #188, opened 2015-07-02
modules: initialize the data for packets sent by generators

cmarcelo:
Fields other than 'val' in {int,float}-generator output were not
initialized, so the packet would contain unitialized data. When another
node used that packet and extracted data from there, it could read
garbage.

This error was flagged by 'make check-fbp-valgrind' in the tests for
filter-repeated.


barbieri:
+2

cmarcelo:
Applied in 4d29fe87a4fcd891aa5baa6dd23002aebb82ef39.

PR #190, opened 2015-07-02
sol-flow-builder: Add a check for port description

anselmolsm:
Issue pointed out by KW.

Avoids a possible NULL pointer dereference in the following line.


barbieri:
while it doesn't hurt, I believe this was a false positive since we get an existing index and since we never append null to that vector, the return will never be null.

anselmolsm:
Yeah, I got that feeling too.
Well, as it doesn't hurt, applied...

PR #195, opened 2015-07-02
Add ldflags to ccode dependency check

rchiossi:
If a config needs to check if a library was compiled with a given
feature using "ccode" mode, ldflags must be passed to the compiler to
properly find the symbols for the fragment.


dorileo:
Ok, like I said before, I got your point but was concerned if we didn't want to split the flags used to compile the tests and the exported by it. But, ok, it seems we no such case.

Having said that, +2

Thanks...

rchiossi:
Merged. Thanks.

PR #196, opened 2015-07-02
builtins.json: we should generate it on build

dorileo:
The builtins.json file was being generated only on install target,
we must generate it while building.


ricardotk:
ACK, go ahead

dorileo:
v2 coming...

PR #199, opened 2015-07-02
Fix foosball.fbp after changes in boolean nodes

anselmolsm:
boolean/* nodes now use array ports. This samples wasn't
properly updated accordingly.


ricardotk:
+1

anselmolsm:
Applied, tks.

PR #200, opened 2015-07-02
build config: Change default log level

anselmolsm:
The default log level was set to be ERROR, but it should
be DEBUG.


ibriano:
It used to be WARNING, not DEBUG.

anselmolsm:
Correct. I was dealing with DBG when I found it. v2: #203

PR #201, opened 2015-07-02
fbp-generator: add support for DECLARE statement v2

ricardotk:
Differences from v1:
    - each generated fbp type has its own type store and the JSONs will be stored in a common type store.
    - changed the generated C code, to split the creation of fbp type into functions, creating them recursively.

Now the generator handles the declaration and usage of new types in the FBP itself.

e.g.
DECLARE=MyType:fbp:my-type.fbp
my_node(MyType) OUT -> ...

In the generated C code now, we split the creation of fbp types into functions, creating them recursively.


ricardotk:
I'll apply the comments and merge it, thanks for the review!

PR #202, opened 2015-07-02
More array ports

lgywata:
More changes to array ports schema: Float and Int types.

### DIFF
--- a/src/modules/flow/float/float.json
+++ b/src/modules/flow/float/float.json
@@ -12,19 +12,11 @@
       "in_ports": [
         {
           "data_type": "float",
-          "description": "First port of addition operation.",
+          "description": "Thirty two ports for addition operation. Indexed from 0 to 31.",
           "methods": {
             "process": "operator_process"
           },
-          "name": "IN0"
-        },
-        {
-          "data_type": "float",
-          "description": "Second port of addition operation.",
-          "methods": {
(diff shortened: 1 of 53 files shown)


lgywata:
@barbieri suggested it and @cmarcelo also mentioned the 32 array ports for operations like addition, multiplication, etc. I sent an email to the list about this saying that I would proceed with the changes if no one else had a different opinion. But, yes, I agree the 32 is arbitrary, but If I'm a Soletta user I'd prefer the 32 ports instead of having to cascade several nodes to have a simple arithmetic operation done. Besides, we can get back here in the future and change it again when dynamic arrays is supported.

ibriano:
It was discussed in the devapp mailing list, which brings up the

ibriano:
It was brought up in the devapp-list, which brings up the soletta-dev that we should be using for these things (unless it's something that can't be discussed in public).
32 is arbitrary, but seems large enough to be sufficient for now. About dynamic array ports, are people seeing a need to have that in the near future? It would require some major reworking of the way we do things right now.

anselmolsm:
@lgywata I see that you updated the tests, remember to also check the samples affected by these changes.

lgywata:
@anselmolsm, you're right. Thanks!

lgywata:
v2 sent with samples included. #220

PR #203, opened 2015-07-02
build config: Change default log level

anselmolsm:
The default log level was set to be ERROR, but it should
be WARNING.


barbieri:
-1, the kconfig is wrong and the option name is misleading. That's the MAXIMUM_LOG_LEVEL, not the default. I'm fixing that in my PR #205

PR #204, opened 2015-07-02
lib/io: set gpio and pwm config api version on usage

bdilly:
They were set to 0 instead of current version.


mbelluzzo:
+2

bdilly:
merged

PR #205, opened 2015-07-02
Kconfig review

barbieri:
Sorry not making various commits, but the last one has a single purpose and I fixed the problems as I saw them, not bothering to split as it wouldn't help much.


barbieri:
merged

PR #206, opened 2015-07-02
Build - Fix node description depencies

ceolin:


ceolin:
close with fixes in https://github.com/solettaproject/soletta/pull/213c

PR #207, opened 2015-07-02
Fix samples that should be using array ports

anselmolsm:
Fix samples that use boolean/* and string/* node types. Update their code to use array ports.


glima:
Ack.

PR #209, opened 2015-07-02
build: Add support for setting the pkg-config to use

vcgomes:
This is useful when cross-compiling, for example, I have a 32-bit
pkg-config installed as i686-pc-linux-gnu-pkg-config configured to look
for things in /usr/i686-pc-linux-gnu/lib.

So, with this patch, I just do the following when I want a 32-bit
version of Soletta:

$ make TOOLCHAIN_PREFIX=i686-pc-linux-gnu- HOST_PREFIX=x86_64-pc-linux-gnu- PKG_CONFIG=i686-pc-linux-gnu-pkg-config menuconfig

$ make TOOLCHAIN_PREFIX=i686-pc-linux-gnu- HOST_PREFIX=x86_64-pc-linux-gnu- PKG_CONFIG=i686-pc-linux-gnu-pkg-config


barbieri:
while I don't see a problem with it, usually you do not need to provide a different pkg-config, just set PKG_CONFIG_LIBDIR and PKG_CONFIG_PATH so it won't use your host.

vcgomes:
Yeah, I know. It is just that the distro I use changed the way it builds packages (everything is a cross-compilation), and I found this way of dealing with multiple trees neat.

dorileo:
awesome, this is useful for yocto as well, we're still fixing and adjusting things for cross compilation...
thanks..
+2

data/scripts/dependency-resolver.py
```diff
@@ -61,17 +61,17 @@ def handle_pkgconfig_check(args, conf):
     ver_match = True
 
     if exact_ver:
-        cmd = "pkg-config --exact-version=%s %s" % (exact_ver, pkg)
+        cmd = "%s --exact-version=%s %s" % (args.pkg_config, exact_ver, pkg)
```
lpereira: Unrelated to this patch, but creating a command string like this is most often wrong, as arguments are not properly escaped.

It would be a better idea to use an argument vector (akin to whatever you pass to any function in the `exec()` family), or, at the very least, pass the version and pkg-name in single quotes (which isn't 100% ideal but should avoid most gotchas).

Doing this will avoid problems if package names or versions contain special characters to the shell.

bdilly:
@vcgomes did you forget to merge it?

PR #210, opened 2015-07-02
Fix builds without log.

barbieri:
If we disable logs, then we shouldn't compile sol-log.c or
sol-log-impl-*.c and we better not declare the log domain structure at
all, neither call the functions to touch it.


barbieri:
merged.

PR #212, opened 2015-07-02
Riot fixes v2

zehortigoza:


src/lib/common/Makefile
```diff
@@ -30,6 +31,7 @@ obj-core-$(PLATFORM_LINUX_MICRO) += \
 obj-core-$(SOCKET_LINUX) += \
     sol-socket-linux.o
 obj-core-$(PLATFORM_RIOTOS) += \
+    sol-log-impl-riot.o \
```
barbieri: I just did a commit to master that changes this. It would break if we disable logs.

Please check at the beginning of the file a block "if (y,$(LOG))" and then add this there.

zehortigoza: Thanks, patch updated.

glima:
:+1:

zehortigoza:
Merged

PR #214, opened 2015-07-02
sol-pin-mux: Add Intel Edison mux description

mbelluzzo:
Also add rules to detect Intel Edison platform.

Tested and the platform was detect and the mux rules successfully applied!


mbelluzzo:
As I told you last PR, that's intended, so the schema will allow it.

mbelluzzo:
Merged

PR #215, opened 2015-07-03
Build random fixes and improvements.

dorileo:


lpereira:
Not sure if the multithreaded checks are actually useful.

I'd wait until they're actually a problem and work on a possibly better alternative (for instance, using `multiprocess.pool.ThreadPool`, and then their `map` and `apply` variants, so that a manageable amount of processes/threads will be used) rather than get the big guns early in the game.

dorileo:
ok guys, v2 coming

PR #216, opened 2015-07-03
fbp-generator: write to a file only if the generation suceed

ricardotk:
Before this patch, we were appending to the file piece by piece,
at the risk of generating the file by half if the generation fails
at some part.

Now we're creating an unnamed temporary file and when the generation
ends, we link this unnamed temporary file to a filepath that the
user provide us.

### DIFF
--- a/src/bin/sol-fbp-generator/main.c
+++ b/src/bin/sol-fbp-generator/main.c
@@ -32,7 +32,11 @@
 
 #include <ctype.h>
 #include <errno.h>
+#include <fcntl.h>
+#include <limits.h>
 #include <stdio.h>
+#include <sys/types.h>
+#include <sys/stat.h>
 #include <unistd.h>
 
 #include "sol-arena.h"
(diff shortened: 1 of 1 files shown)


src/bin/sol-fbp-generator/main.c
```diff
@@ -694,8 +727,20 @@ main(int argc, char *argv[])
         goto fail_resolve;
     }
 
+    fd = open(".", O_TMPFILE | O_RDWR, S_IRUSR | S_IWUSR);
```
lpereira: The generator is a pretty portable piece of software, and `O_TMPFILE` is not only Linux-specific, it's also pretty recent (3.11).

Also, I'm not sure how opening "." is the way to go. What if the process' current working directory sits on a read-only file system? What if the current directory is on a different filesystem than `args.output_file`... will `linkat()` still work?

src/bin/sol-fbp-generator/main.c
```diff
@@ -637,6 +647,29 @@ resolve_nodes(const struct sol_fbp_graph *g, struct type_store *store)
...
+write_to_output_file(void)
+{
+    char fd_path[PATH_MAX];
+    int r;
+
+    r = snprintf(fd_path, PATH_MAX, "/proc/self/fd/%d", fd);
+    if (r < 0 || r >= PATH_MAX)
+        goto failure;
+
+    if (access(args.output_file, F_OK) == 0 && unlink(args.output_file) != 0)
```
lpereira: This line, alongside the call to `linkat()` below, is a TOCTOU.

ricardotk:
Thanks for the comments, I'll send a v2!

PR #217, opened 2015-07-03
build: fix oic generator race condition

dorileo:
The oic generator targets were in race condition since it's a pre
generation dependency and the modules' itself. Both could have their
execution in clash and break the oic module's source code generation;

This patch removes the pre generation condition and leave it up to
the module to directly depend on it and require its generation up
to their build conditioning.


rchiossi:
+1

PR #218, opened 2015-07-03
Doc v2

glima:
Changes since https://github.com/solettaproject/soletta/pull/211:

 * No more kconfig menu for docs -- the rule will be there if the deps are found
 * More dependency checks to achive that

Doxygen output is now a make target.


glima:
@bdilly Now that lives in the mailoop's header. If all is fine besides those issues, I could merge.

bdilly:
@glima go for it

glima:
Cool, merging.

PR #219, opened 2015-07-03
code style: fix an old demand of indenting continuation lines properly

glima:
Like function headers, function invocations, macro declarations, etc.
We always indent to a fixed size, now.


glima:
Yes, I have. Thanks, merging.

PR #221, opened 2015-07-03
Build v3

ceolin:
- v3
 - Fix commit message
 - Keep the resolver builtin as const (and set the conffile to const)
 - Fix how the resolver is set in the build


dorileo:
The series looks good to me +1
Thanks @ceolin

ceolin:
new pullreq with more fixes https://github.com/solettaproject/soletta/pull/226

PR #222, opened 2015-07-03
fbp-generator: write to a temporary file until the generation ends v2

ricardotk:
Differences from v1:
    - start using mkostemp()
    - don't use O_TMPFILE and linkat() and use temporary file and rename instead.
    - use sol_util_strerrora()

Before this patch, we were appending to the output file piece by
piece, at the risk of generating the output file by half if the
generation fails at some part.

Now we're creating a temporary file and when the generation ends,
we rename to a filepath that the user provide us.

### DIFF
--- a/src/bin/sol-fbp-generator/main.c
+++ b/src/bin/sol-fbp-generator/main.c
@@ -32,6 +32,7 @@
 
 #include <ctype.h>
 #include <errno.h>
+#include <fcntl.h>
 #include <stdio.h>
 #include <unistd.h>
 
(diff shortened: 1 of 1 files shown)


src/bin/sol-fbp-generator/main.c
```diff
@@ -645,6 +652,7 @@ main(int argc, char *argv[])
     struct sol_fbp_graph graph = {};
     struct type_store *store;
     struct type_description **descs;
+    char temp_file[] = ".sol-generated.c.tmp-XXXXXX";
```
lpereira: Any reason for these temporary files to be hidden?

ricardotk: There's no reason at all, do you think it would be better if they're not hidden?

lpereira: Makes it easier to spot stale files if something went wrong.

lpereira:
LGTM with comment. If nobody else spots anything, go ahead and merge without passing a hidden file template to `mktemp()`; there's no need to send another revision.

glima:
LGTM too (also in favor of non-hidden name).

anselmolsm:
:+1:  too

PR #223, opened 2015-07-03
flow: use sol_util_strerrora() instead of sol_util_strerror()

ricardotk:
Since sol_util_strerrora() creates a buffer, there's no need to
create it and use sol_util_strerror().


lpereira:
:+1:

PR #224, opened 2015-07-03
parser: use sol_util_strerrora() instead of strerror()

ricardotk:
strerror() should never be used in Soletta.

sol_util_strerrora() is a wrapper for that that's always
threadsafe, regardless of implementation details.


anselmolsm:
LGTM

PR #225, opened 2015-07-03
Expose C API to deal with network, coap and oic

bdilly:
It would be awesome if any of the guys that worked
more on that could review it. Specially because
I've tried to fix some issues found on them.

@ceolin , @acidx , @rchiossi , @vcgomes ^  =D


src/lib/comms/sol-oic-client.c
```diff
@@ -57,6 +56,27 @@
 #define IOTIVITY_CON_REQ_OBS_MID 0x7d44
 #define IOTIVITY_NONCON_REQ_MID 0x7d40
 
+#define OIC_RESOURCE_CHECK_API(ptr, ...) \
```
lpereira: Since exported structures will always have an API version member and these checks will be performed, it might be a good idea to move this code to a general purpose set of macros.

bdilly: Actually it would be nice to use other fields to identify the struct. I'm being lazy on this commit since I'm dealing with lots of structs.

For example, checks for sol_flow_packet_type version give a msg with more context, using type->name.

Maybe could create a macro that receives (ptr, version, msg_fmt, ...), but in VA_ARGS we could receive args for the msg or return val. Not sure about exporting a general check. For NULL_CHECKs it's good since you don't have any fields to identify the problematic struct.

bdilly:
v2 pushed

PR #226, opened 2015-07-03
Build v4

ceolin:
-v4
 - resolver is build only if node description is enabled
 - Fix the build without flow
 - Fix the tests dependencies


dorileo:
The patch looks good, I tested with the possible config combinations and works well...
Just this #ifdef minor issue...

ceolin:
Why we don't always define those variables ? I think it's better than always have to do "if defined && "

dorileo:
It's not up to us, this is the kconfig behavior and since we don't want to patch it we rather stick to its common/default usage.

dorileo:
btw the overall point is, these macros represent 4 states, 1) disabled (not defined symbol), 2) enabled #define MY_VAR y, 3) enabled as module #define MY_VAR m and 4) a string literal #define MY_VAR "my string var value".

ceolin:
Merged with the fixes, thanks for reviewing and helping with this serie.

PR #227, opened 2015-07-03
gpio-common: Make sol_gpio_open fail when config is NULL

anselmolsm:
There are many dereferences of config assuming it is valid.
Also changes the call to the pin mux setup, it should check
for config->dir, not for config.
Issue pointed out by coverity.


ricardotk:
+1

anselmolsm:
Applied with the change suggested by @mbelluzzo . tks.

PR #228, opened 2015-07-03
lib: changes versioned structs regarding api_version

bdilly:
Use uint16_t for api_version (most structs were using it
already).
In some cases, it leads to a hole, so a "reserved" field was
placed.
In some cases struct was changed a bit to reduce holes.


bdilly:
v2 pushed

PR #229, opened 2015-07-03
fbp-to-dot: Abort if sol_init fails

anselmolsm:

### DIFF
--- a/src/bin/sol-fbp-to-dot/main.c
+++ b/src/bin/sol-fbp-to-dot/main.c
@@ -306,7 +306,9 @@ main(int argc, char *argv[])
 
     struct sol_fbp_graph input_graph;
 
-    sol_init();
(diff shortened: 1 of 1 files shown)


ricardotk:
+1

src/bin/sol-fbp-to-dot/main.c
```diff
@@ -350,5 +352,6 @@ main(int argc, char *argv[])
     free(fbpfile);
     free(dotfile);
     sol_shutdown();
+end:
     return 0;
```
anselmolsm: But, of course this return should be != 0 in case of errors....

ricardotk: Yes, I agree... see how generator or runner handles it

anselmolsm:
v2 submitted: #231

PR #231, opened 2015-07-03
fbp-to-dot: Abort if sol_init fails v2

anselmolsm:
Also change the return value when errors occur.

### DIFF
--- a/src/bin/sol-fbp-to-dot/main.c
+++ b/src/bin/sol-fbp-to-dot/main.c
@@ -303,10 +303,13 @@ main(int argc, char *argv[])
     char *fbpfile = NULL;
     char *dotfile = NULL;
     struct sol_file_reader *fr;
+    int r = EXIT_FAILURE;
 
     struct sol_fbp_graph input_graph;
 
-    sol_init();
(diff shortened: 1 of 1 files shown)


src/bin/sol-fbp-to-dot/main.c
```diff
@@ -337,7 +340,8 @@ main(int argc, char *argv[])
 
     if (convert_fbp_to_dot(&input_graph, dotfile)) {
```
ricardotk: I would remove the brackets here and change the if condition to

if (convert_fbp_to_dot(&input_graph, dotfile) != 0)
   r = EXIT_SUCCESS;
else
    fprintf(stderr, "Couldn't convert from fbp to ...

ricardotk:
LGTM with comments,  if nobody else spots anything, go ahead and merge it.

src/bin/sol-fbp-to-dot/main.c
anselmolsm: OK. But in fact, convert_fbp_to_dot returns 0 on success ;)

anselmolsm:
Applied, tks

PR #232, opened 2015-07-03
lib: changes versioned structs regarding api_version

bdilly:
v2 changelog:
  * use bit fields of 0 len

Use uint16_t for api_version (most structs were using it
already).
In some cases, it may lead to a hole, so a unnamed bit fields with
lenght of 0 was placed.
In some cases struct was changed a bit to reduce holes.

### DIFF
--- a/src/lib/common/include/sol-pin-mux-modules.h
+++ b/src/lib/common/include/sol-pin-mux-modules.h
@@ -86,9 +86,10 @@ struct sol_pin_mux_controller {
  * Structure containing the recipes (lists of rules) that should be used
  * to multiplex the pins of a given platform
  */
-#define SOL_PIN_MUX_API_VERSION (1UL)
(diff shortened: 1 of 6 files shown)


mbelluzzo:
Just as an observation, we did some tests to see if this field would break "SOL_PIN_MUX_DECLARE" macro that uses positional initialization by testing some more often used structs (irange) and it appears that didn't break anything. I will try an explicit test on pin_mux just to be sure but I believe its OK

lpereira:
I'd also try using `pahole` or a similar tool to actually ensure that the hole is being reserved. I'd try that in both 32- and 64-bit, maybe using something different than a nameless int (such as a `size_t`), to test different alignment possibilities.

bdilly:
Holes are being reserved, yet I'm closing it but not merging since we decided to let the way it is for now while discussing pros and cons between possible approaches.

PR #233, opened 2015-07-03
sol-flow: Add an API to help the creation of complex nodes

anselmolsm:
The current complex nodes we have - the ones with subflows -
required more than a call to sol_flow_static_new_type. This
patch introduces an auxiliar funcion that does the steps
required to create such nodes.

### DIFF
--- a/src/lib/flow/include/sol-flow.h
+++ b/src/lib/flow/include/sol-flow.h
@@ -509,6 +509,17 @@ struct sol_flow_node_type *sol_flow_static_new_type(
     const struct sol_flow_node_options *opts,
     struct sol_flow_node_options *child_opts));
 
+struct sol_flow_node_type *sol_flow_static_new_container_node_type(
+    const struct sol_flow_node_type **base,
+    const struct sol_flow_static_node_spec nodes[],
+    const struct sol_flow_static_conn_spec conns[],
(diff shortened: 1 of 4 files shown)


anselmolsm:
Questions: 
I am not happy with the function name. Suggestions?
Should this be an internal API, at least for now?

cmarcelo:
-1

Name container is not very good because we already use it. And what you are creating is a static flow, static flows always contains nodes.

The helper itself is a bit awkward, it copy some but not all the members from the original, and then replaces it. A different venue could be changing the Python generator to take the specs and create the appropriate type.

Related: we probably should make static new type take a struct with all those specs. It's not scaling for all those members.

ibriano:
Me not likes. I'd rather play with a way to have the type be declared already using an fbp, since we have exported options now.

anselmolsm:
@cmarcelo The use of "container" was a wrong move, indeed.

The copy / replace thing was introduced when the hardcoded decriptions were removed from C code, then another iteration happened when the hardcoded options (GPIO pins) were removed. In fact, what I consider awkward is the current process to create such nodes - this function doesn't do anything different than that (this probably explains its awkwardness). But, I agree that this helper is not good enough. Better invest in the alternatives you and @ibriano mentioned.
Dropped.

PR #234, opened 2015-07-03
Platform detect

mbelluzzo:
* Add sol_ prefix to .json file
* Search in /etc
* Add json schema
* Update related help text


glima:
Looking OK

src/lib/common/Kconfig
```diff
@@ -306,7 +308,7 @@ menuconfig HAVE_PIN_MUX
           $PREFIX/lib/soletta/modules/pin-mux/<module>.so
 
           Pin mux modules are used based on the platform-detect logic
-          defined in $PREFIX/share/soletta/platform_detect.json or
+          defined in $PREFIX/share/soletta/sol_platform_detect.json or
```
barbieri: no need to use prefix if using a namespace directory.

src/lib/common/sol-platform-detect.c
```diff
@@ -215,9 +217,14 @@ sol_platform_detect(void)
     struct sol_json_token token, key, value, platform_name = { NULL };
     enum sol_json_loop_reason reason;
 
-    json_doc = _json_open_doc(DATADIR "/platform_detect.json", &scanner);
-    if (!json_doc)
-        return NULL;
+    json_doc = _json_open_doc("/etc/" PLATFORM_JSON, &scanner);
```
barbieri: in this case I believe it's better to use /etc/soletta/ and then a name, instead of putting multiple soletta files in the /etc itself. With linux-micro we use PKGSYSCONFDIR, you should use that instead of "/etc" and it should evaluate to "/etc/soletta" (currently does not, a bug I just noticed right now -- @dorileo  please take a look)

src/lib/common/sol-platform-detect.c
```diff
@@ -215,9 +217,14 @@ sol_platform_detect(void)
...
+    if (!json_doc) {
+        json_doc = _json_open_doc(DATADIR PLATFORM_JSON, &scanner);
+        if (!json_doc) {
+            SOL_INF(PLATFORM_JSON " could not be found.");
```
barbieri: please say which paths were looked.

mbelluzzo:
Please see v2 in #266

PR #235, opened 2015-07-06
[v1] worker-threads: change and version API

ulissesf:


cmarcelo:
I think for this amount of callbacks, it is indeed better to use the struct. Added some minor comments about the patch itself.

lpereira:
A struct will also help implement this trick:

    struct func_params {
        int x;
        in y;
    };

    void func(struct func_params params);
    #define func(params)    func((struct func_params) { .x = 42, params })

Which can be used like so:

    func();   /* x = 42, y = 0 */
    func(.x = -1, .y = 42); /* x = -1, y = 42 */
    func(.x = 0); /* x = 0, y = 0 */

ulissesf:
Check new PR #244.

PR #239, opened 2015-07-06
check: allow disabling colors

cmarcelo:
Adds --color and --no-color. If none is provided, then try to figure out
the right choice based on whether standard output is a tty or not. In
practice this automatic detection will handle most cases.

### DIFF
--- a/data/scripts/suite.py
+++ b/data/scripts/suite.py
@@ -34,6 +34,7 @@ from multiprocessing import Manager
 from threading import Thread
 import argparse
 import subprocess
+import sys
 
 WARN = '\033[93m'
 PASS = '\033[92m'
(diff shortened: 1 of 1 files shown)


cmarcelo:
@dorileo @glima

data/scripts/suite.py
```diff
@@ -139,9 +140,21 @@ def run_suite(args):
...
     parser.add_argument("--valgrind-supp", help="Path to valgrind's suppression file", type=str)
+    parser.add_argument("--color", dest="color", help="Disable colored output", action="store_true")
+    parser.add_argument("--no-color", dest="color", help="Force enable colored output", action="store_false")
+    parser.set_defaults(color=None)
 
     args = parser.parse_args()
     if args.valgrind_supp:
         args.valgrind_supp = "--suppressions=%s" % args.valgrind_supp
+    if args.color is None:
+        args.color = sys.stdout.isatty()
```
lpereira: The logging code (sol_log) also checks if `$TERM` is a terminal known to support color (`urxvt`, `xterm-color`, etc) in addition to calling `isatty()`. Not sure if overkill or required.

data/scripts/suite.py
```diff
@@ -139,9 +140,21 @@ def run_suite(args):
     parser.add_argument("--valgrind", help="Path to valgrind, if provided " \
                         "the tests are run with it", type=str)
     parser.add_argument("--valgrind-supp", help="Path to valgrind's suppression file", type=str)
+    parser.add_argument("--color", dest="color", help="Disable colored output", action="store_true")
+    parser.add_argument("--no-color", dest="color", help="Force enable colored output", action="store_false")
```
lpereira: Can't this be a `--color=true`, `--color=auto`, or `--color=false` instead of two options?

data/scripts/suite.py
cmarcelo: For my use cases (mostly keeping logs in the bots output) I don't need to check TERM. It's easy to add later inside this conditional.

data/scripts/suite.py
cmarcelo: OK.

cmarcelo:
I wanted a way to opt out from any specific "smartness". I'll push another version of this PR that also change `check-style.py` so both work the same way.

PR #240, opened 2015-07-06
fbp-generator: remove useless old file removal

ricardotk:
Rename will already replace the old file to the generated file,
there's no need to remove it before.


lpereira:
This removes a harmless TOTCOU, so :+1:.

PR #241, opened 2015-07-06
Expose C API to deal with network, coap and oic

bdilly:
v2 changelog:
  * split comms config per lib
  * handle issues on sol-coap packet_enqueue()


src/samples/coap/simple-server.c
```diff
@@ -210,6 +210,7 @@ light_method_get(const struct sol_coap_resource *resource, struct sol_coap_packe
 }
 
 static struct sol_coap_resource light = {
+    .api_version = SOL_COAP_RESOURCE_API_VERSION,
```
lpereira: While you're at it, please change `oicgen.py` so that the API version is set there as well.

src/lib/comms/sol-coap.c
```diff
@@ -1438,6 +1457,19 @@ sol_coap_server_new(int port)
 }
 
 SOL_API int
+sol_coap_packet_get_buf(struct sol_coap_packet *pkt, uint8_t **buf, uint16_t *len)
```
lpereira: This shouldn't be exposed. The only user of this getter can afford to not call this, as discussed in the previous PR.

src/samples/coap/simple-server.c
bdilly: Looks like oicgen.py don't fill this struct by itself, it uses sol_oic_device_definition_register_resource_type() that calls create_coap_resource().

I'm adding the version on this point, so we should be good about it.

Am I missing any direct usage on oicgen.py?

src/lib/comms/sol-coap.c
bdilly: OK, but for me those comments were pretty much not conclusive.
What approach should be used? Chiossi's suggestion?

bdilly: I've discussed with vcgomes and he'll modify it in such a way accessing buf won't be required.
So I'll wait for this solution and rebase to avoid exposing this function.
Thanks

src/samples/coap/simple-server.c
lpereira: Ah, yes, you're right. Slight brain fart moment here.

bdilly:
Merged adding a not about the exposed function that shouldn't be exposed.
@vcgomes is working on how to avoid this buf usage, but it doesn't seem to be trivial.

PR #242, opened 2015-07-06
fbp-generator: fix printf arg type mismatch on error message

ricardotk:
Pointed out by coverity.


lpereira:
:+1:

PR #243, opened 2015-07-06
Build random fixes 2 v2

dorileo:
v2:
   + removed the multi-thread implementation for dependency-resolver;
   + added new patches;


dorileo:
I'm fixing the issue pointed by semaphore but the review is still valid...

data/jsons/dependencies.json
```diff
@@ -1,4 +1,93 @@
...
+            "defines": [
+                "_GNU_SOURCE 1"
+            ],
+	    "headers": [
+		"<string.h>"
+            ],
+            "fragment": "strndupa(0,0);",
+            "cflags": {
+                "comment": "append to COMMON_CFLAGS so it's used for further checks",
+                "appent_to": "COMMON_CFLAGS",
```
lpereira: appent?

data/jsons/dependencies.json
```diff
@@ -1,4 +1,93 @@
...
+        "FLOWMODULESDIR": "{MODULESDIR}flow/",
+        "PINMUXDIR": "{MODULESDIR}pin-mux/",
+        "LINUXMICROMODULESDIR": "{MODULESDIR}linux-micro/"
+    },
+    "pre-dependencies": [
+	{
+	    "dependency": "gnu_source",
+	    "type": "ccode",
+            "defines": [
+                "_GNU_SOURCE 1"
```
lpereira: If this were an object:
    "defines": {
         "_GNU_SOURCE": "1"
    }

It would be possible to create the "cflags" object automatically below.

data/jsons/dependencies.json
```diff
@@ -1,4 +1,93 @@
...
+            ],
+	    "headers": [
+		"<string.h>"
+            ],
+            "fragment": "strndupa(0,0);",
+            "cflags": {
+                "comment": "append to COMMON_CFLAGS so it's used for further checks",
+                "appent_to": "COMMON_CFLAGS",
+                "value": "-D_GNU_SOURCE=1"
```
lpereira: Can't this be derived from "defines"?

data/jsons/dependencies.json
dorileo: ouch.. thanks.

data/scripts/dependency-resolver.py
```diff
@@ -144,40 +205,125 @@ def handle_python_check(args, conf):
...
+    for k,v in context.get_kconfig().items():
+        output += "config {config}\n{indent}{ktype}\n{indent}default {enabled}\n". \
+                  format(config=k, indent="       ", ktype=v["type"], enabled=v["value"])
     f = open(args.kconfig_output, "w+")
     f.write(output)
     f.close()
 
+def run(args, dep_checks, context):
+    for i in dep_checks:
+        handler = type_handlers.get(i["type"])
```
lpereira: This will raise a `KeyNotFound` exception instead of returning `None`. Pass a `None` as the second parameter to the `get()` method.

data/scripts/dependency-resolver.py
```diff
@@ -108,31 +141,59 @@ def compile_test(source, compiler, cflags, ldflags):
 
     return status
 
-def handle_ccode_check(args, conf):
+def set_makefile_compflags(flags, prefix, suffix):
+    append_to = flags.get("appent_to")
```
lpereira: `appent`?

data/scripts/dependency-resolver.py
```diff
@@ -108,31 +141,59 @@ def compile_test(source, compiler, cflags, ldflags):
...
+def handle_ccode_check(args, conf, context):
     dep = conf["dependency"].upper()
     source = ""
-    for i in conf["headers"]:
-        source += "#include %s\n" % i
 
+    cflags = conf.get("cflags")
+    ldflags = conf.get("ldflags")
+
+    defines = conf.get("defines")
```
lpereira: You can simplify this get + if with:

    for define in conf.get("defines", []):
        source += ...

data/scripts/dependency-resolver.py
```diff
@@ -144,40 +205,125 @@ def handle_python_check(args, conf):
...
+        wanted.append(i)
+
+    flags = " ".join(wanted)
+    success = compile_test(source, args.compiler, "-Werror %s" % flags, None)
+    if success:
+        context.add_append_makefile_var(append_to, flags)
+        return
+
+    supported = []
+    for i in check_cflags:
```
lpereira: Leave `i` for numeric indices. It's better to be explicit here: use `flag`, `cflag` or something along these lines instead.

data/scripts/dependency-resolver.py
```diff
@@ -144,40 +205,125 @@ def handle_python_check(args, conf):
...
+
+def handle_cflags_check(args, conf, context):
+    check_cflags = conf.get("cflags")
+    append_to = conf.get("append_to")
+    source = cstub.format(headers="", fragment="")
+
+    if not append_to or not check_cflags:
+        return
+
+    wanted = []
```
lpereira: If you'd like to copy `check_cflags`, do: `wanted = check_cflags[:]`.

data/scripts/dependency-resolver.py
```diff
@@ -144,40 +205,125 @@ def handle_python_check(args, conf):
...
+        handler(args, i, context)
+
+def vars_expand(origin, dest, maxrec):
+    remaining = {}
+
+    if not maxrec:
+        return
+
+    for k,v in origin.items():
+        if not isinstance(v, str):
```
lpereira: It's better to `isinstance(v, basestring)`, unless you really mean `str`.

data/scripts/template.py
```diff
@@ -0,0 +1,152 @@
...
+    for f in files:
+        lines = f.read()
+        content = "[context]\n%s" % lines
+        handle = configparser.ConfigParser(delimiters=('=','?=',':='))
+        handle.read_string(content)
+        dc = handle["context"]
+        result = dict(list(result.items()) + list(dc.items()))
+    return result
+
+def try_subst(verbatin):
```
lpereira: `verbatim`?

data/scripts/template.py
```diff
@@ -0,0 +1,152 @@
...
+    p = re.compile("^\@.*\@$")
```
lpereira: If you're already going to the trouble of using regexes, why not just compile one that matches `{{foo}}` instead of parsing the text by yourself?

dorileo: possibly...

data/scripts/dependency-resolver.py
```diff
@@ -144,40 +205,125 @@ def handle_python_check(args, conf):
...
+        else:
+            value = "%d" % v
+        context.add_cond_makefile_var(k, v)
+        header += "#define %s %s\n" % (k, value)
+
+    dirname = os.path.dirname(args.definitions_header)
+    if not os.path.exists(dirname):
+        os.makedirs(dirname)
+
+    f = open(args.definitions_header, "w")
```
lpereira: Minor, but you can  use the `with` syntax:

    with open(..., "w") as f:
        f.write(header)

data/scripts/dependency-resolver.py
```diff
@@ -144,40 +205,125 @@ def handle_python_check(args, conf):
...
     f = open(args.kconfig_output, "w+")
     f.write(output)
     f.close()
 
+def run(args, dep_checks, context):
+    for i in dep_checks:
+        handler = type_handlers.get(i["type"])
+        if not handler:
+            print("Could not handle type: %s" % i["type"])
+            continue
```
lpereira: Shouldn't this be a fatal error?

dorileo: depends on the POV, but I prefer to not simply parse it and still handle all the rest. Why do you think it should be a fatal error?

data/scripts/dependency-resolver.py
lpereira: Nevermind; there's no `basestring` in Python 3 anymore.

data/scripts/dependency-resolver.py
lpereira: It's a programmer error: it's better to abort early and break the build during development rather than just print something that can be overlooked and move on.

ibriano: Speaking of errors, don't just print and move on, throw a proper exception so that the build actually stops at the correct point, and showing useful information.

data/jsons/dependencies.json
```diff
@@ -1,4 +1,93 @@
...
+        "SOL_FLOW_DATADIR": "{SOL_DATADIR}flow/",
+        "DESCDIR": "{SOL_FLOW_DATADIR}descriptions/",
+        "SCHEMADIR": "{SOL_FLOW_DATADIR}schema/",
+        "PCDIR": "{LIBDIR}pkgconfig/",
+        "PKGNAME": "soletta",
+        "PKGSYSCONFDIR": "{SYSCONF}",
+        "FLOWMODULESDIR": "{MODULESDIR}flow/",
+        "PINMUXDIR": "{MODULESDIR}pin-mux/",
+        "LINUXMICROMODULESDIR": "{MODULESDIR}linux-micro/"
+    },
```
ibriano: Since we are doing this here... I've seen in several places that we end up with paths formed with // sometimes, and to me, "{SOME_VAR}more/path/" looks a lot more like a typo than "{SOME_VAR}/more/path".

data/jsons/dependencies.json
dorileo: don't think it's worth it, this defines are not propagated to the other tests, since this would be the only use case I think we should leave as it is until we have more uses for that.

data/scripts/dependency-resolver.py
dorileo: yep, probably a change of approach left over, thanks for catching.

data/jsons/dependencies.json
dorileo: This is the way it's since PREFIX may be provided with a trailing slash or not, well, we could be smarted on that but we don't need to change this behavior in this series. I would like to have this series in as soon as possible since it fixes some very important *core* issues. We may handle the stylistic issues later - this definitions are an as is translation to what we had before.

data/jsons/dependencies.json
```diff
@@ -1,4 +1,93 @@
...
+        "INCLUDEDIR": "{PREFIX}/include/",
+        "SOL_DATADIR": "{DATADIR}soletta/",
+        "SOL_INCLUDEDIR": "{INCLUDEDIR}soletta/",
+        "MODULESDIR": "{LIBDIR}soletta/modules/",
+        "SOL_FLOW_DATADIR": "{SOL_DATADIR}flow/",
+        "DESCDIR": "{SOL_FLOW_DATADIR}descriptions/",
+        "SCHEMADIR": "{SOL_FLOW_DATADIR}schema/",
+        "PCDIR": "{LIBDIR}pkgconfig/",
+        "PKGNAME": "soletta",
+        "PKGSYSCONFDIR": "{SYSCONF}",
```
barbieri: PKGSYSCONFDIR implies it's package-specific, thus it is missing the "/soletta". It should be: "{SYSCONF}/soletta"

dorileo:
I'm sending v3, I'm taking in account the last minute comment by @barbieri for a possible v4

PR #244, opened 2015-07-06
[v2] worker-threads: change and version API

ulissesf:
v2:
   - changed following review comments on v1
   - added patch to fix synchronization issue

### DIFF
--- a/src/modules/flow/file/file.c
+++ b/src/modules/flow/file/file.c
@@ -347,6 +347,17 @@ file_writer_worker_thread_iterate(void *data)
 static int
 file_writer_load(struct file_writer_data *mdata)
 {
+    struct sol_worker_thread_spec spec = {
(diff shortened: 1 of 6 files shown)


src/modules/flow/file/file.c
```diff
@@ -347,6 +347,17 @@ file_writer_worker_thread_iterate(void *data)
 static int
 file_writer_load(struct file_writer_data *mdata)
 {
+    struct sol_worker_thread_spec spec = {
+        .api_version = SOL_WORKER_THREAD_SPEC_API_VERSION,
+        .setup = file_writer_worker_thread_setup,
+        .cleanup = file_writer_worker_thread_cleanup,
+        .iterate = file_writer_worker_thread_iterate,
+        .cancel = NULL,
```
lpereira: Minor, but there's no need to specify `cancel` if designated initializers are being used.

ulissesf:
Check new PR #269.

PR #245, opened 2015-07-06
More array ports v3

lgywata:
Changes from V2:
- Fix subtraction operation since it won't work with array ports
- Rename subtraction ports to MINUEND and SUBTRAHEND so the order of operands become clear
- Fix samples diagrams with the new names for ports

### DIFF
--- a/src/modules/flow/float/float.json
+++ b/src/modules/flow/float/float.json
@@ -12,19 +12,11 @@
       "in_ports": [
         {
           "data_type": "float",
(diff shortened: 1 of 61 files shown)


cmarcelo:
I'm not very keen about having `OPERAND` instead of `IN`, but let's see how it goes.

+1 please squash the fix test to the appropriate patch, then feel free to push unless other comments appear in between.

lgywata:
Thanks. Applied.

PR #247, opened 2015-07-06
build: Fix buildopts platform detection

ceolin:
Fix platform linux detection


lpereira:
Rubberstamping: :+1:

lpereira:
BTW, isn't there anything similar to Perl's `use strict`?

ceolin:
Have no idea, for C's code we can pass the flag "-Werror=undef" and the build will fail if we do "#if bogus" and "bogus" is not defined.

ceolin:
merged, thanks!

PR #248, opened 2015-07-06
coap: Fix depending on internal structures from sol_coap_packet

vcgomes:
sol_coap_packet is only defined in coap.h, which is not exported, so we
should avoid depending on it from outside sol-coap.{c,h}.

The remaining usage was re-using the buf field of packet as an
preallocated memory, this proposed fix makes it write directly to the
response packet, avoiding that access.

### DIFF
--- a/src/lib/coap/sol-oic-server.c
+++ b/src/lib/coap/sol-oic-server.c
@@ -687,8 +687,11 @@ _sol_oic_resource_type_handle(
 {
     struct sol_coap_packet *response;
     sol_coap_responsecode_t code = SOL_COAP_RSPCODE_INTERNAL_ERROR;
-    uint8_t *payload;
-    uint16_t payload_len;
+    uint8_t *payload, *response_payload;
+    uint16_t payload_len, response_len;
+    uint8_t response_prefix[] = "{\"oc\":[{\"rep\":";
+    uint8_t response_suffix[] = "}]}";
+    int r;
 
     OIC_SERVER_CHECK(-ENOTCONN);
 
@@ -698,9 +701,21 @@ _sol_oic_resource_type_handle(
         return -1;
     }
 
-    if (!handle_fn) {
-        code = SOL_COAP_RSPCODE_NOT_IMPLEMENTED;
+    if (sol_coap_packet_get_payload(response, &response_payload, &response_len) < 0)
         goto done;
+
+    r = strlen(response_prefix);
+
+    if (response_len <= r)
+        goto done;
+
+    memcpy(response_payload, response_prefix, r);
+    response_payload += r;
+    response_len -= r;
+
+    if (!handle_fn) {
+	    code = SOL_COAP_RSPCODE_NOT_IMPLEMENTED;
+	    goto done;
     }
 
     if (expect_payload) {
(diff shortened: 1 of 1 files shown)


src/lib/coap/sol-oic-server.c
```diff
@@ -698,9 +701,21 @@ _sol_oic_resource_type_handle(
...
     }
 
-    if (!handle_fn) {
-        code = SOL_COAP_RSPCODE_NOT_IMPLEMENTED;
+    if (sol_coap_packet_get_payload(response, &response_payload, &response_len) < 0)
         goto done;
+
+    r = strlen(response_prefix);
+
+    if (response_len <= r)
```
lpereira: Should check for `response_suffix` len as well

src/lib/coap/sol-oic-server.c
```diff
@@ -687,8 +687,11 @@ _sol_oic_resource_type_handle(
 {
     struct sol_coap_packet *response;
     sol_coap_responsecode_t code = SOL_COAP_RSPCODE_INTERNAL_ERROR;
-    uint8_t *payload;
-    uint16_t payload_len;
+    uint8_t *payload, *response_payload;
+    uint16_t payload_len, response_len;
+    uint8_t response_prefix[] = "{\"oc\":[{\"rep\":";
+    uint8_t response_suffix[] = "}]}";
+    int r;
```
lpereira: Calling this `r` and assuming it's the payload length downwards is slightly dangerous, right? `r` is often shorthand for `result` and might be wrongly reused in a future revision.

src/lib/coap/sol-oic-server.c
```diff
@@ -698,9 +701,21 @@ _sol_oic_resource_type_handle(
         return -1;
     }
 
-    if (!handle_fn) {
-        code = SOL_COAP_RSPCODE_NOT_IMPLEMENTED;
+    if (sol_coap_packet_get_payload(response, &response_payload, &response_len) < 0)
         goto done;
+
+    r = strlen(response_prefix);
```
rchiossi: sizeof(response_prefix) - 1 ?

lpereira: Makes no difference here since these are (or should be) `static const`, so the compiler will optimize these out.

src/lib/coap/sol-oic-server.c
```diff
@@ -687,8 +687,11 @@ _sol_oic_resource_type_handle(
 {
     struct sol_coap_packet *response;
     sol_coap_responsecode_t code = SOL_COAP_RSPCODE_INTERNAL_ERROR;
-    uint8_t *payload;
-    uint16_t payload_len;
+    uint8_t *payload, *response_payload;
+    uint16_t payload_len, response_len;
+    uint8_t response_prefix[] = "{\"oc\":[{\"rep\":";
```
lpereira: `static const uint8_t response_prefix[] = ...`

barbieri:
ping?

lpereira:
[**@barbieri**](https://github.com/barbieri) commented on [Jul 29, 2015, 2:40 PM GMT-3](https://github.com/solettaproject/soletta/pull/248#issuecomment-126031473 "2015-07-29T17:40:38Z - Replied by Github Comment Enhancer"):
> ping?

We've had some discussions IRL and this doesn't seem to be the right approach. I don't recall what was the course of action, though.

vcgomes:
Thanks for the comments. Updated PR https://github.com/solettaproject/soletta/pull/639

PR #251, opened 2015-07-06
Fixes

ceolin:
- Separate which is linux specific on utils
 - Add a function to set a flag to a fille descriptor (checking the returns properly)
 - Fix the problem pointed by coverity (unchecked return from fcntl())


ceolin:
i'm fixing now the problem :(

ceolin:
closed https://github.com/solettaproject/soletta/pull/252

PR #252, opened 2015-07-06
Fixes v2

ceolin:
- v2
  - Fix the build (required defines should be in .h)

--------------------------------------------------------------------
 Separate which is linux specific on utils
 Add a function to set a flag to a fille descriptor (checking the returns properly)
  Fix the problem pointed by coverity (unchecked return from fcntl())


src/shared/sol-util-linux.c
```diff
@@ -0,0 +1,292 @@
...
+    if (getcwd(cwd, sizeof(cwd)) == NULL)
+        return -errno;
+
+#ifdef HAVE_SYS_AUXV_H
+    execfn = (char *)getauxval(AT_EXECFN);
+
+    if (!execfn)
+        return -errno;
+#else
+    if (readlink("/proc/self/exe", execfn, sizeof(execfn) - 1) < 0 || !execfn[0])
```
lpereira: `readlink()` does not zero-terminates the string. Do not throw away its return value: it's the number of characters that were copied to `execfn`, or `-1` in case of error.

src/shared/sol-util-linux.c
```diff
@@ -0,0 +1,292 @@
...
+#endif
+
+    if (getcwd(cwd, sizeof(cwd)) == NULL)
+        return -errno;
+
+#ifdef HAVE_SYS_AUXV_H
+    execfn = (char *)getauxval(AT_EXECFN);
+
+    if (!execfn)
+        return -errno;
```
lpereira: I'd just call the `readlink()` path if `getauxval()` fails.

src/shared/sol-util-linux.c
```diff
@@ -0,0 +1,292 @@
...
+    if (r < 0 || r >= (int)sizeof(progname))
+        return r;
+
+    substr = strstr(progname, PREFIX);
+    if (!substr) {
+        return -1;
+    }
+
+    prefix = strndupa(progname, strlen(progname) - strlen(substr));
+    if (!prefix)
```
lpereira: It's undefined what happens when `alloca()` fails, so please don't check. Use `%.*s` in the snprintf() below.

ceolin:
Leandro, this commit is just to split what is linux dependent. I've not changed the code.

I do not see any problem to fix those problems but can we address them on a different commit ?

lpereira:
No problems.

:+1: for the build fix.

zehortigoza:
+2

ceolin:

PR #253, opened 2015-07-06
Use SOL_STR_SLICE_PRINT()

ricardotk:
The macro is there to simplify the operation.


lpereira:
Seems sane. Are these all the places that this macro could be used?

ricardotk:
I found more entries on oircgen.py, thanks!

PR #257, opened 2015-07-07
Some minor stuff for static flow

cmarcelo:
Recovered from my older repo and updated. The struct in particular will be helpful in the near future.


cmarcelo:
@ibriano @ricardotk

ibriano:
Looks good, go for it.

cmarcelo:
Applied with the bool -> void change, and adding missing `;` to the output.

PR #258, opened 2015-07-07
Use name from packet type pointer

lgywata:
Instead of hard coding it.

### DIFF
--- a/src/bin/sol-fbp-runner/runner.c
+++ b/src/bin/sol-fbp-runner/runner.c
@@ -146,27 +146,46 @@ add_simulation_node(struct runner *r, const char *node_type, const char *port_na
 }
 
 struct map {
-    const char *packet_name;
+    const struct sol_flow_packet_type *packet;
     const char *node_type;
     const char *port_name;
 };
 
-static const struct map input_nodes[] = {
-    { "IRange", "gtk/spinbutton", "OUT" },
-    { "DRange", "gtk/slider", "OUT" },
-    { "Any", "gtk/toggle", "OUT" },
-    { "Empty", "gtk/toggle", "OUT" },
-    { "Boolean", "gtk/pushbutton", "OUT" },
-    { "RGB", "gtk/rgb-editor", "OUT" },
-    { "Byte", "byte/editor", "OUT" },
+static struct map input_nodes[] = {
+    { NULL, "gtk/spinbutton", "OUT" },
+    { NULL, "gtk/slider", "OUT" },
+    { NULL, "gtk/toggle", "OUT" },
+    { NULL, "gtk/toggle", "OUT" },
+    { NULL, "gtk/pushbutton", "OUT" },
+    { NULL, "gtk/rgb-editor", "OUT" },
+    { NULL, "byte/editor", "OUT" },
 };
 
-static const struct map output_nodes[] = {
-    { "IRange", "gtk/label", "IN" },
-    { "DRange", "gtk/label", "IN" },
-    { "String", "gtk/label", "IN" },
-    { "Boolean", "gtk/led", "IN" },
+static struct map output_nodes[] = {
+    { NULL, "gtk/label", "IN" },
+    { NULL, "gtk/label", "IN" },
+    { NULL, "gtk/label", "IN" },
+    { NULL, "gtk/led", "IN" },
 };
+
+static void
+init_tables() {
+    struct map *in = input_nodes;
+    struct map *on = output_nodes;
+
+    in[0].packet = SOL_FLOW_PACKET_TYPE_IRANGE;
+    in[1].packet = SOL_FLOW_PACKET_TYPE_DRANGE;
+    in[2].packet = SOL_FLOW_PACKET_TYPE_ANY;
+    in[3].packet = SOL_FLOW_PACKET_TYPE_EMPTY;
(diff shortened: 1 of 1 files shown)


src/bin/sol-fbp-runner/runner.c
```diff
@@ -146,28 +146,47 @@ add_simulation_node(struct runner *r, const char *node_type, const char *port_na
...
-    { "Boolean", "gtk/led", "IN" },
+static struct map output_nodes[] = {
+    { NULL, "gtk/label", "IN" },
+    { NULL, "gtk/label", "IN" },
+    { NULL, "gtk/label", "IN" },
+    { NULL, "gtk/led", "IN" },
 };
 
+static void
+init_tables() {
```
lpereira: Functions without parameters in C should have a `"void"` argument: `init_tables(void)`.

src/bin/sol-fbp-runner/runner.c
```diff
@@ -146,28 +146,47 @@ add_simulation_node(struct runner *r, const char *node_type, const char *port_na
 }
 
 struct map {
-    const char *packet_name;
+    const struct sol_flow_packet_type *packet;
```
lpereira: Maybe call this `packet_type`?

src/bin/sol-fbp-runner/runner.c
```diff
@@ -146,28 +146,47 @@ add_simulation_node(struct runner *r, const char *node_type, const char *port_na
...
+    { NULL, "gtk/label", "IN" },
+    { NULL, "gtk/led", "IN" },
 };
 
+static void
+init_tables() {
+    struct map *in = input_nodes;
+    struct map *on = output_nodes;
+
+    in[0].packet = SOL_FLOW_PACKET_TYPE_IRANGE;
```
lpereira: Maybe check if `in[0].packet` is NULL and only proceed if that happens?

src/bin/sol-fbp-runner/runner.c
```diff
@@ -146,28 +146,47 @@ add_simulation_node(struct runner *r, const char *node_type, const char *port_na
...
 
-static const struct map input_nodes[] = {
-    { "IRange", "gtk/spinbutton", "OUT" },
-    { "DRange", "gtk/slider", "OUT" },
-    { "Any", "gtk/toggle", "OUT" },
-    { "Empty", "gtk/toggle", "OUT" },
-    { "Boolean", "gtk/pushbutton", "OUT" },
-    { "RGB", "gtk/rgb-editor", "OUT" },
-    { "Byte", "byte/editor", "OUT" },
+static struct map input_nodes[] = {
```
lpereira: This can be left as `const`.

src/bin/sol-fbp-runner/runner.c
```diff
@@ -146,28 +146,47 @@ add_simulation_node(struct runner *r, const char *node_type, const char *port_na
...
+    in[3].packet = SOL_FLOW_PACKET_TYPE_EMPTY;
+    in[4].packet = SOL_FLOW_PACKET_TYPE_BOOLEAN;
+    in[5].packet = SOL_FLOW_PACKET_TYPE_RGB;
+    in[6].packet = SOL_FLOW_PACKET_TYPE_BYTE;
+
+    on[0].packet = SOL_FLOW_PACKET_TYPE_IRANGE;
+    on[1].packet = SOL_FLOW_PACKET_TYPE_DRANGE;
+    on[2].packet = SOL_FLOW_PACKET_TYPE_STRING;
+    on[3].packet = SOL_FLOW_PACKET_TYPE_BOOLEAN;
+}
```
lpereira: I've been thinking: maybe if you only reserve the space for `input_nodes[]` and `output_nodes[]`, and initialize it here (inside this function) like so:

    struct map output_nodes[4] = {0};
    void init_tables(void) {
        struct map on[] = {
            { SOL_FLOW_PACKET_TYPE_IRANGE, "gtk/label", "IN" },
            ...
        };
        assert(!output_nodes[0].packet_type);
        static_assert(ARRAY_SIZE(on) == ARRAY_SIZE(output_nodes));
        memcpy(output_nodes, on, sizeof(on));
    }

Note that, since `on[]` isn't `static` and/or `const`, it should be possible to initialize it this way.

src/bin/sol-fbp-runner/runner.c
lgywata: Ack.

src/bin/sol-fbp-runner/runner.c
lgywata: Ack.

src/bin/sol-fbp-runner/runner.c
lgywata: Ack.

src/bin/sol-fbp-runner/runner.c
lgywata: Nice. Thanks.

cmarcelo:
Replace "Instead of hard coding it." in the commit message with "Instead of relying on the packet type name, that is not necessarily unique."

cmarcelo:
By the way, it is still hardcoded, we are just using a better reference to the packet type.

src/bin/sol-fbp-runner/runner.c
cmarcelo: I prefer the original -- just replace `on` with `out`.

PR #259, opened 2015-07-07
build: Fix the build of external modules

ceolin:
It was depending of a header that is not installed.


otaviobp:
Working fine for me

PR #260, opened 2015-07-07
fbp-generator: enable the generation of subflows

ricardotk:
-s option now can be used to generate a subflow (without includes
and main)


ricardotk:
Previously set as +1 from @glima and @ceolin. I just rebased before merge and the comments were vanished from here. merging it...

PR #270, opened 2015-07-07
fbp-generator: start handling directories on JSON arguments

ricardotk:
now it's possible to pass a directory path to -j and it will search
recursively into the directories for '*.json' files.

### DIFF
--- a/src/bin/sol-fbp-generator/main.c
+++ b/src/bin/sol-fbp-generator/main.c
@@ -31,10 +31,12 @@
  */
 
 #include <ctype.h>
+#include <dirent.h>
 #include <errno.h>
 #include <fcntl.h>
 #include <libgen.h>
 #include <stdio.h>
+#include <sys/stat.h>
 #include <unistd.h>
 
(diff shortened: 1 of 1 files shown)


src/bin/sol-fbp-generator/main.c
```diff
@@ -680,6 +682,71 @@ sol_fbp_generator_type_store_load(struct type_store *common_store)
...
+        if (entry->d_name[0] == '.')
+            continue;
+
+        r = snprintf(full_path, sizeof(full_path), "%s/%s", path, entry->d_name);
+        if (r < 0 || r >= (int)sizeof(full_path)) {
+            closedir(dp);
+            return false;
+        }
+
+        handle_json_path(full_path);
```
lpereira: Shouldn't the return value of this function be checked here as well?

ricardotk: Yes it should. Nice catch

lpereira:
Feel free to merge after adding that check.

PR #271, opened 2015-07-07
Packet type descriptions

barbieri:
Declare packet payload members in the type so we can use it in a generic way, done in 'console' node type and also inspector provided in the fbp runner. The JS (duktape) should also benefit from it.

### DIFF
--- a/src/bin/sol-fbp-runner/inspector.c
+++ b/src/bin/sol-fbp-runner/inspector.c
@@ -124,80 +124,111 @@ static void
 inspector_show_packet(const struct sol_flow_packet *packet)
 {
     const struct sol_flow_packet_type *type = sol_flow_packet_get_type(packet);
+    const uint8_t *mem;
+    const struct sol_flow_packet_member_description *desc;
+    bool has_members, single_member;
 
-    if (type == SOL_FLOW_PACKET_TYPE_EMPTY) {
-        fputs("<empty>", stdout);
-        return;
-    } else if (type == SOL_FLOW_PACKET_TYPE_ANY) {
-        fputs("<any>", stdout);
-        return;
-    } else if (type == SOL_FLOW_PACKET_TYPE_ERROR) {
-        int code;
-        const char *msg;
-        if (sol_flow_packet_get_error(packet, &code, &msg) == 0) {
-            fprintf(stdout, "<error:%d \"%s\">", code, msg ? msg : "");
-            return;
-        }
-    } else if (type == SOL_FLOW_PACKET_TYPE_BOOLEAN) {
-        bool v;
-        if (sol_flow_packet_get_boolean(packet, &v) == 0) {
-            fprintf(stdout, "<%s>", v ? "true" : "false");
-            return;
-        }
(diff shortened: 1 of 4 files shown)


cmarcelo:
I don't think we should go this direction now.

First, with the amount of packets we have today, you have to treat
more "member types" than packet types. This might be a win in the
future depending on the nature of packets that we bet are going to be
created. I would say, let's see how packets are used first.

Second, as a form of debugging/inspecting I rather have a to_string()
kind of callback for the packet type. I know it have limitations, but
the packet can give me better results. E.g. irange packet can make its
own to_string() type to omit the boundaries if they are the max/min --
or writing them with the symbolic names.

Third, for custom packets, that contains a handle (either a pointer or
a blob reference), the member type for that will be almost unhelpful
-- at least in the current form.

My suggestion is: add to_string() to the packet type and this will
cover both console and inspector. Do binding for packet types. Packets
that are not recognized by the binding should be treated as opaque
things.

cmarcelo:
Regardless *how* we get the "string representation of a packet for debugging purposes" (via reflection, via callback in the packet type), it's worthwhile to make a function for that, and use it in both console and inspector.

barbieri:
to_string() is basically not helpful -- useless, I thought about it and it adds nothing other than a simple debug. Not worth (even tho my code only change the code that uses it for debug).

Others, such as @bdilly and @ceolin already expressed desire to have code that deals automatically with different types. In that situation to_string() doesn't help, while the description I propose do let one do that. Say you want to receive packet for irange, bool or drange, get the main field (i.e: "val" or we flag a main field) and do something with it, you can easily do it.

@ibriano says that the main types in a flow won't be the basic we provide, he was proposing a complex "key=value" object which is cumbersome in C... but with the given proposal that could be implemented in an efficient way if the user declares such type in JSON and we generate all the structs, setters and getters for him -- using a dynamic complex "key=value" as a last-resort/fallback.

For packets that do contain a handle or deeper structures, it is possible to map using the current descriptions to some extent. I did think about it but either we make the data_type more complex or we go as a route I describe in an email to soletta-dev "RFC: memdesc".

cmarcelo:
to_string() is helpful for the case pointed (debugging / introspection). It also would provide an easy way to print packets as part of debug. And you are right, it is not a the tool for helping out generic node types, but that's not what I suggested.

I'm not familiar with the plans for generic nodes -- maybe that's the missing context, could you elaborate? Flagging a main field could be equivalent to having a getter for the "main value" as another callback, so not clear what the gains are here. Letting the node specify what value to peek could be done in a similar way.

barbieri:
the idea about the generic node is nothing I personally like as we don't allow a list of packet-types per port, we'd have to use "any" which opens for other problems. But as I said it would work with many packet types without really knowing about the packet-type, given that it provide some known fields. Think about Python's duck-typing: I don't care about the type if it provides me the information I need.

cmarcelo:
You can implement duck typing the other way around: add a callback to packet type that can accept a general identifier and returns value for the identifiers that it supports.

barbieri:
this is being worked one using the new memdesc infrastructure.

PR #274, opened 2015-07-07
flow: provide a generic function for deleting node types

cmarcelo:
Some node *types* are created in runtime, and need to be later deleted
and have extra resources disposed. Types created with
sol_flow_static_new_type() are one example, others might be added later.

This patch provide a generic function for deleting those types, so code
that contains collections of node types doesn't need to track which
specific delete function to call. We'll use that as part of
sol-flow-resolver, for the declared types, that might have different
sources.

There's a callback function to handle the per type behavior in the type
structure. Uses of sol_flow_static_del_type() were replace with
sol_flow_node_type_del().


cmarcelo:
Applied in a6e5340d96d2815aef7c9644ad3268823e835276.

PR #276, opened 2015-07-07
Json config try2

tcanabrava:
Second try of the json based config - make check and make check-fbp passes, also a small testcase with a json config written just for it also passed. This version has the following differences regarding v1:
- use sol_util_load_file_string instead of loading the file manually
- use sol_util_str_split instead of splitting the file manually
- fixes a bug on util_load_file_raw where a we could fail after reading the file sucessfully
- use sol_ptr_vector to store the options temporarely


lpereira:
Please close the previous PR when sending a new version.

mbelluzzo:
-1, the way you fixed the "bug" in _raw makes it fall in the 'do' loop in case of error, which is even worst than the original code.

The better chance to avoid regular files to fall in the worst case scenario of this function would be to add the 'do' in the else of stat diff, remove the 'goto end' completely.

Also, this change should not be mixed with json conf changes since it's a improvement that is not even needed to make it work.

tcanabrava:
@mbelluzzo  - will do.

data/schemas/config.schema
```diff
@@ -0,0 +1,28 @@
...
+        },
+
+        "nodetypes" : {
+            "type": "array",
+            "minItems": 1,
+            "uniqueItems": true,
+            "items": { "$ref": "#/definitions/nodetype" }
+        }
+    }
+}
```
mbelluzzo: I think we are using only 2 spaces for indenting .json files (I know, its ugly as hell, but its the way it is).

data/scripts/convert-config-to-json.py
```diff
@@ -0,0 +1,81 @@
+#!/usr/bin/env python3
```
lpereira: If there's no reason to have this script in the tree, as stated in the previous version of this PR, just leave it out.

src/shared/sol-conffile.c
```diff
@@ -58,147 +61,187 @@ sol_conffile_entry_sort_cb(const void *data1, const void *data2)
...
+
+    sol_json_scanner_init_from_token(&scanner, &options_object);
+    SOL_JSON_SCANNER_OBJECT_LOOP (&scanner, &token, &key, &value, reason) {
+        int key_len, value_len;
+        char *tmp;
+        key_len = sol_json_token_get_size(&key) - 2; // remove quotes from the key.
+        value_len = sol_json_token_get_size(&value); // do not remove quotes from the value. (not everything is a string here)
+        len = key_len + value_len + 2; // plus = and \0
+        tmp = malloc( len );
+        snprintf(tmp, len, "%.*s=%.*s\0", key_len, key.start + 1, value_len, value.start);
```
lpereira: Use `asprintf()` instead of `malloc()` + `snprintf()`. If using `snprintf()`, **always** check the return value for truncation.

src/shared/sol-conffile.c
```diff
@@ -58,147 +61,187 @@ sol_conffile_entry_sort_cb(const void *data1, const void *data2)
...
+    sol_ptr_vector_init(&vec_options);
+
+    sol_json_scanner_init_from_token(&scanner, &options_object);
+    SOL_JSON_SCANNER_OBJECT_LOOP (&scanner, &token, &key, &value, reason) {
+        int key_len, value_len;
+        char *tmp;
+        key_len = sol_json_token_get_size(&key) - 2; // remove quotes from the key.
+        value_len = sol_json_token_get_size(&value); // do not remove quotes from the value. (not everything is a string here)
+        len = key_len + value_len + 2; // plus = and \0
+        tmp = malloc( len );
```
lpereira: Memory allocation failures should be checked.

___Also, please:___
   * Run check-style.
   * Read other parts of the code to see how things are done in the project.
   * If you're not familiar with C library functions, read the man pages. Specifically the NOTES section and RETURN VALUES section. Check all return values.

src/shared/sol-conffile.c
```diff
@@ -58,147 +61,187 @@ sol_conffile_entry_sort_cb(const void *data1, const void *data2)
...
-    g_error_free(error);
-    g_key_file_free(keyfile);
-    return NULL;
+    if (file[0] == '/') {
+        config_file_contents = sol_util_load_file_string(file, NULL);
+    } else {
+        for (i = 0; i < search_dirs_size; i++) {
+            char filename[PATH_MAX];
+            size_t size;
+            sprintf(filename, "%s/%s", search_dirs[i], file);
```
lpereira: ...

src/shared/sol-conffile.c
```diff
@@ -209,9 +252,9 @@ sol_conffile_free_entry_vector(struct sol_ptr_vector *pv)
 
     SOL_PTR_VECTOR_FOREACH_IDX (pv, e, i) {
         free((void *)e->id);
-        g_free((void *)e->type);
+        free((void *)e->type);
         if (e->options)
-            g_strfreev((gchar **)e->options);
+            free((char **)e->options);
```
lpereira: ...

src/shared/sol-conffile.c
```diff
@@ -307,7 +347,7 @@ sol_conffile_load_vector(void)
 static void
 sol_conffile_clear_data(void)
 {
-    sol_conffile_free_entry_vector(&sol_conffile_entry_vector);
+    //sol_conffile_free_entry_vector(&sol_conffile_entry_vector);
```
lpereira: Commented code?

src/shared/sol-conffile.c
```diff
@@ -58,147 +61,187 @@ sol_conffile_entry_sort_cb(const void *data1, const void *data2)
...
-    if (fallback_paths && !keyfile) {
-        splitted_paths = g_strsplit(fallback_paths, ";", -1);
-        for (ptr = splitted_paths; *ptr; ptr++) {
-            keyfile = sol_conffile_load_keyfile_from_dirs
-                    (*ptr, full_path);
-            if (keyfile)
+        config_file_contents = sol_conffile_load_keyfile_from_dirs(path, full_path);
+
+    if (fallback_paths && !config_file_contents) {
+        int idx;
```
lpereira: Indices should be `uint16_t`.

src/shared/sol-conffile.c
```diff
@@ -58,147 +61,187 @@ sol_conffile_entry_sort_cb(const void *data1, const void *data2)
...
-    g_key_file_free(keyfile);
-    return NULL;
+    if (file[0] == '/') {
+        config_file_contents = sol_util_load_file_string(file, NULL);
+    } else {
+        for (i = 0; i < search_dirs_size; i++) {
+            char filename[PATH_MAX];
+            size_t size;
+            sprintf(filename, "%s/%s", search_dirs[i], file);
+            config_file_contents = sol_util_load_file_string(file, &size);
```
mbelluzzo: Since you are dealing with regular files for sure, Instead of use `sol_util_load_file_string` I think you should use `sol_file_reader_open` and sol_file_reader API to deal with files that you will parse. `sol_file_reader` API will take advantage of mmap whenever its possible (i.e. when dealing with regular files which is the case) and fallback to `sol_util_load_file_raw` only if mmap isn't possible. Than just use sol_file_reader_get_all and you will have proper slice to feed `sol_json_scanner_init` (slice.data). You can check `_json_open_doc` in 'sol-platform-detect.c' if needed.

Since our parser use silces everywhere, you dont need to `sol_util_load_file_string` anyway since you dont need to enforce the contents to be null terminated (only reason to use _string instead of _raw).

src/shared/sol-conffile.c
```diff
@@ -58,147 +61,187 @@ sol_conffile_entry_sort_cb(const void *data1, const void *data2)
...
+        return pv;
+
+    sol_json_scanner_init_from_token(&obj_scanner, &value);
+    SOL_JSON_SCANNER_ARRAY_LOOP (&obj_scanner, &token, SOL_JSON_TYPE_OBJECT_START, reason) {
         entry = calloc(1, sizeof(*entry));
         SOL_NULL_CHECK(entry, pv);
+        SOL_JSON_SCANNER_OBJECT_LOOP_NEST (&obj_scanner, &token, &key, &value, reason) {
 
-        entry->id = strdup(*groups_ptr);
+    #define DUP_JSON_STR( JSON_VALUE ) \
```
lpereira: Please use a function here instead of a macro.

src/shared/sol-conffile.c
```diff
@@ -58,147 +61,187 @@ sol_conffile_entry_sort_cb(const void *data1, const void *data2)
...
+    sol_json_token_get_size(&JSON_VALUE) >= 2 ? \
+    strndup(JSON_VALUE.start + 1, sol_json_token_get_size(&JSON_VALUE) - 2 ) : NULL
 
-        entry->type = g_key_file_get_string(keyfile,
-            *groups_ptr, "Type", NULL);
+            if (sol_json_token_str_eq(&key, node_name, strlen(node_name))) {
+                entry->id = DUP_JSON_STR(value);
+            }
```
lpereira: `else`

src/shared/sol-conffile.c
```diff
@@ -58,147 +61,187 @@ sol_conffile_entry_sort_cb(const void *data1, const void *data2)
...
-        if (!entry->type) {
-            SOL_DBG("could not find mandatory 'Type' key on group [%s],"
-                " skipping entry", *groups_ptr);
-            free((void *)entry->id);
-            free(entry);
-            continue;
-        }
+            if (sol_json_token_str_eq(&key, node_type, strlen(node_type))) {
+                entry->type = DUP_JSON_STR(value);
+            }
```
lpereira: `else`

ricardotk:
s/try/v on PR message please

src/shared/sol-conffile.c
```diff
@@ -58,147 +61,187 @@ sol_conffile_entry_sort_cb(const void *data1, const void *data2)
...
 
-        entry->options = (const char **)g_key_file_get_string_list(keyfile,
-            *groups_ptr, "Options", NULL, NULL);
+            if (sol_json_token_str_eq(&key, node_options, strlen(node_options))) {
+                sol_conffile_set_entry_options(entry, value);
+            }
```
lpereira: `else { error! }`

src/shared/sol-conffile.c
```diff
@@ -58,147 +61,187 @@ sol_conffile_entry_sort_cb(const void *data1, const void *data2)
...
-        include_group, "Include", &error);
-    if (*include == NULL && error
-        && error->code != G_KEY_FILE_ERROR_KEY_NOT_FOUND) {
-        SOL_DBG("could not read value from %s group: %s",
-            include_group, error->message);
-        g_error_free(error);
-    }
+    struct sol_json_token token, key, value;
+    struct sol_json_scanner include_scanner;
+    enum sol_json_loop_reason reason;
```
lpereira: This variable is never checked. It should *always* be checked.

lpereira:
Please review the comments in this PR and in the previous versions, as there are still a few things that were not fully addressed.

data/schemas/config.schema
tcanabrava: easy to fix, done.

data/scripts/convert-config-to-json.py
tcanabrava: Okay

src/shared/sol-conffile.c
tcanabrava: I was looking for a function that did the 'malloc + snprintf' trick, didn't found it, THANKS <3

src/shared/sol-conffile.c
tcanabrava: check style was run and didn't found *any* issues on that

tcanabrava:
closing for v3

PR #278, opened 2015-07-08
Build random fixes 2 v8

dorileo:
v3:
+ upstream rebased;
+ fixed issue with in sequence reconfiguration issue - pointed by semaphore;
+ fixed appent_to typo - pointed by @lpereira;
+ simplifiled conf.get("defines") + if with conf.get("defines", []) - pointed by @lpereira;
+ changed the use of var i for non index usaged - pointed by @lpereira;
+ removed the wanted var/use from dep-resolver - pointed by @lpereira;
+ s/verbatin/verbatim - pointed by @lpereira;
+ fix issue with PREFIX changes;
+ removed patch "avoid log domain..." - pointed by @glima;

v4:
+ fixed test and bin builds;
+ fixed issue with !HAVE_DECL_IFLA_INET6_ADDR_GEN_MODE check;

v5
+ resorted the bin, test build fixes and the DECL_IFLA otherwise semaphore was pointing an intermediate state breakage;

v6:
+ whole bunch of new stuffs I coudn't keep track of them all;
+ fixed issue pointed by @barbieri on PKGCONFIGDIR;
+ using cflags on ccode on tests (similarly ti what we do with ldflags, the tested is the same used in the output) instead of using defines [] on _GNU_SOURCE test - pointed by @lpereira ;
+ added patch to properly install -gdb.py file - pointed by @barbieri ;

v7:
+ path normalization on dep resolver defines - suggested by @ibriano ;

v8:
+ upstream rebase;


vcgomes:
> But if you feel unhappy with that I can drop it until we have an actual user.

If you are sure that it will be used, leave it.

vcgomes:
:+1:

It has some stuff that helps when building with musl, and just tested it.

glima:
Well, too many lines to read here, but looking good overall.

dorileo:
ok guys, here we go. applied.

PR #279, opened 2015-07-08
tests: fix warnings on two fbp tests

bdilly:
Division by 0 was expected on these tests.

### DIFF
--- a/src/test-fbp/converter-error-expected.txt
+++ b/src/test-fbp/converter-error-expected.txt
@@ -1,0 +1,1 @@
+WRN: ./src/lib/common/sol-types.c:242 sol_irange_division() Division by zero: 10, 0
--- a/src/test-fbp/hub-expected.txt
+++ b/src/test-fbp/hub-expected.txt
@@ -1,0 +1,1 @@
+WRN: ./src/lib/common/sol-types.c:242 sol_irange_division() Division by zero: 10, 0


lpereira:
Rubberstamping. :+1:

dorileo:
+1

src/test-fbp/converter-error-expected.txt
```diff
@@ -0,0 +1 @@
+WRN: ./src/lib/common/sol-types.c:242 sol_irange_division() Division by zero: 10, 0
```
cmarcelo: The problem is that this warning depends on the line number of the code it is in. If someone adds something to sol-types.c completely unrelated, this will change.

Maybe what's missing here is that we need to tweak the log level that those tests work (in `src/test-fbp/run`), since this is not what we expect to production users to see unless debug is enable (and we don't want to test for output of debug messages).

What do you think?

lpereira: Isn't this the case of sending a packet to the error port so it can be captured by a test node?

cmarcelo: The node does send an error packet and the test wait for it.

This warning is in our division function `sol_irange_division`.

lpereira: Tweaking the debug output will work for this case, but maybe that's not a one-size-fits-all solution. Instead of having a verbatim `-expected.txt` file, maybe one with a regex or some other globbing pattern would work?

    WRN: .*sol-types.c.* Division by zero: 10,0

cmarcelo: That's a good improvement for expected file.

But in this case, I don't think we intend to spill that message for the regular user -- unless he asks for it, even less ensure that this is the exact message. The tests here that check the output should mostly check the output we give for regular (non-debugging) uses.

@bdilly what is going to be the default log level for "release builds"? I suggest we change the `src/test-fbp/run` to set this log level.

bdilly: @cmarcelo I'm not sure about that. We may miss some unexpected warnings this way. In this case I really wanted to force an error to create an error packet so nodes that handle error packets could be tested. But usually, if a warning shows up while we're running these tests, it means something gone wrong. It's a test suite after all. We're using it to test node types code and cover our codebase. I'm more compelled to work with a solution like proposed by @lpereira to make sure we're not masking any problems.

cmarcelo:
@bdilly @lpereira

The problem I see is that we are too eager to give warnings to the client code, so when the user actually handle the error, there's still a visible warning around.

In this case, our code already look at the error code from the function, treat it (send a packet of error), this packet is handled by the flow. If it wasn't handled by the flow, we would indeed warning.

So, we taken care, but the warning of the lowlevel function is still there. I remember from the discussion that those warnings inside low level functions are meant to help developer identify issues and they would not be the default for release builds.

I dislike this level of "tracking" of the warnings (adding this internal function warning to expected file), because a completely honest change to the code, like verify parameters before calling the helper function or even using another helper function that doesn't warn, would led to have to deal with the test, even though everything is working correctly (the error packet is being sent, and treated). In the long run this leads to making things very difficult to change, or to wildcards in the expected files that would end up eating output that was really unexpected.

In other words, proper behavior of Soletta doesn't require this warning to appear -- compare to the parser_errors expected files, proper behavior of Soletta require that output.


> But usually, if a warning shows up while we're running these tests, it means something gone wrong.

Yes. Maybe we are mixing debug messages with warnings intended to the user then?

bdilly:
@cmarcelo I agree these warnings should be disabled on releases. So this warning won't popup on client's users face. But test-fbp has nothing to do with releases / users. It's our test suite. It's a thing to be used by developers / QA. All warnings are welcome at this point, imo.

cmarcelo:
`src/test-fbp` tests the behavior we promise to deliver when running FBP. We care about delivering correct line / column information in parser errors, so we test for that. I don't think this particular warning should appear when running that particular FBP. Division by zero already becomes an error packet, and if that error packet isn't handled, we then give a warning.

If `sol_irange_division` insists in warning (i.e. it's telling us we are feeding it with "bad" data), we probably shouldn't be feeding it with divisions by zero then.

If you are going to the route of flexible expected lines, this PR might be helpful https://github.com/solettaproject/soletta/pull/289.

bdilly:
v2 submitted

PR #266, opened 2015-07-07
Platform detect v2

mbelluzzo:
* Applied proposed changes
* Add patch to check platform name that should fix Coverity Issue: 122264

PR #280, opened 2015-07-08
platform-detect: Sanitize platform name

mbelluzzo:
Version "regexless". This should fix coverity bug 122264.

Related to #266

@anselmolsm @barbieri

### DIFF
--- a/src/lib/common/sol-platform-detect.c
+++ b/src/lib/common/sol-platform-detect.c
@@ -266,3 +266,24 @@ end:
     sol_file_reader_close(json_doc);
     return platform;
 }
+
+bool
+sol_platform_invalid_name(const char *name)
+{
+    unsigned int i;
+
+    if (name[0] == '\0')
+        return true;
+
+    for (i = 0; name[i] != '\0'; i++) {
+        // https://xkcd.com/276/
+        if ((name[i] >= 'A' && name[i] <= 'Z') ||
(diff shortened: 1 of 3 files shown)


src/lib/common/sol-platform-detect.c
```diff
@@ -266,3 +266,24 @@ sol_platform_detect(void)
...
+sol_platform_invalid_name(const char *name)
+{
+    unsigned int i;
+
+    if (name[0] == '\0')
+        return true;
+
+    for (i = 0; name[i] != '\0'; i++) {
+        // https://xkcd.com/276/
+        if ((name[i] >= 'A' && name[i] <= 'Z') ||
```
lpereira: `isalpha(name[i])`?

mbelluzzo: "will break the alignment"

lpereira: Unbeatable argument. Patch LGTM.

mbelluzzo: Actually I will change to `isalnum` because this patch is lacking the digit check, so thanks for the catch ... and bye bye xkcd link :-(

lpereira: If that's the only change, merge it directly; no need to issue another PR.

mbelluzzo:
Merged

PR #282, opened 2015-07-08
Part of the fixes required to build with musl-libc

vcgomes:
Building with musl is a nice feature to generate small static binaries, useful for our linux-micro targets.

More changes are needed, this is the first step with the simple stuff.

### DIFF
--- a/src/modules/flow/process/common.h
+++ b/src/modules/flow/process/common.h
@@ -43,14 +43,14 @@ extern struct sol_log_domain _process_log_domain;
 struct subprocess_data {
     pid_t pid;
     struct {
-        int stdin[2];
-        int stdout[2];
-        int stderr[2];
+        int in[2];
+        int out[2];
+        int err[2];
     } pipes;
     struct {
-        struct sol_fd *stdin;
-        struct sol_fd *stdout;
-        struct sol_fd *stderr;
+        struct sol_fd *in;
+        struct sol_fd *out;
+        struct sol_fd *err;
     } watches;
     struct sol_vector write_data;
     struct sol_flow_node *node;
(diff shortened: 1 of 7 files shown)


src/modules/flow/process/common.h
```diff
@@ -43,14 +43,14 @@ extern struct sol_log_domain _process_log_domain;
...
+        int out[2];
+        int err[2];
     } pipes;
     struct {
-        struct sol_fd *stdin;
-        struct sol_fd *stdout;
-        struct sol_fd *stderr;
+        struct sol_fd *in;
+        struct sol_fd *out;
+        struct sol_fd *err;
```
lpereira: Unrelated to the patch, but maybe this could be rewritten as:

    struct {
        int fds[2];
        struct sol_fd *watch;
    } in, out, err;

lpereira:
LGTM

src/modules/flow/process/subprocess.c
```diff
@@ -172,9 +172,9 @@ on_fork(void *data)
  */
 err:
     SOL_WRN("Failed in setup the files descriptors");
-    close(mdata->pipes.stdout[1]);
-    close(mdata->pipes.stderr[0]);
-    close(mdata->pipes.stdin[0]);
+    close(mdata->pipes.out[1]);
```
lpereira: Also unrelated to the patch, but maybe `0` and `1` could be changed to constants, such as `READ_SIDE` or `WRITE_SIDE`?

dorileo:
Awesome to know it required just these minor changes.
+1

vcgomes:
> Awesome to know it required just these minor changes.

It still requires more changes, but I am still working on them.

glima:
Neat, :+1:

ibriano:
Ack

PR #283, opened 2015-07-10
Fix some general warnings

ricardotk:
They showed up after the addition of new flags on build


src/bin/sol-fbp-generator/main.c
```diff
@@ -717,13 +716,13 @@ handle_json_dir(const char *path)
     return true;
 }
 
-const char *
+static const char *
```
lpereira: Argh. This one slipped.

lpereira:
:+1:

PR #284, opened 2015-07-10
Let type outlive builder

cmarcelo:
The goal of this series is to make a simple change: let the type created by the builder be deleted using the now existing `sol_flow_node_type_del()`.

This is useful because the code handling different "kinds" of node types (from FBP, from scripting), can be also generic, and don't care about where the type came from to delete it afterwards.

### DIFF
--- a/src/bin/sol-fbp-runner/runner.c
+++ b/src/bin/sol-fbp-runner/runner.c
@@ -329,10 +329,12 @@ runner_del(struct runner *r)
 {
     if (r->root)
         sol_flow_node_del(r->root);
+    if (r->builder) {
+        sol_flow_node_type_del(r->root_type);
+        sol_flow_builder_del(r->builder);
+    }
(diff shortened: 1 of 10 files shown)


cmarcelo:
@ibriano could you take a look at this one?

@ricardotk the plan is to rebuild scripting node work on top of this. Check out with care the changes to flow parser, we don't store the builders anymore, and always delete the types using the generic function. This should avoid changes that were done to call the right delete function for your scripting nodes.

ricardotk:
@cmarcelo Ok. Since I finished the big changes on the series I'll send it to review before this change

ibriano:
Looks good.

cmarcelo:
Thanks.

Renamed the function according to @glima suggestion. Applied in 04e070c4b3543f665ed02a5f3b1d146af93ec00d and parents.

PR #285, opened 2015-07-10
Soletta meets JS

ricardotk:


src/lib/flow/sol-flow-js.c
```diff
@@ -0,0 +1,1422 @@
...
+    /* Each node keeps its own JavaScript context and global object. */
+    struct duk_context *duk_ctx;
+};
+
+static duk_ret_t
+send_boolean_packet(struct sol_flow_node *node, uint16_t port, duk_context *ctx)
+{
+    bool value = duk_require_boolean(ctx, 1);
+
+    sol_flow_send_boolean_packet(node, port, value);
```
lpereira: Do we want to ignore these errors? Shouldn't an exception be raised if these functions return <0?

src/lib/flow/sol-flow-js.c
```diff
@@ -0,0 +1,1422 @@
...
+        return 0;
+    }
+
+    port = sol_vector_get(&type->ports_out, port_number);
+    if (!port) {
+        SOL_ERR("'%s' invalid port name.", port_name);
+        return 0;
+    }
+
+    if (streq(port->type_name, "boolean"))
```
lpereira: Instead of comparing the port type names, can't we just compare the packet type pointers instead? It's a lot cheaper.

src/lib/flow/sol-flow-js.c
```diff
@@ -0,0 +1,1422 @@
...
+    if (streq(port->type_name, "rgb"))
+        return send_rgb_packet(node, port_number, ctx);
+    if (streq(port->type_name, "string"))
+        return send_string_packet(node, port_number, ctx);
+
+    /* TODO: Create a way to let the user define custom packets. Maybe we could
+     * use the same techniques we do for option parsing, and provide an object
+     * with an array of fields, offsets and values in basic C types. */
+
+    SOL_WRN("Couldn't handle unknown port type %s.", port->type_name);
```
lpereira: Raise an exception here so the JavaScript side is notified of this?

src/lib/flow/sol-flow-js.c
```diff
@@ -0,0 +1,1422 @@
...
+send_error_packet(duk_context *ctx)
+{
+    const char *value_msg = NULL;
+    struct sol_flow_node *node;
+    int value_code;
+
+    value_code = duk_require_int(ctx, 0);
+
+    if (duk_is_string(ctx, 1))
+        value_msg = duk_require_string(ctx, 1);
```
lpereira: Instead of checking if the variable is a string, can't Duktape just coerce the value into a string automatically? (In most cases it's going to be useless, though; objects in JavaScript are often coerced to "[object Object]", which isn't that great for debugging purposes, but a second number might be.)

src/lib/flow/sol-flow-js.c
```diff
@@ -0,0 +1,1422 @@
...
+    if (duk_peval_lstring(mdata->duk_ctx, type->js_content_buf, type->js_content_buf_len) != 0) {
+        SOL_ERR("Failed to read from javascript content buffer: %s", duk_safe_to_string(mdata->duk_ctx, -1));
+        duk_destroy_heap(mdata->duk_ctx);
+        return -1;
+    }
+    duk_pop(mdata->duk_ctx); /* duk_peval_lstring() result */
+
+    duk_push_global_object(mdata->duk_ctx);
+
+    /* "Soletta_node_pointer" is a hidden property. \xFF is used to give one extra level of hiding */
```
lpereira: Is the 0xff thing a standard Duktape API feature? I remember V8 allowing properties to be hidden from enumeration, be read only, etc; does this give the same effect?

src/lib/flow/sol-flow-js.c
```diff
@@ -0,0 +1,1422 @@
...
+    duk_push_string(mdata->duk_ctx, "\xFF" "Soletta_node_pointer");
+    duk_push_pointer(mdata->duk_ctx, node);
+    duk_def_prop(mdata->duk_ctx, -3, DUK_DEFPROP_HAVE_VALUE |
+        DUK_DEFPROP_HAVE_WRITABLE | 0 |
```
lpereira: I don't understand this: bitwise-or'ing with 0 is a no-op.

src/lib/flow/sol-flow-js.c
```diff
@@ -0,0 +1,1422 @@
...
+        return -1;
+    }
+
+    duk_pop_2(mdata->duk_ctx); /* process() result and global_stash */
+
+    return 0;
+}
+
+static int
+string_process(struct sol_flow_node *node, void *data, uint16_t port,
```
lpereira: These process functions have a lot of common boilerplate; can't these be moved to auxiliary functions?

     int string_process(...) {
             if (!process_boilerplate_pre())
                   error;
             ...
             if (!process_boilerplate_post())
                   error;
     }

src/lib/flow/sol-flow-js.c
```diff
@@ -0,0 +1,1422 @@
...
+        p[j]->data_type = port_type_out->type_name;
+        p[j]->array_size = 0;
+        p[j]->base_port_idx = j;
+        p[j]->required = false;
+    }
+
+    return 0;
+
+fail_ports_out_desc:
+    if (j > 0) {
```
lpereira: You can use `i` here as this isn't a nested loop. Also, the `if` isn't necessary, the condition in the `for` loop will guarantee the same thing.

src/lib/flow/sol-flow-js.c
```diff
@@ -0,0 +1,1422 @@
...
+#include <float.h>
+#include <stdio.h>
+
+#include "duktape.h"
+#include "sol-arena.h"
+#include "sol-flow-internal.h"
+#include "sol-flow-js.h"
+#include "sol-str-table.h"
+
+#define PORTS_IN_METHODS_LENGTH 3
```
lpereira: Why not an `enum`?

src/lib/flow/sol-flow-js.c
```diff
@@ -0,0 +1,1422 @@
...
+        duk_pop_3(duk_ctx);
+    }
+
+    duk_pop(duk_ctx); /* in value */
+
+    return true;
+}
+
+static bool
+setup_ports_out(struct duk_context *duk_ctx, struct sol_arena *str_arena, struct sol_vector *ports_out)
```
lpereira: These two functions are very similar; can't they be just one function that's called twice?

src/lib/flow/sol-flow-js.c
```diff
@@ -0,0 +1,1422 @@
...
+
+    js_content_buf = strndup(buf, len);
+    SOL_NULL_CHECK(js_content_buf, false);
+
+    type->js_content_buf = js_content_buf;
+    type->js_content_buf_len = len;
+
+#ifdef SOL_FLOW_NODE_TYPE_DESCRIPTION_ENABLED
+    if (setup_description(type) < 0)
+        SOL_WRN("Failed to setup description");
```
lpereira: Will this cause a problem if other parts of the code assume that `SOL_FLOW_NODE_TYPE_DESCRIPTION_ENABLED` is set and tries to use description in an inconsistent state?

src/lib/flow/sol-flow-parser.c
```diff
@@ -409,8 +423,47 @@ create_fbp_type(
...
+    SOL_NULL_CHECK(client, -ENOSYS);
+    SOL_NULL_CHECK(client->read_file, -ENOSYS);
+
+    filename = sol_arena_strdup_slice(state->arena, contents);
+    err = client->read_file(client->data, filename, &buf, &size);
+    if (err < 0)
+        return -EINVAL;
+
+    result = sol_flow_js_new_type(buf, size);
+    if (!result)
```
lpereira: Will this leak `buf`?

src/lib/flow/sol-flow-parser.c
```diff
@@ -409,8 +423,47 @@ create_fbp_type(
...
+        return -EINVAL;
+
+    if (sol_ptr_vector_append(&state->parser->no_builder_types, result) < 0) {
+        sol_flow_node_type_del(result);
```
lpereira: Ditto

src/lib/flow/sol-flow-parser.c
```diff
@@ -409,8 +423,47 @@ create_fbp_type(
...
+    const struct sol_flow_parser_client *client = state->parser->client;
+    const char *buf, *filename;
+    struct sol_flow_node_type *result;
+    size_t size;
+    int err;
+
+    SOL_NULL_CHECK(client, -ENOSYS);
+    SOL_NULL_CHECK(client->read_file, -ENOSYS);
+
+    filename = sol_arena_strdup_slice(state->arena, contents);
```
lpereira: If `filename` won't be needed outside `create_js_type()`, why not just use `strndupa()`?

src/test-fbp/javascript-expected.txt
```diff
@@ -0,0 +1,32 @@
+open()
+IN_BOOLEAN connect()
```
lpereira: Is this order deterministic?

lpereira:
There are a few things here and there, but overall I like where this is going. The samples are pretty cool.

src/lib/flow/sol-flow-js.c
ricardotk: nice catch, I'll start raising them

src/lib/flow/sol-flow-js.c
ricardotk: comparing packet type pointers worked well, thanks

src/lib/flow/sol-flow-js.c
ricardotk: ack

src/lib/flow/sol-flow-js.c
ricardotk: sendErrorPacket() receives error_code (int) and error_msg (string), the user might not want to pass a valid error_msg, the error_code could be enough. This check is to let the user pass 'null' or 'undefined' when he doesn't want to pass a valid error_msg.

src/lib/flow/sol-flow-js.c
ricardotk: I used the way it is on duktape documentation. I spoke with guys on duktape channel and they told me that it makes sense at the end and not in between, in order to easily comment the arguments. I'll drop the '0's, thanks for pointing this out @lpereira

src/lib/flow/sol-flow-js.c
ricardotk: Yes, you're right. I was trying to auto content the operations on duktape context (pushing and popping things into the stack) into one single function in order to make it easier to understand, but I think it got too stormy.

src/lib/flow/sol-flow-js.c
ricardotk: Since these loops can `goto` fail_ports_in/out_desc at any iteration, I keep the last iteration, in this case `i` and `j`, to iterate through this arrays freeing the values. But yes, I can remove the `if`.

lpereira: Please disregard this comment. The if is necessary as the condition in the `for` statement is different.

src/lib/flow/sol-flow-js.c
ricardotk: ACK

src/lib/flow/sol-flow-js.c
ricardotk: They are indeed, but we have to declare specific variables like `struct flow_js_port_in *port_type;` that only fits for input ports.

src/lib/flow/sol-flow-js.c
ricardotk: I think it won't

src/lib/flow/sol-flow-parser.c
ricardotk: ACK

src/test-fbp/javascript-expected.txt
ricardotk: No, it is not, I'll keep just the fbp output in this file.

ricardotk: I don't think it is, I'll keep just the fbp output in this file then

src/lib/flow/sol-flow-parser.c
ricardotk: No, the ownership of buf belongs to those who implemented client->read_file(), and there is a promise that it must live until the flow-parser API call retuns.

src/lib/flow/sol-flow-parser.c
ricardotk: ditto

ricardotk:
sending v2

PR #288, opened 2015-07-13
Json config try3

tcanabrava:
Third try of the json config:
- replaced keyfile with json to better represent the code
- more checking if functions worked or not
- do not return vector of config entryes, but int in case of error. pass config vectot ptr to the fill function.
- check reason outside of the json loop, not inside.
- goto err in the apropriate places
- last element of the conf vector should be NULL

Tested with both correct and incorrect json to see if things wheren't exploding.


src/shared/sol-conffile.c
```diff
@@ -58,87 +60,196 @@ sol_conffile_entry_sort_cb(const void *data1, const void *data2)
...
-static GKeyFile *
-sol_conffile_load_keyfile_from_dirs(const char *file, char **full_path)
+static char *
+sol_conffile_load_json_from_dirs(const char *file, char **full_path, struct sol_file_reader **file_reader)
 {
-    char *cwd = NULL;
-    GError *error = NULL;
-    GKeyFile *keyfile = NULL;
-    char cwdbuf[PATH_MAX + 1];
+    unsigned int i;
```
lpereira: `ARRAY_SIZE()` returns a `size_t`. Use this type for `i` as well to avoid comparison with types of different sizes.

src/shared/sol-conffile.c
```diff
@@ -146,59 +257,58 @@ sol_conffile_load_keyfile_from_dirs(const char *file, char **full_path)
...
+            char *filename;
+            if (asprintf(&filename, "%s/%s", search_dirs[i], file) <= 0) {
+                SOL_WRN("Couldn't allocate memory for config file.");
+                break;
+            }
+            *file_reader = sol_file_reader_open(file);
+            if (*file_reader)
+                config_file_contents = sol_file_reader_get_all(*file_reader);
+
+            if (strlen(config_file_contents.data) != 0)
```
lpereira: Why not use the `len` field from `sol_str_slice`, if that's already available?

src/shared/sol-conffile.c
```diff
@@ -262,19 +363,22 @@ sol_conffile_fill_vector(const char *path, const char *fallback_paths)
...
-    keyfile = sol_conffile_load_keyfile_from_paths
-            (path, fallback_paths, &full_path);
-    if (!keyfile)
+    config_file_contents = sol_conffile_load_json_from_paths(path, fallback_paths, &full_path, &file_reader);
+    if (!config_file_contents)
         return;
 
-    sol_conffile_get_keyfile_include_paths
-        (keyfile, &include, &include_fallbacks);
+    sol_json_scanner_init(&json_scanner, config_file_contents, strlen(config_file_contents));
```
lpereira: The file size is already known here inside `sol_file_reader`, there's no need to call `strlen()`.

src/shared/sol-conffile.c
```diff
@@ -262,19 +363,22 @@ sol_conffile_fill_vector(const char *path, const char *fallback_paths)
     char *full_path = NULL;
     char *include = NULL;
     char *include_fallbacks = NULL;
-    GKeyFile *keyfile = NULL;
+    char *config_file_contents = NULL;
+    struct sol_file_reader *file_reader;
     struct sol_ptr_vector pv;
+    struct sol_json_scanner json_scanner;
```
lpereira: Do you need to share the `json_scanner`?

lpereira:
Looking better.

Maybe @cmarcelo and @barbieri (which knows more about this part of the code) will have some comments as well.

src/shared/sol-conffile.c
barbieri: not only "there is no need" as it may be wrong as files may not have trailing \0, that would result ing segv.

Please run in valgrind and see, likely it is already pointing some problems

cmarcelo:
Couldn't look at this in detail today, but will check out tomorrow.

tcanabrava:
closed because of a v4

PR #289, opened 2015-07-13
Convert script that run FBP from bash to Python

cmarcelo:
It's already better than the bash version and will make easier to make further changes.


cmarcelo:
@lpereira @bdilly I suspect converting to Python might make easy to implement the flexible expected lines you suggested in the other PR.

bdilly:
@cmarcelo I agree. +1 with a minor comment.
I didn't test it.

cmarcelo:
Changed the code to use more idiomatic check. Applied in 19ecbd0095d0a8a29db71e51fa2e3f8e747a6c4f and parent.

lpereira:
:+1: for using Python instead of shell. A lot more readable.

PR #291, opened 2015-07-13
Initial Contiki port

zehortigoza:

### DIFF
--- a/src/lib/common/Kconfig
+++ b/src/lib/common/Kconfig
@@ -121,6 +121,18 @@ config PLATFORM_RIOTOS
             Soletta will do its best to offer the same experience on
             all platforms, including RIOt.
 
+config PLATFORM_CONTIKI
+	bool "contiki"
+	help
+            This platform is to be used with Contiki Small OS.
+
+            Contiki (http://contiki-os.org/) is an open source operating
+            system for the Internet of Things. Contiki connects tiny low-cost,
+            low-power microcontrollers to the Internet.
+
+            Soletta will do its best to offer the same experience on
+            all platforms, including Contiki.
+
 endchoice
 
 source "src/modules/linux-micro/Kconfig"
@@ -194,6 +206,12 @@ config MAINLOOP_RIOTOS
 	depends on PLATFORM_RIOTOS && HAVE_RIOTOS
 	help
             The mainloop to be used in RIOT platform.
+
+config MAINLOOP_CONTIKI
+	bool "contiki"
+	depends on PLATFORM_CONTIKI
+	help
+            The mainloop to be used in Contiki platform.
 
 endchoice
 
--- a/src/lib/common/include/sol-mainloop.h
+++ b/src/lib/common/include/sol-mainloop.h
@@ -161,6 +161,7 @@ struct sol_main_callbacks {
         return sol_mainloop_default_main(&(CALLBACKS), argc, argv);  \
     }
 
+#ifndef SOL_PLATFORM_CONTIKI
 #define SOL_MAIN_DEFAULT(STARTUP, SHUTDOWN)                              \
     static const struct sol_main_callbacks sol_main_callbacks_instance = { \
         .api_version = SOL_MAIN_CALLBACKS_API_VERSION,                   \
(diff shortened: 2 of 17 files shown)


src/lib/common/Kconfig
```diff
@@ -121,6 +121,18 @@ config PLATFORM_RIOTOS
             Soletta will do its best to offer the same experience on
             all platforms, including RIOt.
 
+config PLATFORM_CONTIKI
+	bool "contiki"
```
barbieri: I think you should check for some contiki header in dependencies.json and depend on it here, otherwise contiki platform will show as a build option even when it is not.

src/lib/common/include/sol-mainloop.h
```diff
@@ -161,6 +161,7 @@ struct sol_main_callbacks {
         return sol_mainloop_default_main(&(CALLBACKS), argc, argv);  \
     }
 
+#ifndef SOL_PLATFORM_CONTIKI
```
barbieri: this is not right, we should have SOL_MAIN_DEFAULT() always defined and working for every platform. I guess for contiki it would have a different implementation, but we should define it.

src/lib/common/sol-event-handler-contiki.h
```diff
@@ -0,0 +1,40 @@
...
+ * OF THIS SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
+ */
+
+#pragma once
+
+#include <stdbool.h>
+
+#include <contiki.h>
+
+bool sol_mainloop_contiki_event_handler_add(const process_event_t *ev, const process_data_t ev_data, void (*cb)(process_event_t ev, process_data_t ev_data, void *user_data), const void *data);
```
barbieri: we're using a convention of user_data being the first parameter to the function call. Also give the parameters of the callback and the "_add()/del()" functions the same, right now one is "user_data" and the other is "data".

src/lib/common/sol-mainloop-contiki.h
```diff
@@ -0,0 +1,66 @@
...
+#include <contiki.h>
+
+#ifdef __cplusplus
+extern "C" {
+#endif
+
+void sol_mainloop_contiki_event_set(process_event_t ev, process_data_t data);
+bool sol_mainloop_contiki_iter(void);
+
+#define SOL_MAIN_DEFAULT(name, strname, setup_func, teardown_func) \
```
barbieri: Either have this macro in sol-mainloop.h or define all mainloop macros in their own files (a "common" is okay since others are shared). It's just weird to see that part being #ifndef and having no clue it is being defined elsewhere (here)

src/lib/common/sol-mainloop-impl-contiki.c
```diff
@@ -0,0 +1,287 @@
...
+void
+sol_mainloop_impl_iter(void)
+{
+    // empty
+}
+
+void *
+sol_mainloop_impl_fd_add(int fd, unsigned int flags, bool (*cb)(void *data, int fd, unsigned int active_flags), const void *data)
+{
+    SOL_CRI("Unsupported");
```
barbieri: I don't think you should implement these "Unsupported". Like we do for other stuff, do not even define it in the header (#ifdef it) so it fails to compile. Nobody should be using fd_add in platforms that do not support it.

src/lib/io/sol-gpio-contiki.c
```diff
@@ -0,0 +1,135 @@
...
+static SOL_LOG_INTERNAL_DECLARE(_log_domain, "gpio");
+
+struct sol_gpio
+{
+    int pin;
+    struct sensors_sensor *button_sensor;
+    bool active_low;
+    struct {
+        void (*cb)(void *data, struct sol_gpio *gpio);
+        void *data;
```
barbieri: keep this pointer as 'const' and only cast it when calling back the cb. Instead of doing the cast when the pointer is assigned to the sol_gpio structure. This way we make it clear we should not mess with it.

zehortigoza:
New version coming, thanks all.

PR #292, opened 2015-07-13
adrafruit: Adafruit LED strip node

edersondisouza:
Node type for LPD8806 RGB led strip controller. A usage sample was also
added.

This code is untested, as I have no access to the actual hardware. I tested the fbp using console as input of some key nodes output, but that's not much more I could do. So this PR is more to validate the code. Thanks!

### DIFF
--- a/src/modules/flow/adafruit/adafruit.c
+++ b/src/modules/flow/adafruit/adafruit.c
@@ -1,0 +1,161 @@
+/*
+ * This file is part of the Soletta Project
+ *
+ * Copyright (C) 2015 Intel Corporation. All rights reserved.
+ *
+ * Redistribution and use in source and binary forms, with or without
+ * modification, are permitted provided that the following conditions
+ * are met:
+ *
+ *   * Redistributions of source code must retain the above copyright
+ *     notice, this list of conditions and the following disclaimer.
+ *   * Redistributions in binary form must reproduce the above copyright
+ *     notice, this list of conditions and the following disclaimer in
+ *     the documentation and/or other materials provided with the
+ *     distribution.
+ *   * Neither the name of Intel Corporation nor the names of its
+ *     contributors may be used to endorse or promote products derived
+ *     from this software without specific prior written permission.
+ *
+ * THIS SOFTWARE IS PROVIDED BY THE COPYRIGHT HOLDERS AND CONTRIBUTORS
+ * "AS IS" AND ANY EXPRESS OR IMPLIED WARRANTIES, INCLUDING, BUT NOT
+ * LIMITED TO, THE IMPLIED WARRANTIES OF MERCHANTABILITY AND FITNESS FOR
+ * A PARTICULAR PURPOSE ARE DISCLAIMED. IN NO EVENT SHALL THE COPYRIGHT
+ * OWNER OR CONTRIBUTORS BE LIABLE FOR ANY DIRECT, INDIRECT, INCIDENTAL,
+ * SPECIAL, EXEMPLARY, OR CONSEQUENTIAL DAMAGES (INCLUDING, BUT NOT
+ * LIMITED TO, PROCUREMENT OF SUBSTITUTE GOODS OR SERVICES; LOSS OF USE,
+ * DATA, OR PROFITS; OR BUSINESS INTERRUPTION) HOWEVER CAUSED AND ON ANY
+ * THEORY OF LIABILITY, WHETHER IN CONTRACT, STRICT LIABILITY, OR TORT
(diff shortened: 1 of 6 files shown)


lpereira:
I wouldn't call this module `adafruit`, as this is the name of a vendor, not a product. It would be better to name it `led-strip/lpd8806` or something similar.

Some of these individually-addressable LED strips are available locally; maybe we could buy a roll and try it?

src/modules/flow/adafruit/adafruit.c
```diff
@@ -0,0 +1,161 @@
...
+#include <errno.h>
+
+
+struct lcd_strip_data {
+    struct sol_spi *spi;
+    int last_set_pixel;
+    int last_set_color;
+    uint16_t pixel_count;
+    uint16_t pixel_array_length;
+    uint8_t *pixels;
```
lpereira: Move the pointer right after `spi` to avoid having a hole in the structure. Changing the type of the two `int` fields to `uint16_t` will also help with that.

src/modules/flow/adafruit/adafruit.c
```diff
@@ -0,0 +1,161 @@
...
+        sol_spi_close(mdata->spi);
+}
+
+static int
+led_strip_controler_open(struct sol_flow_node *node, void *data, const struct sol_flow_node_options *options)
+{
+    struct lcd_strip_data *mdata = data;
+    const struct sol_flow_node_type_adafruit_led_strip_options *opts;
+    uint16_t dataBytes;
+    uint8_t latchBytes;
```
lpereira: `data_bytes` and `latch_bytes`?

src/modules/flow/adafruit/adafruit.c
```diff
@@ -0,0 +1,161 @@
...
+
+    opts = (const struct sol_flow_node_type_adafruit_led_strip_options *)options;
+
+    SOL_INT_CHECK(opts->pixel_count.val, < 0, -EINVAL);
+    mdata->pixel_count = opts->pixel_count.val;
+
+    dataBytes = mdata->pixel_count * 3;
+    latchBytes = (mdata->pixel_count + 31) / 32;
+    mdata->pixel_array_length = dataBytes + latchBytes;
```
lpereira: Since there's possibility of `data_bytes + latch_bytes` overflowing an `uint16_t`, it might be better to save this to an `uint32_t`, and checking if it's over `UINT16_MIN` before attributing this value to `pixel_array_length`.

src/modules/flow/adafruit/adafruit.c
edersondisouza: Hmmm... I actually use '-1' as a 'no-value' on the int fields - so that I can wait for one PIXEL and COLOR input before actually setting the color of a pixel. Should I change that? Would add two bit-fields be better?

src/modules/flow/adafruit/adafruit.c
edersondisouza: ops. android shadow =P

src/modules/flow/adafruit/adafruit.c
edersondisouza: Did you mean `UINT16_MAX`? I think that is worth checking if the `mdata->pixel_count * 3` doesn't overflow as well, don't you think?

src/modules/flow/adafruit/adafruit.c
```diff
@@ -0,0 +1,161 @@
...
+static void
+_set_pixel_color(struct lcd_strip_data *mdata)
+{
+    uint8_t r, g, b;
+    uint8_t *pixel;
+
+    r = 0xff & (mdata->last_set_color << 16);
+    g = 0xff & (mdata->last_set_color << 8);
+    b = 0xff & mdata->last_set_color;
+
```
ceolin: shouldn't  be:

r = 0xff & (mdata->las_set_color >> 16)
g = 0xff & (mdata->las_set_color >> 8)

src/modules/flow/adafruit/adafruit.c
lpereira: Ah, fair enough. Nah, leave them as is; just reorder the struct so that (often) larger fields appear first.

src/modules/flow/adafruit/adafruit.c
lpereira: Yes, checking these things specially when there's user input involved, is always a good idea. Since these values are used for arithmetic, memory allocation, and other usually sensitive areas, it'll never hurt.

src/modules/flow/adafruit/adafruit.c
edersondisouza: Certainly... thanks!

edersondisouza:
v2 https://github.com/solettaproject/soletta/pull/300

PR #294, opened 2015-07-13
build: don't install sol-missing.h header

dorileo:
Since it's an internal header we shouldn't ship it, the header should
be used only by soletta itself.


ibriano:
Ack

lpereira:
:+1:

dorileo:
applied, thanks.

PR #298, opened 2015-07-13

dorileo:
v2:
   + new approach on the dep-resolver side - given all the changed happened on it;

### DIFF
--- a/data/scripts/dependency-resolver.py
+++ b/data/scripts/dependency-resolver.py
@@ -191,16 +191,33 @@ def handle_ccode_check(args, conf, context):
         context.add_kconfig("HAVE_%s" % dep, "bool", "n")
 
 def handle_exec_check(args, conf, context):
-    dep = conf["dependency"].upper()
-    path = which(conf["exec"]) or None
+    dep = conf.get("dependency")
+    exe = conf.get("exec", None)
 
-    context.add_cond_makefile_var(dep, path)
+    if exe is None:
+        print("Could not parse dependency: %s, no exec was specified.")
+        exit(1)
+
+    path = which(exe) or None
+    required = conf.get("required", False)
+
+    if required and path is None:
+        req_label = context.find_makefile_var("NOT_FOUND")
+        req_label += "executable: %s\\n" % exe
+        context.add_append_makefile_var("NOT_FOUND", req_label, True)
+
+    context.add_cond_makefile_var(dep.upper(), path)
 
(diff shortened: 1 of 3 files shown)


data/scripts/dependency-resolver.py
```diff
@@ -191,16 +191,33 @@ def handle_ccode_check(args, conf, context):
         context.add_kconfig("HAVE_%s" % dep, "bool", "n")
 
 def handle_exec_check(args, conf, context):
-    dep = conf["dependency"].upper()
-    path = which(conf["exec"]) or None
+    dep = conf.get("dependency")
+    exe = conf.get("exec", None)
```
lpereira: No need to pass `None` to the `get()` method.

data/scripts/dependency-resolver.py
```diff
@@ -191,16 +191,33 @@ def handle_ccode_check(args, conf, context):
...
+
+    if exe is None:
+        print("Could not parse dependency: %s, no exec was specified.")
+        exit(1)
+
+    path = which(exe) or None
```
lpereira: Why not simply `path = which(exe)`?

data/scripts/dependency-resolver.py
```diff
@@ -191,16 +191,33 @@ def handle_ccode_check(args, conf, context):
...
+        req_label += "executable: %s\\n" % exe
+        context.add_append_makefile_var("NOT_FOUND", req_label, True)
 
-    context.add_cond_makefile_var(dep, path)
+    context.add_cond_makefile_var(dep.upper(), path)
 
 def handle_python_check(args, conf, context):
-    dep = conf["dependency"].upper()
+    dep = conf.get("dependency")
+    required = conf.get("required", False)
```
lpereira: Since `None` is falsy, there's no need to pass `False` to `get()`.

data/scripts/dependency-resolver.py
dorileo: ok

data/scripts/dependency-resolver.py
dorileo: stylish thing.. I'm changing.

data/scripts/dependency-resolver.py
dorileo: sounds right, I'm changing...
thanks.

dorileo:
v3 coming.

PR #300, opened 2015-07-13
led-strip: LPD8806 LED strip node v2

edersondisouza:
v2:
  Module name changed to led-strip;
  Checks for uint16_t overflows;
  Fixed bitwise operation (<< instead of >>);
  Free pixels array;
  Reorder struct members;
  Fix variable names styles.

Node type for LPD8806 RGB led strip controller.
A usage sample was also added.

This code is untested, as I have no access to the actual hardware. I tested the fbp using console as input of some key nodes output, but that's not much more I could do. So this PR is more to validate the code. Thanks!


lpereira:
Looks good. We'll be able to test and fix potential problems whenever access to that hardware is obtained. If this is merged now, it's very unlikely to bitrot, even if untested.

edersondisouza:
Integrated

PR #301, opened 2015-07-13
Initial Contiki port V2

zehortigoza:

Changes from V1:
- including sol-mainloop-contiki.h to the exported headers on lib/common/Makefile like others files on the same directory.
- added check HAVE_CONTIKI to kconfig
- changed order of parameters of contiki-event callback to follow other callbacks
- moved Contiki version of SOL_MAIN_DEFAULT to sol-mainloop.h
- added patch to remove sol_mainloop_impl_fd_add/del and sol_mainloop_impl_child_watch_add/del stub implementation from Small Os

### DIFF
--- a/src/lib/common/include/sol-mainloop-linux.h
+++ b/src/lib/common/include/sol-mainloop-linux.h
@@ -32,24 +32,28 @@
 
 #pragma once
 
-/* implement these functions for your own system so soletta will work */
-
-#define SOL_LOG_DOMAIN &_sol_mainloop_log_domain
-extern struct sol_log_domain _sol_mainloop_log_domain;
-#include "sol-log-internal.h"
-#include "sol-mainloop.h"
-#ifdef SOL_PLATFORM_LINUX
-#include "sol-mainloop-impl-linux.h"
+#ifdef __cplusplus
+extern "C" {
 #endif
 
-int sol_mainloop_impl_init(void);
-void sol_mainloop_impl_run(void);
-void sol_mainloop_impl_quit(void);
-void sol_mainloop_impl_shutdown(void);
+enum sol_fd_flags {
+    SOL_FD_FLAGS_NONE = 0,
+    SOL_FD_FLAGS_IN   = (1 << 0),
+    SOL_FD_FLAGS_OUT  = (1 << 1),
+    SOL_FD_FLAGS_PRI  = (1 << 2),
+    SOL_FD_FLAGS_ERR  = (1 << 3),
+    SOL_FD_FLAGS_HUP  = (1 << 4),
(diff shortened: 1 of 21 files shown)


src/lib/common/include/sol-mainloop-linux.h
```diff
@@ -0,0 +1,59 @@
...
+    SOL_FD_FLAGS_IN   = (1 << 0),
+    SOL_FD_FLAGS_OUT  = (1 << 1),
+    SOL_FD_FLAGS_PRI  = (1 << 2),
+    SOL_FD_FLAGS_ERR  = (1 << 3),
+    SOL_FD_FLAGS_HUP  = (1 << 4),
+    SOL_FD_FLAGS_NVAL = (1 << 5)
+};
+
+struct sol_fd;
+struct sol_fd *sol_fd_add(int fd, unsigned int flags, bool (*cb)(void *data, int fd, unsigned int active_flags), const void *data);
```
lpereira: These are not Linux-specific; they should live in the abstraction layer. If a given platform does not implement these, they should give warnings and/or errors in runtime (or, if at all possible, at compile time, using some `__attribute__` flag).

For instance, if `sol_fd_add()` isn't available, it should return `NULL` and set errno to `ENOSYS`.

zehortigoza: On the previous PR https://github.com/solettaproject/soletta/pull/291#discussion_r34491 @barbieri asked to remove stubs implementation and avoid user only know that there is no implementation to this functions in runtime.

src/lib/common/include/sol-mainloop-linux.h
```diff
@@ -0,0 +1,59 @@
...
+bool sol_fd_del(struct sol_fd *handle);
+
+struct sol_child_watch;
+struct sol_child_watch *sol_child_watch_add(uint64_t pid, void (*cb)(void *data, uint64_t pid, int status), const void *data);
```
lpereira: Unrelated to the patch, but shouldn't `pid` be a `pid_t` instead of `uint64_t`?

src/lib/common/include/sol-mainloop-linux.h
lpereira: Alright, makes sense.

However, I would build these not tied to the platform, but tied to features a platform have. In this case, something such as `HAVE_FILE_DESCRIPTORS` and `HAVE_FORK`. This eases future porting for other POSIX-like operating systems.

zehortigoza: Sounds good, v3 on the way.

zehortigoza:
Version 3 coming.

PR #302, opened 2015-07-13

dorileo:
v2:
   + new approach on the dep-resolver side - given all the changed happened on it;

v3:
   + fixed minor stylish issues pointed by @lpereira;

### DIFF
--- a/data/scripts/dependency-resolver.py
+++ b/data/scripts/dependency-resolver.py
@@ -191,16 +191,33 @@ def handle_ccode_check(args, conf, context):
         context.add_kconfig("HAVE_%s" % dep, "bool", "n")
 
 def handle_exec_check(args, conf, context):
-    dep = conf["dependency"].upper()
-    path = which(conf["exec"]) or None
+    dep = conf.get("dependency")
+    exe = conf.get("exec")
 
-    context.add_cond_makefile_var(dep, path)
+    if not exe:
+        print("Could not parse dependency: %s, no exec was specified.")
+        exit(1)
+
+    path = which(exe)
+    required = conf.get("required")
+
+    if required and not path:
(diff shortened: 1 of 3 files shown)


data/scripts/dependency-resolver.py
```diff
@@ -191,16 +191,33 @@ def handle_ccode_check(args, conf, context):
...
+    path = which(exe)
+    required = conf.get("required")
+
+    if required and not path:
+        req_label = context.find_makefile_var("NOT_FOUND")
+        req_label += "executable: %s\\n" % exe
+        context.add_append_makefile_var("NOT_FOUND", req_label, True)
 
-    context.add_cond_makefile_var(dep, path)
+    context.add_cond_makefile_var(dep.upper(), path)
```
lpereira: If a conditional makefile variable is an invariant, why not convert the string there instead of the call site?

dorileo: AFAIR, dep is being used elsewhere, that's the why I just convert it here.

dorileo: Ok, the problem here is I'm not passing it to string format in line 198

lpereira: I mean: inside `add_cond_makefile_var()`, is there any possibility of the first parameter being used being something written in lower case? If there isn't, you can always convert the first parameter to upper case and avoid the trouble of calling `.upper()` everywhere else it's called.

dorileo: hum, ok I got it.

dorileo: but answering your question, yep, it may be not upper, the defines use this same function and my be lower, in this case it's up to what is set in the json file, it's set as is.

dorileo:
applied, thanks.

PR #303, opened 2015-07-13
Move test meta-data (expected, skipped) inside test FBPs as comments

cmarcelo:
Teach `tools/run-fbp-tests` to read the input FBP files and gather the expected results and outputs from the comments.


cmarcelo:
@lpereira

lpereira:
I like the idea of not having to have separate files: it's obvious, when changing the examples, that the tests must also be updated.

However, it's possible that the commands might be misspelled and I worry that tests will go unnoticed for a while. Maybe adopt a slightly different syntax, with two hash signs (pretty much like Doxygen), and have every command be checked by the script if they start with a given prefix?

    ## TEST-EXPECTS-ERROR
    ## TEST-OUTPTU:
    ## hello, world

Will give an error since `TEST-OUTPTU` is not a known command. The script will know it's a command since the line starts with `##` rather than `#`.

cmarcelo:
Double hash for (checked) commands is a good idea, thanks :-)  (I'll add to v2 of the PR, just let this one around for a little while to gather other comments)

lpereira:
[**@cmarcelo**](https://github.com/cmarcelo) commented on [Jul 13, 2015, 6:31 PM GMT-3](https://github.com/solettaproject/soletta/pull/303#issuecomment-121066041 "2015-07-13T21:31:58Z - Replied by Github Comment Enhancer"):
> Double hash for (checked) commands is a good idea, thanks :-)

The script might give a warning if it finds `# TEST-OUTPUT` as well, so if someone forgets the double hash, they'll be notified.

cmarcelo:
Replaced by https://github.com/solettaproject/soletta/pull/306.

PR #305, opened 2015-07-13
tests: fix warnings on two fbp tests

bdilly:
v2 changelog:
  * use regular expressions instead of hardcoded loc


lpereira:
I recommend coordinating with @cmarcelo and add a `TEST-OUTPUT-REGEX` command.

cmarcelo:
The change that #306 would incur is that you read from a snippet of comment instead of from a file. Things should be similar to what is done for `TEST-OUTPUT`.

bdilly:
a rebased version was pushed

PR #306, opened 2015-07-13
Move test meta-data (expected, skipped) inside test FBPs as comments [v2]

cmarcelo:
Teach tools/run-fbp-tests to read the input FBP files and gather the expected results and outputs from the comments.


cmarcelo:
This version uses the double hash prefix to identify the meta-data (and report unknown commands). The commands now are easier to spot. If we face real issues with using the syntax, we can add more safety guards, for now I think this is a good balance.

tools/run-fbp-tests
```diff
@@ -68,26 +68,38 @@ def collect_tests(tests):
...
-    if os.path.isfile(success_file):
-        with open(success_file) as f:
-            out = f.read()
-    elif os.path.isfile(fail_file):
-        with open(fail_file) as f:
-            out = f.read()
-        code = 1
+    f = open(t)
+    for line in f:
+        if not line.startswith("## "):
```
lpereira: Maybe check if it doesn't start with `"##"` instead, and strip the `cmd`?

cmarcelo: If the line doesn't start with `## ` (with space) we won't treat it as a command.

cmarcelo: `############ LOGIC ##############` shouldn't issue a warning. If we just look without space, we may have to do more checking to see if this is a command (and possibly warn about it) or if it is valid comment that we don't care.

glima:
All LGTM. Thanks for the effort, that will help a lot when writing tests for the string converters :)

tools/run-fbp-tests
lpereira: Alright. Makes sense. Stripping the test command might help, though, in cases `##    TEST-OUTPUT:` is accidentally written.

cmarcelo:
Applied in 679daa78a1033a49808e79af445b171a3eed14d2 and parents.

PR #304, opened 2015-07-13
builder: Fix memory leak

ceolin:
Problem spotted by coverity

PR #307, opened 2015-07-13
flow: Avoid possible leak in sol-flow-builder

anselmolsm:
opts would leak if add_guard returned an error.
Issue pointed out by Coverity (125378).

### DIFF
--- a/src/lib/flow/sol-flow-builder.c
+++ b/src/lib/flow/sol-flow-builder.c
@@ -973,8 +973,9 @@ sol_flow_builder_get_node_type(struct sol_flow_builder *builder)
     err = add_guards(builder);
     if (err < 0) {
         SOL_WRN("Failed to allocate memory for constructing node type");
+        free(opts);
         errno = ENOMEM;
-        return NULL;
+        goto guards_error;
     }
 
     spec->nodes = builder->nodes.data;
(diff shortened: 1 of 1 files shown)


src/lib/flow/sol-flow-builder.c
```diff
@@ -973,8 +973,9 @@ sol_flow_builder_get_node_type(struct sol_flow_builder *builder)
     err = add_guards(builder);
     if (err < 0) {
         SOL_WRN("Failed to allocate memory for constructing node type");
+        free(opts);
         errno = ENOMEM;
-        return NULL;
+        goto guards_error;
```
edersondisouza: `guards_error` label already has a `free(opts)`, so isn't this causing a double free, since there's a `free(opts)` just after the `SOL_WRN...`?

ceolin:
this puillreq is duplicated https://github.com/solettaproject/soletta/pull/304.

I think there is no need to set the spec fields to NULL in case of this error because they were not set at this point.

src/lib/flow/sol-flow-builder.c
anselmolsm: @edersondisouza I added the free(opts) line, then changed my mind to use the goto statement. So, it shouldn't be here. But, as @ceolin also submitted a patch to fix this, I dropped my pr.

PR #309, opened 2015-07-14
Initial Contiki port V3

zehortigoza:

Changes from V2:
- Added option to export mainloop fd and fork watch functions instead of having it on a Linux specific file.
- Removed no needed file sol-contiki-missing.h

### DIFF
--- a/src/lib/common/Kconfig
+++ b/src/lib/common/Kconfig
@@ -332,16 +332,6 @@ menuconfig HAVE_PIN_MUX
 
           If unsure, say Y.
 
-config MAINLOOP_FD
-	bool "Mainloop file descriptors"
-	depends on SOL_PLATFORM_LINUX
-	default y
-
-config MAINLOOP_FORK_WATCH
-	bool "Mainloop fork watch"
-	depends on SOL_PLATFORM_LINUX
-	default y
-
 source "src/modules/pin-mux/intel-galileo-rev-d/Kconfig"
 source "src/modules/pin-mux/intel-galileo-rev-g/Kconfig"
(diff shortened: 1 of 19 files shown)


src/lib/common/Kconfig
```diff
@@ -313,6 +332,16 @@ menuconfig HAVE_PIN_MUX
 
           If unsure, say Y.
 
+config MAINLOOP_FD
```
lpereira: If these are options, the build will break unless every single part of the code that depends on fd watches get properly left out if this is disabled. Too much ado over a oft unused feature.

Instead, I suggest enabling them conditionally using the C pre-processor in the `sol-common-buildopts.h.in` file:

    #if defined(SOL_PLATFORM_LINUX)
    #define SOL_MAINLOOP_FD_ENABLED 1
    #define SOL_MAINLOOP_FORK_WATCH 1
    #endif

That'll still enable code to be built depending on features being available on a platform or not.

I'd go even further and define `SOL_PLATFORM_POSIX` if `SOL_PLATFORM_LINUX` is enabled, and define these if that's the case. This way, if, say, a FreeBSD port comes along, one just defines `SOL_PLATFORM_POSIX`  if `SOL_PLATFORM_FREEBSD` and things should just work.

zehortigoza: Sounds better, done in the patch 'squash to: Remove mainloop stub implementations of Small Os'
If no other changes were requested in this PR I will just squash and merge it.

ulissesf: +1 with @lpereira's suggetion (and if you test it :-)

zehortigoza:
Merged with 'squash to: Remove mainloop stub implementations of Small Os' squashed.

PR #312, opened 2015-07-14
Json config try4

tcanabrava:
1 - removed 'const' from config entry struct 
2 - return sol_str_slice instead of char* in the functions
3 - use sol_str_slice.len instead of strlen(sol_str_slice.data)
4 - memory is finally freed.


tcanabrava:
( missed a commit with 'size_t' instead of 'unsigned int' in a line, already updated this in my local repo)

src/shared/sol-conffile.c
```diff
@@ -58,87 +60,207 @@ sol_conffile_entry_sort_cb(const void *data1, const void *data2)
...
+
+    sol_ptr_vector_init(&vec_options);
+
+    sol_json_scanner_init_from_token(&scanner, &options_object);
+    SOL_JSON_SCANNER_OBJECT_LOOP (&scanner, &token, &key, &value, reason) {
+        int key_len, value_len;
+        char *tmp;
+        key_len = sol_json_token_get_size(&key) - 2; // remove quotes from the key.
+        value_len = sol_json_token_get_size(&value); // do not remove quotes from the value. (not everything is a string here)
+        if (asprintf(&tmp, "%.*s=%.*s", key_len, key.start + 1, value_len, value.start) <= 0) {
```
lpereira: Might be a good idea to check if `key_len` is zero before calling `asprintf()` and abort the loading procedure.

src/shared/sol-conffile.c
```diff
@@ -58,87 +60,207 @@ sol_conffile_entry_sort_cb(const void *data1, const void *data2)
...
-            SOL_DBG("could not find mandatory 'Type' key on group [%s],"
-                " skipping entry", *groups_ptr);
-            free((void *)entry->id);
-            free(entry);
-            continue;
+static int
+sol_conffile_json_to_vector(struct sol_json_scanner scanner, struct sol_ptr_vector *pv)
+{
+    struct sol_conffile_entry *entry = NULL;
+    const char *node_group = "nodetypes";
```
lpereira: `static const char *node_group = "..."`: this way, the array is always kept in the text segment.

src/shared/sol-conffile.c
```diff
@@ -58,87 +60,207 @@ sol_conffile_entry_sort_cb(const void *data1, const void *data2)
...
 
-        sol_ptr_vector_insert_sorted(&pv, entry,
-            sol_conffile_entry_sort_cb);
+    sol_json_scanner_init_from_token(&obj_scanner, &value);
+    SOL_JSON_SCANNER_ARRAY_LOOP (&obj_scanner, &token, SOL_JSON_TYPE_OBJECT_START, reason) {
+        entry = calloc(1, sizeof(*entry));
+        SOL_NULL_CHECK(entry, -ENOMEM);
+        SOL_JSON_SCANNER_OBJECT_LOOP_NEST (&obj_scanner, &token, &key, &value, reason) {
+            if (reason != SOL_JSON_LOOP_REASON_OK) {
+                SOL_DBG("Error: Invalid Json.");
```
lpereira: s/Json/JSON/. JSON is an acronym (JavaScript Object Notation), and all acronyms should be written in capital letters.

src/shared/sol-conffile.c
```diff
@@ -58,87 +60,207 @@ sol_conffile_entry_sort_cb(const void *data1, const void *data2)
...
-sol_conffile_load_keyfile_from_dirs(const char *file, char **full_path)
+static struct sol_str_slice
+sol_conffile_load_json_from_dirs(const char *file, char **full_path, struct sol_file_reader **file_reader)
 {
-    char *cwd = NULL;
-    GError *error = NULL;
-    GKeyFile *keyfile = NULL;
-    char cwdbuf[PATH_MAX + 1];
+    size_t i;
+    struct sol_str_slice config_file_contents;
```
lpereira: Initialize this slice with an empty slice (there's a macro for that in the slice include). This is because, if the `else` below fails, `config_file_contents` will be used uninitialized.

src/shared/sol-conffile.c
```diff
@@ -146,59 +268,60 @@ sol_conffile_load_keyfile_from_dirs(const char *file, char **full_path)
...
+            }
+            *file_reader = sol_file_reader_open(file);
+            if (*file_reader)
+                config_file_contents = sol_file_reader_get_all(*file_reader);
 
-    return keyfile;
+            if (config_file_contents.len != 0)
+                break;
+            else
+                SOL_WRN("Error creating type.");
```
lpereira: "Could not open file ($PATH) while creating type" or something like that will help the user when things don't work as expected.

src/shared/sol-conffile.c
```diff
@@ -146,59 +268,60 @@ sol_conffile_load_keyfile_from_dirs(const char *file, char **full_path)
...
+sol_conffile_load_json_from_paths(const char *path,
     const char *fallback_paths,
-    char **full_path)
+    char **full_path,
+    struct sol_file_reader **file_reader)
 {
-    gchar **ptr;
-    gchar **splitted_paths;
-    GKeyFile *keyfile = NULL;
+    struct sol_str_slice config_file_contents;
```
lpereira: Initialize this as well: there's a chance that this might get uninitialized.

src/shared/sol-conffile.c
```diff
@@ -146,59 +268,60 @@ sol_conffile_load_keyfile_from_dirs(const char *file, char **full_path)
...
+    config_file_contents.data = NULL;
+    config_file_contents.len = 0;
```
lpereira: Ah, I see. There's a macro to initialize a slice, use it instead.

src/shared/sol-conffile.c
tcanabrava: okay

PR #315, opened 2015-07-14
Soletta meets JS v2

ricardotk:
DIfferences from v1:
    - start raising javascript exceptions in error cases
    - start comparing packet type pointers in `send_packet()`
    - remove useless `| 0` in `duk_def_prop()` flags
    - create auxiliary functions to process functions
    - use `strndupa()` instead of `sol_arena_strdup_slice()` in `create_js_type()`
    - small changes on samples


src/lib/flow/sol-flow-js.c
```diff
@@ -0,0 +1,1328 @@
...
+    duk_put_prop_string(mdata->duk_ctx, -2, "sendPacket");
+
+    duk_push_c_function(mdata->duk_ctx, send_error_packet, 2);
+    duk_put_prop_string(mdata->duk_ctx, -2, "sendErrorPacket");
+
+    /* From this point node JS object is always in the top of the stack. */
+    duk_get_prop_string(mdata->duk_ctx, -1, "node");
+
+    if (!setup_ports_methods(mdata->duk_ctx, type->ports_in.len, type->ports_out.len)) {
+        SOL_ERR("Failed to bla TODO: %s", duk_safe_to_string(mdata->duk_ctx, -1));
```
lpereira: :)

ricardotk: It was a test and you're approved! congratz (: :confetti_ball:

ricardotk:
sparsecheckout seems to be the best solution among the others, thanks @bdilly and @dorileo

ricardotk:
I changed to sparse checkout and added a MIT file about the license (required by legal), does anyone have more comments? I'll push it today if no other commentary shows up.

ricardotk:
v3 PR #346

PR #317, opened 2015-07-14
UART: Simplify API

zehortigoza:


src/lib/io/include/sol-uart.h
```diff
@@ -53,36 +53,47 @@ extern "C" {
 
 struct sol_uart;
 
-struct sol_uart *sol_uart_open(const char *port_name);
+enum sol_uart_speed {
+    SOL_UART_SPEED_9600 = 0,
```
lpereira: Better to call this baud rate. Speed is distance over time.

src/lib/io/include/sol-uart.h
```diff
@@ -53,36 +53,47 @@ extern "C" {
...
+    SOL_UART_DATA_BITS_5
+};
+
+enum sol_uart_parity {
+  SOL_UART_PARITY_DISABLE = 0,
+  SOL_UART_PARITY_ENABLE_EVEN,
+  SOL_UART_PARITY_ENABLE_ODD
+};
+
+enum sol_uart_stop_bits {
```
lpereira: I still think that merging these three enums would be a good idea. Something along these lines:

    enum settings {
           DATA_BITS_8 = 1<<0,
           DATA_BITS_7 = 1<<1,
           DATA_BITS_6 = 1<<2,
           DATA_BITS_5 = 1<<3,
           PARITY_NONE = 1<<4,
           PARITY_EVEN = 1<<5,
           PARITY_ODD = 1<<6,
           STOP_BITS_1 = 1<<7,
           STOP_BITS_2 = 1<<8,
           FLOW_CONTROL_XON = 1<<9,
           FLOW_CONTROL_XOFF = 1<<10,
           DEFAULT = DATA_BITS_8 | PARITY_NONE | STOP_BITS_1
    }

Then it's the matter of:

    uart = uart_open("/dev/ttyS0", BAUD_115200, DEFAULT , cb, data);

src/lib/io/include/sol-uart.h
mbelluzzo: +1 to baud rate

src/lib/io/include/sol-uart.h
mbelluzzo: imho follow the same pattern used in pwm and gpio. Leave the 3 (4 in the case) enums as is and have a struct called config where those are set and passed as arg. So we have a consistent API across lib/io

lpereira: I dislike functions with many parameters, since it's too easy to mix their orders (hell, I ended up mixing the order of `memset()` and took me a while to find the bug because the compiler didn't complain!). Parameters being enums will help here, though, so whatever floats @zehortigoza's boat.

dorileo: Yeah, I agree. API argument increase doesn't seem to be an "API simplification", bitmask should do it all at once...

zehortigoza:
Thanks all, new version https://github.com/solettaproject/soletta/pull/329

PR #318, opened 2015-07-15
docs: Remove reference to old configure option

anselmolsm:
There is no '--with-mainloop' anymore.


lpereira:
:+1: 

Maybe we should make exceptions for PRs like these in the future? WebKit had a nice exception for build fixes for instance: they didn't need to go past revision.

anselmolsm:
@lpereira +1 for this exception. Tks.

PR #320, opened 2015-07-15
More robust way to get the rootdir

vcgomes:
Instead of using paths relative to the binary that is being run, this series proposes a way to find the path relative to the libsoletta.so library. This solves the problem when running the binaries under gdb, sudo, and so on.

### DIFF
--- a/data/jsons/dependencies.json
+++ b/data/jsons/dependencies.json
@@ -128,6 +128,20 @@
 	    "fragment": "getauxval(AT_EXECFN);"
 	},
 	{
+	    "dependency": "decl_dladdr",
+	    "type": "ccode",
+	    "defines": [
+		"_GNU_SOURCE"
+	    ],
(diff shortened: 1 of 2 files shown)


ibriano:
Looks good

mbelluzzo:
+1

data/jsons/dependencies.json
```diff
@@ -128,6 +128,20 @@
 	    "fragment": "getauxval(AT_EXECFN);"
 	},
 	{
+	    "dependency": "decl_dladdr",
+	    "type": "ccode",
+            "defines": [
```
lpereira: These seems oddly indented. Patch looks fine otherwise, so before merging ensure they're OK.

vcgomes: Ugh, I did miss that. Thanks for spotting it.

PR #321, opened 2015-07-15
build: create rule for generate svg

bdilly:
Also support parameters on script to set output and fbp-to-dot
paths


bdilly:
v2 is out there

PR #322, opened 2015-07-15
build: split internal tests

dorileo:
We have 2 categories of tests, one testing the public exported API's
and those testing the "internal API's", the current only internal test
implementation is test-coap.

This patch introduces a differentiation for the mentioned categories.
Tests using the public API are now linked to libsoletta.so and uses
only objects exported there, the internal ones have access to any part
of internal symbols, so they can test any part of it.

### DIFF
--- a/src/test/Makefile
+++ b/src/test/Makefile
@@ -1,8 +1,8 @@
 test-$(TEST_ARENA) += test-arena
 test-test-arena-$(TEST_ARENA) := test.c test-arena.c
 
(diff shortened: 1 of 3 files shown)


dorileo:
@vcgomes @cmarcelo this patch implements what we talked yesterday about test suite, in the end I haven't even split into different directories, I just created a different category of tests and I think they can sit together - side by side - in the src/test/ dir....

vcgomes:
Looks good. Did a quick test, all went fine. :+1:

PR #323, opened 2015-07-15
build: generate svg on cheat-sheet rule

bdilly:
v2 changelog:
  * keep make doc dependency on cheat sheet
  * check if sol-fbp-to-dot was built

Also support parameters on script to set output and fbp-to-dot
paths


bdilly:
merged with proposed change

PR #324, opened 2015-07-15
Fix check-style errors

zehortigoza:


lpereira:
:+1: 

For style fixes, you can just push them directly to the repository. No need to create a PR.

PR #325, opened 2015-07-15
Fixes (some issues raised on github)

ceolin:
- Remove EBADR from code
- Fix the build when i2c-dev.h is given by i2c-tools
- Fix the exit() on child process
 - This last fix was based from a old @thiagomacieira patch.


lpereira:
You can also write "Closes #97" in the commit that removed `EBADR`.

I would cherry-pick #95 (eventually rebasing it), and find other places where `_exit()` could be used instead of `exit()`. And, of course, write `Closes #95`.

Besides this, LGTM.

Issue #89, opened 2015-06-25
Can't include both linux/i2c.h and linux/i2c-dev.h if i2c-dev is provided by i2c-tools


thiagomacieira:
Whenever linux/i2c-dev.h is provided by i2c-tools, it is incompatible with linux/i2c.h.

Errors:
```
In file included from src/shared/sol-i2c-linux.c:39:0:
/usr/include/linux/i2c-dev.h:37:8: error: redefinition of ‘struct i2c_msg’
 struct i2c_msg {
        ^
In file included from src/shared/sol-i2c-linux.c:38:0:
/usr/include/linux/i2c.h:68:8: note: originally defined here
 struct i2c_msg {
        ^
In file included from src/shared/sol-i2c-linux.c:39:0:
/usr/include/linux/i2c-dev.h:89:7: error: redefinition of ‘union i2c_smbus_data’
 union i2c_smbus_data {
       ^
In file included from src/shared/sol-i2c-linux.c:38:0:
/usr/include/linux/i2c.h:128:7: note: originally defined here
 union i2c_smbus_data {
       ^
```

See https://build.opensuse.org/package/view_file/openSUSE:13.2/linux-glibc-devel/linux-glibc-devel.spec line 102

glima:
Since I'm starting vacations real soon and I could not reproduce it on time, @ceolin is going to chase the issue instead.

thiagomacieira:
Install OpenSUSE to reproduce. Or at the very least ensure that /usr/linux/linux/i2c-dev.h comes from the i2c-tools package and is unmodified from upstream.

This is most likely an OpenSUSE/i2c-tools bug. I'm asking for a workaround.

ceolin:
@thiagomacieira the following commit does the workaround. I've tested installing the header from i2c-tools manually and worked. Could you confirm it to me close this issue ?

https://github.com/solettaproject/soletta/commit/427f47fb2c7698d06438a243600d5c0c5aa5758d

thiagomacieira:
If you're getting that warning, I guess it fixes the issue, because I'm getting the warning too.

Why do you need to include linux/i2c.h? I didn't need to here.

ceolin:
linux/i2c-dev.h from i2c-tools contains the symbols from linux/i2c.h. But in the common case where i2c-dev.h comes from kernel it's necessary to include linux/i2c.h

thiagomacieira:
Understood.

PR #326, opened 2015-07-15
Fixes v2

ceolin:
-v2
 - Removed duplicated commit (there already is a commit fixing the exit() problem)

This series fix the issues:
* https://github.com/solettaproject/soletta/issues/97
* https://github.com/solettaproject/soletta/issues/89

@lpereira I prefer to put the information about the issues in the pullreq message instead of the put in the commit messages.

### DIFF
--- a/data/scripts/sol-flow-node-type-stub-gen.py
+++ b/data/scripts/sol-flow-node-type-stub-gen.py
@@ -259,7 +259,7 @@ packet_get_%(name)s(const struct sol_flow_packet *packet
     struct %(name_data)s %(name)s;
     int ret;
 
-    SOL_NULL_CHECK(packet, -EBADR);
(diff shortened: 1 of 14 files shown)


lpereira:
The issues will be closed automatically if they're in commit messages. The messages in the PR will only work if it's merged via the GitHub UI, but we decided to fight the tool we're using rather than just embrace it... :P

ceolin:
I do prefer not mess our history with github's information :)

Anyway, about the commits do I have your ack ?

ceolin:
That is exactly the point, when i2c-dev.h is given by i2c-tools it is defined on i2c-dev.
T

barbieri:
+1

ceolin:
merged, thanks

PR #328, opened 2015-07-15
dependency-resolver: Add verbose option.

lpereira:
Set environment variable `V` to a non-zero number or any trueish value
("true", "yes", "on"), and messages will be printed while the checks are
performed.

This variable can be passed to make as well: `make reconf V=1`.


lpereira:
Merged with suggested changes, thanks!

PR #329, opened 2015-07-15
UART: Simplify API V2

zehortigoza:
Changes from V1:
- Changed API to the bitmask approach
- Add 2 more patches: 'UART: Remove transmission queue' and 'UART: Improve sample'

### DIFF
--- a/src/lib/io/include/sol-uart.h
+++ b/src/lib/io/include/sol-uart.h
@@ -53,36 +53,40 @@ extern "C" {
 
 struct sol_uart;
 
-struct sol_uart *sol_uart_open(const char *port_name);
+enum sol_uart_baud_rate {
+    SOL_UART_BAUD_RATE_9600 = 0,
+    SOL_UART_BAUD_RATE_19200,
+    SOL_UART_BAUD_RATE_38400,
+    SOL_UART_BAUD_RATE_57600,
+    SOL_UART_BAUD_RATE_115200
+};
+
+enum sol_uart_settings {
+    SOL_UART_DATA_BITS_8 = (1 << 0),
+    SOL_UART_DATA_BITS_7 = (1 << 1),
+    SOL_UART_DATA_BITS_6 = (1 << 2),
+    SOL_UART_DATA_BITS_5 = (1 << 3),
+    SOL_UART_PARITY_NONE = (1 << 4),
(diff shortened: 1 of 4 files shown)


src/lib/io/include/sol-uart.h
```diff
@@ -53,36 +53,40 @@ extern "C" {
...
+    SOL_UART_STOP_BITS_ONE = (1 << 7),
+    SOL_UART_STOP_BITS_TWO = (1 << 8),
+    SOL_UART_FLOW_CONTROL = (1 << 9),
+    SOL_UART_DEFAULT = (SOL_UART_DATA_BITS_8 | SOL_UART_PARITY_NONE | SOL_UART_STOP_BITS_ONE)
+};
+
+struct sol_uart *sol_uart_open(const char *port_name,
+    enum sol_uart_baud_rate baud_rate, enum sol_uart_settings settings,
+    void (*rx_cb)(void *user_data, struct sol_uart *uart, char read_char),
+    const void *rx_cb_user_data);
```
mbelluzzo: -1
I think it's important to have a consistent API across lib/io. All others APIs use a config structure and I think that uart's API would benefit from this pattern as well.

zehortigoza: Any more opinions? @lpereira and @dorileo like this way, for me both are better than the actual(several getters and setters).

mbelluzzo: No one is defending the current way, but since you are changing, do it the
right way.

For me it doesn't matter that they like, I don't dislike as well compared
whit the current one, but the fact is that neither @lpereira nor @ldorileo
are working directly in those APIs, so unless a good reason for not to
follow the  current pattern (that is being used in PWM, GPIO and AIO) is
given,  the -1 stands.

ibriano: I agree. I'd be fine if it were just the two enum values, but having to pass the callbacks and user data as well it begins to make sense to have a struct passed instead. Especially since we have already changed other interfaces (besides IO) to use that approach.

zehortigoza: Ok, I will update and send another PR.

Thanks

zehortigoza:
New version https://github.com/solettaproject/soletta/pull/360

PR #331, opened 2015-07-15
New version of cheat sheet

bdilly:


tools/generate-svg-from-all-fbps
```diff
@@ -56,11 +56,18 @@ function die() {
...
+       continue
     fi
+
     dir=$(dirname $f)
     mkdir -p $OUTPUT/$dir
+
+    # Copy fbp used to generate svg removing license.
+    # It's going to be used to display code snippets, having 30 lines
+    # of license is annyoing.
+    tail -n +31 $f > $OUTPUT/$f
```
lpereira: Wouldn't it be better to remove, for these purposes, all commented-out lines? It's less hackish than skipping the first few lines, which won't work if, say, a Copyright line is added due to some external contributor work.

bdilly: I don't think so.
Many of the samples explain a lot in comments.
I prefer to keep them.

lpereira: Not sure how to solve this, then. Removing the first few lines will most likely break in the future. Even if there's a single newline as the first character, it'll break the assumption it's 31 lines.

bdilly: I understand, it's a weak solution.
But since we're using a script / template / coping and pasting a header with this size, it seems to work.

doc/node-types-html/js/Category.js
```diff
@@ -1,25 +0,0 @@
-(function(window){
-
-    //name          = The name of the category
-    //description   = The description of the category
-    //id            = The id of the category, used to function as an anchor to the created category
```
ricardotk: Why not `/* */` on commentaries?

bdilly:
I've changed comments and fixed an issue when setting data_types in our script.

tools/generate-svg-from-all-fbps
lpereira: How about using `sed` to remove every line, from `This file is part of the Soletta Project` up to and including `OF THIS SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.`? While still somewhat hackish, I think this is doable and more future-proof.

bdilly: Yeah, I agree this solution is more future-proof. I'm going to test it here.
Thanks

bdilly:
merged

PR #332, opened 2015-07-15
stub-gen: Macro and struct names wrong

edersondisouza:
Generated files struct name and functions changed over time, but
sol-flow-node-type-stub-gen.py has not been updated to reflect this new
naming.

### DIFF
--- a/data/scripts/sol-flow-node-type-stub-gen.py
+++ b/data/scripts/sol-flow-node-type-stub-gen.py
@@ -335,11 +335,11 @@ static int
         print_data_struct(outfile, struct)
         if "options" in data:
             outfile.write("""\
-    const struct %(name_c)s_options *opts;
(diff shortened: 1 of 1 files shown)


lpereira:
LGTM.

It would be a good idea to add a test for this, though: given a JSON file, see if the stub compiles.

bdilly:
@lpereira @edersondisouza we have a dummy.json that would exercise stub. When it was added, it was covering everything, including custom packets.

Actually we used to have even a build rule for that, most probably lost on transition to kbuild.

edersondisouza:
Oh, I see - I failed to look at the script options. Sorry for the noise.

edersondisouza:
Will send a v2 only changing the API check macro name.

edersondisouza:
There's nothing to check there either - maybe I'm too high. I can check the testing stuff sometime.

PR #333, opened 2015-07-15
build: .in files should be regenerated after a .config change

dorileo:
If we don't regenerate the .in files they will end up using the wrong
prefix if it has been reset.


barbieri:
+1

dorileo:
applied.

PR #334, opened 2015-07-15
build: doc targets

dorileo:
Created a new target specific to doxygen, the former doc target now
remains aggregating all the docs targets (i.e doxygen, cheat-sheet).

Removed the builtins.json dependency since we don't need it to generate
the cheat-sheet.

Added a fallback rule for cheat-sheet if solettas is configured without
FBP_TO_DOT.

Also added graphviz dependency to dep-resolver once it's required to run
the svg generator script.


dorileo:
v2 coming.

PR #336, opened 2015-07-16
Simple C Type

barbieri:
This allows creating a `struct sol_flow_node_type` from a single function and a list of ports.

It could be used by things like expressing logic in C from FBP using DECLARE statements (similar to JS, but would need someone to hook it up, call the compiler and so on).

### DIFF
--- a/src/lib/common/include/sol-macros.h
+++ b/src/lib/common/include/sol-macros.h
@@ -54,6 +54,7 @@
 #define SOL_ATTR_SECTION(secname) __attribute__((section(secname)))
 #define SOL_ATTR_USED __attribute__((__used__))
 #define SOL_ATTR_UNUSED __attribute__((__unused__))
+#define SOL_ATTR_SENTINEL __attribute__((sentinel))
 #else
 #define SOL_API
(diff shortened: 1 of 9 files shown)


src/lib/common/include/sol-macros.h
```diff
@@ -64,6 +65,7 @@
 #define SOL_ATTR_NONNULL(...)
 #define SOL_ATTR_SECTION(secname)
 #define SOL_ATTR_USED
```
lpereira: Unrelated to the patch, but `UNUSED` isn't here.

src/lib/common/include/sol-types.h
```diff
@@ -46,6 +46,51 @@ extern "C" {
...
+ * #define mystruct_set_string(st, x) \
+ *    do { \
+ *       st->type = mystruct_type_string; \
+ *       st->value = sol_check_const_str(x); \
+ *    } while (0)
+ * @endcode
+ *
+ * @see sol_check_str()
+ */
+static inline const char *sol_check_const_str(const char *s) { return s; }
```
lpereira: There's a macro to check that already.

barbieri:
New PR #339.

PR #337, opened 2015-07-16
build: doc targets

dorileo:
v2:
   + removed the (left over) cheat-sheet dep from doxygen;
   + fixed echo call with wrong quotes;
   + fixed a type on cheat-sheet graphviz guard warning msg;

Created a new target specific to doxygen, the former doc target now
remains aggregating all the docs targets (i.e doxygen, cheat-sheet).

Removed the builtins.json dependency since we don't need it to generate
the cheat-sheet.

Added a fallback rule for cheat-sheet if solettas is configured without
FBP_TO_DOT.

Also added graphviz dependency to dep-resolver once it's required to run
the svg generator script.

### DIFF
--- a/data/jsons/dependencies.json
+++ b/data/jsons/dependencies.json
@@ -265,6 +265,11 @@
 	    "type": "python",
 	    "pkgname": "jsonschema",
 	    "required": true
+	},
+	{
+	    "dependency": "graphviz",
+	    "type": "exec",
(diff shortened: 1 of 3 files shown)


bdilly:
lgtm
I didn't test it

dorileo:
ok, applied. THanks.

PR #338, opened 2015-07-16
docs: show examples on node types doc

bdilly:
TODO:
  * fix thumbnails (I'm not happy with thumbnails idea, area is too
    small)
  * fix svg visualization: it would be better to move a viewport
    over the image, respecting boundaries. Also it should zoom to
    fit all the area when image is small.


bdilly:
just to make @alponsirenas "happy"...
there is a big todo / fixme list related to this task yet

bdilly:
closing because a new version was pushed

PR #340, opened 2015-07-17
Lsm303 - Accelerometer, magnetometer and compass

edersondisouza:
Something to notice - on the accelerometer module, I've tried to have a separate file for LSM303, but the result code is hackish, with some weird `#includes` to have things working - I've actually create a branch lsm303-multiple-files, if someone is curious to see the result. Anyway, this made me wonder if we should have a better way to handle this case - at least I find annoying having different devices handling on the same file - it won't scale.


edersondisouza:
v2 #345

PR #343, opened 2015-07-17
docs: show examples on node types doc v2

bdilly:
changes from v1:
  * generate thumbnails


tools/generate-svg-from-all-fbps
```diff
@@ -64,10 +64,28 @@ for f in $(find src/ -name \*.fbp); do
...
+
     # Copy fbp used to generate svg removing license.
     # It's going to be used to display code snippets, having 30 lines
     # of license is annyoing.
     awk -v filter=0 -e '{if ($0 ~ "This file is part of the Soletta") filter=1; if (filter==0) print; if ($0 ~ "SUCH DAMAGE") filter=0;}' < $f > $OUTPUT/$f
 
-    $FBP_TO_DOT --fbp=$f --dot=/dev/stdout | dot -Tsvg > $OUTPUT/${f/.fbp/.svg}
+    # Create svg from fbp
+    $FBP_TO_DOT --fbp=$f --dot=$f_dot
+    # It may fail if any module wasn't built
```
lpereira: It might be interesting to adopt [the unnoficial bash strict mode](http://redsymbol.net/articles/unofficial-bash-strict-mode/).

Also, it's often a better idea to just use the following syntax:

    if $FBP_TO_DOT --fbp=$f --dot=$f_dot; then
         generate_img $f_dot
    else
        echo Failed to convert "$f".
        continue
    fi

bdilly:
merged with proposed changes + fix on bs

PR #344, opened 2015-07-17
conffile: Fix memory leak

ceolin:
The error label only treats the entries in the vector. Just releasing the
entry's memory in case the error before its insertion in the vector.


ceolin:
merged, thanks !

PR #345, opened 2015-07-17
Lsm303 v2

edersondisouza:
v2:


lpereira:
By [calculating the angle the direction vector is pointing to](http://gamedev.stackexchange.com/a/49300), it's possible to convert the vector to a cardinal direction. That might be useful, although possibly as another PR.

bdilly:
Seems good overall.

edersondisouza:
v3 #348

PR #346, opened 2015-07-17
Soletta meets JS v3

ricardotk:
Differences from v2:
    - use git modules to get duktape sources ( @dorileo helped me a lot, thanks! )
    - fixed typo in warning message

DIfferences from v1:
    - start raising javascript exceptions in error cases
    - start comparing packet type pointers in `send_packet()`
    - remove useless `| 0` in `duk_def_prop()` flags
    - create auxiliary functions to process functions
    - use `strndupa()` instead of `sol_arena_strdup_slice()` in `create_js_type()`
    - small changes on samples

### DIFF
--- a/data/scripts/git-init-submodules.py
+++ b/data/scripts/git-init-submodules.py
@@ -1,0 +1,130 @@
+#!/usr/bin/env python3
+
+# This file is part of the Soletta Project
+#
+# Copyright (C) 2015 Intel Corporation. All rights reserved.
+#
+# Redistribution and use in source and binary forms, with or without
+# modification, are permitted provided that the following conditions
+# are met:
+#
+#   * Redistributions of source code must retain the above copyright
+#     notice, this list of conditions and the following disclaimer.
+#   * Redistributions in binary form must reproduce the above copyright
+#     notice, this list of conditions and the following disclaimer in
+#     the documentation and/or other materials provided with the
+#     distribution.
+#   * Neither the name of Intel Corporation nor the names of its
+#     contributors may be used to endorse or promote products derived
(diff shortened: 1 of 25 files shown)


data/scripts/git-init-submodules.py
```diff
@@ -0,0 +1,125 @@
...
+                                         shell=True, universal_newlines=True)
+        return output.replace("\n", "").strip(), True
+    except subprocess.CalledProcessError as e:
+        return e.output, False
+
+def submodule_clone(name, url, branch, depth, dest, sparserules):
+    if os.path.exists(dest):
+        return None, True
+
+    print("GIT: clonning submodule %s ..." % name, end="")
```
lpereira: `s/clonning/cloning/`

data/scripts/git-init-submodules.py
```diff
@@ -0,0 +1,125 @@
...
+            config.write(fd)
+
+        content = ""
+        for ln in sparserules.split():
+            content += "%s\n" % ln
+
+        with open(rules_file, "w") as fd:
+            fd.write(content)
+
+    run_command("git submodule init .")
```
lpereira: Does that `.` mean that the script must be run from a particular path? If so, please assert that before doing anything else.

dorileo: That means submodule command will consider all the submodules, seeing that now, I think we should do that out of submodule_clone() call, @ricardotk could you adjust that?

ricardotk: Yes, sure.

ricardotk:
If no other issue raise here I'll push it tomorrow with the proper fixes.

PR #348, opened 2015-07-20
Lsm303 v3

edersondisouza:
v3:
Sending error packets when failing to read sensors.

v2:


src/modules/flow/accelerometer/adxl45.c
```diff
@@ -0,0 +1,401 @@
...
+ * http://www.analog.com/static/imported-files/data_sheets/ADXL345.pdf
+ */
+struct accelerometer_adxl345_data {
+    struct sol_flow_node *node;
+    struct sol_i2c *i2c;
+    struct sol_timeout *timer;
+    double reading[3];
+    uint8_t init_power;
+    unsigned init_sampling_cnt;
+    unsigned pending_ticks;
```
lpereira: Might be a good idea to move `init_power` after the two `unsigned` values to avoid having a 3-byte hole in the `struct`.

src/modules/flow/accelerometer/adxl45.c
```diff
@@ -0,0 +1,401 @@
...
+    num_samples_available = fifo_status & 0x3F;
+
+    if (!num_samples_available) {
+        SOL_INF("No samples available");
+        return;
+    }
+
+    SOL_DBG("%d samples available", num_samples_available);
+
+    {
```
lpereira: Maybe move this to a function?

src/modules/flow/accelerometer/adxl45.c
```diff
@@ -0,0 +1,401 @@
...
+        return false;
+    }
+
+    r = sol_i2c_write_register(mdata->i2c, ACCEL_REG_BW_RATE, &value, 1);
+    if (!r) {
+        SOL_WRN("could not set ADXL345 accel sensor's sampling rate");
+        return false;
+    }
+
+    if (accel_timer_resched(mdata, ACCEL_INIT_STEP_TIME,
```
lpereira: The first and last parameter of this function is always the same; since it's a `static` function, can't the last parameter be removed?

src/modules/flow/accelerometer/adxl45.c
```diff
@@ -0,0 +1,401 @@
...
+
+    r = sol_i2c_write_register(mdata->i2c, ACCEL_REG_POWER_CTL,
+        &mdata->init_power, 1);
+    if (!r) {
+        SOL_WRN("could not set ADXL345 accel sensor's power mode");
+        return false;
+    }
+
+    if (mdata->init_power == 0x00)
+        mdata->init_power = 0xff;
```
lpereira: I would prefer if these magic values were constants.

edersondisouza:
@lpereira  Thanks for the review. Even though I believe they are valid points, adxl45.c is actually old code that was moved to a new file - should I change it?

src/modules/flow/compass/compass.c
```diff
@@ -0,0 +1,149 @@
...
+    double heading;
+    bool has_accel : 1;
+    bool has_mag : 1;
+};
+
+// Normalize a value on an arbitrary range to [-1, 1] range
+static double
+_normalize(double value, double min, double max)
+{
+    return -1.0 + ((value - min) / (max - min)) * (1.0 - (-1.0));
```
lpereira: Writing this as `static double _normalize(double value, double min, double max) { return ((value - min) * 2.0f) / (max - min) - 1.0f; }` is maybe more readable and with lesser instructions, but I'm not really good with floating point.

lpereira:
[**@edersondisouza**](https://github.com/edersondisouza) commented on [Jul 20, 2015, 4:19 PM EDT](https://github.com/solettaproject/soletta/pull/348#issuecomment-123018858 "2015-07-20T20:19:54Z - Replied by Github Comment Enhancer"):
> @lpereira  Thanks for the review. Even though I believe they are valid points, adxl45.c is actually old code that was moved to a new file - should I change it?

Yes, please -- but change in a different commit.

edersondisouza:
v4 on #361

PR #349, opened 2015-07-20
Build random cleanups v2

dorileo:


tools/build/Makefile.vars
```diff
@@ -201,17 +196,8 @@ COMMON_CFLAGS += -include $(abspath $(KCONFIG_AUTOHEADER))
 COMMON_CFLAGS += -include $(abspath $(DEFINITIONS_H))
 COMMON_CFLAGS += $(addprefix -I,$(HEADERDIRS))
 
-## Reduce .eh_frame size.
-COMMON_CFLAGS += -fno-asynchronous-unwind-tables
```
lpereira: Why is this being removed? This reduces the binary size by about 10%.

dorileo:
v3 coming...

tools/build/Makefile.vars
dorileo: look the commit message, it was already moved to dependency-resolver.

PR #352, opened 2015-07-20
Build random cleanups v5

dorileo:
v4:
   + added patch: build: add ldflags check

v5:
   + added patch: dep-resolver: fix flags order on compile_test()


dorileo:
v6 coming...

PR #355, opened 2015-07-20
Build random cleanups v6

dorileo:
v4:
   + added patch: build: add ldflags check

v5:
   + added patch: dep-resolver: fix flags order on compile_test()

v6:
  + added patch "build: rename kdbus to sol_bus" - brought from PR #183


vcgomes:
Looks good. And it works. :+1:

dorileo:
applied, thanks.

PR #356, opened 2015-07-20
Build selective hardware v3

dorileo:
# Changes
v2:
  + fixed indentation;
  + changed the config variables from HARDWARE_ to USE_;
  + pinmux now depends on gpio;
  + changed labels and help according to @lpereira and @mbelluzzo suggestions;
  + added a patch to rename HAVE_PIN_MUX to USE_PIN_MUX so we don't spread the mixed and wrong use of HAVE_ variables;

v3:
  + upstream rebase;
  + dropped patch (since it moved to #355): "build: rename kdbus to sol_bus"
  + dropped patch (it's not required anymore due the - new - networking config options): "	build: general network configuration"

# Rationale

### DIFF
--- a/src/lib/common/Kconfig
+++ b/src/lib/common/Kconfig
@@ -311,6 +311,10 @@ menu "Hardware Options"
 config USE_PWM
 	bool "PWM Support"
 	default y
+
+config USE_SPI
+	bool "SPI Support"
+	default y
 endmenu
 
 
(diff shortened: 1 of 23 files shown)


anselmolsm:
v3 LGTM
@mbelluzzo, @lpereira , anything else?

anselmolsm:
FYI, internal bot is complaining about check-style, 24f1cb9 . Semaphore still not checking this properly...

dorileo:
pushing a v4 to trigger the build bots since I got a smooth rebase and this series was acked over and over...

PR #358, opened 2015-07-21
Runner: Simplify initialization

mbelluzzo:
By adding extra dereference in packet_type is possible to turn
'input_nodes' and 'output_nodes' back to 'static const' structures,
simplifying the initialization and eliminating the need for asserts.

The main reason for this patch was to remove the 'static_assert'
calls because it's not supported by all target platforms (i.e. Galileo).

### DIFF
--- a/src/bin/sol-fbp-runner/runner.c
+++ b/src/bin/sol-fbp-runner/runner.c
@@ -32,7 +32,6 @@
 
 #include <errno.h>
 #include <libgen.h>
-#include <assert.h>
 
 #include "sol-file-reader.h"
 #include "sol-flow-parser.h"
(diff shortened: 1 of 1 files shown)


lpereira:
`static_assert()` is not platform-dependent: it's actually defined in C11, and the toolchain is possibly old enough that it's not available.

It might be a better idea to define `static_assert()` using the old trick of trying to calculate the `sizeof()` of an array with negative size if an assertion fails. Non-tested piece of code that might do the trick (and if it does or if it gets fixed, it should be moved to `sol-missing.h`):

    #if !defined(__USE_ISOC11) && !defined(__cplusplus) && !defined(static_assert)
    #  define static_assert(cond_, msg_)    sizeof(char arr[!!(cond_) - 1])
    #endif

The error message is a little bit cryptic.

mbelluzzo:
You are right about C11 part. When I said that is not support I meant not supported by the toolchain.

But I don't think that is a better idea to keep the old code, I think that this one is good and we also should add static_assert to sol-missing if we end up needing it again.

lgywata:
LGTM.

lpereira:
[**@mbelluzzo**](https://github.com/mbelluzzo) commented on [Jul 21, 2015, 10:32 AM EDT](https://github.com/solettaproject/soletta/pull/358#issuecomment-123340834 "2015-07-21T14:32:40Z - Replied by Github Comment Enhancer"):

> But I don't think that is a better idea to keep the old code, I think that this one is good and we also should add static_assert to sol-missing if we end up needing it again.

Seems like a good compromise.

PR #359, opened 2015-07-21
nodes: AM2315 temperature/humidity sensor

edersondisouza:

### DIFF
--- a/src/modules/flow/am2315/am2315.c
+++ b/src/modules/flow/am2315/am2315.c
@@ -1,0 +1,224 @@
+/*
+ * This file is part of the Soletta Project
+ *
+ * Copyright (C) 2015 Intel Corporation. All rights reserved.
+ *
+ * Redistribution and use in source and binary forms, with or without
+ * modification, are permitted provided that the following conditions
+ * are met:
+ *
+ *   * Redistributions of source code must retain the above copyright
+ *     notice, this list of conditions and the following disclaimer.
+ *   * Redistributions in binary form must reproduce the above copyright
+ *     notice, this list of conditions and the following disclaimer in
+ *     the documentation and/or other materials provided with the
+ *     distribution.
+ *   * Neither the name of Intel Corporation nor the names of its
+ *     contributors may be used to endorse or promote products derived
+ *     from this software without specific prior written permission.
+ *
+ * THIS SOFTWARE IS PROVIDED BY THE COPYRIGHT HOLDERS AND CONTRIBUTORS
+ * "AS IS" AND ANY EXPRESS OR IMPLIED WARRANTIES, INCLUDING, BUT NOT
+ * LIMITED TO, THE IMPLIED WARRANTIES OF MERCHANTABILITY AND FITNESS FOR
+ * A PARTICULAR PURPOSE ARE DISCLAIMED. IN NO EVENT SHALL THE COPYRIGHT
+ * OWNER OR CONTRIBUTORS BE LIABLE FOR ANY DIRECT, INDIRECT, INCIDENTAL,
+ * SPECIAL, EXEMPLARY, OR CONSEQUENTIAL DAMAGES (INCLUDING, BUT NOT
+ * LIMITED TO, PROCUREMENT OF SUBSTITUTE GOODS OR SERVICES; LOSS OF USE,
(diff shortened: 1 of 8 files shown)


src/modules/flow/am2315/am2315.c
```diff
@@ -0,0 +1,224 @@
...
+    }
+
+    // Write a message to read data
+    if (!sol_i2c_write(device->i2c, write_msg, sizeof(write_msg))) {
+        SOL_WRN("Could not read sensor");
+        return false;
+    }
+
+    // Datasheet asks for some delay
+    usleep(2000);
```
lpereira: Not a good idea to use a blocking delay here. Schedule a delay using the main loop.

src/modules/flow/am2315/am2315.c
```diff
@@ -0,0 +1,224 @@
...
+    read_data_size = sol_i2c_read(device->i2c, read_data, sizeof(read_data));
+
+    if (read_data_size != sizeof(read_data)) {
+        SOL_WRN("Could not read sensor data");
+        return false;
+    }
+
+    // CRC from first byte to 6th. iow, do not calculate CRC of CRC bytes
+    crc = _crc16(read_data, 6);
+    if (read_data[0] != AM2315_READ_REG || read_data[1] != AM2315_READ_LEN
```
lpereira: I'd rather split the checks in two blocks: the CRC verification (right after the calculation), and then checking if it's a read with size. This is slightly more readable (one-line conditions are nicer), and the error message can be more contextual ("CRC16 mismatch", etc).

src/modules/flow/am2315/am2315.c
```diff
@@ -0,0 +1,224 @@
...
+
+    return true;
+}
+
+bool
+am2315_read_temperature(struct am2315 *device, float *temperature)
+{
+    if (!_update_readings(device)) return false;
+
+    *temperature = device->temperature / 10.0;
```
lpereira: Instead of hardcoding `10.0`, make it a constant: `TEMPERATURE_SCALE`.

src/modules/flow/am2315/am2315.c
```diff
@@ -0,0 +1,224 @@
...
+
+    device = calloc(1, sizeof(struct am2315));
+    if (!device)
+        goto fail;
+
+    device->i2c = i2c;
+    device->bus = bus;
+    device->slave = slave;
+
+    sol_ptr_vector_append(&devices, device);
```
lpereira: Any reason to not use a `sol_vector` directly?

edersondisouza: Hmmm... Not really, I've just thought that, as I had a pointer do `device`, `sol_ptr_vector` would make more sense. I'll look into sol_vector now =D

edersondisouza: Hmmm... From sol-vector.h:

>(...)pointers returned from sol_vector_get() and sol_vector_append() should be considered invalid after the vector size is modified.

This is a problem: I return a pointer to `am2315_device` that nodes use as reference to device, and it's not a good idea if this pointer is messed up with.

ibriano: On Wed, 22 Jul 2015 10:54:13 -0700, Ederson de Souza wrote:
> > +
> > +    i2c = sol_i2c_open(bus, SOL_I2C_SPEED_10KBIT);
> > +    if (!i2c) {
> > +        SOL_WRN("Failed to open i2c bus");
> > +        return NULL;
> > +    }
> > +
> > +    device = calloc(1, sizeof(struct am2315));
> > +    if (!device)
> > +        goto fail;
> > +
> > +    device->i2c = i2c;
> > +    device->bus = bus;
> > +    device->slave = slave;
> > +
> > +    sol_ptr_vector_append(&devices, device);
> 
> Hmmm... From sol-vector.h:
> 
> >(...)pointers returned from sol_vector_get() and sol_vector_append() should be considered invalid after the vector size is modified.
> 
> This is a problem: I return a pointer to `am2315_device` that nodes use as reference to device, and it's not a good idea if this pointer is messed up with.
> 

Yes, the general rule is: If you need a pointer that others will need to
use, malloc() it and use sol_ptr_vector. If all you care is to have a
list of things that will always be used through that list, avoid the
malloc and use sol_vector.

edersondisouza:
v2 #365

PR #360, opened 2015-07-21
UART: Simplify API V3

zehortigoza:
Changes from V2:
- Change API to be like the API in GPIO and PWM.

### DIFF
--- a/src/lib/io/include/sol-uart.h
+++ b/src/lib/io/include/sol-uart.h
@@ -86,7 +86,7 @@ struct sol_uart_config {
     enum sol_uart_data_bits data_bits;
     enum sol_uart_parity parity;
     enum sol_uart_stop_bits stop_bits;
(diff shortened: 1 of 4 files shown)


mbelluzzo:
Plz, squash everything in a single commit since you are doing most of the subsequent changes in the changes you have introduced in the first commit.

zehortigoza:
The 3th is not changing in any code changed on the first 2 patches, I agree in squash the 1th and 2th but others are not related.

mbelluzzo:
Deal! So squash the first and second and add a decent description to the fourth explaining what was improved in the sample. And plz don't forget to change the tables to "static const" and ad at least a minimal documentation that is something we are trying to improve in Soletta right now.

Besides these minors, it's a nice refactor ;-) thanks!

zehortigoza:
New version https://github.com/solettaproject/soletta/pull/373

PR #361, opened 2015-07-21
Lsm303 v4

edersondisouza:
v4:
Simplifying calculation;
Addressed issues pointed by @lpereira on adxl45.c file

v3:
Sending error packets when failing to read sensors.

v2:


src/modules/flow/compass/compass.c
```diff
@@ -0,0 +1,149 @@
...
+
+    pitch = asin(-mdata->accel.x);
+    if (!sol_drange_val_equal(fabs(pitch), M_PI / 2))
+        roll = asin(mdata->accel.y / cos(pitch));
+    else
+        roll = 0.0;
+
+    mx = mdata->mag.x * cos(pitch) + mdata->mag.z * sin(pitch);
+    my = mdata->mag.x * sin(roll) * sin(pitch) + mdata->mag.y * cos(roll) - mdata->mag.z * sin(roll) * cos(pitch);
+    mz = (-mdata->mag.x) * cos(roll) * sin(pitch) + mdata->mag.y * sin(roll) + mdata->mag.z * cos(roll) * cos(pitch);
```
lpereira: Will the compiler will optimize out all these calls to trig functions, or the results will be cached since `sin()` and `cos()`  are pure?

src/modules/flow/accelerometer/lsm303.c
```diff
@@ -0,0 +1,189 @@
...
+
+#include "accelerometer-gen.h"
+
+#include "sol-i2c.h"
+#include "sol-mainloop.h"
+#include "sol-util.h"
+#include "sol-flow-internal.h"
+
+// LSM303DHLC accelerometer
+// http://www.adafruit.com/datasheets/LSM303DLHC.PDF
```
lpereira: `/* ... */` comments are used in Soletta. Patch series LGTM, but please fix these before merging, no need to issue another PR.

src/modules/flow/compass/compass.c
edersondisouza: Simple test I've done showed that the gcc assembly generated for something in this style, after -O1, kind of cached the values. I've even found out the existence of a 'sincos' instruction =D 
Btw, is your point here that I should save these values instead? Or leave it to the compiler?

lpereira: The compiler seems to be doing a good job, leave it as is.

edersondisouza:
Integrated with code style fixes.

PR #362, opened 2015-07-21
Fixes

mbelluzzo:
Glib says hi to conffile!

Coverity Issue 126578 should be gone.

And tomorrow I'm sure coverity will surprise us again :-)

### DIFF
--- a/src/lib/flow/Kconfig
+++ b/src/lib/flow/Kconfig
@@ -13,7 +13,7 @@ config NODE_DESCRIPTION
 
 config RESOLVER_CONFFILE
 	bool "Resolver conffile"
-	depends on NODE_DESCRIPTION
+	depends on NODE_DESCRIPTION && HAVE_GLIB
 	default y
 
 config INSPECTOR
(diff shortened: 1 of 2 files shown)


src/lib/flow/Kconfig
```diff
@@ -13,7 +13,7 @@ config NODE_DESCRIPTION
 
 config RESOLVER_CONFFILE
 	bool "Resolver conffile"
-	depends on NODE_DESCRIPTION
+	depends on NODE_DESCRIPTION && HAVE_GLIB
```
lpereira: Why?

mbelluzzo: Because resolver-conffile still depends on glib :-(

ulissesf: Ok, but can't we then submit a patch to remove the dependency instead of adding it back again?

mbelluzzo: Why have the build broken until then? The patch should fix resolver-conffile and remove this line. Isn't an one line patch and is breaking galileo build that is being used by QA

PR #364, opened 2015-07-21
shared: Fix conffile memory leak

vcgomes:
On the usage of asprintf(3) here, 'filename' is leaking, both in the
success branch and in the failure branch.

### DIFF
--- a/src/shared/sol-conffile.c
+++ b/src/shared/sol-conffile.c
@@ -299,12 +299,14 @@ sol_conffile_load_json_from_dirs(const char *file, char **full_path, struct sol_
             *file_reader = sol_file_reader_open(file);
             if (*file_reader) {
                 config_file_contents = sol_file_reader_get_all(*file_reader);
-                if (config_file_contents.len != 0)
+                if (config_file_contents.len != 0) {
+                    free(filename);
                     break;
-
+                }
                 sol_file_reader_close(*file_reader);
                 *file_reader = NULL;
             }
+            free(filename);
         }
     }
 


lpereira:
LGTM.

anselmolsm:
:+1:

mbelluzzo:
The problem with filename is a little worst than that. 'filename' is not being used at all after 'asprintf', tha you can think: well, lets just remove it. Than you realize that it should be being used, so lets just use it! Ooopss if you just do that will break conffile because search dirs are being used in wrong way. :-(

vcgomes:
@mbelluzzo Seems you looked at this code for longer than I, so I will drop this PR, and let you fix it ;-) Deal?

mbelluzzo:
Being very honest, given the number of problems we are having with this
code in coverity and such. We should pick someone to refactor this code
with care, it has more problems than expected. Unfortunately my backlog is
far from being empty and with stuff that I'm the only one looking into it.
So if nobody can spend time to take care of this I can queue it, but will
take a while until I can put my hands on this code.

For this pull request, I do recommend to merge a version removing filename
completely until a new version arrives instead of just call free on it.

vcgomes:
I can try to take a look at this, as I am trying to get rid of the glib usage when possible. But this is kind of a side project.

Anyway, I am fine with @mbelluzzo suggestion of removing the filename handling.

@tcanabrava Looking at the log seems that you were the last one to work more extensively on this. Any comments?

lpereira:
[**@mbelluzzo**](https://github.com/mbelluzzo) commented on [Jul 22, 2015, 11:11 AM GMT-3](https://github.com/solettaproject/soletta/pull/364#issuecomment-123735774 "2015-07-22T14:11:08Z - Replied by Github Comment Enhancer"):
> Being very honest, given the number of problems we are having with this
> code in coverity and such. We should pick someone to refactor this code
> with care, it has more problems than expected. 

I agree.

tcanabrava:
I tried to follow the old code and recreate things for json in a sane way - but keeping the old behavior. This wasn't a smart move, I'll try to recreate this code from scratch.

vcgomes:
PR https://github.com/solettaproject/soletta/pull/370 (or its sucessors) will solve this problem. Closing this.

PR #365, opened 2015-07-22
nodes: AM2315 temperature/humidity sensor - v2

edersondisouza:
v2:
Use constant for scale factor;
Fine grained messages for user when getting invalid readings;
Make delay asynchronous - this involved a lot of work from previous version, but I hope everything should be ok.


edersondisouza:
Anyone?

edersondisouza:
v3 on #420

PR #366, opened 2015-07-23
lsm303: Fix switch clause on accelerometer/lsm303

edersondisouza:
It was missing a 'break'. Found by coverity.


lpereira:
Argh, missed that one.

LGTM.

PR #368, opened 2015-07-23
flow: Remove GHashTable usage from resolver-conffile

vcgomes:
This is one of the last usages of glib in the core library. Removing
this hash table usage should have little effect, as the number of
entries is the number of different external node types used.

It should be noted that atexit() is still used, and it should have no
adverse effects even if it the callback is called multiple times.

### DIFF
--- a/src/lib/flow/sol-flow-resolver-conffile.c
+++ b/src/lib/flow/sol-flow-resolver-conffile.c
@@ -30,6 +30,7 @@
  * OF THIS SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
  */
 
+#include <assert.h>
 #include <dlfcn.h>
 #include <glib.h>
 
@@ -77,26 +78,32 @@ resolve_module_type_by_component(const char *component, void (*foreach)(bool (*c
     return ctx.found;
 }
 
-static GHashTable *resolver_conffile_dlopens = NULL;
 struct resolver_conffile_dlopen {
     char *name;
     void *handle;
     void (*foreach)(bool (*cb)(void *data, const struct sol_flow_node_type *type), const void *data);
 };
 
+static struct sol_vector resolver_conffile_dlopens = SOL_VECTOR_INIT(struct resolver_conffile_dlopen);
+
 static void
 resolver_conffile_dlopen_free(struct resolver_conffile_dlopen *entry)
 {
     /* do no dlclose() as some modules may crash due hanging references */
     free(entry->name);
-    free(entry);
 }
 
 static void
 resolver_conffile_clear_data(void)
 {
-    g_hash_table_destroy(resolver_conffile_dlopens);
-    resolver_conffile_dlopens = NULL;
+    struct resolver_conffile_dlopen *entry;
+    int i;
+
+    SOL_VECTOR_FOREACH_IDX (&resolver_conffile_dlopens, entry, i) {
+        free(entry->name);
+    }
+
+    sol_vector_clear(&resolver_conffile_dlopens);
 }
 
 const char MODULE_NAME_SEPARATOR = '/';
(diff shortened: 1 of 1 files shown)


src/lib/flow/sol-flow-resolver-conffile.c
```diff
@@ -77,26 +78,32 @@ resolve_module_type_by_component(const char *component, void (*foreach)(bool (*c
...
 static void
 resolver_conffile_clear_data(void)
 {
-    g_hash_table_destroy(resolver_conffile_dlopens);
-    resolver_conffile_dlopens = NULL;
+    struct resolver_conffile_dlopen *entry;
+    int i;
+
+    SOL_VECTOR_FOREACH_IDX (&resolver_conffile_dlopens, entry, i) {
+        free(entry->name);
```
barbieri: why not use resolver_conffile_dlopen_free()?

src/lib/flow/sol-flow-resolver-conffile.c
```diff
@@ -77,26 +78,32 @@ resolve_module_type_by_component(const char *component, void (*foreach)(bool (*c
...
-    free(entry);
 }
 
 static void
 resolver_conffile_clear_data(void)
 {
-    g_hash_table_destroy(resolver_conffile_dlopens);
-    resolver_conffile_dlopens = NULL;
+    struct resolver_conffile_dlopen *entry;
+    int i;
```
barbieri: unsigned, since vector uses unsigned for 'len'

src/lib/flow/sol-flow-resolver-conffile.c
```diff
@@ -112,6 +119,36 @@ get_module_for_type(const char *type)
     return SOL_STR_SLICE_STR(type, sep -  type);
 }
 
+
+static int
+find_entry_by_match(struct sol_vector *entries,
+    const struct resolver_conffile_dlopen *match)
+{
+    struct resolver_conffile_dlopen *entry;
+    int i;
```
barbieri: unsigned

src/lib/flow/sol-flow-resolver-conffile.c
```diff
@@ -112,6 +119,36 @@ get_module_for_type(const char *type)
     return SOL_STR_SLICE_STR(type, sep -  type);
 }
 
+
+static int
+find_entry_by_match(struct sol_vector *entries,
+    const struct resolver_conffile_dlopen *match)
+{
+    struct resolver_conffile_dlopen *entry;
```
barbieri: const

src/lib/flow/sol-flow-resolver-conffile.c
```diff
@@ -112,6 +119,36 @@ get_module_for_type(const char *type)
...
+    }
+
+    return -1;
+}
+
+static struct resolver_conffile_dlopen *
+find_entry_by_name(struct sol_vector *entries, const char *name)
+{
+    struct resolver_conffile_dlopen *entry;
+    int i;
```
barbieri: const entry, unsigned i.

src/lib/flow/sol-flow-resolver-conffile.c
```diff
@@ -142,19 +169,19 @@ _resolver_conffile_get_module(const char *type)
...
         goto found;
     }
 
-    entry = calloc(1, sizeof(*entry));
-    if (!entry) {
-        SOL_DBG("Could not alloc memory for entry");
-        free(name);
-        return NULL;
+    if (resolver_conffile_dlopens.len == 0) {
+        atexit(resolver_conffile_clear_data);
```
barbieri: man atexit() says that if registered multiple times, it is called once per registration. Then do not use atexit() here, rather do once at the beginning.

barbieri: errr, we do not delete, then len == 0 is that "only once" :-)

src/lib/flow/sol-flow-resolver-conffile.c
```diff
@@ -179,21 +206,18 @@ _resolver_conffile_get_module(const char *type)
...
     ret = resolve_module_type_by_component(type, entry->foreach);
-    SOL_NULL_CHECK_GOTO(ret, wipe_entry);
-
+    SOL_NULL_CHECK_GOTO(ret, error);
     return ret;
 
 error:
+    /* In case 'entry' was not the last one to be added. */
+    index = find_entry_by_match(&resolver_conffile_dlopens, entry);
+    assert(index >= 0);
```
barbieri: I don't get why do we need this.

vcgomes: In the case that there's an associated entry but 'resolve_module_type_by_component()' fails.

src/lib/flow/sol-flow-resolver-conffile.c
vcgomes: If I make 'entry' const in this context I will need to drop its constness, I think it's better to leave this as it is.

vcgomes:
Created PR https://github.com/solettaproject/soletta/pull/369 , with the comments that I didn't respond to fixed. Thanks.

PR #369, opened 2015-07-23
flow: Remove GHashTable usage from resolver-conffile

vcgomes:
Changes from v1:
 - Fixed review comments;

Original message:

This is one of the last usages of glib in the core library. Removing
this hash table usage should have little effect, as the number of
entries is the number of different external node types used.

It should be noted that atexit() is still used, and it should have no
adverse effects even if it the callback is called multiple times.

### DIFF
--- a/src/lib/flow/sol-flow-resolver-conffile.c
+++ b/src/lib/flow/sol-flow-resolver-conffile.c
@@ -30,6 +30,7 @@
  * OF THIS SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
  */
 
+#include <assert.h>
 #include <dlfcn.h>
 #include <glib.h>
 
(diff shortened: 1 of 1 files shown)


barbieri:
+1

src/lib/flow/sol-flow-resolver-conffile.c
```diff
@@ -112,6 +119,36 @@ get_module_for_type(const char *type)
...
+}
+
+static struct resolver_conffile_dlopen *
+find_entry_by_name(struct sol_vector *entries, const char *name)
+{
+    struct resolver_conffile_dlopen *entry;
+    unsigned int i;
+
+    SOL_VECTOR_FOREACH_IDX (entries, entry, i) {
+        if (streq(name, entry->name))
```
lpereira: While this is fine, I wonder if we need a dynamic string table. (Which could use a linear search for the time being, of course.)

lpereira:
:+1:

PR #370, opened 2015-07-24
Rewrite ~ from scratch for the conffile handling

tcanabrava:
The old config file handling made a few assumptions that I overloocked
regarding to how it managed the configuration files. This new version
has the following adjustments:

- Only register the clear_data function to be called atexit once
- Do not free / reload the entry vector for each different node
- Shorter function names for internal functions
- More carefull memory handling
- Keep a cache of different files that where loaded
- Do not use intermediary vector step to fill the entry_vector
- Sub-includes and fallbacks working.

### DIFF
--- a/src/shared/sol-conffile.c
+++ b/src/shared/sol-conffile.c
@@ -43,7 +43,8 @@
 #include "sol-vector.h"
 #include "sol-json.h"
 
-static struct sol_ptr_vector sol_conffile_entry_vector;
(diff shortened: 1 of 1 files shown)


cmarcelo:
-1 (which means: there are things to be looked at or fixed, so don't land this version)

Made some comments, please take a look later.

One suggestion: in the future, if there's a rename that's unrelated to the other changes (but affects many lines of code), make it as a separate patch. This reduces the noise in the real diff where the code happens.

tcanabrava:
closing for v2

PR #371, opened 2015-07-24
Simple

lgywata:


src/bin/sol-fbp-generator/main.c
```diff
@@ -580,11 +582,21 @@ generate_node_type_assignments(const struct fbp_data *data)
 static bool
 generate_create_type_function(struct fbp_data *data)
 {
+    char *function;
+
+    if (args.root_type_function)
+        function = strdup(args.root_type_function);
+    else
+        asprintf(&function, "create_%d_%s_type", data->id, data->name);
```
lpereira: Unchecked `asprintf()`: it's undefined what happens with `function` if it returns `-1`.

src/bin/sol-fbp-generator/main.c
```diff
@@ -580,11 +582,21 @@ generate_node_type_assignments(const struct fbp_data *data)
 static bool
 generate_create_type_function(struct fbp_data *data)
 {
+    char *function;
+
+    if (args.root_type_function)
+        function = strdup(args.root_type_function);
```
lpereira: Null-check here instead of below; use `SOL_INT_CHECK()` for the `asprintf()` case.

src/bin/sol-fbp-generator/main.c
lgywata: ack.

src/bin/sol-fbp-generator/main.c
lgywata: ack. thanks.

PR #372, opened 2015-07-24
Add HTTP client and Thingspeak node types

lpereira:
This PR adds the infrastructure to write HTTP clients. The API has been thought in a way that allows the backend to be changed without any changes; it's currently implemented with CURL, and the main loops have been properly integrated.

The ThingSpeak node types are being added as an example on how to use the HTTP Client APIs. No examples are being provided because API keys are needed; although the service is free, I'm not sure if it's a good idea to publish my test keys.

Each respective commit message contains more information.

I'm looking for feedback on the API, and, of course, for possible leaks and/or race conditions in the HTTP client.

PR #375, opened 2015-07-24
Add HTTP client and Thingspeak node types

lpereira:
Changes from V1 are too many to list. Please see comments in #372 for information.

This PR adds the infrastructure to write HTTP clients. The API has been thought in a way that allows the backend to be changed without any changes; it's currently implemented with CURL, and the main loops have been properly integrated.

The ThingSpeak node types are being added as an example on how to use the HTTP Client APIs. No examples are being provided because API keys are needed; although the service is free, I'm not sure if it's a good idea to publish my test keys.

Each respective commit message contains more information.

I'm looking for feedback on the API, and, of course, for possible leaks and/or race conditions in the HTTP client.


src/lib/datatypes/sol-buffer.c
```diff
@@ -35,48 +35,69 @@
...
+    buf->used = slice.len;
+    return 0;
+}
+
+SOL_API int
+sol_buffer_append_slice(struct sol_buffer *buf, const struct sol_str_slice slice)
+{
+    int err;
+
+    /* Extra room for the ending NUL-byte. */
```
lpereira: SOL_NULL_CHECK() added locally.

src/lib/comms/sol-http-client-impl-curl.c
```diff
@@ -0,0 +1,840 @@
...
+}
+
+
+SOL_API bool
+sol_http_param_add(struct sol_http_param *params,
+    struct sol_http_param_value value)
+{
+    struct sol_http_param_value *ptr;
+
+    if (params->api_version != SOL_HTTP_PARAM_API_VERSION) {
```
lpereira: SOL_NULL_CHECK() added locally.

src/lib/comms/sol-http-client-impl-curl.c
```diff
@@ -0,0 +1,840 @@
...
+    }
+
+    memcpy(ptr, &value, sizeof(value));
+    return true;
+}
+
+SOL_API void
+sol_http_param_free(struct sol_http_param *params)
+{
+    if (params->api_version != SOL_HTTP_PARAM_API_VERSION) {
```
lpereira: SOL_NULL_CHECK() added locally.

src/lib/comms/sol-http-client-impl-curl.c
```diff
@@ -0,0 +1,840 @@
...
+        [SOL_HTTP_METHOD_POST] = CURLOPT_HTTPPOST,
+        [SOL_HTTP_METHOD_HEAD] = CURLOPT_NOBODY,
+    };
+    struct sol_http_param_value *value;
+    struct sol_arena *arena;
+    struct curl_slist *headers = NULL;
+    CURL *curl;
+    uint16_t idx;
+
+    if (!strstartswith(base_uri, "http://")
```
lpereira: SOL_NULL_CHECK() added locally.

src/lib/comms/sol-http-client-impl-curl.c
```diff
@@ -0,0 +1,840 @@
...
+    } else if (!timeout_ms) {
+        /* Timer expired; pump cURL immediately. */
+        return multi_perform_cb(NULL) ? 0 : -1;
+    }
+
+    return 0;
+}
+
+static bool
+http_init(void)
```
lpereira: This function will probably be called `sol_comms_init()` (to be called by `sol_init()`). Shutting down cURL every time is not a good idea as it caches DNS queries and other things. Same for `http_shutdown()`. I'll make another PR for this if this patch series is OK; otherwise, I'll submit the changes in the next revision of this PR.

src/modules/flow/thingspeak/thingspeak.json
```diff
@@ -0,0 +1,149 @@
...
+          {
+            "data_type": "string",
+            "default": "https://api.thingspeak.com/",
+            "description": "Thingspeak endpoint. Defaults to public server",
+            "name": "endpoint"
+          },
+          {
+            "data_type": "int",
+            "default": 1,
+            "description": "Poll interval in seconds. Default to 1 request/second.",
```
barbieri: we're often using milliseconds for timeouts.

src/modules/flow/thingspeak/thingspeak.json
```diff
@@ -0,0 +1,149 @@
...
+      },
+      "out_ports": [
+        {
+          "data_type": "string",
+          "description": "The next command to execute. Freeform string.",
+          "name": "OUT"
+        }
+      ],
+      "private_data_type": "thingspeak_execute_data",
+      "url": "http://solettaproject.org/doc/latest/node_types/iot/thingspeak-execute.html"
```
barbieri: although we need to review the URLs when they are alive, we're using "node_types/MODULE/NODE_TYPE", thus it would be "node_types/thingspeak/thingspeak-execute.html" (s/iot/thingspeak/)

src/modules/flow/thingspeak/thingspeak.json
```diff
@@ -0,0 +1,149 @@
...
+            "data_type": "string",
+            "default": "https://api.thingspeak.com/",
+            "description": "Thingspeak endpoint. Defaults to public server",
+            "name": "endpoint"
+          },
+          {
+            "data_type": "int",
+            "default": -1,
+            "description": "Position for this command to appear in the queue. Default is end of queue",
+            "name": "position"
```
barbieri: is this fixed? Do we need an input port to change this during runtime?

src/modules/flow/thingspeak/thingspeak.json
lpereira: I'll change that. However, to keep the number of requests/second sane, the minimum value will be 1000ms.

src/test/test.h
```diff
@@ -93,3 +94,21 @@ int test_main(struct test *start, struct test *stop, void (*reset_func)(void), i
             exit(-1);                                                   \
         }                                                               \
     } while (0)
+
+#define ASSERT_STR_EQ(str_a, str_b)                                     \
+    do {                                                                \
+        if (strcmp((char *)str_a, (char *)str_b) != 0) {                \
```
barbieri: should we cast here? Giving it an incorrect pointer would crash the test while no compiler errors were emitted. I'd also make these macros smart and check for NULL, printing proper error messages and not crashing :-)

src/modules/flow/thingspeak/thingspeak.json
lpereira: Ideally one would have one node per command to publish. The reason is that there's no easy way to create composite types in FBP without resorting to custom packet types, so having a separate port to change this in runtime would introduce race conditions.

barbieri: ok!

The race issue is something we'll address later. Likely introducing an 'ordering' node type that will receive an array port  and doing an output array port. The 'N' index is only forwarded if 'N-1' was forwarded, this for each cycle. A cycle takes inputs until the last connected array index port is received.

lpereira: I'm thinking more about something similar to Python's struct module, having pack and unpack node types, with an additional port to actually process the data.

src/test/test.h
lpereira: Fixed.

lpereira:
I'm sending another revision and closing this PR.

PR #376, opened 2015-07-25
flow: Fix possible leak in resolver-conffile

anselmolsm:
Issue pointed out by coverity.

### DIFF
--- a/src/lib/flow/sol-flow-resolver-conffile.c
+++ b/src/lib/flow/sol-flow-resolver-conffile.c
@@ -179,7 +179,10 @@ _resolver_conffile_get_module(const char *type)
     }
 
     entry = sol_vector_append(&resolver_conffile_dlopens);
-    SOL_NULL_CHECK(entry, NULL);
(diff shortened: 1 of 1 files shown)


src/lib/flow/sol-flow-resolver-conffile.c
```diff
@@ -179,7 +179,10 @@ _resolver_conffile_get_module(const char *type)
     }
 
     entry = sol_vector_append(&resolver_conffile_dlopens);
-    SOL_NULL_CHECK(entry, NULL);
+    if (!entry) {
```
barbieri: ok, but if you want to keep the warning from `SOL_NULL_CHECK()` you can use the variant `SOL_NULL_CHECK_GOTO()` and put a label to do this cleanup and return NULL

anselmolsm: ack

anselmolsm:
Applied with the change suggested by @barbieri. tks.

PR #379, opened 2015-07-27
build: bring back fbp sample

dorileo:
This patch introduces the basic infra to build the fbp based samples,
new or different requirements may appear to non-basic samples but the
infra must evolve from here.

A clean exception is the c-api that will need an special care, for that,
new patches are comming next.

### DIFF
--- a/Kconfig
+++ b/Kconfig
@@ -63,6 +63,7 @@ endmenu
 menu "Samples"
 source "src/samples/coap/Kconfig"
 source "src/samples/common/Kconfig"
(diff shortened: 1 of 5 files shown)


ricardotk:
+1

lgywata:
LGTM.

dorileo:
There's a situation with -gen.h files that still needs to be sorted out.... I guess we'll have to wait for that.

dorileo:
pull request #413 brings v2...

PR #382, opened 2015-07-27
build: install only current prefix

dorileo:
If we change PREFIX between builds the build_sysroot will be left with
more than a single prefix, the next install will copy all prefixes
previously configured. This patch changes the install target to consoder
only the current PREFIX.


dorileo:
applied with a minor commit msg change... thanks.

PR #383, opened 2015-07-28
Build ldflags to dep resolver

dorileo:


dorileo:
mixed patches .... v2 coming.

PR #384, opened 2015-07-28
build: provide ldflags to dependency resolver

dorileo:
We must provide the environment ldflags to dependency resolver so we run
the tests in a consistently with the environment.


ibriano:
Ack

dorileo:
v3 coming...

PR #389, opened 2015-07-28
Remove warning when using ASSERT_STR_EQ/NE

cmarcelo:


cmarcelo:
@lpereira

barbieri:
-1, this will hide bugs when users give an incorrect pointer.

cmarcelo:
ACK. Will make the appropriate casts in the call site when they are needed.

PR #390, opened 2015-07-28
Add FreeGeoip node type + location packet type

lpereira:
This PR adds a new packet type that holds a geographical location (latitude + longitude, as two doubles). Some services allow tagging data with location information, so this new packet type will come in handy.

As an example on how this could be used, a FreeGeoip client node type is now being provided.


src/modules/flow/freegeoip/Kconfig
```diff
@@ -0,0 +1,6 @@
+config FLOW_NODE_TYPE_FREEGEOIP
+	tristate "Node type: freegeoip"
+	depends on FLOW && HTTP_CLIENT
+	default y
+	help
+		If unsure, say Y.
```
lpereira: Argh, forgot the documentation. I'll silently fix this before merging.

data/schemas/node-type-genspec.schema
```diff
@@ -251,6 +251,24 @@
...
+      "properties": {
+        "data_type": { "enum": [ "location" ] },
+        "default": {
+          "oneOf": [
+            {
+              "type": "object",
+              "properties": {
+                "lat": { "$ref": "#/definitions/float_or_limit" },
+                "lon": { "$ref": "#/definitions/float_or_limit" }
+              },
```
barbieri: how about altitude?

src/lib/flow/include/sol-flow-packet.h
```diff
@@ -135,6 +136,11 @@ struct sol_flow_packet *sol_flow_packet_new_direction_vector_components(double x
 int sol_flow_packet_get_direction_vector(const struct sol_flow_packet *packet, struct sol_direction_vector *direction_vector);
 int sol_flow_packet_get_direction_vector_components(const struct sol_flow_packet *packet, double *x, double *y, double *z);
 
+struct sol_flow_packet *sol_flow_packet_new_location(const struct sol_location *location);
+struct sol_flow_packet *sol_flow_packet_new_location_latlon(double lat, double lon);
```
barbieri: I'd say you can move this (and likely the other packet constructors) to take the structure, not the pointer, and use those macros or inline functions to create the structure from the numbers. So:

```
sol_flow_packet_new_location(const struct sol_location location)
sol_flow_packet_new_location(*my_ptr_to_location);
sol_flow_packet_new_location(SOL_LOCATION(lat, lon));
```

src/modules/flow/freegeoip/freegeoip.c
```diff
@@ -0,0 +1,239 @@
...
+#include <sol-json.h>
+#include <sol-mainloop.h>
+#include <sol-util.h>
+
+struct freegeoip_data {
+    struct sol_flow_node *node;
+    char *endpoint;
+};
+
+#define RESPONSE_CHECK_API(response_, mdata_) \
```
barbieri: move this to sol-http.h? It's being used everywhere.

src/modules/flow/freegeoip/freegeoip.c
```diff
@@ -0,0 +1,239 @@
...
+    FIELD_CITY_NAME = 1<<4,
+    FIELD_ZIP_CODE = 1<<5,
+    FIELD_TIMEZONE = 1<<6,
+    FIELD_LATITUDE = 1<<7,
+    FIELD_LONGITUDE = 1<<8,
+    FIELD_ALL = 0x1ff
+};
+
+static bool
+json_token_to_double(const struct sol_json_token *token, double *output)
```
barbieri: argh, didn't we add the function to parse numbers from json in the sol-json.h? If not, then let's do this ASAP with two variants (int and double), and please no strndupa() in there.

src/modules/flow/freegeoip/freegeoip.c
```diff
@@ -0,0 +1,239 @@
...
+    struct sol_json_scanner scanner;
+    struct sol_json_token token, key, value;
+    enum sol_json_loop_reason reason;
+    enum fields fields = FIELD_ALL;
+    double latitude = 0.0f;
+    double longitude = 0.0f;
+
+    RESPONSE_CHECK_API(response, mdata);
+
+    if (!response->content.used) {
```
barbieri: shouldn't we send an error packet here? Maybe merge this in the following 'if' that checks for response code != 200?

src/modules/flow/freegeoip/freegeoip.c
```diff
@@ -0,0 +1,239 @@
...
+    }
+
+#define FIELD(json_field_, flow_field_) \
+    do { \
+        if (fields & FIELD_ ## flow_field_ && sol_json_token_str_eq(&key, \
+                json_field_, strlen(json_field_))) { \
+            fields &= ~(FIELD_ ## flow_field_); \
+            sol_flow_send_string_packet(mdata->node, \
+                SOL_FLOW_NODE_TYPE_LOCATION_FREEGEOIP__OUT__ ## flow_field_, \
+                strndupa(value.start + 1, value.end - value.start - 2)); \
```
barbieri: we should have a way to send a sol_str_slice... then there is no need to strndupa().

src/modules/flow/freegeoip/freegeoip.c
```diff
@@ -0,0 +1,239 @@
...
+            continue;
+        }
+    }
+
+#undef FIELD
+
+    if (reason != SOL_JSON_LOOP_REASON_OK)
+        return;
+
+    if (!(fields & (FIELD_LATITUDE | FIELD_LONGITUDE))) {
```
barbieri: since you only use 'fields' here and it's used only to send latitude or longitude, what if we initialize them with FP_NAN or even check if both are FP_ZERO?

src/modules/flow/freegeoip/freegeoip.c
lpereira: Here we can afford not using strdndupa(), but in the generic JSON, we can't as there is no warranty the input data is NUL terminated (and strto*() functions require that).

src/modules/flow/freegeoip/freegeoip.c
lpereira: Nope, it's used for all other fields. See the macro.

src/modules/flow/freegeoip/freegeoip.c
```diff
@@ -0,0 +1,239 @@
...
+    int r;
+
+    r = snprintf(json_endpoint, sizeof(json_endpoint), "%s/json/%s",
+        mdata->endpoint, addr ? addr : "");
+    if (r < 0 || r >= (int)sizeof(json_endpoint)) {
+        SOL_WRN("Could not prepare endpoint");
+        return -EINVAL;
+    }
+
+    r = sol_http_client_request(SOL_HTTP_METHOD_GET, json_endpoint,
```
barbieri: uh, oh... what if the node ends before sol_http_client_request() calls back? There is no way to kill the request... we need to allow that, I'd say return a request handle here and provide `sol_http_client_request_cancel()`

src/modules/flow/freegeoip/freegeoip.c
```diff
@@ -0,0 +1,239 @@
...
+        return r;
+
+    return query_addr(mdata, addr);
+}
+
+static void
+freegeoip_close(struct sol_flow_node *node, void *data)
+{
+    struct freegeoip_data *mdata = data;
+
```
barbieri: would need to cancel any pending requests from here.

src/modules/flow/freegeoip/freegeoip.c
barbieri: I recall I've sent a code parsing json numbers as the spec. If there was a bug, just fix that and do not use strtoul/strtod.

Also note that strtod() is locale-dependent, so in pt_BR would use ',' as decimal separator... something not allowed by JSON... likely a problem for our encoding sites as well :-(

src/modules/flow/freegeoip/freegeoip.c
barbieri: I saw the macro. But tha't s what I mean, you initialize fields to "all", clear once it's received... but there is no need for that, in the case you receive more than one field, just send them all in their ports.

The only real use is here to send the location if it was received (since you don't want to send an empty location and not send a location with only latitude or longitude).

lpereira: The main reason was to not blow the stack if the same field was sent more than once, since strndupa() was being used. However, I'm adding a primitive to send a slice, so the fields enum won't be needed anymore.

src/lib/flow/include/sol-flow-packet.h
lpereira: I'll do this but on another series. For now, I'll just make the location packet pass a `const struct sol_location`.

src/modules/flow/freegeoip/freegeoip.c
lpereira: Wouldn't be better: this one assumes flow, an ERROR port, etc. I'll use it maybe one or more times more (with other node types) and think of an alternative.

src/modules/flow/freegeoip/freegeoip.c
lpereira: I'll do this in another PR; it's in my TODO list. (The Thingspeak node types have FIXMEs in place for this.)

data/schemas/node-type-genspec.schema
lpereira: Done.

src/modules/flow/freegeoip/freegeoip.c
lpereira: I'll send a version using strtod() while your version isn't merged. I'll send a followup patch fixing other usages of strtod() in the tree.

src/modules/flow/freegeoip/freegeoip.c
lpereira: Done.

src/modules/flow/freegeoip/freegeoip.c
lpereira: Done.

src/modules/flow/freegeoip/freegeoip.c
lpereira: Done.

lpereira:
Closing PR to send new version.

PR #391, opened 2015-07-28
git-init-submodules: make sure to always complete

dorileo:
Don't ignore if git-init-submodule fails to initialize submodules or is
interrupted by the user and keep trying to complete.

For that we keep the .git/modules-ok control file to be sure if
submodule initialization process is sanitized.

We force a [re]initialization case .git/modules-ok is not found.

### DIFF
--- a/data/scripts/git-init-submodules.py
+++ b/data/scripts/git-init-submodules.py
@@ -34,13 +34,14 @@ import configparser
 import os
 import subprocess
 import sys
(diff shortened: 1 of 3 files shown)


anselmolsm:
:+1:

ricardotk:
:+1:

dorileo:
ok, applied... thanks.

PR #393, opened 2015-07-28
Revert "datatypes: Cast SOL_VECTOR_INIT() appropriately"

dorileo:
This reverts commit 75a18c9e2344fd759d509d31ff42206cb229ab7c.

If we want to use the SOL_VECTOR_INIT() macro with an objects with
static storage initialization this "cast" will be considered as a
"compound literal" resulting in compilation errors for older compilers
(even gcc).

The gcc documentation states the following:

"As a GNU extension, GCC allows initialization of objects with static
storage duration by compound literals (which is not possible in ISO C99,
because the initializer is not a constant)..."

But it seems not to be true, the versions working with that is 5.1.0
on-wards.

### DIFF
--- a/src/lib/datatypes/include/sol-vector.h
+++ b/src/lib/datatypes/include/sol-vector.h
@@ -72,7 +72,7 @@ struct sol_vector {
     uint16_t elem_size;
 };
 
-#define SOL_VECTOR_INIT(TYPE) (struct sol_vector){ .data = NULL, .len = 0, .elem_size = sizeof(TYPE) }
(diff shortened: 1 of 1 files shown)


lpereira:
Is the HTTP client building without the cast?

dorileo:
Yes, seems so, double checking....

vcgomes:
Did an "allyesconfig" build here. Building fine. (I am on the problematic case, gcc 4.9.2 here).

 :+1:

It's kind of sad when a nice patch has to be reverted because of the tools.

cmarcelo:
Applied in 3533bfdae88fd05d0bb2606fc0135cd50139b11f.

PR #394, opened 2015-07-29
test: fix warnings when using ASSERT_STR_EQ/NE

cmarcelo:
The macros expect strictly a 'char *' since it will be used as part of a
fprintf, the implict 'void *' conversion doesn't kick in, so in
test-buffer, ensure we cast appropriatedly the buffer in case of
asserts.

For the reference, GCC 5.2 was giving the following

    warning: format ‘%s’ expects argument of type ‘char *’, but argument 6 has type ‘void *’ [-Wformat=]
                 fprintf(stderr, "%s:%d: %s: Assertion string_different(\"%s\", \"%s\") failed.\n", \

### DIFF
--- a/src/test/test-buffer.c
+++ b/src/test/test-buffer.c
@@ -131,11 +131,11 @@ test_set_slice(void)
     ASSERT(err >= 0);
 
     ASSERT_INT_EQ(buf.used, strlen(backend));
-    ASSERT_STR_EQ(buf.data, backend);
+    ASSERT_STR_EQ((char *)buf.data, backend);
 
     backend[1] = 'a';
-    ASSERT_STR_NE(buf.data, backend);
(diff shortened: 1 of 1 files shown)


cmarcelo:
New version of #389. Casting in the macro might hide problems, so let the call-sites cast.

cmarcelo:
Applied in b3fe03c25ae52c6e2b9d5bff28eab7b6a67f7e63.

barbieri:
After I merged my branch simplectype we got a new macro SOL_TYPE_CHECK() in sol-types.h, this macro will take care of type checking without cast, but does return the value using the requested type, then this issue is gone and no casts are required anymore. See 38bee328f66ecce30c78cd8d514f7a44d2dc4edb

PR #397, opened 2015-07-29
build: fail (return exit 1) if no required deps are found

dorileo:
This patch forces gnu make to return exit code 1 if we can't find
required build dependencies. With that the build bots work properly and
report errors on those conditions.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -30,6 +30,7 @@ warning:
 	$(Q)echo -e "$(NOT_FOUND)"
 	$(Q)echo -e "If you've just installed it, run: make reconf"
 	$(Q)echo -e "For more information/options, run: make help"
+	(false)
 $(warning-targets)
 else
 ifeq ($(HAVE_KCONFIG_CONFIG),)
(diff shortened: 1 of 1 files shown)


cmarcelo:
Why "(false)" with parenthesis instead of "false"?

anselmolsm:
LGTM

WRT false x (false), @dorileo told me that he prefers to always use the parenthesis, because in some cases it is needed. Might not be the case here, but I understand his choice.

dorileo:
Just a matter of habit, there're some cases that without parenthesis the thing will not work as expected so I always enclose it...  it's not necessary in this case specifically, I've locally changed the patch as follows (I'll send the right patch when we get to merge it):

```patch
diff --git a/Makefile b/Makefile
index 4e37e97..7210514 100644
--- a/Makefile
+++ b/Makefile
@@ -30,6 +30,7 @@ warning:
        $(Q)echo -e "$(NOT_FOUND)"
        $(Q)echo -e "If you've just installed it, run: make reconf"
        $(Q)echo -e "For more information/options, run: make help"
+       $(Q)false
 $(warning-targets)
 else
 ifeq ($(HAVE_KCONFIG_CONFIG),)
@@ -37,6 +38,7 @@ warning: $(KCONFIG_GEN)
        $(Q)echo "You need a config file first. Please run a config target, e.g.: make menuconfig"
        $(Q)echo "For a quick default config run: make alldefconfig"
        $(Q)echo "For more information/options run: make help"
+       $(Q)false
 $(warning-targets)
 else
 include $(top_srcdir)tools/build/Makefile.targets

```

cmarcelo:
+1

dorileo:
ok, applied. thanks.

PR #406, opened 2015-07-29
Rewrite ~ from scratch for the conffile handling v2

tcanabrava:
The old config file handling made a few assumptions that I overloocked
regarding to how it managed the configuration files. This new version
has the following adjustments:

- Only register the clear_data function to be called atexit once
- Do not free / reload the entry vector for each different node
- Shorter function names for internal functions
- More carefull memory handling
- Keep a cache of different files that where loaded
- Do not use intermediary vector step to fill the entry_vector
- Sub-includes and fallbacks working.

the "need to fo a thing" typo sneaked into this commit but it's already fixed here locally, don't need to coment on that.

### DIFF
--- a/src/shared/sol-conffile.c
+++ b/src/shared/sol-conffile.c
@@ -43,7 +43,8 @@
 #include "sol-vector.h"
 #include "sol-json.h"
 
-static struct sol_ptr_vector sol_conffile_entry_vector;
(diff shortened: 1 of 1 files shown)


cmarcelo:
+1 please fix those two things I mentioned and push this so we can move forward.

Note: I'm assuming code changes asked in the other revision were done.

tcanabrava:
Commited. 
Closing.

PR #407, opened 2015-07-29
modules/gtk: use type rgb instead of string on led

bdilly:
It shouldn't be trying to parse a string to define a color
if we have a rgb option type already.

@otaviobp reported an issue thanks to that

@glima , when you have some time, please review it.


otaviobp:
I tested this patch and it has fixed the issue I reported.

bdilly:
ok, thanks @otaviobp 
merged

PR #409, opened 2015-07-31
Added code for the Max31855 sensor v5

tcanabrava:
Code is based on the UPM code, it should work as intended.
updated the code to follow up the new sol_spi signatures.

related to v4:

- fixed doing multiple calls to sol_spi_transfer
- added node to the mdata struct so we don't need to malloc/free the callback data.


tcanabrava:
closed because of v6 https://github.com/solettaproject/soletta/pull/432

PR #410, opened 2015-07-31
{float,int}-to-string conversion nodes now do formatting.

glima:
From the new documentation (integer case), verbatim:

  "default": "{val:d}",
  "description": "Format string to be used in the conversion. The syntax
                  is the same as Python's string.format() --
                  https://docs.python.org/3/library/stdtypes.html#str.format
                  --, except that we got no format recursion, no '!'
                  conversions, and the attribute names must be one of
                  the integer fields (val, min, max, step). Numbered
                  fields will get to them in this order. Naturally,
                  there is no field attribute nor field index
                  dereferencing either.",

Tests were also included.


glima:
Sending a new version with these comments fixed and with HAVE_LOCALE support.

PR #411, opened 2015-07-31
Undocument missing.h.

glima:
That file should never have been documented. It was by mistake, since it
was erroneously installed at the time.


lpereira:
:ship:

glima:
Okay, merging. Thanks.

PR #412, opened 2015-07-31
Iio

edersondisouza:


src/modules/flow/iio/iio.json
```diff
@@ -0,0 +1,108 @@
...
+  "name": "iio",
+  "meta": {
+    "author": "Intel Corporation",
+    "license": "BSD 3-Clause",
+    "version": "1"
+  },
+  "types": [
+    {
+      "category": "input/hw",
+      "description": "IIO based gyroscope input node. TODO explain the wonderful world of triggers, buffers and buffer sized 0 and -1.",
```
barbieri: TODO :-D

src/modules/flow/iio/iio.json
```diff
@@ -0,0 +1,108 @@
...
+           "data_type": "int",
+           "description": "IIO buffer size. If -1, buffering is disabled. If 0, will use default buffer size. TODO explain triggers and etc",
+           "default": 0,
+           "name": "buffer_size"
+         },
+         {
+           "data_type": "string",
+           "description": "IIO trigger name. Name of IIO trigger that should be associated to this device for buffered readings. If not set and buffer enabled, will try to use device current trigger, if any. If none, will attempt to create a sysfs trigger",
+           "name": "iio_trigger_name",
+           "default": ""
```
barbieri: you should use null, not an empty string.

src/modules/flow/iio/iio.json
```diff
@@ -0,0 +1,108 @@
...
+      ],
+      "methods": {
+        "close": "gyroscope_close",
+        "open": "gyroscope_open"
+      },
+      "name": "iio/gyroscope",
+      "options": {
+       "members": [
+         {
+           "data_type": "int",
```
barbieri: is this fixed? If it may change according to some setup, ie if user adds more devices, then could we make this a string and accept an integer (checked with strtoul()) and also resolve the number using libudev if it's available? This way we could be more resilient to configuration changes.

src/modules/flow/iio/iio.json
```diff
@@ -0,0 +1,108 @@
...
+             "default": 1,
+             "description": "Scale to be applied to y axis readings. Set use_device_scale_default to false to enable this option",
+             "name": "scale_y"
+         },
+         {
+             "data_type": "float",
+             "default": 1,
+             "description": "Scale to be applied to z axis readings. Set use_device_scale_default to false to enable this option",
+             "name": "scale_z"
+         },
```
barbieri: do we need all of those as individual numbers? Or can we declare the option as a direction-vector and use it here as a single option? Of course we'd need to add support to load direction-vector in options... but we'll need that sooner or later if we add a constant/direction-vector (@glima did you do this constant node and the option parsing?)

src/modules/flow/iio/nodes.c
```diff
@@ -0,0 +1,168 @@
...
+
+    // TODO find appropriate min & max (probably, real min e max outputs)
+    sol_flow_send_direction_vector_packet(mdata->node,
+        SOL_FLOW_NODE_TYPE_IIO_GYROSCOPE__OUT__OUT, &out);
+
+    return;
+
+error:
+    sol_flow_send_error_packet(mdata->node, EIO, errmsg);
+    SOL_WRN(errmsg);
```
barbieri: this is likely to trigger a warning in some compilers, they will require you to "%s", errmsg.

src/modules/flow/iio/nodes.c
```diff
@@ -0,0 +1,168 @@
...
+        sol_iio_set_reader_cb(mdata->device, reader_cb, mdata);
+        sol_iio_device_start_buffer(mdata->device);
+    }
+
+    mdata->node = node;
+
+    return 0;
+
+error:
+    SOL_WRN("Could not open IIO device");
```
barbieri: be more descriptive on what failed, at least say the iio_device that failed.

src/modules/flow/iio/nodes.c
```diff
@@ -0,0 +1,168 @@
...
+        // TODO find appropriate min & max (probably, real min e max outputs)
+        sol_flow_send_direction_vector_packet(node,
+            SOL_FLOW_NODE_TYPE_IIO_GYROSCOPE__OUT__OUT, &out);
+    }
+
+    return 0;
+
+error:
+    sol_flow_send_error_packet(node, EIO, errmsg);
+    SOL_WRN(errmsg);
```
barbieri: same as above, either put the message here or use "%s", errmsg in the SOL_WRN()

src/shared/sol-iio.c
```diff
@@ -0,0 +1,990 @@
...
+    struct sol_iio_device *device;
+    char *name;
+    double scale;
+    int index;
+    int offset;
+
+    int storagebits;
+    int bits;
+    int shift;
+    int offset_in_buffer;
```
barbieri: should all of these be unsigned integers?

src/shared/sol-iio.c
```diff
@@ -0,0 +1,990 @@
...
+craft_filename_path(char *path, size_t size, const char *base, ...)
+{
+    va_list args;
+    int len;
+
+    va_start(args, base);
+    len = vsnprintf(path, size, base, args);
+    va_end(args);
+
+    return len >= 0 && len <= PATH_MAX;
```
barbieri: PATH_MAX here is wrong, you should check `size` instead.

src/shared/sol-iio.c
```diff
@@ -0,0 +1,990 @@
...
+        result = false;
+
+    if (result) {
+        len = strlen(content);
+        result = write(fd, content, len) == (ssize_t)len;
+    }
+
+    free(content);
+    close(fd);
+
```
barbieri: `sol_util_write_file()` ?

src/shared/sol-iio.c
```diff
@@ -0,0 +1,990 @@
...
+{
+    int len;
+    char path[PATH_MAX + 1];
+    struct stat st;
+
+    len = snprintf(path, sizeof(path), SYSFS_TRIGGER_NOW_PATH, trigger_dir);
+
+    // TODO check permissions of file trigger_now too, not only existence
+    return len > 0 && len <= PATH_MAX && !stat(path, &st);
+}
```
barbieri: why not use the `craft_filename_path()` followed by a `stat()`?

src/shared/sol-iio.c
```diff
@@ -0,0 +1,990 @@
...
+    char path[PATH_MAX + 1];
+    char *name;
+    int len;
+    bool result;
+
+    len = snprintf(path, sizeof(path), SYSFS_TRIGGER_NAME_PATH, trigger_dir);
+    if (len < 0 || len > PATH_MAX) {
+        SOL_WRN("Could not open IIO trigger (%s) name on sysfs", trigger_dir);
+        return false;
+    }
```
barbieri: why not `craft_filename_path()`? Either use that everywhere or never, keep a single usage pattern.

src/shared/sol-iio.c
```diff
@@ -0,0 +1,990 @@
...
+
+    len = sol_util_read_file(path, "%ms", &name);
```
barbieri: `sol_util_load_file_string()`

src/modules/flow/iio/nodes.c
lpereira: It's not only a warning, but, if the user has any possibility of changing the printf() formatting string, it's also a security vulnerability. (Not a problem in this case, though, but alas.)

src/shared/sol-iio.c
```diff
@@ -0,0 +1,990 @@
...
+     * by opening all triggers on /sys/bus/iio/devices
+     * and checking name by name =/ */
+
+    dir = opendir(SYSFS_DEVICES_PATH);
+    if (!dir) {
+        SOL_WRN("No IIO devices available");
+        return false;
+    }
+
+    while (ent = readdir(dir), ent != NULL) {
```
barbieri: @lucasdemarchi  any better idea?

src/modules/flow/iio/nodes.c
```diff
@@ -0,0 +1,168 @@
...
+    struct sol_iio_channel *channel_x;
+    struct sol_iio_channel *channel_y;
+    struct sol_iio_channel *channel_z;
+    bool buffer_enabled : 1;
+};
+
+static void
+reader_cb(struct sol_iio_device *device, struct sol_iio_buffer *buffer, void *data)
+{
+    const char errmsg[] = "Could not read channel buffer values";
```
lpereira: Since this string is never changed, it might be a better idea to use pointer to a char array instead: `static const char *errmsg = "Could ..."`.

src/shared/sol-iio.c
```diff
@@ -0,0 +1,990 @@
...
+        if (strstr(ent->d_name, "trigger") == ent->d_name) {
```
barbieri: we now have `strstartswith()` in sol-util.h

src/shared/sol-iio.c
```diff
@@ -0,0 +1,990 @@
...
+    char path[PATH_MAX + 1];
+    int len;
+
+    len = snprintf(path, sizeof(path), CURRENT_TRIGGER_DEVICE_PATH, device->device_id);
+    if (len < 0 || len > PATH_MAX) {
+        SOL_WRN("Could not open device current_trigger file");
+        return false;
+    }
+
+    if (!write_file(path, "%s", trigger_name)) {
```
barbieri: `sol_util_file_write()`

src/shared/sol-iio.c
```diff
@@ -0,0 +1,990 @@
+#include <ctype.h>
```
lpereira: Missing license header?

src/shared/sol-iio.c
edersondisouza: `sol_util_write_file()` uses a `vfprintf` that was hiding some `write` errors (like -EINVAL). That's why I created this one.

src/shared/sol-iio.c
edersondisouza: Probably because I've created it at the end, and failed to replace all other usages. Will do it.

src/shared/sol-iio.c
```diff
@@ -0,0 +1,990 @@
...
+    /* Set device current trigger */
+    len = asprintf(&trigger_name, "sysfstrig%d", i);
+    if (len < 0)
+        goto error;
+    result = set_current_trigger(device, trigger_name);
+
+    if (result)
+        device->trigger_name = strdup(trigger_name);
+
+    free(trigger_name);
```
barbieri: you could change this logic to avoid mallocs():

```c
if (result)
    device->trigger_name = trigger_name;
else
   free(trigger_name);
```

src/shared/sol-iio.c
edersondisouza: Good to know.

src/shared/sol-iio.c
```diff
@@ -0,0 +1,990 @@
...
+};
+
+struct sol_iio_buffer {
+    size_t size;
+    uint8_t *data;
+};
+
+#define DEVICE_PATH "/dev/iio:device%d"
+#define SYSFS_DEVICES_PATH "/sys/bus/iio/devices"
+#define BUFFER_ENABLE_DEVICE_PATH "/sys/bus/iio/devices/iio:device%d/buffer/enable"
```
lpereira: Most of these have the same prefix (`"/sys/bus/iio/devices/"`); it could be easier to have a `#define BUS_PREFIX` with that and use it elsewhere. Might be used even for other constants, such as `CHANNEL_RAW_PATH`, `CHANNEL_SCAN_ENABLE_PATH`, and so on. Makes changing these things easier (even changing one define only for testing purposes: say you mock a filesystem with the values and change the root #define to point to that path).

src/shared/sol-iio.c
```diff
@@ -0,0 +1,990 @@
...
+    char *name = NULL;
+    int len;
+
+    len = snprintf(path, sizeof(path), CURRENT_TRIGGER_DEVICE_PATH, device->device_id);
+    if (len < 0 || len > PATH_MAX) {
+        SOL_WRN("Could not open IIO device%d trigger on sysfs", device->device_id);
+        return false;
+    }
+
+    len = sol_util_read_file(path, "%ms", &name);
```
barbieri: `sol_util_load_file_string()`

src/shared/sol-iio.c
```diff
@@ -0,0 +1,990 @@
...
+    } else
+        device->trigger_name = name;
+
+    return check_manual_triggering(device);
+}
+
+static void
+set_buffer_size(struct sol_iio_device *device, int buffer_size)
+{
+    char path[PATH_MAX + 1];
```
barbieri: you don't need +1 here.

src/shared/sol-iio.c
```diff
@@ -0,0 +1,990 @@
...
+        return false;
+
+    if (!write_file(path, "%d", enabled))
+        return false;
+
+    return true;
+}
+
+static size_t
+calc_buffer_size(struct sol_iio_device *device)
```
barbieri: const struct...

src/shared/sol-iio.c
```diff
@@ -0,0 +1,990 @@
...
+{
+    struct sol_iio_channel *channel;
+    int i;
+    size_t size = 0;
+
+    SOL_PTR_VECTOR_FOREACH_IDX (&device->channels, channel, i) {
+        size += channel->storagebits;
```
barbieri: do we need to account any kind of padding or alignment?

src/shared/sol-iio.c
lpereira: Use `readdir_r()`.

src/shared/sol-iio.c
lpereira: `if (!streq(name, "trigger")) `

src/shared/sol-iio.c
```diff
@@ -0,0 +1,990 @@
...
+    }
+
+    return size / 8;
```
barbieri: you need to round-up otherwise this will miss bits:
```c
storagebits = 3;
storagebits / 8 == 0;
required_storage_bytes = 1;
```

src/shared/sol-iio.c
```diff
@@ -0,0 +1,990 @@
...
+
+static bool
+device_reader_cb(void *data, int fd, unsigned int active_flags)
+{
+    struct sol_iio_device *device = data;
+    bool result = true;
+    size_t len = calc_buffer_size(device);
+    uint8_t *buffer = NULL;
+
+    buffer = calloc(1, len);
```
barbieri: allocate this buffer only once and keep it until close() or callback is reset.

src/shared/sol-iio.c
```diff
@@ -0,0 +1,990 @@
...
+}
+
+static bool
+set_current_trigger(struct sol_iio_device *device, const char *trigger_name)
+{
+    char path[PATH_MAX + 1];
+    int len;
+
+    len = snprintf(path, sizeof(path), CURRENT_TRIGGER_DEVICE_PATH, device->device_id);
+    if (len < 0 || len > PATH_MAX) {
```
lpereira: `len >= PATH_MAX`. You have to count for the `\0`.

src/shared/sol-iio.c
edersondisouza: `storagebits` should be the total number of bits that a channel use. `bits` is the one without paddings.

src/shared/sol-iio.c
```diff
@@ -0,0 +1,990 @@
...
+    struct sol_iio_device *device = data;
+    bool result = true;
+    size_t len = calc_buffer_size(device);
+    uint8_t *buffer = NULL;
+
+    buffer = calloc(1, len);
+    if (!buffer) {
+        SOL_WRN("Could not allocate memory for read buffer");
+        result = false;
+    }
```
barbieri: it's better to return NULL or just SOL_NULL_CHECK(buffer);

src/shared/sol-iio.c
```diff
@@ -0,0 +1,990 @@
...
+        type++;
+    } else
+        channel->is_signed = true; /* TODO check if signed is the default */
+
+    channel->shift = 0;
+    len = sscanf(type, "%d/%d>>%d", &channel->bits, &channel->storagebits, &channel->shift);
+
+    free(original_type);
+
+    if (len < 0)
```
lpereira: Only return `true` if `len` is `3`, as not all items might have been correctly converted.

src/shared/sol-iio.c
```diff
@@ -0,0 +1,990 @@
...
+    char path[PATH_MAX + 1];
+    int len;
+    char *channel_pure_name = NULL;
+    struct sol_iio_device *device = channel->device;
+
+#define GET_SCALE(_name) \
+    do { \
+        len = snprintf(path, sizeof(path), CHANNEL_SCALE_PATH, device->device_id, \
+            _name); \
+        if (len > 0 && sol_util_read_file(path, "%lf", &channel->scale) > 0) { \
```
lpereira: Not checking for snprintf() overflow.

src/shared/sol-iio.c
```diff
@@ -0,0 +1,990 @@
...
+static bool
+read_channel_index_in_buffer(struct sol_iio_channel *channel)
+{
+    struct sol_iio_device *device = channel->device;
+    char path[PATH_MAX + 1];
+    int len;
+
+    len = snprintf(path, sizeof(path), CHANNEL_SCAN_INDEX_PATH,
+        device->device_id, channel->name);
+    if (len < 0 || sol_util_read_file(path, "%d", &channel->index) < 0)
```
lpereira: Not checking for snprintf() overflow.

src/shared/sol-iio.c
```diff
@@ -0,0 +1,990 @@
...
+        if (ret < 0) {
+            if (errno == EINTR)
+                continue;
+            else if (errno == EAGAIN || errno == EWOULDBLOCK)
+                break;
+            else {
+                result = false;
+                break;
+            }
+        }
```
barbieri: we do have a code like this in `sol-util-linux.c`, but it's not exported. As this block is pretty common I'd say export it and use it here (and do other patches to use it in all other usages like sol-i2c-linux, sol-uart-linux, sol-network-linux, evdev, keyboard, unix-socket, stdin and subprocess)

src/shared/sol-iio.c
edersondisouza: Actually 2 is also valid, we may not have the `%d` after the `>>`.

src/shared/sol-iio.c
```diff
@@ -0,0 +1,990 @@
...
+sol_iio_add_channel(struct sol_iio_device *device, const char *name)
+{
+    char path[PATH_MAX + 1];
+    struct sol_iio_channel *channel;
+    struct stat st;
+    int len;
+
+    len = snprintf(path, sizeof(path), CHANNEL_RAW_PATH, device->device_id,
+        name);
+    if (len < 0 || len > PATH_MAX || stat(path, &st)) {
```
lpereira: Check if `len >= PATH_MAX`, you have to account for the terminating NUL character.

Since you're calling stat here, what happens if, during program bootup the channel is there, but is removed due to a kernel module being removed? Will the program crash? It's often a bad idea to use stat() if you plan to open the file later (and assume a bunch of things if stat() or access() do not fail).

src/shared/sol-iio.c
```diff
@@ -0,0 +1,990 @@
...
+
+    if (device->buffer_enabled) {
+        SOL_WRN("device%d is using buffer - use iio_read_buffer_channel_value() instead",
+            device->device_id);
+        return false;
+    }
+
+    len = snprintf(path, sizeof(path), CHANNEL_RAW_PATH, device->device_id,
+        channel->name);
+    if (len < 0) {
```
lpereira: snprintf() overflow not checked.

src/shared/sol-iio.h
```diff
@@ -0,0 +1,194 @@
...
+ */
+
+struct sol_iio_device;
+struct sol_iio_channel;
+struct sol_iio_buffer;
+
+/**
+ * Callback to read data from buffer
+ */
+typedef void (*sol_iio_reader_cb)(struct sol_iio_device *device, struct sol_iio_buffer *buffer, void *data);
```
barbieri: 1 - we're not doing typedefs;
2 - void *data should be the first parameter so it matches other usages;
3 - buffer should be const, we're not allowing users to change it.

src/shared/sol-iio.c
```diff
@@ -0,0 +1,990 @@
...
+
+    len = snprintf(path, sizeof(path), DEVICE_PATH, device->device_id);
+    if (len < 0 || len > PATH_MAX) {
+        SOL_WRN("Could not open IIO device%d: Device does not exist", device->device_id);
+        return false;
+    }
+
+    fd = open(path, O_RDONLY | O_CLOEXEC | O_NONBLOCK);
+    if (fd < 0) {
+        SOL_WRN("Could not open IIO device%d: Device does not exist", device->device_id);
```
barbieri: it's better to %s and use `path` in this error message, followed by sol_util_strerrnoa(). It may be the case that the device does exist, but you don't have permission and the error message will be misleading.

src/shared/sol-iio.c
lpereira: This one is buggy: it doesn't account for `EINTR`, `EAGAIN`, and short writes. If `sol_util_write_file()` was hiding a check, it might be a better idea to make one that doesn't and make `sol_util_write_file()` use it.

src/shared/sol-iio.c
barbieri: also, noticed this pattern of snprintf() + stat() is used in other cases, then rename this function and use it everywhere, such as in sol_iio_open()

src/shared/sol-iio.c
lpereira: He does, since the snprintf() checks all around check if `len > PATH_MAX` instead of `len >= PATH_MAX`. Both are correct, but I prefer the one without `+ 1` since that's the one that's being used by the rest of the code.

src/shared/sol-iio.c
```diff
@@ -0,0 +1,990 @@
...
+    if (buffer_size > -1) {
+        if (trigger_name && trigger_name[0] != '\0') {
+            if (!set_current_trigger(device, trigger_name)) {
+                SOL_WRN("Could not set device current trigger");
+                goto error;
+            }
+        }
+
+        if (!check_trigger(device)) {
+            SOL_WRN("No trigger available for device");
```
barbieri: add context for this error message: which device failed?

src/shared/sol-iio.c
lpereira: So, in case that `len == 2`, return true, but forcefully set `shift` to a sane value.

src/shared/sol-iio.c
```diff
@@ -0,0 +1,990 @@
...
+
+    original_type = type;
+    /* Form of type is [be|le]:[s|u]bits/storagebits[>>shift] */
+    if (strstr(type, "be:")) {
+        channel->little_endian = false;
+        type += 3;
+    } else if (strstr(type, "le:")) {
+        channel->little_endian = true;
+        type += 3;
+    } else
```
barbieri: use `strstartswith()` since it's clear of the purpose and will work in all situations, while `strstr()` will break your usage if the type form changes in some way.

src/shared/sol-iio.c
```diff
@@ -0,0 +1,990 @@
...
+    channel_name_len = strlen(channel->name);
+    if (channel_name_len > 2) {
+        channel_name_suffix = channel->name + channel_name_len - 2;
+        if (streq(channel_name_suffix,  "_x") ||
+            streq(channel_name_suffix,  "_y") || streq(channel_name_suffix,  "_z")) {
+
+            channel_pure_name = malloc(channel_name_len - 1);
+            strncpy(channel_pure_name, channel->name, channel_name_len - 2);
+            channel_pure_name[channel_name_len - 2] = '\0';
+            return channel_pure_name;
```
barbieri: `strndup()` will save you 3 lines ;-)

src/shared/sol-iio.c
```diff
@@ -0,0 +1,990 @@
...
+ * This function shall return the name without <x|y|z> components.
+ * The form <type>[Y][_modifier] is also common (Y is a number), and this function
+ * remove numbers too, in an attempt to get the 'pure' name.
+ * TODO there are other esoteric combinations - for them, if we care about,
+ * we will probably need API calls to get the 'scale' and 'offset' file names */
+static char *
+channel_get_pure_name(struct sol_iio_channel *channel)
+{
+    size_t channel_name_len, i;
+    char *channel_pure_name, *channel_name_suffix, *original_channel_pure_name;
```
barbieri: move these variables to deeper scope to avoid confusion.

src/shared/sol-iio.c
```diff
@@ -0,0 +1,990 @@
...
+                    *channel_pure_name++ = channel->name[i];
+                else if (isdigit(channel->name[i])) {
+                    modified = true;
+                    continue;
+                }
+            }
+            if (modified)
+                return original_channel_pure_name;
+            else
+                free(original_channel_pure_name);
```
barbieri: return NULL here would make it clear.

src/shared/sol-iio.c
```diff
@@ -0,0 +1,990 @@
...
+            _name); \
+        if (len > 0 && sol_util_read_file(path, "%lf", &channel->scale) > 0) { \
+            goto end; \
+        } \
+    } while (0)
+
+    GET_SCALE(channel->name);
+    /* No scale. If channel has x, y, z or Y component, look for scale file without it */
+    channel_pure_name = channel_get_pure_name(channel);
+    GET_SCALE(channel_pure_name);
```
barbieri:
```c
if (channel_pure_name) {
   GET_SCALE(channel_pure_name);
   free(channel_pure_name);
}
```

src/shared/sol-iio.c
```diff
@@ -0,0 +1,990 @@
...
+void
+sol_iio_close(struct sol_iio_device *device)
+{
+    int i;
+    struct sol_iio_channel *channel;
+
+    SOL_PTR_VECTOR_FOREACH_REVERSE_IDX (&device->channels, channel, i) {
+        iio_del_channel(channel);
+        sol_ptr_vector_del(&device->channels, i);
+    }
```
barbieri: it's better to loop the vector and delete contents, then free the array at once with `sol_ptr_vector_clear()` than to delete one vector element at time, which will trigger lots of `realloc()`

src/shared/sol-iio.c
```diff
@@ -0,0 +1,990 @@
...
+                iio_buffer.size = len;
+                iio_buffer.data = buffer;
+                device->reader_cb(device, &iio_buffer, device->reader_cb_data);
+            }
+        }
+    }
+
+    free(buffer);
+
+    return result;
```
barbieri: returning `false` will invalidate the `sol_fd` created with `sol_fd_add()` stored in `device->fd_handler`

src/shared/sol-iio.c
```diff
@@ -0,0 +1,990 @@
...
+    if (len < 0 || len > PATH_MAX || stat(path, &st)) {
+        SOL_WRN("Could not find channel [%s] for device%d", name,
+            device->device_id);
+        return NULL;
+    }
+
+    channel = calloc(1, sizeof(struct sol_iio_channel));
+    SOL_NULL_CHECK(channel, NULL);
+
+    channel->name = strdup(name);
```
barbieri: you could avoid 2 mallocs here by using a single memory:

```c
struct sol_iio_channel {
   ... members ...;
   char name[];
};

channel = calloc(1, sizeof(struct sol_iio_channel) + strlen(name) + 1);
memcpy(channel->name, name, strlen(name));

```

src/shared/sol-iio.c
```diff
@@ -0,0 +1,990 @@
...
+bool
+sol_iio_read_channel_value(struct sol_iio_channel *channel, double *value)
+{
+    int len, raw_value;
+    char path[PATH_MAX + 1];
+    struct sol_iio_device *device = channel->device;
+
+    if (device->buffer_enabled) {
+        SOL_WRN("device%d is using buffer - use iio_read_buffer_channel_value() instead",
+            device->device_id);
```
barbieri: why can't you do it for the user instead of doing a warning?

src/shared/sol-iio.c
```diff
@@ -0,0 +1,990 @@
...
+    *value = (raw_value + channel->offset) * channel->scale;
+    return true;
+}
+
+void
+sol_iio_set_reader_cb(struct sol_iio_device *device, sol_iio_reader_cb reader_cb, void *data)
+{
+    device->reader_cb = reader_cb;
+    device->reader_cb_data = data;
+}
```
barbieri: you could create/delete the buffer memory here  instead of each callback of `sol_fd_add()`.

Also, do we want the `sol_fd_add()` even if there is no read callback associated? If not, then you can delete/add the fd monitor here.

src/shared/sol-iio.c
```diff
@@ -0,0 +1,990 @@
...
+        return false;
+    }
+
+    *value = (raw_value + channel->offset) * channel->scale;
+    return true;
+}
+
+void
+sol_iio_set_reader_cb(struct sol_iio_device *device, sol_iio_reader_cb reader_cb, void *data)
+{
```
barbieri: this is an API, then `SOL_NULL_CHECK()`

barbieri: and `const void *data`

src/shared/sol-iio.c
```diff
@@ -0,0 +1,990 @@
...
+    device->reader_cb = reader_cb;
+    device->reader_cb_data = data;
+}
+
+static int
+calc_channel_offset_in_buffer(struct sol_iio_channel *channel)
```
barbieri: const

src/shared/sol-iio.c
```diff
@@ -0,0 +1,990 @@
...
+        if (channel_itr->index < channel->index)
+            offset += channel->storagebits;
+    }
+
+    return offset;
+}
+
+bool
+sol_iio_read_buffer_channel_value(struct sol_iio_channel *channel, struct sol_iio_buffer *buffer, double *value)
+{
```
barbieri: `const struct sol_iio_buffer...`

src/shared/sol-iio.c
```diff
@@ -0,0 +1,990 @@
...
+    uint64_t data = 0;
+    int64_t s_data = -1;
+    int i, j, offset_bytes, storage_bytes;
+    bool negative = false;
+
+    /* Leve this calc to the first read, as now we know how much channels are enabled */
+    if (channel->offset_in_buffer == -1)
+        channel->offset_in_buffer = calc_channel_offset_in_buffer(channel);
+
+    if (channel->offset_in_buffer + channel->storagebits > (int)buffer->size * 8) {
```
barbieri: i guess you need this cast `(int)` because of signed x unsigned. Make `offset_in_buffer` and `channel->storagebits` all unsigned and you're free of cast.

also, if using uint32_t variant, then you can even compare to `UINT32_MAX` in the `if()` below that checks for `offset_in_buffer` being calculated or not.

src/shared/sol-iio.c
```diff
@@ -0,0 +1,990 @@
...
+        SOL_WRN("Invalid read on buffer - is buffer complete?");
+        return false;
+    }
+
+    if (channel->storagebits > 64) {
+        // XXX is such a thing even possible?
+        SOL_WRN("Could not read channel - more than 64 bits");
```
barbieri: would help to print the `storagebits` so we can better debug it. And even better, do it when `storagebits` is read the first time and not here on every callback.

src/shared/sol-iio.c
edersondisouza: `iio_read_buffer_channel_value()` needs the buffer sent on the callback. Is it worth reading from the last buffer we got? Aren't we hiding the truth from user?

src/shared/sol-iio.c
```diff
@@ -0,0 +1,990 @@
...
+        for (i = storage_bytes - 1, j = 0; i >= 0; i--, j += 8) {
+            data |= buffer->data[i + offset_bytes] << j;
+        }
+    }
+
+    data >>= channel->shift;
+
+    /* Remove the top useless bits */
+    data <<= channel->storagebits - channel->bits;
+    data >>= channel->storagebits - channel->bits;
```
barbieri: I guess you could calculate a mask when storagebits + shift + bits are read, then use a single `&` and `>>`.

src/shared/sol-iio.c
```diff
@@ -0,0 +1,990 @@
...
+    /* If signed and msb is 1, we have a negative number */
+    if (channel->is_signed) {
+        negative = data & (1 << (channel->bits - 1));
+    }
+
+    if (negative) {
+        s_data = data | (s_data << channel->bits);
+        *value = (s_data + channel->offset) * channel->scale;
+    } else
+        *value = (data + channel->offset) * channel->scale;
```
barbieri: I'd do as a multiply:
```c
*value = (negative ? -1 : 1) * (data + channel->offset) * channel->scale;
```

src/shared/sol-iio.c
```diff
@@ -0,0 +1,990 @@
...
+
+    return true;
+}
+
+bool
+sol_iio_device_trigger_now(struct sol_iio_device *device)
```
barbieri: API checks should do `SOL_NULL_CHECK()`

src/shared/sol-iio.h
```diff
@@ -0,0 +1,194 @@
...
+
+/**
+ * Add reading channel.
+ *
+ * @param device IIO handle of device to which channel is being add
+ * @param name Name of channel. Eg 'in_anglvel_x'.
+ *
+ * @return A new IIO channel handle
+ */
+struct sol_iio_channel *sol_iio_add_channel(struct sol_iio_device *device, const char *name);
```
barbieri: Based on the code here and the pain we went with other I/O I'd say we should use a `struct sol_iio_channel_config` structure and give all the configurations at open time, this will also save some public API functions and likely your logic can be simplified.

Remember to version the struct with a `api_version`.

src/shared/sol-iio.h
```diff
@@ -0,0 +1,194 @@
...
+ * @param buffer_size size of reading buffer. If 0, will use device default;
+ * if -1 will disable buffer - and all reads will be performed on channel files
+ * on sysfs.
+ * @param trigger_name Name of IIO trigger that will be used on this device. If NULL
+ * or empty, will try to use device current trigger. If device has no current trigger,
+ * will create a 'sysfs_trigger' and use it.
+ *
+ * @return A new IIO handle
+ */
+struct sol_iio_device *sol_iio_open(int device_id, int buffer_size, const char *trigger_name);
```
barbieri: similar to what I said in `sol_iio_add_channel()` we better have a `struct sol_iio_config` with all the configuration at once instead of multiple calls to setup buffer, callback, etc.

src/shared/sol-iio.h
```diff
@@ -0,0 +1,194 @@
...
+ * be performed if there's data on buffer. A trigger is used to send values
+ * to buffer. When buffer is ready, the callback set here will be called
+ * and data can be read using sol_iio_read_buffer_channel_value().
+ *
+ * @param device IIO handle of device
+ * @param reader_cb sol_iio_reader_cb that will be called when there's data on
+ * buffer.
+ * @param data User data to be sent to reader callback.
+ */
+void sol_iio_set_reader_cb(struct sol_iio_device *device, sol_iio_reader_cb reader_cb, void *data);
```
barbieri: `const void *data`, otherwise passing literals and constants will trigger a warning.

src/shared/sol-iio.c
```diff
@@ -0,0 +1,990 @@
...
+    int offset_in_buffer;
+
+    bool little_endian : 1;
+    bool is_signed : 1;
+};
+
+struct sol_iio_buffer {
+    size_t size;
+    uint8_t *data;
+};
```
barbieri: I'd change this to `struct sol_buffer`, it contains an extra `reserved` field that we don't need, but saves us from defining yet-another structure.

src/shared/sol-iio.c
barbieri: indeed! run `tools/check-license.sh` and `tools/check-style.py`

src/shared/sol-iio.c
```diff
@@ -0,0 +1,990 @@
...
+
+    return offset;
+}
+
+bool
+sol_iio_read_buffer_channel_value(struct sol_iio_channel *channel, struct sol_iio_buffer *buffer, double *value)
+{
+    uint64_t data = 0;
+    int64_t s_data = -1;
+    int i, j, offset_bytes, storage_bytes;
```
lpereira: Since you shift with `j` and `i`, it might be a better idea to make them unsigned.

src/shared/sol-iio.c
barbieri: As I said in other parts you should be keeping the buffer anyways, for performance reason. And here you can help the user, so he doesn't need to account himself 2 code paths...

src/shared/sol-iio.c
edersondisouza: That's curious. I'm running it but it says nothing about this file. Does it still works? (Added license locally, btw.)

barbieri: likely these tools are "too smart" and only check files added, and likely you already added it before and is changing the commit, so it wouldn't detect. @anselmolsm ?

src/modules/flow/iio/nodes.c
edersondisouza: Probably, more warnings were issued before this one, on `sol-iio.c`. Here, probably I should say that 'couldn't create node', and reason is 'could not open iio device %d' - or %s, if not only integers ids will be accepted.

src/modules/flow/iio/nodes.c
edersondisouza: Curiosity: why?

src/shared/sol-iio.c
edersondisouza: Hmmm... don't we 'lose' one char using only 'PATH_MAX'?

src/shared/sol-iio.c
edersondisouza: But macro `GET_SCALE` already frees it (actually `goto end`, where there's a free); maybe I could rewrite this code to be more clear.

src/shared/sol-iio.c
edersondisouza: Should I return `true` even on failure, then? Or should I do something else, like running `sol_fd_add()` again?

src/shared/sol-iio.c
edersondisouza: It won't crash: each and every place where a file is read/written is likely to `return false` or something of the like (if it crashes, is a bug). The idea would be 'to fail early'. Do you think I should not do like this, then? Fail only when really using?

src/shared/sol-iio.c
edersondisouza: Actually, only `j` is used for shift. `i` is just a counter/indexer =D

src/modules/flow/iio/nodes.c
edersondisouza: Curiosity: is this an issue even though `errmsg` is `static const`?

lpereira: Probably not, as the compiler will most likely generate the same code if it were `printf("%s", foo)` or `printf(foo)`, considering `static const char *foo="bar"`. Since `foo` does not contain any formatting characters, it would be OK.

However, it's basically free to be on the safe side: it's more about the practice of doing things correctly all the time and not thing about exceptions rather than thing about the exceptions all the time.

src/shared/sol-iio.c
lpereira: PATH_MAX is 4096 under Linux. We'll be fine.

src/shared/sol-iio.c
lpereira: It's usually a better idea to open the file anyway and fail if it couldn't be opened. In the cases you'll need the file size, just use fstat() with the open file descriptor; this avoids race conditions. Speaking of race conditions, keeping one of these files from /sys open will also avoid the kernel module from being removed, so you don't need to change the assumptions while the program is running.

edersondisouza: So, opening and closing all of them but one? Or opening the one that I plan to keep open and using `stat` on others?

src/shared/sol-iio.c
edersondisouza: But data is complement of 2, I need to extend the signal if negative, that's why I do `s_data = data | (s_data << channel->bits);`.

src/modules/flow/iio/nodes.c
lpereira: When you declare as `errmsg[]`, it's stored in the `.rodata` section and then (if it's not static) copied from there to the call frame every time this function is called. A variable being `const` doesn't really make it read-only/constant: it's just an instruction for the compiler to warn/fail if one tries to write to it without removing the const with a cast.

If it's declared as `*errmsg`, however, it's not copied from the `.rodata` section; it merely points there. The const here is to the pointer, not the pointee.

Since this variable is not exported, the compiler will most likely treat this as `const char * const errmsg = "foo"`: both the pointer and the pointee will be stored in the `.rodata` section. Using `static` will not help much but it's another hint that this won't change.

src/shared/sol-iio.c
lpereira: Open one, keep it open. Stat the others; if the stat fails, close the open one and return error.

src/shared/sol-iio.c
edersondisouza: I've tried using it, but it is a pain, really. I have to account for empty strings and '\n' on files. =/

edersondisouza: Is it a must using it, or can I leave it as is?

src/shared/sol-iio.c
edersondisouza: To make that, it would be something on these lines (checking the errors and keep trying to write the whole string), or something completely different?

barbieri: keep the same idea, try to write everything, handling EINTR (got a signal) or EAGAIN (non-blocking), with a limited amount of retries so we don't busy wait forever.

src/shared/sol-iio.c
barbieri: You want to read a single line as a string, excluding "\n"? You could create another function to do it, or leave as is -- now I understand your point.

src/shared/sol-iio.c
barbieri: `device->fd_handler = NULL; return false`

src/shared/sol-iio.c
barbieri: the multiplication will do it right for you, given data is already an integer.

edersondisouza: As far as I tested, it doesn't work. How would it convert if data is an `uint64_t`? The bit pattern on the last n bits, say `0000 ... 1111 1111` would represent `-1` as an `int8_t`, but is `255` on the `uint64_t`; if multiplied by `-1`, yields to `-255`.

src/shared/sol-iio.c
edersondisouza: Sounds good. I'm thinking, however, of doing this as a separate PR, after this one goes in, as it will touch different places. Is it ok?

edersondisouza:
v2 on #441

PR #413, opened 2015-07-31
Build bring back fbp based samples v2

dorileo:
# Changes
v2:
  + new patch: fbp-generator: Make sure to #include required -gen.h;
  + new patch: node-type: move log_init() implementation to -gen.c;
  + new patch: build: move samples compiled bin to build/stage/;
  + since now sol-fbp-generator does the #include for node types - removed the -include build directive for every -gen.h on patch: build: bring back fbp sample;

### DIFF
--- a/Kconfig
+++ b/Kconfig
@@ -63,6 +63,7 @@ endmenu
 menu "Samples"
 source "src/samples/coap/Kconfig"
 source "src/samples/common/Kconfig"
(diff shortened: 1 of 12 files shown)


vcgomes:
Now I don't need those annoying '-include' when compiling my external node type.

:+1:

dorileo:
ok, applied... thanks.

PR #414, opened 2015-07-31
string module: new uppercase/downcase nodes.

glima:
Naturally, they will convert an input string to either down or uppercase
versions.

### DIFF
--- a/src/modules/flow/string/string.c
+++ b/src/modules/flow/string/string.c
@@ -30,12 +30,13 @@
  * OF THIS SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
  */
 
+#include <ctype.h>
+#include <errno.h>
+
 #include "string-gen.h"
 
 #include "sol-flow-internal.h"
-
-#include <sol-util.h>
-#include <errno.h>
+#include "sol-util.h"
 
 struct string_data {
     int n;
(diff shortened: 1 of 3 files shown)


src/modules/flow/string/string.c
```diff
@@ -427,4 +428,50 @@ string_split(struct sol_flow_node *node, void *data, uint16_t port, uint16_t con
...
+    int r;
+
+    r = sol_flow_packet_get_string(packet, &value);
+    SOL_INT_CHECK(r, < 0, r);
+
+    cpy = strdup(value);
+    SOL_NULL_CHECK(value, -ENOMEM);
+
+    for (ptr = cpy; *ptr; ptr++)
+        *ptr = tolower(*ptr);
```
barbieri: This is mostly okay, but only works with ASCII. Try with "ç", "á" or the heavy-metal test: turkish (http://www.i18nguy.com/unicode/turkish-i18n.html), they will all fail since we're converting base on each byte, and for "ç"/"á" they are multibyte and for turkish there are other variations.

Time has come to detect icu.h and use its functions to do such case change, as well as new nodes to convert encoding. I'd say that our current strings should all be UTF-8, then we could add BLOB -> UTF-8 using an icu node with encoding as option.

glima: Mmm, okay. Let me start looking at ICU/iconv then. Thanks for the feedback.

glima:
v2 coming, closing this one.

PR #417, opened 2015-08-01
git-init-submodule: check for single branch checkout support

dorileo:
Git supports since version 1.7.10 single branch checkouts which in turns
will do a shallow checkout. Since we've seen people using ubuntu 12.04
LTS and once it ships git 1.7.9 we would check for the feature and do a
full checkout if it can't do single branch checkouts.

### DIFF
--- a/data/scripts/git-init-submodules.py
+++ b/data/scripts/git-init-submodules.py
@@ -32,6 +32,7 @@
 
 import configparser
 import os
+import re
 import subprocess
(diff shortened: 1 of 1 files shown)


dorileo:
Well, I think it's not enough to work around issues with ubuntu 12.04, I'm working on that.

dorileo:
I think we'll not want to support so old OS'es like this...

PR #418, opened 2015-08-03
Introduce named options to replace array of strings usage

cmarcelo:


cmarcelo:
@barbieri could you take a look at this?

src/lib/flow/sol-flow-node-options.c
```diff
@@ -633,33 +707,154 @@ options_from_strv(const struct sol_flow_node_options_description *desc, struct s
...
-sol_flow_node_options_new_from_strv(const struct sol_flow_node_type *type, const char *const *strv)
+SOL_API int
+sol_flow_node_options_new(
+    const struct sol_flow_node_type *type,
+    const struct sol_flow_node_named_options *named_opts,
+    struct sol_flow_node_options **out_opts)
 {
-    struct sol_flow_node_options *opts;
+#ifndef SOL_FLOW_NODE_TYPE_DESCRIPTION_ENABLED
+    SOL_WRN("does not work if compiled with --disable-flow-node-type-description");
```
lpereira: Is the message still valid after the move to kbuild?

cmarcelo:
Replaced by #428.

PR #419, opened 2015-08-03
tools: teach calc-lib-size to disable LOG from the build

cmarcelo:
By default, calc-lib-size will disable the log from the build. Also
slightly improve the output.

### DIFF
--- a/tools/calc-lib-size
+++ b/tools/calc-lib-size
@@ -46,16 +46,20 @@ Show size of the library compiled in different configurations
 
 OPTIONS
     -a             show size of all modules in addition to the library
+    -l             do not disable log from the build
     -h             display this help message
 " 1>&2;
     exit 1;
 }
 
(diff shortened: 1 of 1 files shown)


tools/calc-lib-size
```diff
@@ -65,11 +69,21 @@ for arg in "$@"; do
...
+    echo "# logs enabled"
+fi
+
+echo
 
 function calc_size() {
-    make alldefconfig
+    echo "# building with CFLAGS='$@'"
+    make alldefconfig > /dev/null
     make clean > /dev/null
```
barbieri: usually you clean before you change config, that's to avoid missing files that do not belong in the new configuration.

cmarcelo:
Applied in 6663cff219c65609d153562cf2e19ef403fe95b5 changing the order of the clean call.

PR #420, opened 2015-08-03
nodes: AM2315 temperature/humidity sensor v3

edersondisouza:
v3:
No more sending cached data: if more reading requests happen before the two second interval has passed, all requests will be answered at the end of the interval.
Avoid some 'callocs' by storing callback data on `am2315` struct.
Output data only in Kelvin - and sending a sol_drange instead of a raw value.

v2:
Use constant for scale factor;
Fine grained messages for user when getting invalid readings;
Make delay asynchronous - this involved a lot of work from previous version, but I hope everything should be ok.

### DIFF
--- a/src/lib/flow/Kconfig
+++ b/src/lib/flow/Kconfig
@@ -30,6 +30,7 @@ menu "Node Types"
 	depends on FLOW
 source "src/modules/flow/accelerometer/Kconfig"
 source "src/modules/flow/aio/Kconfig"
+source "src/modules/flow/am2315/Kconfig"
 source "src/modules/flow/app/Kconfig"
 source "src/modules/flow/boolean/Kconfig"
 source "src/modules/flow/byte/Kconfig"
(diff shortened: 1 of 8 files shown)


edersondisouza:
Someone?

edersondisouza:
@glima I recall you having issues with the 'burst' of packets after the two second interval. Are you sure that only one packet after N ticks is ok? Anything else?

glima:
Oh, I'm fine with 1 or more, I guess

glima:
If you document it explicitly

zehortigoza:
You will need update this PR to use the new I2C API, when you send the new version ping me and I will review.

edersondisouza:
On 12-08-2015 10:39, José Roberto de Souza wrote:
> You will need update this PR to use the new I2C API, when you send the
> new version ping me and I will review.
>
> —
> <https://github.com/solettaproject/soletta/pull/420#issuecomment-130307034>.
>
ack

edersondisouza:
v4 on #508

Issue #335, opened 2015-07-16
sol-flow-static shouldn't always duplicate options.


barbieri:
`sol-flow-static.c:flow_node_open()` is always calling `sol_flow_node_get_options()` for each children options, this effectively allocates memory and copies the options, then need to call `sol_flow_node_free_options()` right after that -- already bad per se.

However, if the type doesn't specify `new_options()` callback, an `empty_defaults` is used. This is bad, since the original options was available and were not used.

I got to this when creating a type manually and not specifying the `new_options()` callback, as it's not required for my use case (simplectype -- to be sent as PR soon).

I'm not changing this as it's not clear to me why options needed to be duplicated at all. Even in the case of a builder, where you create new options from a textual description, those should be stored in the builder-generated type itself when such type is created.

I also see but do not understand the purpose of `child_opts_set()` callback that is called per-instance of flow, but stores the information per-type (class). It looks wrong, it should be pushing that information to node and should only be used if we're getting an option on a parent flow/node that should be redirected to a children node -- not my case, where I have no exported options in my parent.

To support exporting options, that will result in options being sent/propagated to children, I believe that the most efficient and clear way is making exported options a separate vector we have as options for static flows (ie: `struct sol_flow_static_node_options`) that we iterate in a way similar to what we do for ports:

## sol-flow-static.h defines
```c
struct sol_flow_static_node_options {
   const struct sol_flow_node_options base;
#define SOL_FLOW_STATIC_NODE_OPTIONS_API_VERSION 1
  const struct sol_flow_node_options {
     uint16_t node;
     const struct sol_flow_node_options *opts;
  } children_options;
};

struct sol_flow_static_option_spec {
   uint16_t node;
   /* some stuff to define how to export, like members or the whole struct
   * using whole struct here, so only taking the node index.
   * we could take the member index within the options description.
   */   
};
```

## sol-flow-static.c uses
```c
struct sol_flow_node_options *
get_exported_options(const struct flow_static_type *type, uint16_t node_idx, struct sol_flow_static_node_options *flow_opts) {
   if (type->exported_options && flow_opts && flow_opts->children_options) {
      uint16_t i;
      /* ensure node option is exported */
      for (i = 0; type->exported_options[i]->node != UINT16_MAX; i++) {
          if (type->exported_options[i]->node == node_idx) {
             uint16_t j;
             for (j = 0; flow_opts->children_options[j].node != UINT16_MAX; i++) {
[...]

PR #421, opened 2015-08-03
Reduce the amount of allocations and code for options

cmarcelo:
Improvements to how we deal with `struct sol_flow_node_options`.

The second patch ensures that we simply pass thru the options from the spec if nothing is being exported, as discussed in #335.


cmarcelo:
The third patch still doesn't touch JS and Simple CType, I'll look at them now but this is good enough for a review.

cmarcelo:
@ibriano could you take a look?

ibriano:
LGTM

PR #422, opened 2015-08-03
sol-aio: Add C API to handle Analog I/O (V3)

mbelluzzo:
Tested on RIOT using dummy native API and Galileo.

Applied suggested changes from V2 and V1

### DIFF
--- a/src/lib/common/Kconfig
+++ b/src/lib/common/Kconfig
@@ -310,6 +310,10 @@ config SOL_BUS
 	default n
 
 menu "Hardware Options"
+config USE_AIO
+	bool "AIO Support"
+	default y
+
 config USE_PWM
 	bool "PWM Support"
 	default y
--- a/src/lib/io/sol-aio-common.c
+++ b/src/lib/io/sol-aio-common.c
@@ -1,0 +1,60 @@
+/*
+ * This file is part of the Soletta Project
+ *
+ * Copyright (C) 2015 Intel Corporation. All rights reserved.
+ *
+ * Redistribution and use in source and binary forms, with or without
+ * modification, are permitted provided that the following conditions
+ * are met:
+ *
+ *   * Redistributions of source code must retain the above copyright
+ *     notice, this list of conditions and the following disclaimer.
+ *   * Redistributions in binary form must reproduce the above copyright
+ *     notice, this list of conditions and the following disclaimer in
+ *     the documentation and/or other materials provided with the
+ *     distribution.
+ *   * Neither the name of Intel Corporation nor the names of its
+ *     contributors may be used to endorse or promote products derived
+ *     from this software without specific prior written permission.
+ *
+ * THIS SOFTWARE IS PROVIDED BY THE COPYRIGHT HOLDERS AND CONTRIBUTORS
+ * "AS IS" AND ANY EXPRESS OR IMPLIED WARRANTIES, INCLUDING, BUT NOT
+ * LIMITED TO, THE IMPLIED WARRANTIES OF MERCHANTABILITY AND FITNESS FOR
+ * A PARTICULAR PURPOSE ARE DISCLAIMED. IN NO EVENT SHALL THE COPYRIGHT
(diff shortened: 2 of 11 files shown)


mbelluzzo:
forget the branches name ;-)

src/lib/common/Kconfig
```diff
@@ -310,6 +310,10 @@ config SOL_BUS
 	default n
 
 menu "Hardware Options"
+config USE_AIO
+	bool "AIO Support"
```
lpereira: Is AIO the fabled POSIX Asynchronous IO? Might be a good idea to give this a more descriptive string (`bool "Analog I/O (AIO) support"`).

src/lib/io/sol-aio-linux.c
```diff
@@ -0,0 +1,154 @@
...
+};
+
+static bool
+_aio_open_fd(struct sol_aio *aio)
+{
+    int len;
+    char path[PATH_MAX];
+
+    len = AIO_PATH(path, aio->device, aio->pin);
+    if (len < 0 || len > PATH_MAX)
```
lpereira: `>= PATH_MAX`

src/lib/io/sol-aio-common.c
```diff
@@ -0,0 +1,60 @@
...
+{
+    struct sol_aio *aio;
+
+    SOL_LOG_INTERNAL_INIT_ONCE;
+
+    aio = sol_aio_open_raw(device, pin, precision);
+#ifdef HAVE_PIN_MUX
+    if (aio && sol_pin_mux_setup_aio(device, pin)) {
+        SOL_WRN("Pin Multiplexer Recipe for aio device=%d pin=%d found, \
+            but couldn't be applied.", device, pin);
```
lpereira: This will print a bunch of spaces in the middle of the string. Is this what you want?

src/lib/io/sol-aio-linux.c
```diff
@@ -0,0 +1,154 @@
...
+    }
+
+    len = AIO_DEV_PATH(path, device);
+    if (len < 0 || len > PATH_MAX || stat(path, &st)) {
+        SOL_WRN("aio #%d,%d: aio device %d does not exist", device, pin, device);
+        return NULL;
+    }
+
+    len = AIO_PATH(path, device, pin);
+    if (len < 0 || len > PATH_MAX || stat(path, &st)) {
```
lpereira: Please don't use `stat()` followed by `open()`: this is a TOCTOU. Use `open()` directly and keep the file open for the duration of the program. This will ensure that the kernel module controlling this particular device isn't removed while the program is running.

src/lib/io/sol-aio-linux.c
```diff
@@ -0,0 +1,154 @@
...
+
+int32_t
+sol_aio_get_value(const struct sol_aio *aio)
+{
+    unsigned int val;
+
+    SOL_NULL_CHECK(aio, -1);
+
+    if (!aio->fp) {
+        if (!_aio_open_fd((struct sol_aio *)aio)) {
```
lpereira: This is not opening an `fd`, but an `fp`.

src/lib/common/Kconfig
mbelluzzo: I liked the suggested string, but as a side comment, if it was Asynchronous IO it wouldn't be in Hardware Options anyway ;-)

lpereira: Yeah, I know. :)

It's used in other parts as well (flow node types, for instance), and it might be a good idea to use the same explanation there.

src/lib/io/sol-aio-linux.c
mbelluzzo: It may be a good Idea to change this for the export part of gpio as well. Today we first use stat to see if its exported and open after that. We should try to open, try to export if fails and try open again. Is that right?

src/lib/io/sol-aio-linux.c
mbelluzzo: Is counter really necessary in this case? scope will be totally local and yet I will use _tmp instead of tmp

mbelluzzo:
Closing because V4 at #438

PR #423, opened 2015-08-03
Build more fbp based samples

dorileo:


dorileo:
@cmarcelo could you take a look at the patch touching the sol-fbp-generator.c ?
thanks...

dorileo:
v2 coming...

PR #425, opened 2015-08-04
JSON number parsing

barbieri:
Parse numbers from JSON tokens as both signed and unsigned integers of 32 or 64 bits, as well as a double float point precision.

### DIFF
--- a/src/lib/parsers/sol-json.c
+++ b/src/lib/parsers/sol-json.c
@@ -34,10 +34,13 @@
 #include <errno.h>
 #include <stdlib.h>
 #include <string.h>
+#include <locale.h>
 
 #include "sol-json.h"
 #include "sol-log.h"
 #include "sol-util.h"
+#include <float.h>
+#include <math.h>
 
 static bool
 check_symbol(struct sol_json_scanner *scanner, struct sol_json_token *token,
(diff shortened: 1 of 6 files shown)


src/lib/parsers/sol-json.c
```diff
@@ -138,6 +141,193 @@ check_number(struct sol_json_scanner *scanner, struct sol_json_token *token)
...
+     * pt_BR it will use ',' as fractional/decimal separator and we do
+     * not want that as JSON uses '.' in the same way as the 'C'
+     * locale. Then force 'C' and save oldlocale, restoring it before
+     * we return.
+     */
+    oldlocale = setlocale(LC_NUMERIC, NULL);
+    if (oldlocale && !streq(oldlocale, "C")) {
+        oldlocale = strdupa(oldlocale);
+        setlocale(LC_NUMERIC, "C");
+    }
```
lpereira: The problem in doing this is that this isn't thread safe. We don't have threads now (apart from the worker threads), but we might in the future and this might bite us back. It might be a good idea to use `localeconv()`, get its `decimal_point` character, and if it's not `"."`, substitute it in the passed string before calling `strtod()`.

src/test/test-json.c
```diff
@@ -167,4 +169,416 @@ test_json(void)
...
+        token.start = buf;
+        token.end = buf + strlen(itr->str);
+        retval = sol_json_token_get_int64(&token, &value);
+        if (itr->expected_return == 0 && retval == 0) {
+            if (itr->reference == value) {
+                SOL_DBG("OK: parsed '%s' as %" PRIi64, itr->str, value);
+            } else {
+                SOL_WRN("FAILED: parsed '%s' as %" PRIi64 " where %" PRIi64
+                    " was expected", itr->str, value, itr->reference);
+                FAIL();
```
lpereira: Why `FAIL()` does not take a `printf()-style` formatting string + parameters?

src/shared/sol-util.h
```diff
@@ -160,11 +183,35 @@ struct sol_vector sol_util_str_split(const struct sol_str_slice slice, const cha
 static inline int
 sol_util_size_mul(size_t elem_size, size_t num_elems, size_t *out)
 {
-#ifdef HAVE_UMULL_OVERFLOW
-    if (__builtin_umull_overflow(elem_size, num_elems, out))
+#ifdef HAVE_BUILTIN_MUL_OVERFLOW
+    if (__builtin_mul_overflow(elem_size, num_elems, out))
```
lpereira: I overlooked this version. Nice.

src/lib/parsers/sol-json.c
barbieri: Confirmed and `setlocale()` shouldn't be called from threads... I guess the solution will be to document that in the function body.

I'll implement the extra optimization of only overriding the locale if `localeconv()` reports `decimal_point != '.'`

barbieri:
v2 as #440

PR #426, opened 2015-08-04
Build more fbp based samples v2

dorileo:
## Changes
v2:
  + simplified the patch: sol-fbp-generator: add ERROR port support

### DIFF
--- a/src/bin/sol-fbp-generator/main.c
+++ b/src/bin/sol-fbp-generator/main.c
@@ -84,6 +84,12 @@ struct declared_fbp_type {
     int id;
 };
 
+struct port_description error_port = {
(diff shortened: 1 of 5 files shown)


dorileo:
@cmarcelo you're review is still valid here :)

cmarcelo:
+1 with the comment on splitting the rules for the samples that have their own subdir.

Whether add configuration for each sample one or not, I think the major benefit was pointed out by Dilly, which would be to keep the necessary dependencies for each sample separated. If we don't, to ensure the samples will build and work once built, we'll have to merge all the dependencies in the single config.

I even wonder if the Kconfig piece itself could be part of the sample directory, making it all in one place.

dorileo:
v3 coming...

PR #427, opened 2015-08-04
Test suite improvements

cmarcelo:


cmarcelo:
Not critical, but the last patch fix a longstanding issue: the script always return success, even if one tests fail.

lpereira:
Why not just use `multiprocessing.Pool`? It's as simple as:

    from multiprocessing import Pool

    with Pool(processes=os.cpu_count()) as p:
        results = p.map(run_one_test, all_tests_list)

Then it's the matter of defining `run_one_test()` to return `(test_name, True)` or `(test_name, False)` if a particular test fails.

cmarcelo:
@lpereira 

`multiprocessing` seems to be useful when you want to Python code running in parallel. In our case, the bulk of work is already done forked, so it would just mean one extra fork to run Python first for each worker.

I also like the fact that we can easily return more stuff than the true/false, e.g. we return log pieces, so the final process merge them in the log without races.

I see how `Pool` is nice, but would be better if we had one of this for `threading`.

lpereira:
[**@cmarcelo**](https://github.com/cmarcelo) commented on [Aug 4, 2015, 12:23 PM GMT-3](https://github.com/solettaproject/soletta/pull/427#issuecomment-127647863 "2015-08-04T15:23:15Z - Replied by Github Comment Enhancer"):

> I see how `Pool` is nice, but would be better if we had one of this for `threading`.

In fact, we have: `from multiprocessing.pool import ThreadPool`. It works like the process-based `Pool`, but uses threads.

lpereira:
[**@cmarcelo**](https://github.com/cmarcelo) commented on [Aug 4, 2015, 12:23 PM GMT-3](https://github.com/solettaproject/soletta/pull/427#issuecomment-127647863 "2015-08-04T15:23:15Z - Replied by Github Comment Enhancer"):
> 
> I also like the fact that we can easily return more stuff than the true/false, e.g. we return log pieces, so the final process merge them in the log without races.

As a suggestion, you could return named tuples, with keys for the test status, log excerpts, test name, and maybe something else.

cmarcelo:
After looking at the code I'm not convinced we need `multiprocessing`. The threaded version of pool seems to use the mechanisms in multiprocess with a dummy backend to not create new processes.

Agreed to keep the code as is and revisit next time we make major changes here.

cmarcelo:
Applied in 67d32716610b3e3bcea30f452976cb64a181d8f0 and parents.

PR #428, opened 2015-08-04
Introduce named options to replace array of strings usage [v2]

cmarcelo:


cmarcelo:
Rebased.
Added the extra type information in the error when possible.
Made a separate patch to update the error messages (unsupported function) to not refer to autotools.

This replaces #418.

src/lib/flow/sol-flow-builder.c
```diff
@@ -1135,14 +1077,49 @@ sol_flow_builder_add_node_by_type(struct sol_flow_builder *builder, const char *
...
+
+            for (i = 0; i < extra_opts.count; i++) {
+                struct sol_flow_node_named_options_member *dst = named_opts.members + named_opts.count + i;
+                struct sol_flow_node_named_options_member *src = extra_opts.members + i;
+
+                *dst = *src;
+                if (dst->type == SOL_FLOW_NODE_OPTIONS_MEMBER_STRING) {
+                    dst->string = strdup(src->string);
+                    if (!dst->string)
+                        goto end;
```
lpereira: Set `r = -errno` before going to `end`?

src/lib/flow/sol-flow-node-options.c
```diff
@@ -478,104 +476,176 @@ direction_vector_parse(const struct sol_flow_node_options_member_description *me
...
 
+static int(*const options_parse_functions[]) (const char *, struct sol_flow_node_named_options_member *) = {
+    NULL,
+    boolean_parse,
+    byte_parse,
+    irange_parse,
+    drange_parse,
+    rgb_parse,
+    direction_vector_parse,
+    string_parse,
```
lpereira: Might be a good idea to index these by `enum sol_flow_node_options_member_type`:

     [FOO_BOOLEAN] = boolean_parse,
     [FOO_BYTE] = byte_parse,
     [FOO_STRING] = string_parse,

src/lib/flow/sol-flow-node-options.c
```diff
@@ -478,104 +476,176 @@ direction_vector_parse(const struct sol_flow_node_options_member_description *me
...
+        *s = strdup(m->string);
+    default:
+        /* Not reached. */
+        /* TODO: Create ASSERT_NOT_REACHED() macro.*/
+        break;
+    }
+}
+
+static const struct sol_str_table member_str_to_type[] = {
+    SOL_STR_TABLE_ITEM("boolean", SOL_FLOW_NODE_OPTIONS_MEMBER_BOOLEAN),
```
lpereira: Maybe you can join this table with `options_parse_functions[]`? By making it a `string` -> `function ptr` mapping, and indexing it with the `enum sol_flow_node_options_member_type`, only one table will need to be maintained.

The `sol_str_table` will need to be changed to have a function that also gives the index.

PR #430, opened 2015-08-04
dependency-resolver: produce stable output

michaelolbrich:
Without this, the order of the config entries changes every time the
dependency resolver is executed. As a result the order of the variables in
the .config file changes as well. This is rather annoying if the .config
file is stored elsewhere (e.g. in another git repository).

Enforce a defined order by sorting the items. This way the files only
change when with actual dependency changes.

Signed-off-by: Michael Olbrich <m.olbrich@pengutronix.de>


cmarcelo:
+1

@dorileo could you take a look?

dorileo:
applied, thanks...

PR #431, opened 2015-08-04
build: sol-bus uses glib so depend on it

michaelolbrich:
Also update the dependencies for PLATFORM_SYSTEMD. It selects SOL_BUS so it
must depend on the same symbols as SOL_BUS. Otherwise kconfig may be unable
to resolve the dependencies.

Signed-off-by: Michael Olbrich <m.olbrich@pengutronix.de>


michaelolbrich:
Ok, force pushes are a bad idea. We loose the comments :-/
I cannot drop the systemd -> sd-bus dependency dependency. It's realy needed. Anything else is fixed in #462 so I'll just close this.

PR #434, opened 2015-08-04
Added UV sensor family si114x

tcanabrava:
Code based on the UPM project.

First try on the si114x hardware family.
Code based on the upm project.

I disliked how the sol_i2c_write_register works, since it receives it's data as pointer I can't do 
sol_i2c_write_register(mdata->context, REG_UCOEF2, SOME_ENUM_HERE, 1), having to create a variable to hold that enum and pass it to the function - maybe we should create variants for uint8_t, uint16_t and uint8_t* types?


tcanabrava:
close because of v2

PR #435, opened 2015-08-04
Introduce named options to replace array of strings usage [v3]

cmarcelo:


cmarcelo:
Minor fixes and changes from last review. Decided not to merge the tables in this series.

New patch added in the middle to initialize early the `options_size` of builder types. This ensure that it will be in fact initialized, since other patches trigger code path that uses it. The bot failures seem to be caused by trying to allocate memory with size 0.

cmarcelo:
Replaces #428.

cmarcelo:
Hmmm. Still something going on wrong, trying to verify what's failing.

The added patch still stands on its own (I added a similar change in the other series already).

cmarcelo:
The issue was an invalid cast. "builder" vs "builder_type_data". The extra commit of this series was changed in v4 to include that fix.

The problem wasn't triggered in my environments (more than one!) because the value being read was "ok" (it was zero, which led to my first suspicion above). But in the bots setup, the value ended up being too large, so indeed an "couldn't allocate memory".

PR #436, opened 2015-08-04
String upper/lower v2

glima:
Changes since v1:

The ascii-only code is now a fallback only if ICU is not found on the system at build time. Two new deps are now declared, locale and icu, then. If the latter is not found, a CPP #warning line is output to inform the user of the limitations of the string nodes in this form.

If this is ok, I'll proceed adding i18n support on other string operations throughout the code. Soletta now expects that fbp files are encoded in utf-8, not just plain ASCII.

### DIFF
--- a/README.md
+++ b/README.md
@@ -383,9 +383,10 @@ gpio/writer (a single boolean OUT port):
 
 Flow can be created directly in C using low-level primitives from
 sol-flow.h or using a higher-level API in sol-flow-builder.h. An
-alternative is to write the flow in a domain-specific language "FBP"
-that is easier to express. As an example imagine one wants to blink an
-LED linked to GPIO pin 123 every 200ms:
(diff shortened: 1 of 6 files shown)


glima:
Fuck, my tests do not expect not having ICU :/

glima:
I'll have to lower the required version, it seems (<3 Ubuntu).

README.md
```diff
@@ -499,3 +500,6 @@ By making a contribution to this project, I certify that:
     maintained indefinitely and may be redistributed consistent with
     this project or the open source license(s) involved.
 ```
+
+<a name="footnote_01">1</a>: Soletta expects that FBP files are
+utf-8-encoded
```
lpereira: Might be a good idea to validate that at FBP parse time, if ICU is found.

PR #437, opened 2015-08-04
{float,int}-to-string conversion nodes now do formatting v2

glima:
Changes since v1:

- @dorileo 's comments taken in account
- it now depends on #436 's HAVE_LOCALE dependency check
- log_init() changed to account for @dorileo 's fixes on logs

### DIFF
--- a/src/modules/flow/converter/string-format.c
+++ b/src/modules/flow/converter/string-format.c
@@ -1,0 +1,2329 @@
+/*
+ * This file is part of the Soletta Project
+ *
+ * Copyright (C) 2015 Intel Corporation. All rights reserved.
+ *
+ * Redistribution and use in source and binary forms, with or without
+ * modification, are permitted provided that the following conditions
+ * are met:
+ *
+ *   * Redistributions of source code must retain the above copyright
+ *     notice, this list of conditions and the following disclaimer.
+ *   * Redistributions in binary form must reproduce the above copyright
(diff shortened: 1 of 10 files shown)


src/modules/flow/converter/string-format.c
```diff
@@ -0,0 +1,2329 @@
...
+ * THEORY OF LIABILITY, WHETHER IN CONTRACT, STRICT LIABILITY, OR TORT
+ * (INCLUDING NEGLIGENCE OR OTHERWISE) ARISING IN ANY WAY OUT OF THE USE
+ * OF THIS SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
+ */
+
+/* This file has code excerpts (string formatting module) extracted
+ * from cpython project (https://hg.python.org/cpython/), that comes
+ * under the PSFL license. The string formatting code was adapted here
+ * to Soletta data types. The entire text for that license is present
+ * in this directory */
```
lpereira: IANAL but I believe you should also state the original copyright/authorship here, not only state which license this code has been based upon.

glima: @anselmolsm ^ Can you help with this, please?

anselmolsm: The original copyright line is missing ;)

glima:
Thanks everyone. Copyright note added and now merging.

PR #438, opened 2015-08-04
sol-aio: Add C API to handle Analog I/O (V4)

mbelluzzo:
Applied changes suggested by @lpereira

### DIFF
--- a/src/lib/io/sol-aio-linux.c
+++ b/src/lib/io/sol-aio-linux.c
@@ -1,0 +1,150 @@
+/*
+ * This file is part of the Soletta Project
+ *
+ * Copyright (C) 2015 Intel Corporation. All rights reserved.
+ *
+ * Redistribution and use in source and binary forms, with or without
+ * modification, are permitted provided that the following conditions
+ * are met:
+ *
+ *   * Redistributions of source code must retain the above copyright
+ *     notice, this list of conditions and the following disclaimer.
+ *   * Redistributions in binary form must reproduce the above copyright
+ *     notice, this list of conditions and the following disclaimer in
+ *     the documentation and/or other materials provided with the
+ *     distribution.
+ *   * Neither the name of Intel Corporation nor the names of its
+ *     contributors may be used to endorse or promote products derived
+ *     from this software without specific prior written permission.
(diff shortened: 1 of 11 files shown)


src/lib/io/sol-aio-linux.c
```diff
@@ -0,0 +1,150 @@
...
+{
+    char path[PATH_MAX];
+
+    if (!AIO_PATH(path, aio->device, aio->pin))
+        return false;
+
+    aio->fp = fopen(path, "re");
+    if (!aio->fp)
+        return false;
+    setvbuf(aio->fp, NULL, _IONBF, 0);
```
edersondisouza: Why is this needed?

mbelluzzo: Because there is no need for buffering in this case.

glima:
Other than that, LGTM

src/lib/io/sol-aio-linux.c
edersondisouza: Sounds fair, but on iio isn't the buffer located on `/dev/iio:deviceX`? I mean, is this really necessary? (If so, I may need to change something on my pr #441)

edersondisouza: Or is this not related to the iio buffers?

ibriano: On Wed, 05 Aug 2015 12:34:30 -0700, Ederson de Souza wrote:
> > +    int pin;
> > +    unsigned int mask;
> > +};
> > +
> > +static bool
> > +_aio_open_fp(struct sol_aio *aio)
> > +{
> > +    char path[PATH_MAX];
> > +
> > +    if (!AIO_PATH(path, aio->device, aio->pin))
> > +        return false;
> > +
> > +    aio->fp = fopen(path, "re");
> > +    if (!aio->fp)
> > +        return false;
> > +    setvbuf(aio->fp, NULL, _IONBF, 0);
> 
> Or is this not related to the iio buffers?
> 

No, it's the internal buffer in FILE*

mbelluzzo:
Inc. V5 with power on/off management.

PR #439, opened 2015-08-04
Introduce named options to replace array of strings usage [v4]

cmarcelo:


cmarcelo:
Replaces #435.

cmarcelo:
Applied in 61d873e35c8473ee5cf9be13a1e81df9042ce262 and parents.

PR #440, opened 2015-08-04
JSON number parsing (v2)

barbieri:
Parse numbers from JSON tokens as both signed and unsigned integers of 32 or 64 bits, as well as a double float point precision.

Changes since v1 (PR #425):
 * changed sol_util_strtodn() to take a boolean `use_locale`. If so, the current locale is allowed (as in standard `strtod()`), otherwise `strtod_l()` is used with a "C" locale. In the case where `strtod_l()` is not available but there is locale support we translate the string exchanging '.' and `lc->decimal_point`, either in-place using the copy (tmpbuf) or creating yet-another copy if `lc->decimal_point` is more than one byte.
 * enable test-util.c to be compiled;
 * fix `test_size_mul()`, at least on my 32 bits machine, `SIZE_MAX` was odd and the test was failing since when multiplying by 2 it was not getting the same value;

### DIFF
--- a/src/shared/sol-util.c
+++ b/src/shared/sol-util.c
@@ -37,9 +37,38 @@
 #include <stdlib.h>
 #include <string.h>
 #include <sys/types.h>
+#define HAVE_LOCALE 1 /* TODO: remove me when glima's patch gets in */
+#ifdef HAVE_LOCALE
+#include <locale.h>
+#endif
 
 #include "sol-util.h"
+#include "sol-log.h"
 #include "sol-str-slice.h"
+
+#ifdef HAVE_LOCALE
+static locale_t c_locale;
+static void
+clear_c_locale(void)
+{
+    freelocale(c_locale);
+    c_locale = NULL;
+}
+
+static bool
+init_c_locale(void)
+{
+    if (c_locale)
+        return true;
+
+    c_locale = newlocale(LC_ALL_MASK, "C", NULL);
+    if (!c_locale)
+        return false;
+
+    atexit(clear_c_locale);
+    return true;
+}
+#endif
 
 void *
 sol_util_memdup(const void *data, size_t len)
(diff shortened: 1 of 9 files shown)


src/shared/sol-util.c
```diff
@@ -37,10 +37,39 @@
 #include <stdlib.h>
 #include <string.h>
 #include <sys/types.h>
+#define HAVE_LOCALE 1 /* TODO: remove me when glima's patch gets in */
+#ifdef HAVE_LOCALE
+#include <locale.h>
```
lpereira: Isn't `locale.h` part of C89?

src/shared/sol-util.c
```diff
@@ -52,6 +81,104 @@ sol_util_memdup(const void *data, size_t len)
...
+     * http://www.netlib.org/fp/dtoa.c), we take the short path to
+     * call libc.
+     */
+    tmpbuf = strndupa(nptr, len);
+
+    errno = 0;
+#ifdef HAVE_LOCALE
+    if (!use_locale) {
+#ifdef HAVE_STRTOD_L
+        value = strtod_l(tmpbuf, &tmpbuf_endptr, c_locale);
```
lpereira: I'd provide a `strtod_l()` in `sol-missing.h`.

src/shared/sol-util.c
```diff
@@ -52,6 +81,104 @@ sol_util_memdup(const void *data, size_t len)
...
+                        if (*src == '.')
+                            *dst = *lc->decimal_point;
+                        else if (*src == *lc->decimal_point)
+                            *dst = '.';
+                    }
+                } else {
+                    char *transl;
+                    unsigned count = 0;
+
+                    for (src = tmpbuf; src < src_end; src++) {
```
lpereira: I don't think there'll be more than one decimal separator. Having said that, I believe that just using `strstr()` to find `lc->decimal_point` in the to-be-parsed string, copying whatever's before that, copying the decimal point, then copying whatever was after the decimal point, will be faster and shorter. The length of `transl` can be just `len + strlen(lc->decimal_point) + 1`.

src/shared/sol-util.c
barbieri: AFAIK this can be disabled and some small OS will not provide it anyway.

src/shared/sol-util.c
barbieri: It's not that simple or useful to just emulate `strtod_l()` in general.  In this patch we mimic a well known locale, "C", and take short paths. If we say we support users giving any `locale_t`, not only "C", then we must do much more work.

src/shared/sol-util.c
barbieri: I did this in my first version, but it may cause us problems to set the "endptr" in the same way `strtod()` would, thus I'm doing the "costly" version that replaces them all -- better safe than sorry, and it shouldn't impact much if there is a single '.'

What I could do is to use count in the second loop and stop earlier if all was done, resulting in a memcpy() of the rest. Not sure it would matter much.

barbieri:
merged with removed TODOS since @glima's patch with `HAVE_LOCALE` was merged.

lpereira:
I'll issue another PR tomorrow fixing users of strtod() in the code base.

PR #441, opened 2015-08-05
Iio v2

edersondisouza:
v2:
List of changes is too big to detail - but I should have addressed all issues raised on v1, or at least answered then. Highlights:
 - using config struct on API - avoiding lots of calls on unknown order on setup;
 - API calls to read data simplified - buffer reading is now implicit when buffer is enabled, but this only makes sense on 'data reader' callback.
 - using direction-vector type of scale and offset options on node - needed a small fix on generator;
 - `sol_util_vwrite_file` rewritten - it was hiding errors.
Issue about allowing a string to be used to identify a device will be addressed on a future PR, on my TODO list.


src/shared/sol-iio.c
```diff
@@ -0,0 +1,1094 @@
...
+    if (!check_file_existence(SYSFS_TRIGGER_SYSFS_ADD_TRIGGER)) {
+        SOL_WRN("No 'iio_sysfs_tigger' under '/sys/bus/iio/devices'."
+            " Missing 'modprobe iio-trig-sysfs'?");
+        return false;
+    }
+
+    /* TODO check available trigger indice instead of using a random one */
+    /* Create new trigger */
+    i = rand();
+    if (sol_util_write_file(SYSFS_TRIGGER_SYSFS_ADD_TRIGGER, "%d", i) < 0)
```
barbieri: I recall @lpereira said to just write to the `SYSFS_TRIGGER_SYSFS_ADD_TRIGGER` without prior `check_file_existence()`. To print the same message you can check for `-ENOENT`.

src/shared/sol-iio.c
```diff
@@ -0,0 +1,1094 @@
...
+    while (result) {
+        ssize_t ret;
+
+        ret = read(fd, device->buffer, device->buffer_size);
+        if (ret < 0) {
+            if (errno == EINTR)
+                continue;
+            else if (errno == EAGAIN || errno == EWOULDBLOCK)
+                break;
+            else {
```
barbieri: what about that idea to export the function from `sol-util-linux.c` that does this inner loop to read data?

Also, here you should retry if `EAGAIN || EWOULDBLOCK` (actually they are the same), with a upper bound. You do so because more data may be available right before you exit... then it's not worth to go back to the loop, instead try again couple of times.

BUT, AFAIU you're going to change this to use `mmap()`, in such case you just use the `device_reader_cb()` to be notified about more data, the read would act directly on the mmape'd memory -- no loops at all.

src/shared/sol-iio.c
```diff
@@ -0,0 +1,1094 @@
...
+                result = false;
+                break;
+            }
+        }
+
+        if ((size_t)ret == device->buffer_size) {
+            if (device->reader_cb) {
+                device->reader_cb((void *)device->reader_cb_data, device);
+            }
+        }
```
barbieri: again, if you keep the `read()` version you should account partial reads and change parameters of subsequent read requests.

src/shared/sol-iio.c
```diff
@@ -0,0 +1,1094 @@
...
+     * to it */
+    r = craft_filename_path(path, sizeof(path), DEVICE_NAME_PATH,
+        device->device_id);
+    if (!r) {
+        SOL_WRN("Could not open IIO device%d name file", device->device_id);
+        goto error;
+    }
+    device->name_fd = open(path, O_RDONLY | O_CLOEXEC);
+    if (device->name_fd == -1) {
+        SOL_WRN("Could not open IIO device%d name file", device->device_id);
```
barbieri: would help debug to print out the path you're opening.

src/shared/sol-iio.c
edersondisouza: About using `sol-util-linux.c`, I thought about doing on another PR that would touch all other places you've suggested (to avoid touching many unrelated files on this PR), but I can do it by steps (issuing another PR only for changing other files). 
About the `mmap` and high speed stuff, it isn't upstream yet, it's only on analogdevices linux (https://github.com/analogdevicesinc/linux). Not sure if it worth doing now.

src/shared/sol-iio.h
```diff
@@ -0,0 +1,141 @@
...
+
+struct sol_iio_channel_config {
+#define SOL_IIO_CHANNEL_CONFIG_API_VERSION (1)
+    uint16_t api_version;
+    double scale; /**< Channel scale, to be applied to raw readings. -1 uses device default. Some devices share scale among all channels, so changing one will change all. If, in this case, different channels set different scales the result is unknown. */
+    bool use_custom_offset; /**< If true, will use user defined offset on member #offset of this struct */
+    int offset; /**< Channel offset, to be added to raw readings. Some devices share offset among all channels, so changing one will change all. If, in this case, different channels set different offsets the result is unknown. */
+};
+
+#define SOL_IIO_CHANNEL_CONFIG_INIT { .scale = -1.0, .use_custom_offset = false }
```
barbieri: missing api_version here, no?

edersondisouza: I'm not sure, really. This is for initialising default values, shouldn't the user know which `api_version` [s]he's using?

src/shared/sol-iio.c
barbieri: Ok about doing the sol-util-linux in another PR.

As for `mmap()`, too bad :-( What are we going to do? Try `mmap()` and see if it succeeds? What happens in the other cases? Maybe it's not as fast, but you get the kernel doing emulation for you (ie: mmap works, but its not zero-copy).

src/shared/sol-iio.h
barbieri: This check is to link/sync between user version and library version. Then you provide a number you got at compile time (user version), distribute the binary. The libsoletta may be updated in the system and your binary not, in that case these numbers will get out-of-sync and your check will raise an issue.

Then, yes, you should put the .api_version = SOL_IIO_CHANNEL_CONFIG_API_VERSION in that macro.

src/shared/sol-iio.h
```diff
@@ -0,0 +1,141 @@
...
+    int buffer_size; /**< size of reading buffer. 0: use device default; -1: disable buffer and readings will be performed on channel files on sysfs. */
+    int sampling_frequency; /**< Device sampling frequency. -1 uses device default */
+};
+
+struct sol_iio_channel_config {
+#define SOL_IIO_CHANNEL_CONFIG_API_VERSION (1)
+    uint16_t api_version;
+    double scale; /**< Channel scale, to be applied to raw readings. -1 uses device default. Some devices share scale among all channels, so changing one will change all. If, in this case, different channels set different scales the result is unknown. */
+    bool use_custom_offset; /**< If true, will use user defined offset on member #offset of this struct */
+    int offset; /**< Channel offset, to be added to raw readings. Some devices share offset among all channels, so changing one will change all. If, in this case, different channels set different offsets the result is unknown. */
```
barbieri: would be nice to place `use_custom_offset` at the end.

src/shared/sol-iio.c
edersondisouza: I can test if it doesn't work gracefully; weird thing is - I can't test if it does work. There are a bunch of `ioctl` to be performed to enable this, so I'll probably have to skip them and jump to the `mmap`.

edersondisouza:
v3 on #464

PR #442, opened 2015-08-05
Build more fbp based samples v3

dorileo:
## Changes
v2:
   + simplified the patch: sol-fbp-generator: add ERROR port support

v3:
   + changed error_port to static;
   + split make/kconfig as requested by @cmarcelo;
   + removed the menu entries and left a single one "flow samples" as suggested/discussed by/with @mbelluzzo and @bdilly;
   + new patches fixing issues found in the process of bringing samples back to like (i.e updated APIs changes, ports naming changes etc);

## Obs
This patch set is growing larger than I first intended, there's still some other issues I found (i.e sol-fbp-generator doing wrong things on generating some data structures, header resolution and so on). I'm leaving from here but plan to retake as soon as possible to fix the issues and include the remaining samples.

At least it was a good exercise to catch few - hidden - errors. :)


cmarcelo:
ACK with comments on one patch. If generetor doesn't support not using the namespace anymore, we can deal in a different patch after this series.

About the set growing, many of those patches you can consider my +2 and apply in isolation if this series continue: change node type, ERROR port support, update to use array ports, fix segfault.

dorileo:
applied, thanks.

PR #443, opened 2015-08-05
Make I2C asynchronous

zehortigoza:
The patches starting with squash will be squashed to I2C: Make API asynchronous, leaving this way to make easy each author check his code port.


zehortigoza:
@edersondisouza @glima @bdilly Could you guys review your nodes ports?

PR #444, opened 2015-08-05
build: dep-resolver set HAVE_* symbols

dorileo:
Instead of calculating the HAVE_* symbols for exec dependencies whenever
required this patch instructs dependency-resolver.py to generate those
symbols into Makefile.gen.


rchiossi:
+1

dorileo:
applied, thanks.

PR #445, opened 2015-08-05
generate-svg: show svg being generated

dorileo:
We're currently blind on what the cheat-sheet target is doing since it
takes so long to complete we don't have any user feedback on what it's
going on. This patch makes it more verbose to tell us what it's doing.


dorileo:
ok

dorileo:
changed and applied... thanks.

PR #446, opened 2015-08-05
Coverity open issues

mbelluzzo:
Coverity is green, but was listing 3 open issues to me. This pullreq should close them.

Please see commit messages for more explanation on some issues.


barbieri:
+1, I'd avoid changing the var name in the simplectype patch so the diff is smaller, but that's ok.

mbelluzzo:
I changed only the ordering, I can fix that before merge.

mbelluzzo:
merged

PR #448, opened 2015-08-05
Reduce the amount of allocations and code for options [v2]

cmarcelo:
Improvements to how we deal with struct sol_flow_node_options.

The second patch ensures that we simply pass thru the options from the spec if nothing is being exported, as discussed in #335.

### DIFF
--- a/data/scripts/sol-flow-node-type-gen.py.in
+++ b/data/scripts/sol-flow-node-type-gen.py.in
@@ -812,8 +812,6 @@ static void
 }
 """)
 
-    new_options_func = "NULL"
-    free_options_func = "NULL"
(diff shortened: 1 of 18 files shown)


cmarcelo:
@ibriano could you take another look? Patches 1 and 2 are unmodified. Patch 3 got some code changes due to the addition of named_opts.

This is rebased, deals with JS (currently doesn't uses options) and with simplectype.

cmarcelo:
Replaces #421.

ibriano:
Still looks good

cmarcelo:
Applied in 12b9675d431b68b4e38c89df51249ea9e04cbddd and parents.

PR #449, opened 2015-08-05
Now all string nodes are UTF-8-aware.

glima:
Tests expect that ICU is present, as usual.


glima:
@barbieri please fix the tests for semaphoreci :)

glima:
I missed more translations to ICU schema there, wait a minute. Also I've had complaints about the excess of ifdefs -- I might give separating the implementations a try.

glima:
Pushing a new version.

PR #450, opened 2015-08-05
sol-aio: Add C API to handle Analog I/O (V5)

mbelluzzo:
Diff from V4:
* tracking devices that are power on/off on riot impl.


mbelluzzo:
Merged, thanks!

PR #451, opened 2015-08-05
pwm: Track device power state on RIOT

mbelluzzo:
Since a device can be shared by N nodes using different channels, we can
only power off the device when no other node is using it, so we need to
keep track of this information.

### DIFF
--- a/src/lib/io/sol-pwm-riot.c
+++ b/src/lib/io/sol-pwm-riot.c
@@ -37,6 +37,7 @@
 #include "sol-log-internal.h"
 #include "sol-pwm.h"
 #include "sol-util.h"
+#include "sol-vector.h"
 
 #include "periph/pwm.h"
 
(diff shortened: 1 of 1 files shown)


src/lib/io/sol-pwm-riot.c
```diff
@@ -55,6 +63,54 @@ struct sol_pwm {
...
+
+    SOL_VECTOR_FOREACH_IDX (&_dev_ref, ref, i) {
+        if (ref->device == device) {
+            found = true;
+            break;
+        }
+    }
+
+    if (!found) {
+        ref = (struct dev_ref *)sol_vector_append(&_dev_ref);
```
barbieri: no casts

src/lib/io/sol-pwm-riot.c
```diff
@@ -55,6 +63,54 @@ struct sol_pwm {
...
+_power_on(const int device)
+{
+    uint16_t i;
+    bool found = false;
+    struct dev_ref *ref;
+
+    SOL_VECTOR_FOREACH_IDX (&_dev_ref, ref, i) {
+        if (ref->device == device) {
+            found = true;
+            break;
```
barbieri: could ref++ and return right here.

mbelluzzo:
It must be CIPA's influence, incoming v2

PR #452, opened 2015-08-05
test: add ASSERT_SIZE_T_EQ/ASSERT_SIZE_T_NE() macros

dorileo:
This patch adds the macros ASSERT_SIZE_T_EQ() and ASSERT_SIZE_T_NE() so
we can check for size_t variables.

### DIFF
--- a/src/test/test.h
+++ b/src/test/test.h
@@ -74,27 +74,39 @@ int test_main(struct test *start, struct test *stop, void (*reset_func)(void), i
         }                                                               \
     } while (0)
 
-#define ASSERT_INT_EQ(expr_a, expr_b)                                   \
+#define ASSERT_NUMBER_EQ(type, expr_a, expr_b)                          \
     do {                                                                \
-        const int __value_expr_a = expr_a;                              \
-        const int __value_expr_b = expr_b;                              \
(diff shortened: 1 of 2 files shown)


barbieri:
+1, but we can make the macro use the same as @lucasdemarchi did for sol-log.h `SOL_INT_CHECK()`

src/test/test.h
```diff
@@ -74,28 +74,40 @@ int test_main(struct test *start, struct test *stop, void (*reset_func)(void), i
...
         }                                                               \
     } while (0)
 
-#define ASSERT_INT_NE(expr_a, expr_b)                                   \
+#define ASSERT_NUMBER_NE(type, expr_a, expr_b)                          \
     do {                                                                \
-        const int __value_expr_a = expr_a;                              \
-        const int __value_expr_b = expr_b;                              \
+        const type __value_expr_a = expr_a;                              \
+        const type __value_expr_b = expr_b;                              \
```
lpereira: Use `__builtin_types_compatible_p()` to assert that `typeof(expr_a)` is compatible with `typeof(expr_b)`.

You can also use `_Generic` to pick the formatting string.

Then there's no need to have different macros.

PR #455, opened 2015-08-06
Fix for sol-fbp-generator

lgywata:
Types that have subflows must call their init_type() function
before being assigned to the node->type.


cmarcelo:
+2

PR #456, opened 2015-08-06
OS-supported feature based dependencies and RIOT

ibriano:
Introducing a novel way to mark the dependencies of a feature in our build
system, so we don't need to use PLATFORM_* to know if something is or not
supported. Instead, when kbuild starts, it will detect which OS is building
for based on that the dependency resolver says, and choose an OS-specific
file to include that selects the features we have support for that OS.

Then come some changes necessary to make Soletta build for upstream RIOT.


src/lib/comms/Makefile
```diff
@@ -1,49 +1,38 @@
...
+    sol-network-riot.o
 
-    ifeq (y,$(SOL_PLATFORM_LINUX))
-    obj-comms-$(COMMS) += \
-        sol-network-linux.o
-    endif
-endif
+obj-networking-$(PLATFORM_LINUX) += \
+    sol-network-linux.o \
+    sol-socket-linux.o
```
barbieri: please take the time and make these files follow the other implementation pattern "sol-XXXX-impl-linux".

src/lib/io/Kconfig
```diff
@@ -1,3 +1,54 @@
...
+config USE_I2C
+    bool "I2C Support"
+    depends on FEATURE_HW_I2C
+    default y
+
+config USE_GPIO
+    bool "GPIO Support"
+    depends on FEATURE_HW_GPIO
+    default y
+
```
barbieri: Please take the time to document all the hw access, try wikipedia's short description or something that can clarify to users that may not know what it is.

src/lib/io/Makefile
```diff
@@ -1,122 +1,70 @@
...
-ifeq (y,$(PLATFORM_RIOTOS))
-ifeq (y, $(USE_AIO))
-obj-io-$(IO) += \
+obj-io-aio-$(USE_AIO) := \
+    sol-aio-common.o
+obj-io-aio-$(PLATFORM_RIOTOS) += \
     sol-aio-riot.o
-endif # USE_AIO
+obj-io-aio-$(PLATFORM_LINUX) += \
+    sol-aio-linux.o
```
barbieri: as in the other comment, take the time to rename to "-impl"

ibriano:
I'll address those comments separately. Merged.

PR #459, opened 2015-08-06
sol-util: implement overflow checks for mul/add operations

dorileo:
This patch implements a overflow check fallback - for cases where
__builtin_mul_overflow() and __builtin_add_overflow() are not available.

The lack of overflow check was resulting in test-json failure - since
we're checking for potential errors on string -> number conversion.

### DIFF
--- a/src/shared/sol-util.h
+++ b/src/shared/sol-util.h
@@ -47,6 +47,11 @@
 #include "sol-util-linux.h"
 #endif
 
+/** constants used to calculate mul/add operations overflow */
+#define OVERFLOW_TYPE(type) ((type)1 << (sizeof(type) * 4))
+#define OVERFLOW_UINT64 OVERFLOW_TYPE(uint64_t)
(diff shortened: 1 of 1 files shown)


src/shared/sol-util.h
```diff
@@ -207,6 +215,9 @@ sol_util_uint64_mul(const uint64_t a, const uint64_t b, uint64_t *out)
     if (__builtin_mul_overflow(a, b, out))
         return -EOVERFLOW;
 #else
+    if ((a >= OVERFLOW_UINT64 || b >= OVERFLOW_UINT64) &&
+        a > 0 && SIZE_MAX / a < b)
```
lpereira: `size_t` may be defined to 32-bit. Please use `UINT64_MAX` from `stdint.h`

dorileo: all right... I knew we'd end up with this comment :)

dorileo:
fixed the comments and applied... thanks.

PR #460, opened 2015-08-06
Only enable locale support if glib have freelocale()

zehortigoza:
newlocale() and freelocale() are available since glibc 2.10,
on the Intel Galileo SDK we have glibc 2.00.

### DIFF
--- a/data/jsons/dependencies.json
+++ b/data/jsons/dependencies.json
@@ -227,7 +227,7 @@
 	    "headers": [
 		"<locale.h>"
 	    ],
-	    "fragment": "setlocale(LC_ALL, NULL);"
+	    "fragment": "freelocale(NULL);"
 	},
 	{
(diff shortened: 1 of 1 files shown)


lpereira:
Please fix the commit message: it's glibc, not glib (different libraries).

data/jsons/dependencies.json
```diff
@@ -227,7 +227,7 @@
 	    "headers": [
 		"<locale.h>"
 	    ],
-	    "fragment": "setlocale(LC_ALL, NULL);"
+	    "fragment": "freelocale(NULL);"
```
barbieri: hum, this is not the way to go. @glima need only `setlocale()`, and I guess other places will do as well.

Add a new check for `newlocale` and use `HAVE_NEWLOCALE` at the call site.

PR #461, opened 2015-08-06
mqtt

rchiossi:
@vcgomes @lpereira could you guys take a look?

### DIFF
--- a/src/lib/comms/include/sol-mqtt.h
+++ b/src/lib/comms/include/sol-mqtt.h
@@ -1,0 +1,393 @@
+/*
+ * This file is part of the Soletta Project
+ *
+ * Copyright (C) 2015 Intel Corporation. All rights reserved.
+ *
+ * Redistribution and use in source and binary forms, with or without
+ * modification, are permitted provided that the following conditions
+ * are met:
+ *
+ *   * Redistributions of source code must retain the above copyright
+ *     notice, this list of conditions and the following disclaimer.
+ *   * Redistributions in binary form must reproduce the above copyright
+ *     notice, this list of conditions and the following disclaimer in
+ *     the documentation and/or other materials provided with the
+ *     distribution.
+ *   * Neither the name of Intel Corporation nor the names of its
+ *     contributors may be used to endorse or promote products derived
+ *     from this software without specific prior written permission.
+ *
+ * THIS SOFTWARE IS PROVIDED BY THE COPYRIGHT HOLDERS AND CONTRIBUTORS
+ * "AS IS" AND ANY EXPRESS OR IMPLIED WARRANTIES, INCLUDING, BUT NOT
+ * LIMITED TO, THE IMPLIED WARRANTIES OF MERCHANTABILITY AND FITNESS FOR
+ * A PARTICULAR PURPOSE ARE DISCLAIMED. IN NO EVENT SHALL THE COPYRIGHT
+ * OWNER OR CONTRIBUTORS BE LIABLE FOR ANY DIRECT, INDIRECT, INCIDENTAL,
+ * SPECIAL, EXEMPLARY, OR CONSEQUENTIAL DAMAGES (INCLUDING, BUT NOT
+ * LIMITED TO, PROCUREMENT OF SUBSTITUTE GOODS OR SERVICES; LOSS OF USE,
+ * DATA, OR PROFITS; OR BUSINESS INTERRUPTION) HOWEVER CAUSED AND ON ANY
+ * THEORY OF LIABILITY, WHETHER IN CONTRACT, STRICT LIABILITY, OR TORT
+ * (INCLUDING NEGLIGENCE OR OTHERWISE) ARISING IN ANY WAY OUT OF THE USE
+ * OF THIS SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
+ */
+
+#pragma once
+
+#include <stdbool.h>
+#include <stdlib.h>
+
+#ifdef __cplusplus
+extern "C" {
+#endif
+
+/**
+ * @file
+ * @brief Routines to handle MQTT protocol.
+ *
+ * Wrapper library for MQTT communication using the mosquitto MQTT
+ * library.
+ */
+
+/**
+ * @defgroup MQTT
+ * @ingroup Comms
+ *
+ * MQTT is a machine-to-machine (M2M)/"Internet of Things" connectivity
+ * protocol. It was designed as an extremely lightweight
+ * publish/subscribe messaging transport. It is useful for connections
+ * with remote locations where a small code footprint is required
+ * and/or network bandwidth is at a premium. For example, it has been
+ * used in sensors communicating to a broker via satellite link, over
+ * occasional dial-up connections with healthcare providers, and in a
+ * range of home automation and small device scenarios. It is also ideal
+ * for mobile applications because of its small size, low power usage,
+ * minimised data packets, and efficient distribution of information to
+ * one or many receivers
+ *
+ * @{
+ */
+
+/**
+ * @struct sol_mqtt
+ *
+ * @brief MQTT Object
+ *
+ * @see sol_mqtt_connect
+ *
+ * This object is the abstraction of a MQTT session. This is the base
+ * structure for all MQTT operations and is obtained through the
+ * sol_mqtt_connect API.
+ */
+struct sol_mqtt;
+
+/**
+ * @struct sol_mqtt_message
+ *
+ * @brief MQTT Message
+ *
+ * This object is the abstraction of a MQTT message and is the base for
+ * publishing and receiving data to/from the broker.
+ */
+struct sol_mqtt_message;
+
+/**
+ * @typedef sol_mqtt_qos
+ *
+ * @brief MQTT QOS level for message delivery
+ */
+typedef enum {
+    /**
+     * The message is delivered according to the capabilities of the
+     * underlying network. No response is sent by the receiver and no
+     * retry is performed by the sender. The message arrives at the
+     * receiver either once or not at all.
+     */
+    SOL_MQTT_QOS_AT_MOST_ONCE = 0,
+
+    /**
+     * This quality of service ensures that the message arrives at the
+     * receiver at least once. A QoS 1 PUBLISH Packet has a Packet
+     * Identifier in its variable header and is acknowledged by a PUBACK
+     * Packet.
+     */
+    SOL_MQTT_QOS_AT_LEAST_ONCE = 1,
+
+    /**
+     * This is the highest quality of service, for use when neither
+     * loss nor duplication of messages are acceptable. There is an
+     * increased overhead associated with this quality of service.
+     */
+    SOL_MQTT_QOS_EXACTLY_ONCE = 2,
+} sol_mqtt_qos;
+
+/**
+ * @enum sol_mqtt_conn_status
+ *
+ * @brief Connection status
+ */
+enum sol_mqtt_conn_status {
+    /**
+     * Disconnected due to unexpected reasons.
+     */
+    SOL_MQTT_DISCONNECTED = -1,
+
+    /**
+     * Succesfully conencted to the broker
+     */
+    SOL_MQTT_CONNECTED = 0,
+
+    /**
+     * MQTT protocol rejected by the broker
+     */
+    SOL_MQTT_WRONG_PROTOCOL = 1,
+
+    /**
+     * Client ID rejected by the broker
+     */
+    SOL_MQTT_ID_REJECTED = 2,
+
+    /**
+     * Broker anavailable at provided host
+     */
+    SOL_MQTT_UNAVAILABLE = 3,
+};
+
+/**
+ * @struct sol_mqtt_config
+ *
+ * @brief Server Configuration and callback handlers
+ */
+struct sol_mqtt_config {
+#define SOL_MQTT_CONFIG_API_VERSION (1)
+    /**
+     * Should always be set to SOL_MQTT_CONFIG_API_VERSION
+     */
+    uint16_t api_version;
+
+    /**
+     * If set, the broker will drop all messages and subscriptions when
+     * the client disconnects. Must be set if no client id is provided.
+     */
+    bool clean_session;
+
+    /**
+     * Time interval between PING messages that should be sent by the
+     * broker to the client.
+     */
+    time_t keepalive;
+
+    /**
+     * NULL terminated string that should be used as client ID. If
+     * not set, clean_session must be set to true.
+     */
+    char *client_id;
+
+    /**
+     * A message that the broker should send when the client disconnects.
+     */
+    struct sol_mqtt_message *will;
+
+    /**
+     * @brief On connect callback
+     *
+     * @param mqtt MQTT Object
+     * @param data User provided data
+     *
+     * @see sol_mqtt_connect
+     *
+     * Callback called when a connect request has been processed
+     */
+    void (*connect) (struct sol_mqtt *mqtt, void *data);
+
+    /**
+     * @brief On disconnect callback
+     *
+     * @param mqtt MQTT Object
+     * @param data User provided data
+     *
+     * @see sol_mqtt_connect
+     *
+     * Callback called when the client has disconnected from the
+     * broker.
+     */
+    void (*disconnect) (struct sol_mqtt *mqtt, void *data);
+
+    /**
+     * @brief On publish callback
+     *
+     * @param mqtt MQTT Object
+     * @param data User provided data
+     *
+     * @see sol_mqtt_connect
+     *
+     * Callback called when a publish request has been processed.
+     */
+    void (*publish) (struct sol_mqtt *mqtt, void *data);
+
+    /**
+     * @brief On message callback
+     *
+     * @param mqtt MQTT Object
+     * @param message Message received from the broker
+     * @param data User provided data
+     *
+     * @see sol_mqtt_connect
+     *
+     * Callback called when a message is received from the broker.
+     * This does not include PING messages, only messages incoming due
+     * to publish request from other clients or the broker itself.
+     *
+     * The memory associated to the message object will be freed after
+     * the callback returns.
+     */
+    void (*message) (struct sol_mqtt *mqtt, const struct sol_mqtt_message *message, void *data);
+
+    /**
+     * @brief On subscribe callback
+     *
+     * @param mqtt MQTT Object
+     * @param data User provided data
+     *
+     * @see sol_mqtt_connect
+     *
+     * Callback called when a subscribe request has been processed.
+     */
+    void (*subscribe) (struct sol_mqtt *mqtt, void *data);
+
+    /**
+     * @brief On unsubscribe callback
+     *
+     * @param mqtt MQTT Object
+     * @param data User provided data
+     *
+     * @see sol_mqtt_connect
+     *
+     * Callback called when a unsubscribe request has been processed.
+     */
+    void (*unsubscribe) (struct sol_mqtt *mqtt, void *data);
+};
+
+/**
+ * @brief Connect to a MQTT broker
+ *
+ * @param host The host address of the MQTT broker
+ * @param port The host port to connect to
+ * @param config Configuration and callbacks
+ * @param data User data provided to the callbacks
+ *
+ * @return New mqtt object on sucess, NULL otherwise
+ */
+struct sol_mqtt *sol_mqtt_connect(const char *host, int port, const struct sol_mqtt_config *config, void *data);
+
+/**
+ * @brief Reestablish the connection to the MQTT broker
+ *
+ * @param mqtt MQTT Object
+ *
+ * @return 0 on success, -EINVAL otherwise
+ */
+int sol_mqtt_reconnect(struct sol_mqtt *mqtt);
+
+/**
+ * @brief Disconnect from MQTT Broker
+ *
+ * @param mqtt MQTT Object;
+ *
+ * @see sol_mqtt_connect
+ *
+ * Terminate the connection to the broker and free the resources
+ * associated to the mqtt object.
+ */
+void sol_mqtt_disconnect(struct sol_mqtt *mqtt);
+
+/**
+ * @brief Get connection status
+ *
+ * @param MQTT Object
+ *
+ * @see sol_mqtt_conn_status
+ *
+ * @return sol_mqtt_conn_status code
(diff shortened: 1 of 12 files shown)


src/lib/comms/include/sol-mqtt.h
```diff
@@ -0,0 +1,393 @@
...
+     * Time interval between PING messages that should be sent by the
+     * broker to the client.
+     */
+    time_t keepalive;
+
+    /**
+     * NULL terminated string that should be used as client ID. If
+     * not set, clean_session must be set to true.
+     */
+    char *client_id;
```
barbieri: const

src/lib/comms/include/sol-mqtt.h
```diff
@@ -0,0 +1,393 @@
...
+
+    /**
+     * A message that the broker should send when the client disconnects.
+     */
+    struct sol_mqtt_message *will;
```
barbieri: const? not clear the ownership and modifications of this.

src/lib/comms/include/sol-mqtt.h
```diff
@@ -0,0 +1,393 @@
...
+ * @brief Connect to a MQTT broker
+ *
+ * @param host The host address of the MQTT broker
+ * @param port The host port to connect to
+ * @param config Configuration and callbacks
+ * @param data User data provided to the callbacks
+ *
+ * @return New mqtt object on sucess, NULL otherwise
+ */
+struct sol_mqtt *sol_mqtt_connect(const char *host, int port, const struct sol_mqtt_config *config, void *data);
```
barbieri: const void data...

Also, I/O configs are passing the user data inside the config itself.. so better move this pointer there.

src/lib/comms/include/sol-mqtt.h
```diff
@@ -0,0 +1,393 @@
...
+     * @brief On unsubscribe callback
+     *
+     * @param mqtt MQTT Object
+     * @param data User provided data
+     *
+     * @see sol_mqtt_connect
+     *
+     * Callback called when a unsubscribe request has been processed.
+     */
+    void (*unsubscribe) (struct sol_mqtt *mqtt, void *data);
```
barbieri: In soletta we're giving user data as the first parameter for callbacks, see `sol-gpio.h`

src/lib/comms/include/sol-mqtt.h
```diff
@@ -0,0 +1,393 @@
...
+/**
+ * @brief Get connection status
+ *
+ * @param MQTT Object
+ *
+ * @see sol_mqtt_conn_status
+ *
+ * @return sol_mqtt_conn_status code
+ */
+int sol_mqtt_get_connection_status(struct sol_mqtt *mqtt);
```
barbieri: const

src/lib/comms/include/sol-mqtt.h
```diff
@@ -0,0 +1,393 @@
...
+ * @brief Ask the Broker to be subscribed to a given topic
+ *
+ * @params mqtt MQTT Object
+ * @params topic Topic to subscribe
+ * @param topic_len Topic length in bytes
+ * @param qos MQTT QOS that should be used by the subscribe message
+ *
+ * @return 0 on success, -EINVAL otherwise
+ */
+int sol_mqtt_subscribe(struct sol_mqtt *mqtt, char *topic, size_t topic_len, sol_mqtt_qos qos);
```
barbieri: const char *topic

Is topic null-terminted? If so, can we remove `topic_len`?

src/lib/comms/include/sol-mqtt.h
```diff
@@ -0,0 +1,393 @@
...
+ * @param topic Topic where the message will be published
+ * @param topic_len Length in bytes of the topic
+ * @param payload Payload of the message
+ * @param payload_len Length in bytes of the payload
+ * @param qos MQTT QOS that should be used when sending the message
+ * @param retain If the message should be retained by the broker
+ *
+ * @return New message object on success, NULL otherwise
+ */
+struct sol_mqtt_message *sol_mqtt_message_new(char *topic, size_t topic_len, void *payload,
```
barbieri: const char *topic

barbieri: const void *payload? Also you could use the `const struct sol_buffer` to pass payload. If you make it clear it's const, then you're not doing any references of changes, just using both memory and size in a single structure.

src/lib/comms/include/sol-mqtt.h
```diff
@@ -0,0 +1,393 @@
...
+        size_t payload_len, sol_mqtt_qos, bool retain);
+
+/**
+ * @brief Free the memory associated to a given message
+ *
+ * @param message MQTT message to be copied
+ *
+ * @return Copy of message on success, NULL otherwise
+ */
+struct sol_mqtt_message *sol_mqtt_message_copy(struct sol_mqtt_message *message);
```
barbieri: const

src/lib/comms/include/sol-mqtt.h
```diff
@@ -0,0 +1,393 @@
...
+/**
+ * @brief Get the topic of a given MQTT message
+ *
+ * @param message MQTT message to get the topic from
+ * @param buf Buffer to write the topic to
+ * @param size Size in bytes of buf
+ *
+ * @return Number of bytes written to buf
+ */
+size_t sol_mqtt_get_topic(const struct sol_mqtt_message *message, char *buf, size_t size);
```
barbieri: this API is kinda awkward... can we return a pointer to an internal buffer? Like:

`const char *sol_mqtt_get_topic(msg, size_t *size)`

src/lib/comms/include/sol-mqtt.h
```diff
@@ -0,0 +1,393 @@
...
+
+/**
+ * @brief Get address and size of the message payload
+ *
+ * @param payload Upon return, will be set to the address of the payload memory
+ * @param len Upon return, will be set to the size in bytes of the payload memory.
+ *
+ * @return 0 on success, -EINVAL otherwise
+ */
+int sol_mqtt_get_payload(const struct sol_mqtt_message *message, void **payload, size_t *len);
```
barbieri: is a null payload valid? We could rework this api to make it better. If NULL payload is valid, considering the proposal for `get_topic()` above, we could do:

`const void *sol_mqtt_get_payload(msg, size_t *size)`

with `SIZE_MAX` being error.

src/lib/comms/sol-mqtt.c
```diff
@@ -0,0 +1,579 @@
+/*
```
barbieri: `sol-mqtt-impl-mosquitto.c`

src/lib/comms/sol-mqtt.c
```diff
@@ -0,0 +1,579 @@
...
+    struct sol_fd *socket_write;
+
+    struct sol_timeout *connect_timeout;
+    struct sol_timeout *disconnect_timeout;
+    struct sol_timeout *publish_timeout;
+    struct sol_timeout *subscribe_timeout;
+    struct sol_timeout *unsubscribe_timeout;
+    struct sol_timeout *message_timeout;
+
+    const struct sol_mqtt_config *config;
```
barbieri: if we're keeping the handle, we must document that, so users don't allocate this on a stack that may go away.

IMO we should keep a copy (ie: inline `struct sol_mqtt_config` and `memcpy()` it).

src/lib/comms/include/sol-mqtt.h
rchiossi: Nop, we are not. Check sol-coap.h

src/lib/comms/sol-mqtt.c
```diff
@@ -0,0 +1,579 @@
...
+        goto remove;
+
+    return true;
+
+remove:
+    if (active_flags & SOL_FD_FLAGS_OUT)
+        mqtt->socket_write = NULL;
+    else
+        mqtt->socket_read = NULL;
+
```
barbieri: I don't think this logic is right, but I didn't review in depth. Are you sure `active_flags` will be `SOL_FD_FLAGS_OUT` when we get an error from `mqtt->socket_write`? If that's not the case, then you'd be clearing `mqtt->socket_read` while `return false` would delete the `mqtt->socket_read`.

I'd create 2 functions, one for `mqtt->socket_read()` and the other for `mqtt->socket_write()` and avoid messing up.

src/lib/comms/sol-mqtt.c
```diff
@@ -0,0 +1,579 @@
...
+    return false;
+}
+
+/*
+ * When mosquitto calls the a user provided callback, it's internal lock is held
```
lpereira: s/it's/its/

src/lib/comms/sol-mqtt.c
```diff
@@ -0,0 +1,579 @@
...
+            mqtt->config->_cb(mqtt, mqtt->data); \
+        return false; \
+    }
+
+CALLBACK_WRAPPER(connect);
+CALLBACK_WRAPPER(disconnect);
+CALLBACK_WRAPPER(publish);
+CALLBACK_WRAPPER(subscribe);
+CALLBACK_WRAPPER(unsubscribe);
+
```
lpereira: `#undef CALLBACK_WRAPPER`

src/lib/comms/sol-mqtt.c
```diff
@@ -0,0 +1,579 @@
...
+    mqtt->socket_fd = mosquitto_socket(mqtt->mosq);
+    if (mqtt->socket_fd == -1) {
+        SOL_WRN("Unable to get socket file descriptor");
+        goto socket_error;
+    }
+
+    mqtt->socket_read = sol_fd_add(mqtt->socket_fd, SOL_FD_FLAGS_IN | SOL_FD_FLAGS_PRI,
+            sol_mqtt_event_loop, mqtt);
+    SOL_NULL_CHECK(mqtt->socket_read, NULL);
```
lpereira: Memory leaks on failure.

src/lib/comms/sol-mqtt.c
```diff
@@ -0,0 +1,579 @@
...
+
+    mqtt->socket_fd = mosquitto_socket(mqtt->mosq);
+    if (mqtt->socket_fd == -1) {
+        SOL_WRN("Unable to get socket file descriptor");
+        return -EINVAL;
+    }
+
+    mqtt->socket_read = sol_fd_add(mqtt->socket_fd, SOL_FD_FLAGS_IN | SOL_FD_FLAGS_PRI,
+            sol_mqtt_event_loop, mqtt);
+
```
lpereira: Check if watcher could be created.

src/lib/comms/sol-mqtt.c
barbieri: I mean, in the one for `read`:

```c
r = mosquito_loop_read(mqtt->mosq, 1);
r |= mosquitto_loop_misc(mqtt->mosq); /* NO IDEA WHAT IT DOES */
if (!r) {
   mqtt->socket_read = NULL;
   return false;
}
return true;
```

And for `write`:
```c
r = mosquitto_loop_write(mqtt->mosq, 1);
r |= mosquitto_loop_misc(mqtt->mosq); /* NO IDEA WHAT IT DOES */
if (!r || !mosquitto_want_write(mqtt->mosq)) {
   mqtt->socket_read = NULL;
   return false;
}
return true;
```

If this brings and ordering issue due read/write in mosquitto, then use a single `fd_add` and handler, flagging what we're doing:

```c
r = 0;
if (active_flags & SOL_FD_FLAGS_IN)
   r |= mosquitto_loop_read(mqtt->mosq, 1);
if (active_flags & SOL_FD_FLAGS_OUT)
  r |= mosquitto_loop_write(mqtt->mosq, 1);

r |= mosquitto_loop_misc(mqtt->mosq);

if (!r) {
   mqtt->fd_handler = NULL;
   return false;
}

wants_write = mosquitto_want_write(mqtt->mosq);
if (mqtt->watching_write && !wants_write) {
    mqtt->fd_handler = sol_fd_handler_add(mqtt->fd, SOL_FD_FLAGS_IN... );
    mqtt->watching_write = false;
    return false; /* delete the previous handler that also did write */
} else if (!mqtt->watching_write && wants_write) {
    mqtt->fd_handler = sol_fd_handler_add(mqtt->fd, SOL_FD_FLAGS_IN | SOL_FD_FLAGS_OUT... );
    mqtt->watching_write = true;
    return false; /* delete the previous handler that did not write */
}
```

Alternatively we can introduce a fd_set_flags() to allow changing the flags... should be more efficient.

src/lib/comms/sol-mqtt.c
```diff
@@ -0,0 +1,579 @@
...
+    struct sol_mqtt *mqtt = data;
+
+    SOL_NULL_CHECK(mqtt);
+
+    mqtt->connection_status = rc;
+
+    if (mqtt->connect_timeout || !mqtt->config || !mqtt->config->connect)
+        return;
+
+    mqtt->connect_timeout = sol_timeout_add(0, sol_mqtt_on_connect_wrapper, mqtt);
```
barbieri: why do we need to call from a timer? Is this being called from a thread?

hint: SOL_DBG() and other logs will show if running from a different thread.

src/lib/comms/sol-mqtt.c
```diff
@@ -0,0 +1,579 @@
...
+SOL_API int
+sol_mqtt_publish(struct sol_mqtt *mqtt, struct sol_mqtt_message *message)
+{
+    int r;
+
+    CHECK_INIT(-EINVAL);
+    SOL_NULL_CHECK(mqtt, -EINVAL);
+    SOL_NULL_CHECK(message, -EINVAL);
+
+    r = mosquitto_publish(mqtt->mosq, &message->id, message->topic, message->payload_len, message->payload,
```
barbieri: bummer... you request message->topic_len and just ignore it :-D

src/lib/comms/sol-mqtt.c
```diff
@@ -0,0 +1,579 @@
...
+            (int) message->qos, message->retain);
+
+    if (r != MOSQ_ERR_SUCCESS) {
+        SOL_WRN("Unable to publish to '%s'", message->topic);
+        return -EINVAL;
+    }
+
+    if (!mqtt->socket_write) {
+        mqtt->socket_write = sol_fd_add(mqtt->socket_fd, SOL_FD_FLAGS_OUT, sol_mqtt_event_loop, mqtt);
```
barbieri: isn't it clear to `mosquito_want_write() && mqtt->socket_write()`?

src/lib/comms/sol-mqtt.c
```diff
@@ -0,0 +1,579 @@
...
+{
+    struct sol_mqtt_message *message;
+
+    SOL_NULL_CHECK(topic, NULL);
+    SOL_NULL_CHECK(payload, NULL);
+
+    message = calloc(1, sizeof(struct sol_mqtt_message));
+    SOL_NULL_CHECK(message, NULL);
+
+    message->topic = calloc(1, topic_len + 1);
```
barbieri: you can calloc a single blob if we can't change the topic.

src/lib/comms/sol-mqtt.c
```diff
@@ -0,0 +1,579 @@
...
+    SOL_NULL_CHECK_GOTO(message->topic, topic_error);
+    memcpy(message->topic, topic, topic_len);
+
+    message->payload = sol_util_memdup(payload, payload_len);
```
barbieri: this is bad. What if we use `sol_blob` and take a reference? Then we release the reference whe it's not needed anymore.

src/lib/comms/sol-mqtt.c
```diff
@@ -0,0 +1,579 @@
...
+    size_t s;
+
+    SOL_NULL_CHECK(message, 0);
+    SOL_NULL_CHECK(buf, 0);
+
+    s = strlen(message->topic);
+    if (s > size)
+        s = size;
+
+    strncpy(buf, message->topic, s);
```
barbieri: bummer... not ensuring \0

src/lib/comms/include/sol-mqtt.h
barbieri: we should fix `sol-coap.h`, me looks at @lpereira and @vcgomes

src/lib/comms/include/sol-mqtt.h
rchiossi: It's onyl needed during connect. mosquitto will keep an internal copy of the relevant data.

src/lib/comms/include/sol-mqtt.h
rchiossi: ack

src/lib/comms/include/sol-mqtt.h
rchiossi: I thought about it, just kept it separate to keep the API similar to what we have elsewhere. I'm ok with moving it to config.

src/lib/comms/include/sol-mqtt.h
rchiossi: ack

src/lib/comms/include/sol-mqtt.h
rchiossi: This is only required when connect is called. Mosquitto will keep an internal copy of this, but our code doesn't really need it after that. Maybe it would be better to move it to the connect API, since it's connection specific rather than config specific.

src/lib/comms/include/sol-mqtt.h
rchiossi: indeed. Will change the API.

src/lib/comms/include/sol-mqtt.h
rchiossi: Makes sense, will take a look.

src/lib/comms/include/sol-mqtt.h
rchiossi: If we are getting the topic of a message we got on the on_message callback, the message, along with its topic will be freed once the callback returns, so the user would have to make a copy anyway. I find it weird that in that case you would get an object with a lifetime that's not clear.

src/lib/comms/include/sol-mqtt.h
rchiossi: A NULL payload should be valid since for publish messages, the payload is optional. Still, this API mimics the coap API which have similar bahavior. (see sol_coap_packet_get_payload).

src/lib/comms/sol-mqtt.c
rchiossi: Yeah, that was its name at first, but I got a feedback that we should move to a different name after we got a second implementation. I'm ok with the rename though.

src/lib/comms/sol-mqtt.c
rchiossi: Makes sense. I'll rework.

src/lib/comms/sol-mqtt.c
rchiossi: I'll look for the set_flags() solution. Besides that, I'm not convinced you can call the write_loop() without calling the read_loop() first due to mosquitto internal state being dependent on this order. I'll need to take a closer look at mosquitto code.

src/lib/comms/sol-mqtt.c
rchiossi: :)

src/lib/comms/sol-mqtt.c
rchiossi: ack

src/lib/comms/sol-mqtt.c
rchiossi: indeed. Dorileo also spotted this one.

src/lib/comms/sol-mqtt.c
rchiossi: ack

src/lib/comms/sol-mqtt.c
rchiossi: As we talked irl, this is like this because mosquitto is not reentrant, so we need to let it return to release the internal lock. There is a comment about this a couple lines above.

src/lib/comms/sol-mqtt.c
rchiossi: topic is a NULL terminated string here. We allocate one extra byte for the \0 when the message is created.

src/lib/comms/sol-mqtt.c
rchiossi: ack

src/lib/comms/sol-mqtt.c
rchiossi: ack

src/lib/comms/sol-mqtt.c
rchiossi: I'll look into it, may need some rework in other parts of the code.

src/lib/comms/sol-mqtt.c
rchiossi: well, whoever gets it, will get its length, also, ensuring \0 could possibly trim out the last byte since we don't control the dst buffer. Anyway, will change everywhere just to use/expect a NULL terminated string.

src/lib/comms/include/sol-mqtt.h
barbieri: that's the case with most APIs, particularly those in Soletta -- so there is no issue, but it would be good to document that the memory is valid until the caller returns, the main loop runs and other conditions that may change it.

It's just a better pattern to return, eventually you just want to print and exit. Also, you can't know the size to properly allocate a buffer.. returning a "const char*" will let user do strdup(), strlen() and friends.

src/lib/comms/include/sol-mqtt.h
barbieri: I'm not sure why CoAP is still like that. I've talked to @vcgomes many times that since CoAP is fixed to a UDP message and that size is known, the whole UDP packet should be allocated and filled with CoAP headers, the remaining bits are the payload. So the user would get the packet payload, not send it to Soletta.

@vcgomes didn't you change that? reasons?

rchiossi: But that's what happens. The pointer is used only to return the value. It is expected that the user provides an "empty" pointer that will be set to the payload upon return.

src/lib/comms/include/sol-mqtt.h
rchiossi: It is documented that the message object will live until the end of the callback, but there is no way to document that get_topic() returned value will live until the end of the callback since that will only be true if the user created code calls get_topic() inside on_message callback. I can document that the topic will live as much as the message it came from. It works, but I believe it may still be confusing for some users.
Anyway, I'll take this approach, let see how it goes.

PR #462, opened 2015-08-07
Reduce glib

barbieri:

### DIFF
--- a/src/lib/common/sol-mainloop-common.c
+++ b/src/lib/common/sol-mainloop-common.c
@@ -38,6 +38,17 @@
 #include "sol-mainloop-impl.h"
 #include "sol-mainloop-common.h"
 
+struct sol_mainloop_source_common {
+    const struct sol_mainloop_source_type *type;
+    const void *data;
+    bool ready;
+    bool remove_me;
+};
+
+static bool source_processing;
+static unsigned source_pending_deletion;
+static struct sol_ptr_vector source_vector = SOL_PTR_VECTOR_INIT;
+
 static bool run_loop;
 
 static bool timeout_processing;
@@ -58,9 +69,12 @@ timeout_compare(const void *data1, const void *data2)
 
 #ifdef PTHREAD
 
+static struct sol_ptr_vector source_v_process = SOL_PTR_VECTOR_INIT;
 static struct sol_ptr_vector timeout_v_process = SOL_PTR_VECTOR_INIT;
 static struct sol_ptr_vector idler_v_process = SOL_PTR_VECTOR_INIT;
 
+#define SOURCE_PROCESS source_v_process
+#define SOURCE_ACUM source_vector
 #define TIMEOUT_PROCESS timeout_v_process
 #define TIMEOUT_ACUM timeout_vector
 #define IDLER_PROCESS idler_v_process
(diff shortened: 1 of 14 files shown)


src/lib/common/sol-mainloop-common.c
```diff
@@ -296,6 +347,177 @@ sol_mainloop_common_idler_first(void)
...
+static inline void
+source_cleanup(void)
+{
+    struct sol_mainloop_source_common *source;
+    uint16_t i;
+
+    if (!source_pending_deletion)
+        return;
+
+    // Walk backwards so deletion doesn't impact the indices.
```
lpereira: Every single use of backwards-loop with sol-vector has this comment. I wonder if something akin to C++'s `std::remove_if` would be feasible with some macro trickery.

src/lib/common/sol-mainloop-common.c
```diff
@@ -296,6 +347,177 @@ sol_mainloop_common_idler_first(void)
...
+
+    return found;
+}
+
+bool
+sol_mainloop_common_source_get_next_timeout(struct timespec *timeout)
+{
+    bool r;
+
+    sol_mainloop_impl_lock();
```
lpereira: A few months ago I proposed a way of creating something similar to Java's `synchronized` blocks. Maybe it's time to unearth that?

ibriano:
Seems fine

src/lib/common/sol-mainloop-common.c
barbieri: could be, but this one is really an exception in usage since we don't want to keep it locked for long time, see the other functions.

This one is not even used, I just added it for completeness, only the _locked variant is used at the moment.

ulissesf: My opinion, but I'm not a big fan of those ```synchronized``` blocks. Not sure it's worth the effort even.

src/lib/common/sol-mainloop-common.c
```diff
@@ -296,6 +347,177 @@ sol_mainloop_common_idler_first(void)
...
+        struct timespec cur;
+        bool r;
+
+        if (source->remove_me)
+            continue;
+
+        if (!source->type->get_next_timeout)
+            continue;
+
+        r = source->type->get_next_timeout((void *)source->data, &cur);
```
ulissesf: @barbieri, can this somehow deadlock trying to get the mainloop lock again? Just asking since I didn't check everything on your commit. Executing callbacks with the mainloop lock held is something I avoided since it does easily leads to deadlocks. It might be in this case we're safe, not sure.

barbieri: it can... indeed. In theory the get_next_timeout() shouldn't be doing anything like changing the main loop, but who knows what people are doing.

I got a similar deadlock from calling dispose with the lock held, then sol-bus.c was deleting the sol_fd and it dead-locked.

ulissesf: Yeah, you shouldn't call any callback with this lock held. If the code has its own lock we might get even nastier deadlocks with the locks not being taken in the same order and so on.

barbieri:
Replaced by #477

PR #464, opened 2015-08-07
Iio v3

edersondisouza:
v3:
 - Exposing `sol_util_fill_buffer()` and using it;
 - More debug information;
 - Removed check for `/dev/iio:deviceX` existence on `sol_iio_open`: it should exist as sysfs was already checked - if it happens to not exist, an error will happen when trying to open the buffer;
 - Not checking `SYSFS_TRIGGER_SYSFS_ADD_TRIGGER` before writing;
 - `.min` and `.max` of packet sent adjusted to great and least values obtained from sample.

I intend to a issue new PR expanding use of `sol_util_fill_buffer()`.
I did *not* advanced on `mmap()` stuff. Analysing carefully libiio code showed that it just doesn't make sense without kernel patches: every detail about the mapping is gotten from kernel via `ioctl()` requests that are not implemented on upstream (there is actually a queue of buffers). When they cannot use the 'high speed', they fallback to normal `read()` (and `write()`).

v2:
List of changes is too big to detail - but I should have addressed all issues raised on v1, or at least answered then. Highlights:

    using config struct on API - avoiding lots of calls on unknown order on setup;
    API calls to read data simplified - buffer reading is now implicit when buffer is enabled, but this only makes sense on 'data reader' callback.
    using direction-vector type of scale and offset options on node - needed a small fix on generator;
    sol_util_vwrite_file rewritten - it was hiding errors. Issue about allowing a string to be used to identify a device will be addressed on a future PR, on my TODO list.


edersondisouza:
Merged. Besides fixing @glima issue, I also addressed a new warning raised by sanitizer by applying the following diff:

    -s_data = data | (s_data << channel->bits);
    +s_data = data | ~channel->mask;
 
I've also changed channel->mask to be a `uint64_t`.

PR #465, opened 2015-08-07
sol-fbp-generator: fix segfault

dorileo:
This patch is complementary for 4a5798f5f2cc2d45d6b8adda797cabe32ec723ab

Also we should initialize opts_as_string and check for it and avoid
stack initialization issues.


ricardotk:
ACK. Just remembert to fix identation before merging.

anselmolsm:
:+1:
The semaphore error is due to check-style. Fix the indentation and it's good to go.

dorileo:
applied - with the indentation fix. thanks.

PR #466, opened 2015-08-07
modules/flower-power: Add module with 3 node types

bdilly:
Parrot Flower Power is a wireless plant monitor that measures
and analyses the four elements crucial to plant's health:
sunlight, temperature, soil moisture and fertilizer.

Initial version provides node types to fetch data via
HTTP, to filter packets per id, and to get measures
values from packets.

Details can be obtained from its website,
<http://developer.parrot.com/flowerpower.html/>

@lpereira could you please take a look on this commit?


src/modules/flow/flower-power/flower-power.c
```diff
@@ -0,0 +1,709 @@
...
+#include <errno.h>
+#include <float.h>
+
+
+struct packet_type_flower_power_packet_data {
+    struct sol_drange fertilizer;
+    struct sol_drange light;
+    struct sol_drange temperature;
+    struct sol_drange water;
+    /* TODO: maybe convert it to int */
```
lpereira: Do we have a timestamp paket type?

src/modules/flow/flower-power/flower-power.c
```diff
@@ -0,0 +1,709 @@
...
+
+    if (!found_locations) {
+        SOL_DBG("No plants found on response");
+        return;
+    }
+
+    sol_json_scanner_init_from_token(&locations_scanner, &value);
+    SOL_JSON_SCANNER_ARRAY_LOOP (&locations_scanner, &token,
+        SOL_JSON_TYPE_OBJECT_START, reason) {
+        struct sol_drange fertilizer = {0, -DBL_MAX, DBL_MAX, DBL_MIN};
```
lpereira: Maybe we should have a `SOL_DRANGE_VALUE()` macro?

src/modules/flow/flower-power/flower-power.c
```diff
@@ -0,0 +1,709 @@
...
+        struct sol_drange water = {0, -DBL_MAX, DBL_MAX, DBL_MIN};
+        struct sol_drange temperature = {0, -DBL_MAX, DBL_MAX, DBL_MIN};
+        struct sol_drange light = {0, -DBL_MAX, DBL_MAX, DBL_MIN};
+        char *id = NULL, *timestamp = NULL;
+
+        SOL_JSON_SCANNER_OBJECT_LOOP_NEST (&locations_scanner, &token,
+            &key, &value, reason) {
+            if (sol_json_token_str_eq(&key, "fertilizer",
```
lpereira: I believe there's a macro in `sol-json.h` that helps these comparisons with constant strings (specify them only once, etc).

src/modules/flow/flower-power/flower-power.c
```diff
@@ -0,0 +1,709 @@
...
+
+#include <sol-http-client.h>
+#include <sol-json.h>
+#include <sol-util.h>
+
+#include <errno.h>
+#include <float.h>
+
+
+struct packet_type_flower_power_packet_data {
```
lpereira: Do we need a packet type just for flower power? Because then we'll need conversion nodes to extract individual values. Sort of cumbersome.

src/modules/flow/flower-power/flower-power.c
```diff
@@ -0,0 +1,709 @@
...
+    packet_type_flower_power->light = in->light;
+    packet_type_flower_power->temperature = in->temperature;
+    packet_type_flower_power->water = in->water;
+
+    return 0;
+}
+
+#define PACKET_TYPE_FLOWER_POWER_PACKET_TYPE_API_VERSION (1)
+
+static const struct sol_flow_packet_type _PACKET_TYPE_FLOWER_POWER = {
```
lpereira: Do we really need a new packet type? Adding a new packet type for limited uses such as this is not a good idea IMO, considering that converter nodes will be needed to extract information from them, and that we lose things like console debugging or the inspector.

src/modules/flow/flower-power/flower-power.c
```diff
@@ -0,0 +1,709 @@
...
+
+static void
+http_get_close(struct sol_flow_node *node, void *data)
+{
+    struct http_get_data *mdata = data;
+
+    free(mdata->client_id);
+    /* Zero memory used to store client secret and password */
+    if (mdata->client_secret) {
+        memset(mdata->client_secret, 0, strlen(mdata->client_secret) + 1);
```
lpereira: Not sure if this `memset()` will be called in high optimization levels, given that `free()` is called below. Maybe you need `explicit_bzero()`?

(Regardless, there's no need to zero out the terminating NUL.)

src/modules/flow/flower-power/flower-power.c
```diff
@@ -0,0 +1,709 @@
...
+            size_t auth_len = strlen(AUTH_START);
+            /* value is between double quotes */
+            size_t token_size = value.end - value.start + auth_len - 1;
+
+            free(mdata->token);
+            mdata->token = malloc(token_size);
+            SOL_NULL_CHECK(mdata->token);
+
+            strcpy(mdata->token, AUTH_START);
+            memcpy(mdata->token + auth_len, value.start + 1,
```
lpereira: Is `token_size` enough to hold `AUTH_START` and the value? Maybe using `asprintf()` here would be a better idea

src/modules/flow/flower-power/flower-power.c
lpereira: Ah, nevermind. Saw the filter stuff.

src/modules/flow/flower-power/flower-power.c
```diff
@@ -0,0 +1,709 @@
...
+    }
+
+    sol_json_scanner_init(&scanner, response->content.data,
+        response->content.used);
+    SOL_JSON_SCANNER_OBJECT_LOOP (&scanner, &token, &key, &value, reason) {
+        if (!sol_json_token_str_eq(&key, "access_token",
+                strlen("access_token")))
+            continue;
+        else {
+            size_t auth_len = strlen(AUTH_START);
```
lpereira: Make this `const`, then deindent the whole block. The `token_size` variable won't be needed if `asprintf()` is used to build `mdata->token`.

src/modules/flow/flower-power/flower-power.c
bdilly: no, but anyway it would be part of a flower power packet

src/modules/flow/flower-power/flower-power.c
```diff
@@ -0,0 +1,709 @@
...
+    struct filter_data *mdata = data;
+    struct sol_drange fertilizer, light, temperature, water;
+    char *id, *timestamp;
+    int r;
+
+    r = packet_get_packet_type_flower_power(packet, &id, &timestamp,
+        &fertilizer, &light, &temperature, &water);
+    SOL_INT_CHECK(r, < 0, r);
+
+    if ((!*id) || (!mdata->id)) {
```
lpereira: Are these parenthesis required?

src/modules/flow/flower-power/flower-power.c
bdilly: Yeah, and the problem is that just splitting it on different ports is bad, because these measures (water, temperature, ...) don't mean anything without location identifier and timestamp... so it's really required.

src/modules/flow/flower-power/flower-power.c
lpereira: The get-value node could send a timestamp packet. It's pretty useful to have nodes to compare timestamps, calculate the difference, etc.

src/modules/flow/flower-power/flower-power.c
```diff
@@ -0,0 +1,709 @@
...
+    if (!opts->client_id || !strlen(opts->client_id) ||
+        !opts->client_secret || !strlen(opts->client_secret)) {
+        SOL_ERR("Valid client id and secret are required.");
+        return -EINVAL;
+    }
+
+    mdata->client_id = strdup(opts->client_id);
+    SOL_NULL_CHECK(opts->client_id, -ENOMEM);
+    mdata->client_secret = strdup(opts->client_secret);
+    SOL_NULL_CHECK(opts->client_secret, -ENOMEM);
```
lpereira: `mdata->client_id` leaking

src/modules/flow/flower-power/flower-power.c
bdilly: It is, auth_len is AUTH_START len, and end - start - 2 is for value without double quotes.

src/modules/flow/flower-power/flower-power.c
```diff
@@ -0,0 +1,709 @@
...
+    const struct packet_type_flower_power_packet_data *in = input;
+    struct packet_type_flower_power_packet_data *packet_type_flower_power = mem;
+
+    SOL_NULL_CHECK(in->id, -EINVAL);
+    packet_type_flower_power->id = strdup(in->id);
+    SOL_NULL_CHECK(packet_type_flower_power->id, -ENOMEM);
+
+    SOL_NULL_CHECK(in->timestamp, -EINVAL);
+    packet_type_flower_power->timestamp = strdup(in->timestamp);
+    SOL_NULL_CHECK(packet_type_flower_power->timestamp, -ENOMEM);
```
lpereira: `pkt->id` leaking

src/modules/flow/flower-power/flower-power.c
lpereira: Also, the client secret might still be in the memory after the node type has been initialized (in the options struct that's passed to the open function).

src/modules/flow/flower-power/flower-power.c
bdilly: I don't have a string to fill asprintf, I only have a memory region that goes from value.start + 1 until value.end - 1. I'm confortable using memcpy for that.

I agree I don't need to recalculate auth_len, I'm going to move it out this loop.

src/modules/flow/flower-power/flower-power.c
lpereira: Yes, but you received this from the network and you're blindly accepting that the contents are within the spec.

src/modules/flow/flower-power/flower-power.c
bdilly: I didn't know about explicit_bzero(), it seems the right approach.
Thanks

src/modules/flow/flower-power/flower-power.c
lpereira: Using `asprintf()` will be cleaner (and shorter!).

We might need a function to get strings from JSON tokens as well, the `value.start + 1` and `(value.end - value.start) - 2` are too obscure.

bdilly:
@edersondisouza  there is an extra slash in the end, sorry.

lpereira:
Looks good. Not much of a big fan of custom packet types but in this case it seemed inevitable. Maybe we need a struct packet type so that it's understood by the inspector? It would be useful in other places.

src/modules/flow/flower-power/flower-power.c
bdilly: mdata->client_id is freed at http_get_close.
How are you getting this leak?

Anyway, I saw an issue on this code block, it's checking for opts->client_id not null instead of mdata->client_id.

src/modules/flow/flower-power/flower-power.c
bdilly: pkt->id is freed at packet_type_flower_power_packet_dispose()

How are you getting this leak warnings? They don't seem correct.

But I found out that timestamp is leaking.

src/modules/flow/flower-power/flower-power.c
bdilly: Ah, yeah.
A family of node types working with timestamps could be great indeed.
Let's add them to our node types ideas list =)

src/modules/flow/flower-power/flower-power.c
bdilly: Do you mean SOL_JSON_TOKEN_STR_LITERAL_EQ?

I'm just following the same approach you have done on freegeoip, 
you check stuff using:
        if (sol_json_token_str_eq(&key, "latitude", strlen("latitude")))
       ...
       else if (sol_json_token_str_eq(&key, "longitude", strlen("longitude")))

etc

src/modules/flow/flower-power/flower-power.c
bdilly: client_secret must be in memory while this node is in use, because it's required to create a new token all the time a new username is used (or token expires (not handled)).

I'm just zeroing it after we're done with it.

src/modules/flow/flower-power/flower-power.c
lpereira: I'm not running the code, I'm just reading the patch. If ` strdup(opts->client_secret)` fails, the `strdup(opts->client_id)` is not freed.

src/modules/flow/flower-power/flower-power.c
lpereira: Right, in the node data struct. But what about the struct that's passed down to the node open function? How's the sensitive data is handled there?

(I mean, it's kind of a moot point to protect this, given that the password must be sent in the clear -- although through HTTPS -- and needs to be in the memory anyway, but reducing the amount of copies in memory of sensitive data is always a good thing.)

src/modules/flow/flower-power/flower-power.c
bdilly: This check is wrong, fixed.

bdilly:
Merged with most suggestions applied.
Some stuff was marked as TODO and we'll work on that soon.
Thanks

src/modules/flow/flower-power/flower-power.c
barbieri: +1 on timestamp. it's useful to have it clear in a separated type and packet, not try to stick that information in irange.

src/modules/flow/flower-power/flower-power.c
barbieri: I recall we had that... even handling the escape, if any, no?

src/modules/flow/flower-power/flower-power.c
barbieri: You can use https://github.com/solettaproject/soletta/blob/master/src/lib/datatypes/include/sol-str-slice.h#L71 SOL_STR_SLICE_PRINT(str_slice) in printf()-like functions.

Then a macro that converts the token to slice and use the slice :-)

barbieri:
@lpereira extra packet types are not bad, they are good. The bad part is having to do all the cumbersome stuff... like converters.

My patchset #271 was about adding the introspection you said, but @cmarcelo is always against it saying we're going the wrong route, too much stuff, etc...

My idea was to merge something in those lines and then add support to node-type-gen.py to get packet description from the JSON, generating the packet type and all the C functions and converter nodes for you.

PR #467, opened 2015-08-07
Added UV sensor family si114x

tcanabrava:
Code based on the UPM project.
- fixed incorrect timer return
- fixed documentation
- misc fixes


src/modules/flow/si114x/si114x.c
```diff
@@ -0,0 +1,370 @@
...
+#include "sol-mainloop.h"
+#include "sol-util.h"
+#include "sol-i2c.h"
+#include "sol-types.h"
+#include "sol-flow-internal.h"
+
+/*
+ * SI114X registers
+ */
+enum _SI114X_REG {
```
lpereira: Why uppercase? Why prefixed with `_`?

src/modules/flow/si114x/si114x.c
```diff
@@ -0,0 +1,370 @@
...
+    REG_PARAM_READ    = 0x2e,
+
+    REG_CHIP_STAT     = 0x30,
+
+    REG_ANA_IN_KEY0   = 0x3b,
+    REG_ANA_IN_KEY1   = 0x3c,
+    REG_ANA_IN_KEY2   = 0x3d,
+    REG_ANA_IN_KEY3   = 0x3e
+};
+typedef enum _SI114X_REG SI114X_REG_T;
```
lpereira: Typedefs for enums are not used elsewhere in Soletta.

src/modules/flow/si114x/si114x.c
```diff
@@ -0,0 +1,370 @@
...
+{
+    SOL_NULL_CHECK(cb, -EINVAL);
+
+    if (mdata->timer)
+        sol_timeout_del(mdata->timer);
+
+    mdata->timer = sol_timeout_add(timeout_ms, cb, cb_data);
+    SOL_NULL_CHECK(mdata->timer, -ENOMEM);
+
+    return 0;
```
lpereira: The return value of this function isn't checked where it's called. Should it?

src/modules/flow/si114x/si114x.c
```diff
@@ -0,0 +1,370 @@
...
+
+static int
+si114x_process(struct sol_flow_node *node, void *data, uint16_t port, uint16_t conn_id,
+    const struct sol_flow_packet *packet)
+{
+    struct si114x_data *mdata = data;
+    uint16_t word;
+    double value;
+
+    if (sol_i2c_read_register(mdata->context, REG_AUX_UVINDEX0, (uint8_t *)&word, 2) < 0) {
```
lpereira: Maybe store the return value and use %s and sol_util_strerrora() in the call to send_error_packet()?

src/modules/flow/si114x/si114x.c
```diff
@@ -0,0 +1,370 @@
...
+        return false;
+    }
+    mdata->fully_initialized = true;
+    return false;
+}
+
+static bool
+setup_device_pt2(void *d)
+{
+    struct si114x_data *mdata = (struct si114x_data *)d;
```
lpereira: Unneeded cast.

src/modules/flow/si114x/si114x.c
```diff
@@ -0,0 +1,370 @@
...
+    mdata->timer = sol_timeout_add(timeout_ms, cb, cb_data);
+    SOL_NULL_CHECK(mdata->timer, -ENOMEM);
+
+    return 0;
+}
+
+static bool
+setup_device_pt3(void *d)
+{
+    struct si114x_data *mdata = (struct si114x_data *)d;
```
lpereira: Unneeded cast

src/modules/flow/si114x/si114x.c
```diff
@@ -0,0 +1,370 @@
...
+si114x_open(struct sol_flow_node *node, void *data, const struct sol_flow_node_options *options)
+{
+    struct si114x_data *mdata = data;
+
+    SOL_FLOW_NODE_OPTIONS_SUB_API_CHECK(options, SOL_FLOW_NODE_TYPE_LIGHT_SENSOR_SI114X_OPTIONS_API_VERSION, -EINVAL);
+
+    mdata->fully_initialized = false;
+    mdata->opts = (struct sol_flow_node_type_light_sensor_si114x_options *)options;
+    mdata->uv_calibration.data_raw = mdata->opts->calibration.val;
+    mdata->context = sol_i2c_open(mdata->opts->bus.val, mdata->opts->speed.val);
```
lpereira: What happens if sol_i2c_open() fails here?

src/modules/flow/si114x/si114x.c
```diff
@@ -0,0 +1,370 @@
...
+ * SI114X registers
+ */
+enum _SI114X_REG {
+    REG_PART_ID       = 0x00,
+    REG_REV_ID        = 0x01,
+    REG_SEQ_ID        = 0x02,
+    REG_INT_CFG       = 0x03,
+    REG_IRQ_ENABLE    = 0x04,
+
+    // these two are not documented in the datasheet,
```
lpereira: Please use /* comments */ instead of // comments

src/modules/flow/si114x/si114x.c
```diff
@@ -0,0 +1,370 @@
...
+static int
+si114x_open(struct sol_flow_node *node, void *data, const struct sol_flow_node_options *options)
+{
+    struct si114x_data *mdata = data;
+
+    SOL_FLOW_NODE_OPTIONS_SUB_API_CHECK(options, SOL_FLOW_NODE_TYPE_LIGHT_SENSOR_SI114X_OPTIONS_API_VERSION, -EINVAL);
+
+    mdata->fully_initialized = false;
+    mdata->opts = (struct sol_flow_node_type_light_sensor_si114x_options *)options;
+    mdata->uv_calibration.data_raw = mdata->opts->calibration.val;
```
edersondisouza: Are you sure that, for the user, grouping the calibration data on an integer is the way to go? Datasheet appears to inform it separately.

src/modules/flow/si114x/si114x.json
```diff
@@ -0,0 +1,65 @@
...
+            "name": "address"
+          },
+          {
+            "data_type": "int",
+            "description": "Sensor reading speed",
+            "name": "speed"
+          },
+          {
+            "data_type": "int",
+            "description": "uv-calibration",
```
edersondisouza: You could set the default one here, don't you think? But please, note my first comment about this.

src/modules/flow/si114x/si114x.json
```diff
@@ -0,0 +1,65 @@
...
+            "name": "calibration"
+          }
+        ],
+        "version": 1
+      },
+      "in_ports": [
+        {
+          "data_type": "any",
+          "description": "Signal the s2p chip that a read must be performed",
```
edersondisouza: s2p?

src/modules/flow/si114x/si114x.json
```diff
@@ -0,0 +1,65 @@
...
+          "methods": {
+            "process": "si114x_process"
+          },
+          "name": "TICK"
+        }
+      ],
+      "out_ports": [
+        {
+          "data_type": "float",
+          "description": "The current uv reading [1 - 11] in the uv index scale.",
```
edersondisouza: Is there a link to point the user to, about this uv index?

src/modules/flow/si114x/si114x.c
edersondisouza: I think you should check `fully_initialized` variable here, before issuing a read, don't you think? Maybe implement a pending counter to dispatch it?

src/modules/flow/si114x/si114x.c
```diff
@@ -0,0 +1,370 @@
...
+
+    data = 0xff;
+    if (!sol_i2c_write_register(mdata->context, REG_IRQ_STATUS, &data, 1)) {
+        SOL_WRN("Couldn't reset the REG_IRQ_STATUS register");
+        return false;
+    }
+
+    // send a reset
+    data = CMD_RESET;
+    if (!sol_i2c_write_register(mdata->context, REG_COMMAND, &data, 1)) {
```
lpereira: If lots of I2C devices end up having initialization steps like this, it might be a good idea to devise a way of writing to multiple registers, akin to what `writev()` does:

     struct sol_i2c_register_vec register_init_vec[] = {
         { .register = REG_COMMAND, ... },
         { .register = REG_IRQ_STATUS, ... },
         { .register = REG_INT_CFG, ... },
     };
     sol_i2c_write_registerv(mdata->context, register_init_vec, ARRAY_SIZE(register_init_vec));

lpereira:
Please don't refer to the past when writing commit messages. Prefer to use something along the lines of "Add UV sensor family SI114X".

src/modules/flow/si114x/si114x.c
```diff
@@ -0,0 +1,370 @@
...
+    setup_device(mdata);
+    return 0;
+}
+
+static void
+si114x_close(struct sol_flow_node *node, void *data)
+{
+    struct si114x_data *mdata = data;
+
+    sol_timeout_del(mdata->timer);
```
edersondisouza: You must check if there's a timeout active before deleting it. And you must also nullify this reference after returning false from timeouts callbacks, to avoid double deleting them.

src/modules/flow/si114x/si114x.c
glima: Also we're not spacing enum value entries like that (we don't enforce the other way yet, but...)

src/modules/flow/si114x/si114x.c
tcanabrava: because I didn't found a rule to use typedef so used what I know it was right - but since your other review is to not use typedefs like this I'll remove those.

src/modules/flow/si114x/si114x.c
tcanabrava: it's easyer to read, but if you guys want I can make the enums to have only one space.

src/modules/flow/si114x/si114x.c
tcanabrava: datasheet uses uchar values, we use int32 - that's the only reason I used it like this.
I wanted to be able to do something like "options = INIT_UV | INIT_IR" on the config one day.
if I add all 4 necessary parameters on the config this will get too big:
_(si114x:first_register=0x1,second_register=0x4,third=register=0x5,fourth_register=0x10) OUT

Too, too big.

src/modules/flow/si114x/si114x.json
tcanabrava: it's in the datasheet.

tcanabrava:
closed because of v3
https://github.com/solettaproject/soletta/pull/485

PR #469, opened 2015-08-07
common: Rename 'platform name' to 'board name'

mbelluzzo:
Since is the board name that is expected and not the platform
name. This should avoid misinterpretations that were already happening.

The detection system was automatically setting the board name to
'riotos' when RIOT was found but it should be the name of where RIOT
is supposed to run (or empty if not known) that should be used e.g.
'intel-galileo-rev-d'.

Since there is no demand for a set of board related API's yet, instead
of change 'sol-platform-name-get' to 'sol-board-name-get' and have it
moved to it's on file and header file that would need to be installed, I
decided to name the function 'sol-platform-get-board-name' and keep it
on 'sol-platform' files.

============================ Pull Req Note ===============
This was suggested by k-s sometime as a low prio, but since I realized that some misusage was already happening, priority got higher.

### DIFF
--- a/data/jsons/board_detect.json
+++ b/data/jsons/board_detect.json
@@ -1,6 +1,6 @@
 {
-  "$schema": "http://solettaproject.github.io/soletta/schemas/platform-detect.schema",
-  "platforms": [
+  "$schema": "http://solettaproject.github.io/soletta/schemas/board-detect.schema",
+  "boards": [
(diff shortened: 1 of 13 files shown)


lpereira:
LGTM

mbelluzzo:
Land Patch

Go Vacation

PR #470, opened 2015-08-10
build system updates

dorileo:
## Rationale
This pull request brings 2 patches, 1 to generate make dependency chains with gcc -MM and friends, and one another to make sure to remove python generated code with dependency-resolver tests.


dorileo:
v2 coming....

PR #471, opened 2015-08-10
build system updates - V2

dorileo:
## Changes
v2:
  + added new patch: "build: add soletta's specific help listing"

## Rationale
This pull request brings 2 patches, 1 to generate make dependency chains with gcc -MM and friends, and one another to make sure to remove python generated code with dependency-resolver tests.


dorileo:
@barbieri it seems not to be working, would involve some time to update it or see what's missing, I can drop it until we have it working - if we want it to.

barbieri:
just drop it

dorileo:
ok, applied with the requested changes...

PR #474, opened 2015-08-10
Make I2C asynchronous V2

zehortigoza:
Changes from V1:
- All transactions now return a sol_i2c_pending instead of a void *
- Some minor fixes on Grove LCD

All the ports patches was squashed to ' I2C: Make API asynchronous'


zehortigoza:
Merged

PR #477, opened 2015-08-10
Reduce glib v2

barbieri:

Changes since v1 (#462):
 * rebased to master, changes mostly due @ibriano's Kconfig/Makefile cleanups.
 * release the lock before calling users's `get_next_timeout()`, done in the same fashion as other loops, we steal the vector and replaces it with a temporary vector with the lock held, then loop in the stolen vector without locks, then take the lock and update the vector. The issue was pointed out by @ulissesf 
 * remove sol-bus dependency on Glib (Makefile)
 * remove sol-bus.c comment mentioning glib integration.

### DIFF
--- a/src/lib/common/Makefile
+++ b/src/lib/common/Makefile
@@ -20,8 +20,8 @@ obj-log-$(PLATFORM_CONTIKI) += \
 
 obj-core-$(SOL_BUS) += \
     sol-bus.o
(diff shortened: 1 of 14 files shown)


ulissesf:
Apart from minor doc strings in one patch it LGTM.

barbieri:
Replaced with #497

PR #479, opened 2015-08-10
Valgrind

tcanabrava:
Fix valgrind issues related to supression of the curl issues, and a big timeout for tests that where failing because valgrind makes execution time 40x slower.


tcanabrava:
closing because of https://github.com/solettaproject/soletta/pull/499

PR #480, opened 2015-08-10
Fix conf file options requirement.

cabelitos:
Not all nodes requires options. If one creates a conf file with the following entry:

/* Section of a conf file */
{
 "name": "MyBtn"
 "type": "gtk/pushbutton"
}

The sol-fbp-runner would exit with an error code, because it could not
find the options for MyBtn.

@cmarcelo Could you please take a look?


cabelitos:
Closing. V2 available.

PR #482, opened 2015-08-10
build: bring back flto flags

dorileo:
## Changes
v2:
handle only -O[s12345\ ]

## Rationale
This patch re enables the -flto flags based on optimization level
provided flags.


barbieri:
+1

dorileo:
self -1
we're lacking -lubsan and with that lto-wrapper will fail... I'm sending a v3 including a patch to fix the issue.

PR #483, opened 2015-08-10
build: bring back -fsanitize=address|undefined flags

dorileo:
This patch brings back the -fsanitize flags. It adds the CC_SANITIZE
item to Kconfig and the sanitize type dependent to it. The default (like
previously with autotools) is -fsanitize=undefined, unless disabled or
non supported.


ibriano:
LGTM

dorileo:
ok, applied.

PR #484, opened 2015-08-10
Fix conf file options requirement.

cabelitos:
Not all nodes requires options. If one creates a conf file with the following entry:

/* Section of a conf file */
{
 "name": "MyBtn"
 "type": "gtk/pushbutton"
}

The sol-fbp-runner would exit with an error code, because it could not
find the options for MyBtn.


cmarcelo:
+2

Tip: when adding new versions of PRs, it's the usual practice to make a reference to the previous PR (so previous discussions can be easily read) and also add some kind of version in the name of the PR.

cmarcelo:
This PR replaces #480.

cabelitos:
Sorry about that, next time I will reference the older PR.

cabelitos:
Merged.

PR #485, opened 2015-08-10
Add UV sensor family si114x

tcanabrava:
Code based on the UPM project, following the datasheet
specification for the si1145/6/7 products, it does NOT
the datasheet for the si1141/2/3, that are completely
different products with the same family name.

Differences from v2:
- correct the enums spacing
- remove option for uv calibration because there's no explanation on the datasheet how it works
- fixed all issues pointed by the other developers.


tcanabrava:
old pr, new one opened.

PR #486, opened 2015-08-10
modules/flower-power: expose custom packet

bdilly:
It'll be required from other modules.

@cabelitos check if it's good for your use case, please


cabelitos:
I think you forgot to expose the PACKET_TYPE_FLOWER_POWER

bdilly:
new version pr 488

PR #488, opened 2015-08-11
modules/flower-power: expose custom packet

bdilly:
It'll be required from other modules.

v2 changelog:
 * expose PACKET_TYPE_FLOWER_POWER


bdilly:
coding style fixed

@cabelitos has tested it and now it's good

merged

PR #489, opened 2015-08-11
Check if glibc have newer locale functions

zehortigoza:
On 'json: add number parsing.'[b9dc3ad4] we started to use freelocale()
and newlocale() that are functions only available on glibc 2.10 or newer
but in our Galileo SDK the available version is 2.10.


barbieri:
+1, but I'd test for newlocale... just to be consistent with the name ;-)

zehortigoza:
Change to newlocale and merged.

PR #490, opened 2015-08-11
Fill buffer

edersondisouza:
Using `sol_util_fill_buffer()` instead of raw `read()` on many applicable places. I've tried to not just blindly replace, but definitely review must take into account my lack of familiarity with the code (iow, I may have screwed everything up). One special doubt is about 'keyboard' - as `sol_util_fill_buffer()` is linux only, I made it depend on linux - however, I'm not sure about it. If opposed, I see no problem on dropping that patch.
All patches are separated by file touched - but I can squash them.

### DIFF
--- a/src/lib/comms/sol-network-linux.c
+++ b/src/lib/comms/sol-network-linux.c
@@ -52,6 +52,7 @@
 #include "sol-missing.h"
 #include "sol-network.h"
 #include "sol-util.h"
+#include "sol-util-linux.h"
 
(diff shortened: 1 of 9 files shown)


edersondisouza:
Besides @cmarcelo comment, is there any other issue? If not, I may fix it and merge this series tomorrow.

barbieri:
not on my side.

edersondisouza:
Fixed issue and integrated.

PR #491, opened 2015-08-11
Build Contiki without config

zehortigoza:
With this patches we don't need anymore any .config file, we can just call make alldefconfig and build to Contiki over native platform.


zehortigoza:
New version coming.

PR #492, opened 2015-08-11
Fix memory leaks on http-client

vcgomes:
This series also introduces sol_ptr_vector_remove() which removes a pointer from a sol_ptr_vector as this seems to be a common operation.

The strategy for fixing the memory leak is keeping references to the existing connections, and using the procedure recommended in the curl_multi_cleanup(3) man page for cleaning them up:

> Cleans  up  and  removes  a whole multi stack. It does not free or touch any individual
> easy handles in any way - they still need to be closed individually,  using  the  usual
> curl_easy_cleanup(3) way. The order of cleaning up should be:
> 1 - curl_multi_remove_handle(3) before any easy handles are cleaned up
> 2 -  curl_easy_cleanup(3)  can now be called independently since the easy handle is no
>   longer connected to the multi handle
> 3 - curl_multi_cleanup(3) should be called when all easy handles are removed

### DIFF
--- a/src/lib/comms/sol-http-client-impl-curl.c
+++ b/src/lib/comms/sol-http-client-impl-curl.c
@@ -52,12 +52,13 @@ void sol_http_client_shutdown(void);
 static struct {
     CURLM *multi;
     struct sol_timeout *multi_perform_timeout;
+    struct sol_ptr_vector connections;
     long timeout_ms;
-    unsigned int fds;
     int ref;
 } global = {
     .timeout_ms = 100,
-    .ref = 0
+    .ref = 0,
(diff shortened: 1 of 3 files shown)


vcgomes:
I don't have a real test for this. @bdilly could you see if this works for the flower-power HTTP nodes?

vcgomes:
@dorileo, yeah, I really think that it was not needed in the first place. It is looking like legacy from previous rewrites.

bdilly:
Flower power is still working fine.

vcgomes:
Fixed comments, new PR https://github.com/solettaproject/soletta/pull/507

Closing this. Thanks.

PR #494, opened 2015-08-11
sol-util: start using sol-util-$(PLATFORM) v4

ricardotk:
Differences from v3:
- do not use impl header and impl functions
- remove sol-util-linux header and put the functions on sol-util header to be implemented on whichever platform

Differences from v2:
- maintain util on libshared.mod instead of creating util.mod

Differences from v1:
- fixed kbuild makefile

Since we were using a lot of platform #ifdefs in sol-util.c


src/shared/sol-util.h
```diff
@@ -116,6 +115,15 @@ sol_util_int_compare(const int a, const int b)
...
 int sol_util_timespec_get_realtime(struct timespec *t);
+int sol_util_write_file(const char *path, const char *fmt, ...) SOL_ATTR_PRINTF(2, 3);
+int sol_util_vwrite_file(const char *path, const char *fmt, va_list args) SOL_ATTR_PRINTF(2, 0);
+int sol_util_read_file(const char *path, const char *fmt, ...) SOL_ATTR_SCANF(2, 3);
+int sol_util_vread_file(const char *path, const char *fmt, va_list args) SOL_ATTR_SCANF(2, 0);
+void *sol_util_load_file_raw(const int fd, size_t *size) SOL_ATTR_WARN_UNUSED_RESULT;
+char *sol_util_load_file_string(const char *filename, size_t *size) SOL_ATTR_WARN_UNUSED_RESULT;
+int sol_util_get_rootdir(char *out, size_t size) SOL_ATTR_WARN_UNUSED_RESULT;
+int sol_util_fd_set_flag(int fd, int flag) SOL_ATTR_WARN_UNUSED_RESULT;
+ssize_t sol_util_fill_buffer(const int fd, char *buffer, const size_t buffer_size, size_t *size_read);
```
cmarcelo: I think these should be in a `sol-util-file.h` or something like. And we don't even use / compile those for the other platforms that won't support it.

The other part of the patch: splitting the implementation of util (time stuff) in specific files is good.

cmarcelo:
+1 but I think need one last tweak.

ricardotk:
#537

PR #495, opened 2015-08-11
build: add headers-m support

dorileo:
It seems we've missed the headers-m identification/parsing, this patch
adds that and fixes issues with flower-power module.


vcgomes:
:+1:

Works fine.

dorileo:
applied, thanks.

PR #497, opened 2015-08-11
Reduce glib v3

barbieri:

Changes since v2 (#477):
 * `sol-mainloop.h` now documents `struct sol_mainloop_source_type` members in separated lines, similar to `sol-worker-thread.h` as requested by @ulissesf 
 * `sol-mainloop-impl-posix.c:fd_process()`:
   * `sources_ready` implies in non-block `ppoll()`, thus set `use_ts=true` with a zero-timeout
   * call `sol_mainloop_common_source_prepare()` before `fd_prepare()`, this allows source's prepare to create `struct sol_fd` to listen to its internal file descriptors before soletta prepares its own `pollfds` array.
   * only call `sol_mainloop_common_source_check()` once at the end, do not interleave it with own callback dispatches;
   * call `sol_mainloop_common_source_dispatch()` if `sol_mainloop_common_source_check() || sources_ready`. 
 * added `sol_mainloop_common_source_shutdown()` and changed the code to call it after threads are finished and before fd/idler/timeouts are destroyed, this allows sources to delete their own `struct sol_fd` at their type's `dispose()`
 * added `sol-glib-integration.h` and `sol_glib_integration()`, with test;
 * gtk now uses `sol_glib_integration()` and can be used with posix mainloop!

Changes since v1 (#462):
 * rebased to master, changes mostly due @ibriano's Kconfig/Makefile cleanups.
 * release the lock before calling users's `get_next_timeout()`, done in the same fashion as other loops, we steal the vector and replaces it with a temporary vector with the lock held, then loop in the stolen vector without locks, then take the lock and update the vector. The issue was pointed out by @ulissesf 
 * remove sol-bus dependency on Glib (Makefile)
 * remove sol-bus.c comment mentioning glib integration.

### DIFF
--- a/README.md
+++ b/README.md
@@ -259,6 +259,11 @@ for in-depth information, the summary of environment variables to use:
 
 	make CFLAGS="-O0 -ggdb3" # disable optimizations
 
+Users of `GMainLoop` (ie: gtk, libsoup) should use the function
+`sol_glib_integration()` from the header file `sol-glib-integration.h`
+to make sure they will work with Soletta, regardless of how libsoletta
+was compiled.
+
 #### POSIX (kconfig: Core library -> Mainloop -> posix)
 
 If GLib is too big then you can use a simpler implementation based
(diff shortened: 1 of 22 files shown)


barbieri:
merged.

PR #498, opened 2015-08-11
modules/flower-power: make packet API more flexible

bdilly:
When creating a new packet passing components, some components
can be passed as NULL and default measures will be used in this case.

Remove redundant checks on function that create a new packet
given a flower power data struct.

@otaviobp : I've talked to ceolin about the possible
issue with packet getter, but it seems to be behaving
just like all the other packet types.

If memory was alloced for strings, they would need to be
freed by getter() users... It wouldn't be nice.

Also, for our cases, these pointers are used in the same
scope where packet is got. If later usage is required,
user is responsible for duplicating it.

Same is done for error packet (it gives a pointer for
an error msg that stop existing when packet dies).


bdilly:
merged, thanks

PR #500, opened 2015-08-11
Build Contiki without config V2

zehortigoza:
Changes from V1:
* Removed 'Disable sanitizer on RIOT and Contiki'  we will solve this in another way
* Added 3 more patches, to enable build on a real device using newlib.
  * Only build flow Javascript support on Linux
  * Converter: Rename MIN and MAX macros
  * Contiki: Fix undeclared SSIZE_MAX


glima:
LGTM on the rest.

zehortigoza:
ping @ibriano

zehortigoza:
Merged

PR #504, opened 2015-08-12
Fix issues in freeing memory after calling sol_flow_send_packet

otaviobp:
Documentation states that sol_flow_send_packet is responsable in releasing
packet memory. But in a few places method sol_flow_packet_del was called
if sol_flow_send_packet returns a non zero value.

And, as far as I could get, sol_flow_send_packet was not releasing memory if
parent->send method was failing.

I am not familiar with sol-flow code, but as far as I could trace, the only
implementation of parent->send method was located in sol-flow-static, and when
send was returning a non zero value, the packet memory was not being released.


barbieri:
+1

ceolin:
both patches are pretty good !

ceolin:
Please just make a double check if there is any other place checking if the send failed to release the memory.

PR #506, opened 2015-08-12
Enable flow on RIOT by default

ibriano:
Enable flow, and the necessary fixes to make that happen with alldefconfig.


ibriano:
V2 in #519

PR #509, opened 2015-08-12
String replace

glima:
This adds a new string/replace node, which acts akin to python's string.replace() method.


glima:
Now fixing the error on semaphoreci...

glima:
My !HAVE_BUILTIN_MUL_OVERFLOW version of the new function is wrong, looking at that.

glima:
All fixed, now issuing a v2

PR #510, opened 2015-08-12
Fix issues in freeing memory after calling sol_flow_send_packet - v2

otaviobp:
Add commit to free string in correct position in sol_flow_send_error_packet
Added to this pull req because it is related to the fixes proposed.


otaviobp:
This pr replaces #504

ceolin:
The patches are pretty good, go ahead !

PR #511, opened 2015-08-12
Build bring back flto flags v4

dorileo:
## Changes
- v2:
   + handle only -O[s12345\ ]
- v3:
   + added new patch: "build: [sanitizer] add -lubsan"
- v4:
   + added new patch: "build: compiler options"
   + changed the libubsan linking to use the -fsanitize option so the compiler knows exactly what to do - also it keeps consistent with what we used to have in the autotools era

## Rationale
This patch re enables the -flto flags based on optimization level provided flags.

Signed-off-by: Leandro Dorileo leandro.maciel.dorileo@intel.com


cmarcelo:
+1 on the two first patches.

The one with config I added a comment, I rather not see config for each optimization level and so on. I saw we give the common cases (debug or release build) and the config_cflags/ldflags.

dorileo:
2 first patches were applied, the 3rd was sent to my postponed-patches.

PR #516, opened 2015-08-12
mainloop: allow flags to be set on fd watch

rchiossi:
Add a new API to allow flags to be set on a fd watch. This API reduces
the amount of allocations needed since even when a new watch needs to
be created, the associated data can be reused.

### DIFF
--- a/src/lib/common/include/sol-mainloop.h
+++ b/src/lib/common/include/sol-mainloop.h
@@ -142,6 +142,8 @@ enum sol_fd_flags {
 struct sol_fd;
 struct sol_fd *sol_fd_add(int fd, unsigned int flags, bool (*cb)(void *data, int fd, unsigned int active_flags), const void *data);
 bool sol_fd_del(struct sol_fd *handle);
+bool sol_fd_set_flags(struct sol_fd *handle, unsigned int flags);
+unsigned int sol_fd_get_flags(const struct sol_fd *handle);
 #endif
 
 #ifdef SOL_MAINLOOP_FORK_WATCH_ENABLED
--- a/src/lib/common/sol-mainloop-impl-glib.c
+++ b/src/lib/common/sol-mainloop-impl-glib.c
@@ -130,15 +130,20 @@ sol_mainloop_impl_idle_del(void *handle)
     return g_source_remove((uintptr_t)handle);
 }
 
-struct fd_wrap_data {
+struct sol_fd_glib {
     bool (*cb)(void *data, int fd, unsigned int active_flags);
     const void *data;
+    int fd;
+    unsigned int flags;
+    gint id;
+
+    int refcnt;
 };
 
 static gboolean
 on_fd(gint fd, GIOCondition cond, gpointer data)
 {
-    struct fd_wrap_data *wrap_data = data;
+    struct sol_fd_glib *fd_handle = data;
     unsigned int sol_flags = 0;
 
     if (cond & G_IO_IN) sol_flags |= SOL_FD_FLAGS_IN;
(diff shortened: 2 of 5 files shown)


src/lib/common/include/sol-mainloop.h
```diff
@@ -142,6 +142,8 @@ enum sol_fd_flags {
 struct sol_fd;
 struct sol_fd *sol_fd_add(int fd, unsigned int flags, bool (*cb)(void *data, int fd, unsigned int active_flags), const void *data);
 bool sol_fd_del(struct sol_fd *handle);
+bool sol_fd_set_flags(struct sol_fd *handle, unsigned int flags);
```
barbieri: I'd add a helper (static inline) to unset flags, it would do the `set_flags(get_flags() & ~unset_flags`

src/lib/common/sol-mainloop-impl-glib.c
```diff
@@ -130,15 +130,20 @@ sol_mainloop_impl_idle_del(void *handle)
     return g_source_remove((uintptr_t)handle);
 }
 
-struct fd_wrap_data {
+struct sol_fd_glib {
     bool (*cb)(void *data, int fd, unsigned int active_flags);
     const void *data;
+    int fd;
```
barbieri: starting with glib 2.36 we have, maybe move to use it and rely on newer glib? Or do it in an ifdef?
```c
GLIB_AVAILABLE_IN_2_36
gpointer             g_source_add_unix_fd    (GSource        *source,
                                              gint            fd,
                                              GIOCondition    events);
GLIB_AVAILABLE_IN_2_36
void                 g_source_modify_unix_fd (GSource        *source,
                                              gpointer        tag,
                                              GIOCondition    new_events);
GLIB_AVAILABLE_IN_2_36
void                 g_source_remove_unix_fd (GSource        *source,
                                              gpointer        tag);
GLIB_AVAILABLE_IN_2_36
GIOCondition         g_source_query_unix_fd  (GSource        *source,
                                              gpointer        tag);
```

src/lib/common/sol-mainloop-impl-glib.c
```diff
@@ -130,15 +130,20 @@ sol_mainloop_impl_idle_del(void *handle)
     return g_source_remove((uintptr_t)handle);
 }
 
-struct fd_wrap_data {
+struct sol_fd_glib {
```
barbieri: do not rename it if it's not needed.

src/lib/common/sol-mainloop-impl-glib.c
```diff
@@ -169,14 +191,63 @@ sol_mainloop_impl_fd_add(int fd, unsigned int flags, bool (*cb)(void *data, int
...
+    fd_handle->refcnt++;
+    if (!g_source_remove(fd_handle->id))
+        return false;
+
+    if (flags & SOL_FD_FLAGS_IN) glib_flags |= G_IO_IN;
+    if (flags & SOL_FD_FLAGS_OUT) glib_flags |= G_IO_OUT;
+    if (flags & SOL_FD_FLAGS_PRI) glib_flags |= G_IO_PRI;
+    if (flags & SOL_FD_FLAGS_ERR) glib_flags |= G_IO_ERR;
+    if (flags & SOL_FD_FLAGS_HUP) glib_flags |= G_IO_HUP;
+    if (flags & SOL_FD_FLAGS_NVAL) glib_flags |= G_IO_NVAL;
```
barbieri: create a function to convert to and from glib flags.

src/lib/common/sol-mainloop-impl-posix.c
```diff
@@ -786,6 +786,28 @@ sol_mainloop_impl_fd_del(void *handle)
...
 
+bool
+sol_mainloop_impl_fd_set_flags(void *handle, unsigned int flags)
+{
+    struct sol_fd_posix *fd_handle = handle;
+
+    SOL_NULL_CHECK(handle, false);
+
+    fd_handle->flags = flags;
+
```
barbieri: @ulissesf remembered we may need to lock the main loop and notify the main thread so `poll()` will run again if modified the fd from a different thread.

src/lib/common/sol-mainloop-impl-glib.c
rchiossi: This matches the name used by the posix implementation. Also, this has been re purposed - it has the same initial fields, but it is not used in the same way. It used to be a wrapper, now it's the object returned to the user.

src/lib/common/sol-mainloop-impl-glib.c
rchiossi: Support both? Just the new one? your call..

dorileo: yep, please if changing this ifdef with ```GLIB_CHECK_VERSION()``` yocto has packaged 2.21.

barbieri: argh... then just keep your current version :-/

dorileo: nah! mentioned wrong version, but I see it as reasonable the ifdef.

PR #517, opened 2015-08-12
Buffer review

barbieri:
This series reviews `sol-buffer.h` so it's better and easier to use.

I plan to wait #490 to be merged and change that and more APIs to use `struct sol_buffer` instead of manually manipulating buffers, that is error prone.


barbieri:
Replaced with #522

PR #518, opened 2015-08-12
Fix make check fbp issues v2

tcanabrava:
Differences from v1:
- Added a test for a wrong shift operation
- Removed curl from valgrind ignore file
- Fixed a bunch of thigns found by address sanitizer
- don't set -fno-sanitizer if compiler doesn't support it
- rework "TEST-SKIP-VALGRIND"


tcanabrava:
merged.

PR #519, opened 2015-08-12
Enable flow on RIOT by default

ibriano:
Enable flow, and the necessary fixes to make that happen with alldefconfig.

Changes since v1:
 - Rebased, some changes gone since Ze's patch-set included them already.


glima:
All LGTM.

ibriano:
Merged

PR #520, opened 2015-08-12
test: improve test-mainloop-linux

cmarcelo:
The test-mainloop-linux does various checks for timeout and idle
executions, and expects to later receive a SIGTERM from its child, which
is handled and causes the mainloop to be interrupted.

Instead of using a usleep() to decide when to send SIGTERM, the child
will wait the parent to ask for it. Parent does it after all the other
checks are done. Child waits blocked on read() and when it's time,
parent closes the pipe.

Fixes one of the bots failing sometimes when running this test.

### DIFF
--- a/src/test/test-mainloop-linux.c
+++ b/src/test/test-mainloop-linux.c
@@ -48,12 +48,24 @@ static int read_magic[2];
 static int timeout_count;
 static int idler_count1, idler_count2;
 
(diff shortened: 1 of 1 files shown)


cmarcelo:
Applied in d0f2415ffe41c0b972d671f32fd0efb7b0c27dbd.

PR #524, opened 2015-08-13
string-icu: Memory issues

edersondisouza:
Possible double free on string_change_case();
Memory leak on string_concat() and string_change_case().


edersondisouza:
Integrated

PR #526, opened 2015-08-13
Exercise some utf8 chars on the string converter nodes.

glima:


glima:
Since this is uber-simple, it got in with another PR which was acked.

PR #527, opened 2015-08-13
Make sanitize depend on Linux

zehortigoza:
To this feature works on RIOT and Contiki we should add the sanitize
compiler flags when the OS build system is generating the final binary
contaning the OS + Soletta + application.
Until we don't have some infrascructure to get the compiler flags that
Solleta need we will keep it disabled.


cmarcelo:
+1

@dorileo @ibriano could either of you take a look?

ibriano:
Does the job

cmarcelo:
Minor: last sentence of the commit message is a bit confusing. It's not even clear to me we want those sanitizers in RIOT / Contiki.

cmarcelo:
They add runtime checks. We probably want to disable them in the small targets.

dorileo:
LGTM.

dorileo:
@cmarcelo it doesn't make sense to disable dep-resolver tests on a given condition (be it a platform, os or whatever) the dep-resolver runs before the configuration selection, i.e when you run make menuconfig in a fresh cloned tree the dep-resolver will be run before launching the menuconfig to generate the dep-resolver's cache. I don't think we're saving too much by avoiding some tests.

cmarcelo:
@dorileo I meant, if I get this right, we'll never enable FEATURE_CC_SANITIZE, even if we manage to separate the build flags on small os side. Just the commit message seems confusing. "Runtime checks" was refering to what the sanitizer does.

Suggestion:

> Contiki and RIOT builds don't fully support this and we probably won't need these extra runtime checks that sanitizer adds to the code for those targets.

zehortigoza:
Merged with cmarcelo message commit.

PR #530, opened 2015-08-13
Move ICU helper macros into functions.

glima:
And fix coverity issues on this module once and for all (at least
against a local coverity run).

"goto considered harmful"


glima:
Cool, thanks. Merging.

PR #522, opened 2015-08-12
Buffer review v2

barbieri:
Changes since v1 (#517):
 * changed from `bool fized_size` to `enum sol_buffer_flags`;
 * renamed `reserved` to `capacity` member to allow a nice flag name;
 * removed the `sol_buffer_fini_nofree()`, it's handled by a flag now;
 * added `sol_buffer_at()` and `sol_buffer_at_end()`;
 * added `sol_buffer_insert_slice()` to insert a slice at a given position;
 * added `sol_buffer_insert_printf()` to insert a formatted string at a given position;
 * added `sol_buffer_get_slice_at()`

PR #532, opened 2015-08-13
Buffer review v3

barbieri:
Changes since v2 (#522):
 * introduced a `va_copy()` in `sol_buffer_append_printf()` so we can call `vsnprintf()` twice. Also replace the loop with two explicit calls to make sure tools understand we only call these twice, and do it with the original `args` and the other with `args_copy`. Should fix the build bot failure.

Changes since v1 (#517):
 * changed from `bool fized_size` to `enum sol_buffer_flags`;
 * renamed `reserved` to `capacity` member to allow a nice flag name;
 * removed the `sol_buffer_fini_nofree()`, it's handled by a flag now;
 * added `sol_buffer_at()` and `sol_buffer_at_end()`;
 * added `sol_buffer_insert_slice()` to insert a slice at a given position;
 * added `sol_buffer_insert_printf()` to insert a formatted string at a given position;
 * added `sol_buffer_get_slice_at()`

### DIFF
--- a/src/lib/datatypes/include/sol-buffer.h
+++ b/src/lib/datatypes/include/sol-buffer.h
@@ -160,6 +160,14 @@ sol_buffer_get_slice(const struct sol_buffer *buf)
     return SOL_STR_SLICE_STR(buf->data, buf->used);
 }
 
+static inline struct sol_str_slice
+sol_buffer_get_slice_at(const struct sol_buffer *buf, size_t pos)
+{
+    if (!buf || buf->used < pos)
+        return SOL_STR_SLICE_STR(NULL, 0);
+    return SOL_STR_SLICE_STR((char *)buf->data + pos, buf->used - pos);
+}
+
 /* Appends the 'slice' into 'buf', reallocating if necessary. */
 int sol_buffer_append_slice(struct sol_buffer *buf, const struct sol_str_slice slice);
 
--- a/src/lib/datatypes/include/sol-buffer.h
+++ b/src/lib/datatypes/include/sol-buffer.h
@@ -76,20 +76,57 @@ extern "C" {
  * that will be de-allocated twice.
  */
 
-struct sol_buffer {
-    void *data;
-    size_t reserved, used;
+enum sol_buffer_flags {
+    /**
+     * default flags: buffer may be resized and memory will be free'd
+     * at the end.
+     */
+    SOL_BUFFER_FLAGS_DEFAULT = 0,
+    /**
+     * fixed capacity buffers won't be resized, sol_buffer_resize()
+     * will fail with -EPERM.
+     */
+    SOL_BUFFER_FLAGS_FIXED_CAPACITY = (1 << 0),
+    /**
+     * no free buffers won't call @c free(buf->data) at
+     * sol_buffer_fini().
+     */
+    SOL_BUFFER_FLAGS_NO_FREE = (1 << 1),
+    /**
+     * buffers where the @c buf->data is not owned by sol_buffer, that
+     * is, it can't be resized and free() should not be called at it
+     * at sol_buffer_fini().
+     */
+    SOL_BUFFER_FLAGS_MEMORY_NOT_OWNED = (SOL_BUFFER_FLAGS_FIXED_CAPACITY | SOL_BUFFER_FLAGS_NO_FREE)
 };
 
-#define SOL_BUFFER_EMPTY (struct sol_buffer){.data = NULL, .reserved = 0, .used = 0 }
+struct sol_buffer {
+    void *data;
+    size_t capacity, used;
+    enum sol_buffer_flags flags;
+};
+
+#define SOL_BUFFER_INIT_EMPTY (struct sol_buffer){.data = NULL, .capacity = 0, .used = 0, .flags = SOL_BUFFER_FLAGS_DEFAULT }
+#define SOL_BUFFER_INIT_FLAGS(data_, size_, flags_) (struct sol_buffer){.data = data_, .capacity = size_, .used = 0, .flags = flags_ }
 
 static inline void
 sol_buffer_init(struct sol_buffer *buf)
 {
     assert(buf);
     buf->data = NULL;
-    buf->reserved = 0;
+    buf->capacity = 0;
     buf->used = 0;
+    buf->flags = SOL_BUFFER_FLAGS_DEFAULT;
+}
+
+static inline void
+sol_buffer_init_flags(struct sol_buffer *buf, void *data, size_t data_size, enum sol_buffer_flags flags)
+{
+    assert(buf);
+    buf->data = data;
+    buf->capacity = data_size;
+    buf->used = 0;
+    buf->flags = flags;
 }
 
 static inline void
@@ -97,10 +134,11 @@ sol_buffer_fini(struct sol_buffer *buf)
 {
     if (!buf)
         return;
-    free(buf->data);
+    if (!(buf->flags & SOL_BUFFER_FLAGS_NO_FREE))
+        free(buf->data);
     buf->data = NULL;
     buf->used = 0;
-    buf->reserved = 0;
+    buf->capacity = 0;
 }
 
 int sol_buffer_resize(struct sol_buffer *buf, size_t new_size);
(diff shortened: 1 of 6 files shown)


barbieri:
since semaphore test passed I want to merge this now, okay @cmarcelo ?

my idea is to add couple of more patches on top, like documentation and some helpers to avoid manual `memcpy()` and memmove()` manipulations. Until then this patchset enables people to move file reader and string nodes (formatting) to use `sol_buffer`. This means @glima and @edersondisouza should look at it and adapt their code trying to use it and feedback on improvements that would help their side.

The APIs I plan to add later are:
```c
size_t
sol_buffer_get_slice_position(const struct sol_buffer *buf, const struct sol_str_slice slice)
{
   CHECK_SLICE(buf, &slice, SIZE_MAX);
   return (const char *)slice.data - (const char *)buf->data;
}

void *
sol_buffer_at_slice(const struct sol_buffer *buf, const struct sol_str_slice slice)
{
   CHECK_SLICE(buf, &slice, NULL);
   return (void *)slice.data;
}

/* memset() to 0 the slice within buf. slice.data and slice.data+slice.len must be within buf->data and buf->data+buf->used */
int sol_buffer_clear_slice(struct sol_buffer *buf, const struct sol_str_slice slice)
{
   CHECK_SLICE(buf, &slice, -EINVAL);
   memset(sol_buffer_at_slice(buf, slice), 0, slice.len);
   return 0;
}

/* delete (remove) the slice within buf. This means memmove() on whatever is after the slice to its start. slice.data and slice.data+slice.len must be within buf->data and buf->data+buf->used */
int sol_buffer_del_slice(struct sol_buffer *buf, const struct sol_str_slice slice)
{
   char *ptr;
   size_t pos;
   CHECK_SLICE(buf, &slice, -EINVAL);
   ptr = sol_buffer_at_slice(buf, slice);
   pos = sol_buffer_get_slice_position(buf, slice);
   memmove(ptr, ptr + slice.len, buf->used - pos - slice.len);
   /* TODO: ensure trailing \0? */
   /* TODO: shrink buffer? */
   return 0;
}

/* replace one slice with another. The orig slice must be within buf, while the new can be external. orig.data and orig.data+slice.len must be within buf->data and buf->data+buf->used */
int sol_buffer_replace_slice(struct sol_buffer *buf, const struct sol_str_slice orig, const struct sol_str_slice replacement)
{
  /* TODO: handle more efficiently than del + insert, save orig and replacement offset as their data members will change during realloc(). Error if orig and replacement overlap, or do a temporary copy. */
}
```
These makes me think if all the slices should really enforce trailing '\0', I don't like it for binary buffers... but replicating the APIs to have one version with and another without sucks. Then maybe add a flag `SOL_BUFFER_FLAGS_NULL_TERMINATED`? If so we could move the "+ 1" logic inside `sol_buffer_ensure()` and have a internal function to do the null-termination.

barbieri:
ah, another function would be:

```c
int sol_buffer_trim(struct sol_buffer *buf) {
    return sol_buffer_resize(buf, buf->used);
}
```

This would be used to finalize the buffer, making it the exact size it should be. Usually to be called when no further resizes to be done.

glima:
* Gustavo Sverzut Barbieri <notifications@github.com> [2015-08-14 06:13:46 -0700]:

> since semaphore test passed I want to merge this now, okay @cmarcelo ?
> 
> my idea is to add couple of more patches on top, like documentation and some helpers to avoid manual `memcpy()` and memmove()` manipulations. Until then this patchset enables people to move file reader and string nodes (formatting) to use `sol_buffer`. This means @glima and @edersondisouza should look at it and adapt their code trying to use it and feedback on improvements that would help their side.

Ack. I'll jump in this adaptation soon.

> 
> The APIs I plan to add later are:
> ```c
> size_t
> sol_buffer_get_slice_position(const struct sol_buffer *buf, const struct sol_str_slice slice)
> {
>    CHECK_SLICE(buf, &slice, SIZE_MAX);
>    return (const char *)slice.data - (const char *)buf->data;
> }
> 
> void *
> sol_buffer_at_slice(const struct sol_buffer *buf, const struct sol_str_slice slice)
> {
>    CHECK_SLICE(buf, &slice, NULL);
>    return (void *)slice.data;
> }
> 
> /* memset() to 0 the slice within buf. slice.data and slice.data+slice.len must be within buf->data and buf->data+buf->used */
> int sol_buffer_clear_slice(struct sol_buffer *buf, const struct sol_str_slice slice)
> {
>    CHECK_SLICE(buf, &slice, -EINVAL);
>    memset(sol_buffer_at_slice(buf, slice), 0, slice.len);
>    return 0;
> }
> 
> /* delete (remove) the slice within buf. This means memmove() on whatever is after the slice to its start. slice.data and slice.data+slice.len must be within buf->data and buf->data+buf->used */
> int sol_buffer_del_slice(struct sol_buffer *buf, const struct sol_str_slice slice)
> {
>    char *ptr;
>    size_t pos;
>    CHECK_SLICE(buf, &slice, -EINVAL);
>    ptr = sol_buffer_at_slice(buf, slice);
>    pos = sol_buffer_get_slice_position(buf, slice);
>    memmove(ptr, ptr + slice.len, buf->used - pos - slice.len);
>    /* TODO: ensure trailing \0? */
>    /* TODO: shrink buffer? */
>    return 0;
> }
> 
> /* replace one slice with another. The orig slice must be within buf, while the new can be external. orig.data and orig.data+slice.len must be within buf->data and buf->data+buf->used */
> int sol_buffer_replace_slice(struct sol_buffer *buf, const struct sol_str_slice orig, const struct sol_str_slice replacement)
> {
>   /* TODO: handle more efficiently than del + insert, save orig and replacement offset as their data members will change during realloc(). Error if orig and replacement overlap, or do a temporary copy. */
> }
> ```
> These makes me think if all the slices should really enforce trailing '\0', I don't like it for binary buffers... but replicating the APIs to have one version with and another without sucks. Then maybe add a flag `SOL_BUFFER_FLAGS_NULL_TERMINATED`? If so we could move the "+ 1" logic inside `sol_buffer_ensure()` and have a internal function to do the null-termination.

I like the new flag, maybe we should try it.

> 
> 
> ---
> https://github.com/solettaproject/soletta/pull/532#issuecomment-131101297

-- 
Gustavo Lima Chaves

cmarcelo:
+2 for this PR. I'll change arena to make use of buffers.

As for the next steps:

- `sol_buffer_at_slice`: slice has all the information needed, why not just use slice.data (or at least be a slice operation)?
- Clear slice should also be a slice operation.

Demanding buffer in all those operations will make us having to pass both buffer and slice around (kind of like we passed void *, size around -- which was the point of creating slice in the first place).

The flag seems a good idea.

barbieri:
Because of what is a slice and the memory management.

Slices lifetime are subject to the lifetime of a buffer's internal `buf->data`. All non-const operations will invalidate slices, then we should double check. 

Also we don't want to modify slices, at least not str-slices. That's because them to be usable with const memory and we should never modify them, that also allows compiler to detect we're assigning a literal to a const char and put it in a read-only section.

However I understand your point and I did think about it before. We could introduce a buffer slice, that would be mutable and references the actual buffer. Or make buffers recursive, only the parent would allocate memory, appending to an inner buffer would cause the grand-parent to realloc, move the sibling buffers, then make space for the inner buffer and set that data. In that PoV it would work and would be nice... but it's not a common API and the cost to implement that right internally is huge, so I'm not convinced we should do it.

My docummentation patch will mark all the functions that invalidate existing slices, and mention that for `sol_buffer_get_slice()` and variants.

That said, I think the proposed API that operates on buffers and double check the slice is contained in that buffer is easy and safe. Particularly for things like replacing slices, a tricky operation to do right in all cases, if we rely on users doing all those conditions at call site, we are very likely to get errors.

barbieri:
Merged

cmarcelo:
Replying to myself for the record:

> I'll change arena to make use of buffers.

It's not trivial because the clients of arena need stable pointers and realloc defeats that.

cmarcelo:
Gustavo said:

> Also we don't want to modify slices, at least not str-slices.

Fair, this is a reason for not having clear as a slice operation. I was considering a more relaxed vision of a slice as a simple replacement for ptr+size. Things like "do something funny here" would be able to take a slice instead of a ptr+size. But I don't recall now cases in which this would help.

For `slice_get_at` I still think should be a slice operation. If you pass an invalid slice, you are doing the wrong thing anyway, you only will be able to access a subset of it. Otherwise the client code will simply do this locally (it has data and size), because is not always convenient to pass the buffer around (and sometimes you won't even have it).

PR #537, opened 2015-08-14
sol-util: start using sol-util-$(PLATFORM) v5

ricardotk:
Differences from v4:
- rename sol-util-linux to sol-util-file

Differences from v3:
- do not use impl header and impl functions
- remove sol-util-linux header and put the functions on sol-util header to be implemented on whichever platform

Differences from v2:
- maintain util on libshared.mod instead of creating util.mod

Differences from v1:
- fixed kbuild makefile

Since we were using a lot of platform #ifdefs in sol-util.c


cmarcelo:
+1

It would be useful for the history to explain why the rename.

barbieri:
:+1:

PR #538, opened 2015-08-14
Build duktape cleanup

dorileo:
## Rationale

This series does:
   - remove the submodule initialization from build system;
   - avoid failing if duktape is not present - we simply disable it;
   - add a mechanism to provide DUKTAPE_SRC variable so we can do external duktape clones and use it from there - this is a distro building requirement, namely clear linux;

Besides the feature set of changes the series also brings a patch to fix an -gen.h installation bug;

### DIFF
--- a/README.md
+++ b/README.md
@@ -17,6 +17,9 @@ targets.
 
  * [General Information](#general-information)
  * [Building from Source](#building-from-source)
+   * [Submodules](#submodules)
+       * [Duktape](#duktape)
+   * [Building](#building)
  * [Debug](#debug)
  * [Libraries](#libraries)
   * [Common](#common)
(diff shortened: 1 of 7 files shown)


barbieri:
overall the patches are okay per se, but I really dislike the huge special treatment to handle the duktape case.

Particularly considering its usage at this moment -- none as the JS support is too barebones and nobody is actually working on it. No timers, no XmHttpRequest, no time bounded execution (then it can block the main loop).

I guess I'll work into making FBP understand external modules and we add duktape as a separate module generated by a separate project, similar to SML that provides Soletta modules from outside. Then in that project we include duktape and do the mess all in there.

dorileo:
v2 coming...

PR #539, opened 2015-08-14
Fix coverity error report on linux-micro/console.

glima:
Also fix dangling usage of a variable.


glima:
@vcgomes, I almost split them, but being so small, I retreated. Merging, then. Thanks.

PR #540, opened 2015-08-14
IIO Address

edersondisouza:
A tentative to address iio devices other than its id. Accepting name, absolute path on sysfs and for i2c, a relative path or even instructions to create the device.
I'm specially concerned about the instructions: it's a relative simple comma-separated string, that I started using for tests and survived to final form of this patch, reason: not sure what other format use (json?) that kept things simple. Last commit on this series (the one which changes the node) have examples of strings on its message. It's simple, but you, strings...


edersondisouza:
#567

PR #541, opened 2015-08-14
build fixes

ceolin:
Problems found when building using multilib

### DIFF
--- a/src/lib/common/Makefile
+++ b/src/lib/common/Makefile
@@ -53,8 +53,6 @@ obj-core-$(PLATFORM_CONTIKI) += \
 obj-core-$(PLATFORM_LINUX) += \
     sol-platform-linux-common.o \
     sol-board-detect.o
-obj-core-$(PLATFORM_LINUX)-extra-cflags += $(SYSTEMD_CFLAGS)
-obj-core-$(PLATFORM_LINUX)-extra-ldflags += $(SYSTEMD_LDFLAGS)
 
 obj-core-$(PLATFORM_SYSTEMD) += \
     sol-platform-impl-systemd.o
--- a/src/lib/common/sol-log-impl-linux.c
+++ b/src/lib/common/sol-log-impl-linux.c
@@ -38,7 +38,7 @@
 #include <unistd.h>
 #include <syslog.h>
 
-#ifdef HAVE_SYSTEMD
+#ifdef PLATFORM_SYSTEMD
 #include <systemd/sd-journal.h>
 #endif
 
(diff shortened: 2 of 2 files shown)


glima:
+1

src/lib/common/Makefile
```diff
@@ -53,8 +53,6 @@ obj-core-$(PLATFORM_CONTIKI) += \
 obj-core-$(PLATFORM_LINUX) += \
     sol-platform-linux-common.o \
     sol-board-detect.o
-obj-core-$(PLATFORM_LINUX)-extra-cflags += $(SYSTEMD_CFLAGS)
-obj-core-$(PLATFORM_LINUX)-extra-ldflags += $(SYSTEMD_LDFLAGS)
```
barbieri: just removing it doesn't look right. Maybe add a obj-core-$(PLATFORM_SYSTEMD)... ?

also, we need to link to systemd if either platform-systemd || sd-bus is used.

barbieri: ouch, just read more of that Makefile and right below it there is a block:
```
obj-core-$(PLATFORM_SYSTEMD)-extra-cflags += $(SYSTEMD_CFLAGS) $(UDEV_CFLAGS)
obj-core-$(PLATFORM_SYSTEMD)-extra-ldflags += $(SYSTEMD_LDFLAGS) $(UDEV_LDFLAGS)
```

But we may still fail if building linux-micro on a platform where sd-dbus is to be used.

ceolin: It is not a problem due:

obj-core-$(SOL_BUS)-extra-cflags += $(SYSTEMD_CFLAGS)
obj-core-$(SOL_BUS)-extra-ldflags += $(SYSTEMD_LDFLAGS)

dorileo: actually it's not a problem due to:

```make
obj-core-$(PLATFORM_SYSTEMD)-extra-cflags += $(SYSTEMD_CFLAGS) $(UDEV_CFLAGS)
obj-core-$(PLATFORM_SYSTEMD)-extra-ldflags += $(SYSTEMD_LDFLAGS) $(UDEV_LDFLAGS)
```

ceolin: the point is, if you disable systemd and enable sd-bus the target that will link with systemd is the SOL_BUS isn't ? Or are you saying that if disable systemd it still exists obj-core-$(PLATFORM_SYSTEMD) ?

dorileo: all right missed the sd-bus... @ceolin it's right..

barbieri: +1 then

ceolin:
merged, thanks guys

PR #545, opened 2015-08-14
Add USE_ICU

cmarcelo:
First patch is kind of independent, but touches the string module, so it's here.

I'm not sure `USE_ICU` is in the right place, though.

@glima @dorileo @ibriano thoughts?

Glima told me the replace functions are implemented in a separate C file because they have a different license from the rest of the project. Otherwise we would probably inline them in `string-ascii.c` and `string-icu.c`.


cmarcelo:
Removed the redundant inclusion of `unicode/uchar.h` and rebased.

Applied in 4157352c84a2d2b8e9a13e8981f5c27c47ceac97 and parent.

PR #547, opened 2015-08-14
Add Contiki demo application

zehortigoza:


src/samples/flow/contiki-demo/contiki-demo.fbp
```diff
@@ -0,0 +1,60 @@
...
+# This is a trash disposer that was write to show Soletta working on a board
+# running Contiki.
+# The behavior here is pretty simple, you press the button it make the servo
+# motors to spin 180 degrees, causing the trash being thrown away, after
+# 3 seconds it return to the initial position.
+
+base_servo(servo-motor/controller:chip=0,duty_cycle_range=min:1000|max:2000,period=20000,pin=0)
+top_servo(servo-motor/controller:chip=0,duty_cycle_range=min:1000|max:2000,period=20000,pin=1)
+button(gpio/reader:active_low=false,edge_falling=true,edge_rising=true,pin=15)
+led(gpio/writer:active_low=false,pin=16)
```
barbieri: IMO we should have this file not named after contiki, just have a contiki+board ".conf" file and make it also run on desktop with another conffile using gtk and we may provide one for galileo+linux using some of our reference sensors.

Then we have one more test to show ".conf" capabilities.

What we could do, and here @dorileo may be of help, is to get this to be converted to C and compiled only if a certain ".conf" exists.

Say we declare a `BOARD_CONFFILE` in our Kconfig and Makefile. This FBP->C would rely on existence of `BOARD_CONFFILE` in the FBP source directory (if relative) or an absolute path. If that conffile doesn't exist, this sample is skipped.

This allow this test to be compiled and tested on other OS and boards such as Galileo with Groove Starter Kit.

src/samples/flow/contiki-demo/contiki-demo.fbp
```diff
@@ -0,0 +1,60 @@
...
+button_active OUT -> IN _(converter/empty-to-int:output_value=2000) OUT -> DUTY_CYCLE base_servo
+button_active OUT -> IN _(converter/empty-to-int:output_value=2000) OUT -> DUTY_CYCLE top_servo
+button_active OUT -> IN _(converter/empty-to-boolean:output_value=true) OUT -> ENABLED timer_reset
+button OUT -> IN _(boolean/filter) TRUE -> IN led
+
+# after 3 seconds of button pressed
+timer_reset OUT -> IN _(converter/empty-to-int:output_value=1000) OUT -> DUTY_CYCLE base_servo
+timer_reset OUT -> IN _(converter/empty-to-int:output_value=1000) OUT -> DUTY_CYCLE top_servo
+timer_reset OUT -> IN _(converter/empty-to-boolean:output_value=false) OUT -> IN led
+timer_reset OUT -> IN _(converter/empty-to-boolean:output_value=false) OUT -> ENABLED timer_reset
```
barbieri: missing newline

zehortigoza:
New version coming, thanks all for the input.

PR #554, opened 2015-08-17
Remove get_ports_counts() from sol_flow_node_type

cmarcelo:
We know the sizes upfront and our design implies that these values won't change for an existing type, so set them sooner in a more convenient way. As a bonus, fix a longstanding TODO item.


ibriano:
Ack

cmarcelo:
Applied in 6ea341534de4a33239883bd538f84307c5748bad and parents.

PR #557, opened 2015-08-17
Accept node types (and options) in sol-fbp-runner

cmarcelo:


bdilly:
It seems ok.
However, I didn't test it.

cmarcelo:
Applied in 25103c715810ea005ed4ddc4d3df7249ba8caf9e and parent.

It's easier for more people to test this when landed, so we can flush out any remaining bugs.

PR #558, opened 2015-08-17
gtk: make the window scrollable

cmarcelo:
This is useful when simulating types/flows in the runner that have array
ports.

### DIFF
--- a/src/modules/flow/gtk/window.c
+++ b/src/modules/flow/gtk/window.c
@@ -62,6 +62,7 @@ struct window *
 window_new(void)
 {
     struct window *w;
+    GtkWidget *scrolled_win;
 
     gtk_init(NULL, NULL);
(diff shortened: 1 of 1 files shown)


cmarcelo:
> Do you need to set a default size to make it scrollable?
> Isn't it auto finding a size enough to make everything visible?

No, it will be scrollable, the default size is not necessary. However, the initial size doesn't take into account the contents being scrolled -- in part because of the way scrollable window widget works -- so we start with a supersmall window. That's why I changed the default.

> Does it means that even if we run a sample with a single button it's going to create a window 600x800?

Yes.

I'll look a bit more, to see if we can get the preferred width / height and set to the window to that (still capped at some height).

cmarcelo:
Made a v2 that is slightly better but still somehow arbitrary wrt the height.

PR #559, opened 2015-08-17
Build duktape cleanup - v2

dorileo:
## Changes
  + v2 - (since: #538)
    + integrated patch: **"build: make sure we have Makefile.gen before actually building"**;
    + dependency-resolver will always resolve filesystem files with abs path;
    + fix issues in Makefile.rules on handling:
      - out-of-tree objects with absolute path;
      - in-tree objects with absolute path;
    + added patch: **"duktape: change the submodule settings to use mirror"**;
    + added patch: **"git-init-submodule: get rid of it"**;

## Rationale

This series does:
   - remove the submodule initialization from build system;
   - avoid failing if duktape is not present - we simply disable it;
   - add a mechanism to provide DUKTAPE_SRC variable so we can do external duktape clones and use it from there - this is a distro building requirement, namely clear linux;

Besides the feature set of changes the series also brings a patch to fix an -gen.h installation bug;


bdilly:
ack

dorileo:
applied

PR #560, opened 2015-08-17
samples/contiki: Simplify servo demo

bdilly:
Reusing these anonymous converters (using a explanatory name)
reduces the amount of nodes and connections and make it easier to read.

@zehortigoza please take a look on it. I can't test it here,
hopefully it works. But even if don't, I believe it's a good
start to simplify this sample.


zehortigoza:
Nice, less 100 on the generated file.

zehortigoza:
Thanks @bdilly I had improved the behavior of the demo and used your ideas on the new version #575

PR #563, opened 2015-08-17
gtk: make the window scrollable [v2]

cmarcelo:
This is useful when simulating types/flows in the runner that have array
ports.

### DIFF
--- a/src/modules/flow/gtk/window.c
+++ b/src/modules/flow/gtk/window.c
@@ -62,6 +62,7 @@ struct window *
 window_new(void)
 {
     struct window *w;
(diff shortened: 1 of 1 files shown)


cmarcelo:
Differences from v1: the width is not hardcoded but calculated based on preferred from the grid. The height is the preferred of the grid as well, but capped at 600 (when the scrollbar appears). User can resize the window if prefers.

cmarcelo:
Replaces #558.

cmarcelo:
Applied in 193df768f1d4176bb8827ca62bd66a0a8d040f57.

PR #564, opened 2015-08-17
Console changes

tcanabrava:
Try to simplify the string handling on the error codes,
also fire an error packet ( and not a warning ) if a invalid node type is sending requests to the console.


cmarcelo:
@tcanabrava I ended up making a v2 of your patch, I hope it's not a problem.

PR #566, opened 2015-08-18
tools: toggle-config

cmarcelo:
Command line way to toggle a config option. Example usage

    $ tools/toggle-config use_icu
    # USE_ICU is not set

    Toggling and running make oldconfig

    USE_ICU=y

and the reverse also works

    $ tools/toggle-config use_icu
    USE_ICU=y

    Toggling and running make oldconfig

    # USE_ICU is not set

### DIFF
--- a/tools/toggle-config
+++ b/tools/toggle-config
@@ -1,0 +1,96 @@
+#!/bin/bash
+
+# This file is part of the Soletta Project
+#
+# Copyright (C) 2015 Intel Corporation. All rights reserved.
+#
+# Redistribution and use in source and binary forms, with or without
+# modification, are permitted provided that the following conditions
+# are met:
+#
+#   * Redistributions of source code must retain the above copyright
+#     notice, this list of conditions and the following disclaimer.
+#   * Redistributions in binary form must reproduce the above copyright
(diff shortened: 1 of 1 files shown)


cmarcelo:
@dorileo @vcgomes could you look? Maybe there's a way to do this directly in kconfig that I'm missing.

barbieri:
I think that just toggle is not that helpful. It would help to be able to get or set to a given value. toggle being an option.

Could you make it a:

```sh
   tools/kconfig-option --get use_icu
   tools/kconfig-option --set use_icu y
   tools/kconfig-option --set use_icu n
   tools/kconfig-option --set use_icu m # not really for icu, but you got the idea
   tools/kconfig-option --toggle use_icu # if y or m, then n. Otherwise then y? or m?
```

vcgomes:
I agree with @barbieri, perhaps adding a `--force` option would be an acceptable middle ground?

cmarcelo:
Nice suggestion, I'll try a v2 later and ping back. :-)

I disagree about toggle, though. I was in two situations already that I wanted to try/test something with an USE_ flag both enabled and disabled.

PR #567, opened 2015-08-18
IIO Address v2

edersondisouza:
v2:
Now accepts a list of 'commands' to create a device.
Commits of 'resolving' and 'creation' squashed, as now they are treated inside of command list
Sample updated to show this feature.
i2c api accepts `dev_number` as a number.
i2c api returns -EEXIST when device already exists. No more combination of boolean return and int to find out the truth.

A tentative to address iio devices other than its id. Accepting name, absolute path on sysfs and for i2c, a relative path or even instructions to create the device.
I'm specially concerned about the instructions: it's a relative simple comma-separated string, that I started using for tests and survived to final form of this patch, reason: not sure what other format use (json?) that kept things simple. Last commit on this series (the one which changes the node) have examples of strings on its message. It's simple, but you, strings...

### DIFF
--- a/src/lib/io/include/sol-i2c.h
+++ b/src/lib/io/include/sol-i2c.h
@@ -37,6 +37,7 @@
 #include <sys/types.h>
 
 #include <sol-macros.h>
+#include <sol-buffer.h>
 
 #ifdef __cplusplus
 extern "C" {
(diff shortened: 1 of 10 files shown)


edersondisouza:
Anyone? Specially @barbieri and @lucasdemarchi

barbieri:
+1

edersondisouza:
Integrated.

PR #569, opened 2015-08-18
sol-iio: Make creation of sysfs triggers more reliable

edersondisouza:
Still guessing a number, but re-trying in case trigger name already
exists.
Not going on a deterministic order because we may end up failing many
times before creating a trigger - e.g., creating ten new triggers
starting from '1' will end up with 45 failed tries.
Even we save a global 'last trigger id', we could conflict with existing
ones.
Using normal rand() as this is *not* cryptography related.


edersondisouza:
Typo fixed and integrated. Tks!

PR #570, opened 2015-08-18
dep-resolver: log dependency tests

dorileo:
This patch changes dependency-resolver.py to log the dependency test's
source code, run command, output and so on to .config.log file to ease
the process of debugging - distros ' packaging issues.

### DIFF
--- a/data/scripts/dependency-resolver.py
+++ b/data/scripts/dependency-resolver.py
@@ -46,6 +46,7 @@ class DepContext:
     def __init__(self):
         self.kconfig = {}
         self.makefile_vars = {}
+        self.log_content = []
 
     def add_kconfig(self, k, t, v):
         self.kconfig[k] = {"value": v, "type": t}
@@ -80,10 +81,23 @@ class DepContext:
             return var["value"]
         return ""
 
-def run_command(cmd):
+    def add_log_content(self, line):
+        self.log_content.append(line)
+
+    def write_log_content(self, path):
+        if not self.log_content:
+            return
+
+        with open(path, "w+") as f:
+            for line in self.log_content:
+                f.write("%s\n" % line)
+
+def run_command(cmd, context):
     try:
         output = subprocess.check_output(cmd, stderr=subprocess.STDOUT,
                                          shell=True, universal_newlines=True)
+        context.add_log_content("Running command: %s" % cmd)
+        context.add_log_content("Command output:\n%s" % (output if output else "None"))
         return output.replace("\n", "").strip(), True
     except subprocess.CalledProcessError as e:
         return e.output, False
@@ -99,17 +113,17 @@ def handle_pkgconfig_check(args, conf, context):
 
     if exact_ver:
         cmd = "%s --exact-version=%s %s" % (args.pkg_config, exact_ver, pkg)
-        result, status = run_command(cmd)
+        result, status = run_command(cmd, context)
         if not status:
             ver_match = False
     elif atleast_ver:
         cmd = "%s --atleast-version=%s %s" % (args.pkg_config, atleast_ver, pkg)
-        result, status = run_command(cmd)
+        result, status = run_command(cmd, context)
         if not status:
             ver_match = False
     elif max_ver:
         cmd = "%s --max-version=%s %s" % (args.pkg_config, max_ver, pkg)
-        result, status = run_command(cmd)
+        result, status = run_command(cmd, context)
         if not status:
             ver_match = False
 
@@ -119,8 +133,8 @@ def handle_pkgconfig_check(args, conf, context):
         cflags_cmd = "%s --cflags %s" % (args.pkg_config, pkg)
         ldflags_cmd = "%s --libs %s" % (args.pkg_config, pkg)
 
-        cflags, cflags_stat = run_command(cflags_cmd)
-        ldflags, ldflags_stat = run_command(ldflags_cmd)
+        cflags, cflags_stat = run_command(cflags_cmd, context)
+        ldflags, ldflags_stat = run_command(ldflags_cmd, context)
 
         if cflags_stat:
             context.add_cond_makefile_var("%s_CFLAGS" % dep, cflags)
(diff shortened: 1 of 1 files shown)


data/scripts/dependency-resolver.py
```diff
@@ -80,10 +81,23 @@ def find_makefile_var(self, v):
...
+        self.log_content.append(line)
+
+    def write_log_content(self, path):
+        if not self.log_content:
+            return
+
+        with open(path, "w+") as f:
+            for line in self.log_content:
+                f.write("%s\n" % line)
+
```
barbieri: do not cache this to be later written to the file. Write to the file and flush, so if it stops for some reason (ie: exception, user's ^C) we can have parts of the log in the disk.

data/scripts/dependency-resolver.py
```diff
@@ -80,10 +81,23 @@ def find_makefile_var(self, v):
...
+def run_command(cmd, context):
     try:
         output = subprocess.check_output(cmd, stderr=subprocess.STDOUT,
                                          shell=True, universal_newlines=True)
+        context.add_log_content("Running command: %s" % cmd)
+        context.add_log_content("Command output:\n%s" % (output if output else "None"))
```
barbieri: not enough.

as per above, write directly to the file. And write the command to be executed BEFORE executing it. Also write the `os.environ` sorted by key as it impacts how the execution happens.

It would be nice to get more information about the errors... maybe `subprocess.check_output()` is not enough, maybe it is, just need to log from the exception. But the exit code and error messages would be nice to have, if possible separate stdout from stderr, prefixing lines.

data/scripts/dependency-resolver.py
```diff
@@ -80,10 +81,23 @@ def find_makefile_var(self, v):
...
         return output.replace("\n", "").strip(), True
     except subprocess.CalledProcessError as e:
         return e.output, False
```
barbieri: would be nice to log the errors.

data/scripts/dependency-resolver.py
```diff
@@ -141,7 +155,8 @@ def compile_test(source, compiler, cflags, ldflags):
     output = "%s-bin" % f.name
     cmd = "{compiler} {cflags} {src} -o {out} {ldflags}".format(compiler=compiler,
             cflags=cflags, ldflags=ldflags or "", src=f.name, out=output)
-    out, status = run_command(cmd)
+    context.add_log_content("Compiling source code:\n%s" % source)
```
barbieri: is this printing the actual source or just the temporary location? It would be nice to print the source.

data/scripts/dependency-resolver.py
```diff
@@ -240,7 +255,8 @@ def handle_python_check(args, conf, context):
     f.close()
 
     cmd = "%s %s" % (sys.executable, f.name)
-    output, status = run_command(cmd)
+    context.add_log_content("Testing python code:\n%s" % source)
```
barbieri: source or location? as above.

data/scripts/dependency-resolver.py
```diff
@@ -257,9 +273,13 @@ def handle_python_check(args, conf, context):
 def test_file_path(path, files):
     files_len = len(files)
     for curr in files:
+        context.add_log_content("Lookingup file: %s, at %s" % (curr, path))
         exists = os.path.exists(os.path.join(path, curr))
```
barbieri: could simplify in a single line, giving `exists` to the log.

data/scripts/dependency-resolver.py
```diff
@@ -285,8 +306,10 @@ def handle_filesystem_check(args, conf, context):
     if found_path:
         context.add_kconfig("HAVE_%s" % dep.upper(), "bool", "y")
         context.add_cond_makefile_var("%s_PATH" % dep.upper(), found_path)
+        context.add_log_content("%s found" % dep)
     else:
         context.add_kconfig("HAVE_%s" % dep.upper(), "bool", "n")
+        context.add_log_content("%s found" % dep)
```
barbieri: found? but you set the kconfig to n. Maybe it's better to just print:

`"kconfig: HAVE_%s bool %s" % (dep.upper(), setting)`

also, it's better to do this from inside `add_kconfig()`

data/scripts/dependency-resolver.py
```diff
@@ -395,9 +419,12 @@ def run(args, dep_checks, context):
             exit(1)
 
         result = handler(args, dep, context)
+        context.add_log_content("#############################")
         if verbose:
             print("found." if result else "not found.")
 
+    context.write_log_content(".config.log")
```
barbieri: as above, just log from the beginning. You can move the old log to `.config.log.old` if you want to keep it.

data/scripts/dependency-resolver.py
dorileo: fair enough...

data/scripts/dependency-resolver.py
dorileo: ok

data/scripts/dependency-resolver.py
dorileo: ok

data/scripts/dependency-resolver.py
dorileo: it's printing the source code itself...

data/scripts/dependency-resolver.py
dorileo: source...

data/scripts/dependency-resolver.py
dorileo: do you mean the handle_python_check() implementation overall?

data/scripts/dependency-resolver.py
dorileo: Ok, I just didn't include the kconfig key:value pairs just because it's written to Kconfig.gen file any ways...

data/scripts/dependency-resolver.py
dorileo: all right.

data/scripts/dependency-resolver.py
barbieri: the log could be done in a single python call:
```py
         exists = os.path.exists(os.path.join(path, curr))
         context.add_log_content("Lookingup file: %r at %r: %s" % (curr, path, exists))
```

dorileo: humm! ok, I see..

dorileo:
v2 coming....

PR #571, opened 2015-08-18
Generator fixes [v2]

cmarcelo:
Fixes two problems pointed in #562: first about invalid code being generated from conffile option values, and second the repeated include headers.

Differences from v1:
- In the first patch, also fix the fact that we shouldn't generate includes for types that come from `DECLARE`, they will be declared/defined in the generated C file.


cmarcelo:
Replaces #568.

vcgomes:
The series looks good.

ricardotk:
:+1:

cmarcelo:
Applied in 7999f632d3124e47daced181397bb7fdcad612bd.

PR #572, opened 2015-08-18
Build general improvements

dorileo:


dorileo:
v2 coming... hopefully no build breaks this time...

PR #573, opened 2015-08-18
Build general improvements v2

dorileo:
## Changes
  + v2 (since #572):
     - added patch: **	check-license: ignore makefiles**;
     - added patch: **	build: generate soletta.pc on build time**;

PR #574, opened 2015-08-18
Build general improvements v3

dorileo:
## Changes
  + v2 (since #572):
     - added patch: **	check-license: ignore makefiles**;
     - added patch: **	build: generate soletta.pc on build time**;
  + v3 (since #573):
     - changed the Makefile ignore rule in: **check-license: ignore makefiles**;


ibriano:
Seems fine

PR #575, opened 2015-08-18
Add a trash disposer flow application V2

zehortigoza:
Changes from V1:
- Change the example behavior, now the top servo only spin 250ms after the button is pressed, this way the trash only will fall after the base servo reach the destination.
- Using @bdilly ideas of improvement from #560 
- Moving servo and pins configuration to a conf file


bdilly:
:ok_hand:

zehortigoza:
Fixed comments above and merged.
Thanks all

PR #578, opened 2015-08-19
dep-resolver: log dependency tests

dorileo:
## Changes
  + v2 (since #570):
    - fixed the issues pointed by @barbieri in #570 review;
    - changed the messages being printed with ```print()``` to use the introduced log facilities;
    - fixed an issue with the filesystem's check return value;
    - turned the printed messages more autoconf like;

## Rationale
This patch changes dependency-resolver.py to log the dependency test's
source code, run command, output and so on to .config.log file to ease
the process of debugging - distros ' packaging issues.

We also fix an issue on filesystem check return.


barbieri:
:+1:

dorileo:
applied...

PR #582, opened 2015-08-19
modules/temperature: add converter node type

bdilly:
It converts temperatures between Kelvin, Celsius, Fahrenheit and
Rankine.


bdilly:
@glima I understand your idea regarding naming schema. As we've discussed in IRC channel with @edersondisouza , the point on having schema as model/node was to support better cases when a model has multiple readers (one of them has x/temperature and x/humidity). Also it makes even more sense for me when we think about thirdparty modules, when people wouldn't be able to add nodes to temperature/ module.

bdilly:
merged

PR #585, opened 2015-08-19
Regexp

glima:
New string/regexp-search node. regexp-substitute will now follow that.


glima:
Thanks, @dorileo and @edersondisouza , now opening a v2.

PR #586, opened 2015-08-19
build: build/gen everything out of src dir

dorileo:
## Changes
  + v2 (since #583):
    - also include ```-I$(build_stagedir)/lib/flow``` and ```-I$(build_stagedir)/lib/common```

## Rationale
We don't want to have anything being built or generated inside the src
dir, this patch removes all the remaining resources being generated
there.

The only remaining resource being generated in the source dir is the
oic.json - which in my option should remain there.

See: https://github.com/solettaproject/soletta/issues/561


barbieri:
:+1:

cmarcelo:
+1

dorileo:
applied..

PR #587, opened 2015-08-19
Fix addition and multiplication for int and float

bdilly:
broken when array ports was added


lgywata:
+2

bdilly:
Thanks for reviewing it @lgywata 
Merged

PR #590, opened 2015-08-20
Regexp [v2]

glima:
Changes since v1:

Errors pointed by @edersondisouza fixed (leak, bad define removal, etc), warning messages now make use of new @dorileo 's schema (we just need that after the config phase now), new help string for the string nodes at the config phase.

### DIFF
--- a/data/jsons/dependencies.json
+++ b/data/jsons/dependencies.json
@@ -243,6 +243,12 @@
       "atleast-version": "52.1"
     },
     {
+      "dependency": "libpcre",
+      "type": "pkg-config",
+      "pkgname": "libpcre",
+      "atleast-version": "8.12"
+    },
+    {
       "dependency": "isatty",
       "type": "ccode",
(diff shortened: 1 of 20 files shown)


glima:
There's also a new commit fixing misuse of  sol_flow_send_error_packet throughout the code.

edersondisouza:
+1

edersondisouza:
@glima Remember to update wiki about dependences when this series goes in.

glima:
Ack! Should you raise your review to +2 and I'll do it right away :)
Also, I removed some redundant nested ifdefs there.

edersondisouza:
Hehe. lgtm, but not sure about build stuff that @dorileo is the one who knows =D

dorileo:
I didn't check in depth the added C code, but what concerns build system it's ok.
+1

glima:
Alright, merging.

PR #593, opened 2015-08-20
tools/run-fbp-tests: support python 2 and 3

bdilly:
With a few changes and imports it seems to be working fine.


anselmolsm:
@bdilly interesting, I didn't know about it.
But, what about the line 1 which refers to python3?

cmarcelo:
+1

Probably we should refer to "#!/usr/bin/env python" only?

anselmolsm:
@cmarcelo yep, this should work, I think.

bdilly:
@cmarcelo @anselmolsm you're right.
change done
merged

PR #594, opened 2015-08-20
comms:http:client: Make possible cancel a request

ceolin:
- Delete idler and watch on connection destroy


ceolin:

PR #595, opened 2015-08-20
sol-conffile: Remove quotes from strings.

cabelitos:
When an options is a string, the quotes should be removed from it.


bdilly:
uhmm... please take a look on #596

bdilly:
Closing it since we agreed to move with #596

PR #596, opened 2015-08-20
sol-conffile: Remove quotes from strings

bdilly:
When an options is a string, the quotes should be
removed from it.


bdilly:
@cabelitos please test it and see if it solves your issues.

@barbieri pointed out that we're missing a function to handle escape codes:

escapable_chars[] = {'\\', '\"', '/', '\b', '\f', '\n', '\r', '\t'};

so I've added that to my todo list. But based on your previous PR, I believe it's not impacting you.

glima:
We already have that in our parser, don't we?

cabelitos:
@glima Yes, there's a unesacape_string function.

Btw, this patch fix the problem for me.
Should I close my pullreq?

bdilly:
@glima @cabelitos yeah, we have it in our flow parser indeed.
So maybe it's just a matter of exposing it in our json parser lib.
I'll take a look on that soon.
Thanks

Merged

PR #597, opened 2015-08-21
Iio raw buffer - and other improvements

edersondisouza:
When buffer is enabled, it's possible to get a `sol_str_slice` with channel buffer raw contents. Useful specially for channel whose storage is longer than 64 bits.
Also:
Fixed Kconfig;
Useless wrong attribution (-1 to an ` unsigned int`);
Read channel value, when buffer is disabled, into a `int64_t` var.


glima:
Makes sense to me (but others with more familiarity with iio might want to ack too).

edersondisouza:
Integrated. Tks!

PR #598, opened 2015-08-21
Add support for flow metatype modules

cmarcelo:
Metatypes are the implementations behind the `DECLARE=...` extension we have in our FBP files. We currently have two metatypes: `fbp` (which will remain builtin) and `js`. This PR adds support for metatype modules (like we have type modules for the node types), and make `js` one of those.

One important consequence is that JS code now doesn't need to live inside the flow subsystem.


cmarcelo:
@barbieri could you take a look at the PR?

@vcgomes I think you might want to take a look at "shared: add sol_lib_loader" in particular.

cmarcelo:
@dorileo could you look at "build: function that decides..."?

cmarcelo:
Rebased with the fixes and sent PR #613 for the bots to consume. After they are happy I'll apply that version. Thanks.

PR #599, opened 2015-08-21
datatypes/vector: fix memset arithmetics

bdilly:
Issue reported by coverity:

Suspicious implicit sign extension: "n" with type "unsigned short"
(16 bits, unsigned) is promoted in "v->elem_size * n" to type "int"
(32 bits, signed), then sign-extended to type "unsigned long" (64 bits,
unsigned). If "v->elem_size * n" is greater than 0x7FFFFFFF, the
upper bits of the result will all be 1.


bdilly:
new version #601

PR #601, opened 2015-08-21
datatypes/vector: fix memset arithmetics [v2]

bdilly:
changelog:
  * added changes suggested by @cmarcelo regarding checks
  of overflow on grow()

Issue reported by coverity:

Suspicious implicit sign extension: "n" with type "unsigned short"
(16 bits, unsigned) is promoted in "v->elem_size * n" to type "int"
(32 bits, signed), then sign-extended to type "unsigned long" (64 bits,
unsigned). If "v->elem_size * n" is greater than 0x7FFFFFFF, the
upper bits of the result will all be 1.

### DIFF
--- a/src/lib/datatypes/sol-vector.c
+++ b/src/lib/datatypes/sol-vector.c
@@ -45,10 +45,13 @@ sol_vector_init(struct sol_vector *v, uint16_t elem_size)
     v->elem_size = elem_size;
 }
 
+#define SIZE_MULTIPLICATION_OVERFLOW(var0, var1) \
+    ((var0 > 0) && (var1 > 0) && (var0 > SIZE_MAX / var1))
+
 static int
 sol_vector_grow(struct sol_vector *v, uint16_t amount)
 {
-    unsigned int new_cap, old_cap;
+    size_t new_cap, old_cap;
     uint16_t new_len;
 
     if (v->len > UINT16_MAX - amount)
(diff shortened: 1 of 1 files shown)


barbieri:
@edersondisouza is right, use that function.

cmarcelo:
+1 with ederson suggestion.

bdilly:
I wasn't aware of this function. Perfect
Thanks

barbieri:
hum.. actually, reading this code and the coverity message I wonder why `uint16_t * uint16_t` is promoted to signed integer, that's the case if it says `v->elem_size * n` is promoted to integer.

Actually, given both are `uint16_t` they shouldn't overflow even if `sizeof(size_t) == sizeof(uint32_t)` (32 bits machine). I guess this warning is kinda moot, maybe we just need to adapt the types before doing the multiplication to `uint32_t` (or `size_t`) before doing the multiplication and no extra checks needs to be done.

Could someone with a local coverity tool test if a cast similar to what was done in this patch for `sol_vector_append()` will fix it?

vcgomes:
From this ( https://www.securecoding.cert.org/confluence/display/c/INT02-C.+Understand+integer+conversion+rules ) the C standard mandates that both operands are promoted to `signed int`s (as it can hold all values of an `uint16_t`).

bdilly:
merged with @edersondisouza suggestion

PR #603, opened 2015-08-21
Make it possible for the user to disable libpcre dependency lookup.

glima:
Also move HAVE_ICU usage back to USE_ICU.

As requested by Caio, who is dealing with with builds for small OSes and still has no easy way of skipping these (host) dependency auto-detections.

### DIFF
--- a/src/modules/flow/string/Kconfig
+++ b/src/modules/flow/string/Kconfig
@@ -1,7 +1,15 @@
-# This is the solely user of ICU. If gets wider use, move up in the hierarchy.
+# This is the solely user of ICU. If gets wider use, move up in the
+# hierarchy.
 config USE_ICU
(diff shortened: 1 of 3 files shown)


barbieri:
+1

but as a note to @cmarcelo, if you're building for small OS and it was detecting pcre.h, then you're leaking host stuff in the target build. Check if your CFLAGS and LDFLAGS are correct, such as `--sysroot=` or `-nostdinc`. If the check uses pkg-config, set its envvars accordingly: https://autotools.io/pkgconfig/cross-compiling.html

cmarcelo:
+2

@glima, thanks.

@barbieri, thanks. I'll re-check that.

glima:
Thanks, merging.

PR #604, opened 2015-08-21
build: make sure we've kconfig/conf built

dorileo:
Since we're now calling kconfig/conf the rule should make sure it's
built. This patch does that by implicitly declaring it as its dependency.


barbieri:
:+1:

PR #605, opened 2015-08-21
build: add Requires.private to soletta.pc

dorileo:
## Changes
  + v2 (since #602):
     - fixed issues @barbieri pointed on #602;

## Rationale
This patch introduces a new private-requires-y and private-requires-m
variables to declare which pkg dependencies we're using, with that we
can add that to our distributed soletta.pc file.

For that we're also considering env vars in the template parser so we
can easily export&get useful information while processing template
files.


barbieri:
:+1:

dorileo:
merged...

PR #606, opened 2015-08-21
build: sol_config.h should be regenerated with .config-cache

dorileo:
If .config-cache was regenerated the sol_config.h should be as well. A
common test case is, make:

$ make alldefconfig && make

add new dependencies to dependencies.json then:
$ make reconf

sol_config.h should be regenerated. This patch fixes the issue.


barbieri:
:+1:

dorileo:
applied...

PR #609, opened 2015-08-24
build: split static and shared lib build

dorileo:
This patch changes how shared and static libraries are built. Currently
we build both shared and static, with this patch we'll build only one of
them. The user can choose to build static or library using kconfig
interfaces, the default being shared.

If static is chosen the .so external modules are not built, everything
goes builtin - actually it's done before that way, but since we were
building the shared library together we should also generate the
external modules so it could be used by the shared library users.

Next step - as requested by Jesus - is to split the flags tested by
dependency-resolver, so we don't introduce compilation flags specific to
shared libraries and get the static build affected.

### DIFF
--- a/Kconfig
+++ b/Kconfig
@@ -12,9 +12,22 @@ config BASE_OS
 source "tools/build/Kconfig.$BASE_OS"
 
 menu "Basic"
+choice LIB_OUTPUT
+	prompt "Soletta library output"
+	default SHARED_LIBRARY
+
+config SHARED_LIBRARY
+	bool "Shared"
+	depends on FEATURE_DYNAMIC_LINKER
+
+config STATIC_LIBRARY
+	bool "Static"
+endchoice
+
 # backward compatibility with autotools vars
 config ENABLE_DYNAMIC_MODULES
 	bool
+	depends on SHARED_LIBRARY
 	default n
 
 config CC_SANITIZE
@@ -36,11 +49,6 @@ config CC_SANITIZE_ADDRESS
 	bool "address"
 	depends on HAVE_SANITIZE_ADDRESS
 endchoice
-
-config SHARED_LIBRARY
-    bool "Build shared library"
-    default y
-    depends on FEATURE_DYNAMIC_LINKER
 
 config MODULES
 	bool "Enable loadable module support"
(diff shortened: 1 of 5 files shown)


dorileo:
note to self: maybe explicitly disable loadable modules instead of simply get rid of rules...

ibriano:
Seems fine, but I don't understand why you need to make the rules conditional. If we are building a static library, then MODULES will be off and all the module options will either by y or n, so those rules should never run anyway. Or am I misunderstanding something?

ricardotk:
+1 I've tested and its working

dorileo:
@ibriano yep, that what I meant in the "note to self"... I'll change that.

barbieri:
I wonder if we should only do one or another. It is common for libraries to produce both, so why do we need to limit to one?

Kconfig
```diff
@@ -12,9 +12,22 @@ config BASE_OS
...
+	depends on FEATURE_DYNAMIC_LINKER
+
+config STATIC_LIBRARY
+	bool "Static"
+endchoice
+
 # backward compatibility with autotools vars
 config ENABLE_DYNAMIC_MODULES
 	bool
+	depends on SHARED_LIBRARY
```
barbieri: this is not true. You could have a static library (libsoletta.a) that still calls `dlopen()` and loads modules as `.so`.

What you really depend on is the `FEATURE_DYNAMIC_LINKER`

dorileo:
@barbieri we should compile different compiled objects using different sets of flags, for that we would end up compiling the same objects twice for a common use case where we want only one of them.

barbieri:
I don't see why compile different objects with different flags. Only the last step, linking, is different. One is doing an archive (`libsoletta.a`) without actually resolving symbols, while the other does the linking into a shared ELF object (`libsoletta.so`).

dorileo:
just for the record - maybe I want to include this information in the commit msg - we're doing so because we faced some problems with some embedded platforms building soletta as static and using the -fPIC flag, for that we decided to split things up so we don't run the risk of having this kind of problem, now and in the future.

barbieri:
+1

dorileo:
applied with the changes suggested by @ibriano

PR #610, opened 2015-08-24
File fix v2

edersondisouza:
v2:
- Checking for size overflow
- New commit with another check for size overflow


edersondisouza:
Force pushed (since no one reviewed yet) a minor change to first commit: comparison on `sol_buffer_at_end`.

edersondisouza:
Integrated. Tks!

PR #611, opened 2015-08-24
modules/flower-power improvements

bdilly:


src/lib/parsers/sol-json.c
```diff
@@ -289,28 +289,17 @@ sol_json_token_get_double(const struct sol_json_token *token, double *value)
      * of bytes.
      */
 
-    *value = sol_util_strtodn(token->start, &endptr,
+    errno = 0;
+    tmp_value = sol_util_strtodn(token->start, &endptr,
         sol_json_token_get_size(token), false);
+    SOL_INT_CHECK(errno, != 0, -errno);
```
barbieri: why do you want to change this? I tried to provide a "best value" in the case `strtod()` fails, like trying to set it to `-DBL_MAX` if -infinity.

Also, I recall having to set errno if value was `isinf()` or `isnan()`, could you confirm that still happens? Otherwise it would be returning unusable values with return 0.

bdilly: OK, after some discussion we decided to keep the code as it is now and handle errors in module side.
The main point of divergence was that inf and nan are acceptable values for a float, but shouldn't be passed via JSON.

bdilly:
closing since new version was pushed #614

PR #612, opened 2015-08-24
Regexp replace

glima:
New node + a fixes commit.


glima:
Alright, adding

glima:
I'll change the option for no-match scenarios to forward the packet instead, also adding a new commit doing that on string/replace. So wait for a v2 soon.

PR #613, opened 2015-08-24
Add support for flow metatype modules [v2]

cmarcelo:
Metatypes are the implementations behind the DECLARE=... extension we have in our FBP files. We currently have two metatypes: fbp (which will remain builtin) and js. This PR adds support for metatype modules (like we have type modules for the node types), and make js one of those.

One important consequence is that JS code now doesn't need to live inside the flow subsystem.

Differences from v1:
- Various fixes of issues pointed by Ederson and Vinicius.
- Removed first_load (and the result struct) after Vinicius suggestion.

Replaces https://github.com/solettaproject/soletta/pull/598. Uploaded as a PR to see whether the bot is fine.


cmarcelo:
Applied in 446bc7e43c97c7a54cb809980e76fb8552d11259 and parents.

PR #614, opened 2015-08-24
modules/flower-power improvements [v2]

bdilly:
Changelog:
 * drop changes on json parser and handle errors on module
 * cancel pending http requests
 * renames on init macro


cabelitos:
+1

bdilly:
Yeah, on i2c I would name it "handle" or "operation".
Naming a struct with an adjective instead of a noun doesn't seems clear to me, specially with one that is a temporary state (on some cases the connection won't be pending anymore, as it's the case of the finish_cb).
Also you're not sure what is pending. For http-client I was able to guess it was a connection, but for i2c I had to read the documentation.

ceolin:
so go ahead, you're working closer this api than I. We still can change the i2c api soon :)

bdilly:
merged with proposed change

PR #616, opened 2015-08-24
build: turn optional the use of rpath

dorileo:
We don't need to force the users to have rpath installed to have the
build working - and all the tooling.

This patch makes the use of rpath optional.


barbieri:
didn't test, but looks right

dorileo:
just got an error with the bots... fixing.

PR #617, opened 2015-08-24
Wallclock: Add option to send initial packets.

cabelitos:
For some applications sending an initial packet may be necessary.

### DIFF
--- a/src/modules/flow/wallclock/wallclock.c
+++ b/src/modules/flow/wallclock/wallclock.c
@@ -192,20 +192,17 @@ err:
     return -errno;
 }
 
-static bool
-wallclock_do(void *data)
+static void
+wallclock_update_time(enum sol_flow_node_wallclock_type type,
+    struct wallclock_timer *timer)
 {
-    struct sol_flow_node *node = data, *n;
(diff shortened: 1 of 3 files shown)


barbieri:
is it a problem if we send the initial packet? If no, then always send.

cabelitos:
I talked to @bdilly about that and we concluded that leaving this as an option is the best thing to do,
because the user may want to be notified only once every minute.
If we send the initial packet by default, one can be notified twice in one minute. (In case that the user is using wallclock/minute)

bdilly:
She won't be notified twice in the same minute, but the delta between initial packet and next packet may be much smaller then a minute. She can receive two packet in a timeframe smaller than a minute.
Is it a problem? Depends on the usage. I think that leaving it as an option is the safer approach for now, since we lack use cases.

barbieri:
ok

bdilly:
commit seems ok

cabelitos:
@bdilly I tried to express what you said, maybe I was not clear enough :/
Anyway, thanks.

cabelitos:
Merged

PR #618, opened 2015-08-24
build: turn optional the use of rpath - v2

dorileo:
## Changes
  + v2 (since #616):
    - Changed exec tests to export HAVE_* vars to Kconfig.gen instead of Makefile.gen so we can use in the dependency chain;
    - call export to propagate the LD_LIBRARY_PATH to test suite script - ubuntu's make version requirement;

## Rationale
We don't need to force the users to have rpath installed to have the
build working - and all the tooling.

This patch makes the use of rpath optional.

### DIFF
--- a/Kconfig
+++ b/Kconfig
@@ -85,6 +85,16 @@ config PREFIX
             further resources, such as loadable module support (see
             MODULES configuration help).
 
+config RPATH
+	bool "Uses rpath for local builds"
+	depends on HAVE_CHRPATH
+	help
+            This option defines if rpath ought be used to look up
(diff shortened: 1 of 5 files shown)


dorileo:
I'm just wondering, I think it would be good to have rpath enabled by default of chrpath is available.

glima:
It's more beautiful than setting LD_LIBRARY_PATH by hand, indeed. Ack on path and comment.

barbieri:
+1

dorileo:
ok, applied. Thanks.

PR #619, opened 2015-08-24
fbp-generator: fix options handling

bdilly:
It was overwriting values set on OPTIONS_DEFAULT, for
omitted option fields with 0.

@cmarcelo please take a look on this fix


bdilly:
uhm.. good catch
change done to fix it too
merged

PR #620, opened 2015-08-25
Fix coverage report

cmarcelo:
`make coverage` was failing after the change to make JS metatype a
module. First, test-javascript now may depend on modules, so `check`
should depend on them. Second, duktape is now compiled from the module
directory, so fix the removal from the coverage results to include them
specifically.


cmarcelo:
@ricardotk @dorileo

ricardotk:
:+1:

cmarcelo:
Applied in 90af0e9438e681383318bb77821e001d3c74cedb.

PR #621, opened 2015-08-25
Remove old tests-fbp directory

cmarcelo:
Follow our current convention:

- "Tests" are focused on ensuring that the functionality works, and must
  run automatically (or have a description of steps to run), and a way
  to verify whether they work. E.g. src/test/, src/test-fbp/,
  src/test-stubgen.

- "Samples" are focused on showing how Soletta can be used, they contain
  comments / explanations to help understanding how things
  work. E.g. src/samples/.

Of course we want the samples to be always working (there will be a
procedure for the QA to test them), and the tests to not be
unreadable. But there is a difference in what they accomplish. A sample
doesn't need to address ALL the options available (in the same sample)
of a feature, while this kind of coverage is welcome in the tests.

The two removed files didn't fit either category (no comments, no
instructions, neither automatic ran). For those one-shot development
tests of a node type that doesn't deserve an automatic test, we now have
a way to use sol-fbp-runner to load a node type with options

      sol-fbp-runner -t timer -o interval=3000


cmarcelo:
@bdilly @anselmolsm

bdilly:
@cmarcelo @glima  it isn't about deserving tests... every node type deserves them. =)

But for some node types it's hard to implement them. wallclock/* is a good example. And it's the only reason it was on tests-fbp instead of test-fbp.

Also it isn't on samples because it doesn't exemplify a valid use case for Soletta. It isn't complex enough. It would be easy to document it / add instructions. But I don't think it would be enough to justify having a sample only for that.

bdilly:
Anyway, despite this discussion I agree with merging it.

cmarcelo:
Applied in ee877f7d89f55b44725847ba4dc0d3c4374dd9e0.

cmarcelo:
@bdilly one suggestion for those cases is to add a description of what is expected when running them and put them in a folder named `src/manual-test-fbp/` or similar. So that later we can actually verify whether the test is broken or not.

One nice thing is that we could still run `sol-fbp-runner -c` (the option to check only the syntax) on those manual tests.

PR #622, opened 2015-08-25
build: move test outputs to build dir

dorileo:
In the sense of removing all the built code from src/ dir we left the
tests binaries, this patch moves tests output to build/stage/ as well.


cmarcelo:
+1

dorileo:
applied, thanks.

PR #623, opened 2015-08-25
Regexp replace [v2]

glima:
Changes since v1:

- forward_on_no_match added to string/replace-regexp
- a new commit now adding the forward_on_no_match logic to the string/replace node also


glima:
The CI tool's cry is a style issue -- it will be fixed prior to pushing.

glima:
Sending a v3, again increasing patch count.

PR #624, opened 2015-08-25
build: compiler options

dorileo:
This patch re-organizes the compiler option into their own menu entry,
adds a release/debug build type and also includes the CONFIG_CFLAGS and
CONFIG_LDFLAGS configuration alternatives.

### DIFF
--- a/Kconfig
+++ b/Kconfig
@@ -30,26 +30,6 @@ config ENABLE_DYNAMIC_MODULES
 	depends on SHARED_LIBRARY
 	default n
 
-config CC_SANITIZE
-	bool "Compiler sanitize"
-	depends on FEATURE_CC_SANITIZE && (HAVE_SANITIZE_UNDEFINED || HAVE_SANITIZE_ADDRESS)
-	default y
-
-choice CC_SANITIZE_TYPE
-	prompt "Compiler sanitize type"
-	depends on CC_SANITIZE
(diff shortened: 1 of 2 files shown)


dorileo:
This patch is a respin for **build: compiler options"** of  #511

The only thing I haven't changed/done is about the suggestion made by @glima and @cmarcelo to: **"given a provided ```CFLAGS``` variable propagate it to ```CONFIG_CFLAGS```"**, if the user wants to have this flags persisted across make call it should be set to ```CONFIG_CFLAGS``` using the config facilities - namely kconfig - and get it persisted, or maybe export it via shell (which we support since early days).

Technically speaking, to implement it we would end up adding a third cflags variable (i.e EXTRA_CFLAGS) to make it possible, which I think more confusing than beneficial. So for the user point of view I think it's easier to say i.e use menuconfig to set ```CONFIG_CFLAGS``` and you have your preferred compilation flags persisted across make calls or export ```CFLAGS``` before anything.

cmarcelo:
+1

@dorileo thanks for looking into this. I agree with your explanation about the requested case, it makes sense not to have it. Whoever needs to change it programatically (my case included) can use the `.config` directly.

glima:
Fair for me too.

dorileo:
ok, applied. Thanks..

PR #627, opened 2015-08-25
Make sol-coap work on RIOT

ibriano:
This is more an RFC than an actual pull request because I'm far from
being happy with the sol-socket API.


ibriano:
New version in #641

PR #628, opened 2015-08-25
Fix things related to location packet type and stub generator

bdilly:


bdilly:
@glima it doesn't seem to have an use case yet, but at least it was fully supported. So at least it's working.

"location" was just allowed to be written (in schema + gen perspective), but nobody using such option type would compile...

edersondisouza:
@glima Actually iio/gyroscope uses direction-vector options, or did I misunderstood the remark?

glima:
Ok, @edersondisouza and @bdilly  :)

PR #630, opened 2015-08-26
Add -I option to sol-fbp-runner

lgywata:
Now runner is able to search for fbp files according to paths
passed to it through -I option.


lgywata:
thanks. Applied.

PR #632, opened 2015-08-26
Build modules name lookup and linux micro prefix - v2

dorileo:
## Changes
  + v2 (since #631):
    - Fixed issue with samples dep handling;

## Rationale
This pull request changes how modules are identified - so we can have multiple modules with same name in different categories. Also as a side effect changes how inter-modules dependencies are handled. With these 2 first changes in hand we can finally remove linux-micro prefix from linux-micro modules names.


ibriano:
I have to admit to being a bit lost in those make functions, but overall, it looks good.

barbieri:
```
If no prefix is provided the build system will get the first one it
finds - there's no guarantee the build system will get the right one if
```

I don't like it. We should make all modules explicit with their dirs, including flow-meta, pinmux...

dorileo:
ok, applied. Thanks.

PR #633, opened 2015-08-26
build: Use encoding=UTF-8

anselmolsm:
sol-flow-node-type-gen fails on systems not using UTF-8 for locale.
The generation step fails while processing freegeoip.json,
which contains a non-ASCII character. This patch changes this generator
to open files using enconding=UTF-8, so it is not affected by
the system locale.


barbieri:
+1

anselmolsm:
Applied, tks.

Issue #607, opened 2015-08-21
random: integer errors reported by -fsanitize=undefined


barbieri:
```
$ ./build/stage/samples/flow/misc/random-numbers 
...
src/modules/flow/random/random.c:90:42: runtime error: signed integer overflow: 666 * 1812433253 cannot be represented in type 'int [624]'
src/modules/flow/random/random.c:73:13: runtime error: left shift of 513147066 by 7 places cannot be represented in type 'int'
src/modules/flow/random/random.c:74:13: runtime error: left shift of 395728058 by 15 places cannot be represented in type 'int'
...
```

This is running on a 32bits machine, but shouldn't matter as the type is just "int" (not long/off_t/size_t).

assigned to @lpereira as he introduced this code as an alternative to missing `rand()` in libC.

lpereira:
Taking a look

lpereira:
Fixed in HEAD.

anselmolsm:
Fixed by #637

PR #634, opened 2015-08-26
flow/random: Store state as `unsigned int` to avoid UB in 32-bit

lpereira:
This has been reported on Issue #607, where undefined behavior
sanitizer would produce runtime errors when generating random numbers
on a 32-bit machine.

Move state to `unsigned int` so that the overflow behavior is defined,
and, where signed data is expected, shift value right by 1 bit before
casting to an integer in order to kill the bit that would be the sign
bit.

### DIFF
--- a/src/modules/flow/random/random.c
+++ b/src/modules/flow/random/random.c
@@ -48,46 +48,59 @@
  * article about Mersenne Twister (MT19937). */
 
 struct random_node_data {
-    int state[624];
+    unsigned int state[624];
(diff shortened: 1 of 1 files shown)


barbieri:
looks ok, but I still think we should use platform's function whenever they exist.

lpereira:
Alright, I'll make a second version tomorrow.

lpereira:
Closing to submit V2.

PR #635, opened 2015-08-27
build: properly unquote CONFIG_CFLAGS and CONFIG_LDFLAGS

cmarcelo:
When using CONFIG_CFLAGS containing spaces, it would be appended to the
command line keeping the quotes. The issue is that `patsusbt` operates
on words, but not assuming quotes mean anything.

    $(patsubst "%",%,$(CONFIG_CFLAGS))

was evaluating

    "-Wno-undef
    -Wno-missing-include-dirs"

and neither matched.

Using two patsubst, one for "% and another for %" works in some cases
but would break in the case we had quotes inside one of the flags.

This patch uses `echo`, and ensures that it is called only once for each
flag by creating the UNQUOTED_ variables using `:=` (single assignment).


cmarcelo:
@dorileo

cmarcelo:
The last sentence in the commit message tried to explain why I did it. To elaborate:

The issue is that `COMMON_CFLAGS` is recursively expanded, so doing without the extra `UNQUOTED_` variable we would be calling out to the shell everytime this variable is used. I've even tried to force the `COMMON_CFLAGS` to not be recursively expanded, but ended up with failures that I didn't want to debug in detail.

https://www.gnu.org/software/make/manual/html_node/Flavors.html

Another small benefit of having the variables is that I don't need to write a comment explaining "unquoting the config cflags" ;-)

dorileo:
Ok, the "to not call $(shell) every time the var is expanded" makes sense to me.
LGTM.

cmarcelo:
Applied in d97adb17b2ec3c255096886d185e0089d8d9b78a.

PR #636, opened 2015-08-27
Persistence

edersondisouza:


src/modules/flow/persistence/Kconfig
```diff
@@ -0,0 +1,3 @@
+config FLOW_NODE_TYPE_PERSISTENCE
+	tristate "Node type: persistence"
+	default m
```
barbieri: depends on filesystem or efivars being available, no? Or at least linux at this time.

src/modules/flow/persistence/efivarfs-storage.c
```diff
@@ -0,0 +1,167 @@
...
+
+struct efi_var {
+    uint32_t attributes;
+    uint8_t data[];
+};
+
+int
+efivars_write(const char *name, void *data, size_t size)
+{
+#ifdef FEATURE_FILESYSTEM
```
barbieri: I don't think you can have efivars without filesystem. As above, just move this to Kconfig.

when we have other storage types for systems such as contiki/riot, then we add more conditions to Kconfig and conditionally compile objects in Makefile. Just the core persistence.c will need to ifdef.

src/modules/flow/persistence/efivarfs-storage.c
```diff
@@ -0,0 +1,167 @@
...
+#define SOLETTA_EFIVARS_GUID "076027a8-c791-41d7-940f-3d465869f821"
+#define EFIVARFS_VAR_PATH "/sys/firmware/efi/efivars/%s-" SOLETTA_EFIVARS_GUID
+
+struct efi_var {
+    uint32_t attributes;
+    uint8_t data[];
+};
+
+int
+efivars_write(const char *name, void *data, size_t size)
```
barbieri: `sol_buffer` instead of data/size pair?

src/modules/flow/persistence/efivarfs-storage.c
```diff
@@ -0,0 +1,167 @@
...
+    if (r < 0 || r >= PATH_MAX) {
+        SOL_WRN("Could not create path for efivars persistence file [%s]", path);
+        return -EINVAL;
+    }
+
+    efi_var = calloc(1, sizeof(uint32_t) + size);
+    SOL_NULL_CHECK(efi_var, -ENOMEM);
+
+    efi_var->attributes = EFIVARS_DEFAULT_ATTR;
+    memcpy(efi_var->data, data, size);
```
barbieri: do we need to write as a single call? Otherwise we could skip allocating `efi_var` only to do that.

Also, it is confusing to see `sizeof(uint32_t)` in there, it's better to write `sizeof(struct efi_var)`, given your declaration it should be `sizeof(uint32_t)`.

src/modules/flow/persistence/efivarfs-storage.c
```diff
@@ -0,0 +1,167 @@
...
+
+    file = fopen(path, "w+e");
+    if (!file) {
+        SOL_WRN("Could not open persistence file [%s]", path);
+        free(efi_var);
+        return -errno;
+    }
+
+    /* TODO should short writes be considered? If so, something on sol-util-file? */
+    r = fwrite(efi_var, sizeof(uint32_t) + size, 1, file);
```
barbieri: `fwrite()` already handles `EINTR` and friends, so if it cannot write the size, it is an error.

src/modules/flow/persistence/efivarfs-storage.c
```diff
@@ -0,0 +1,167 @@
...
+    return 0;
+
+#else
+    SOL_WRN("No file system available");
+    return -ENOSYS;
+#endif
+}
+
+int
+efivars_read(const char *name, void *data, size_t size)
```
barbieri: change the API to receive a `sol_buffer`?

Then again, if partial reads are allowed, you read one buffer for the `sizeof(uint32_t)` that is the attributes field, then you read directly to buffer argument the actual variable contents.

src/modules/flow/persistence/persistence.c
```diff
@@ -0,0 +1,472 @@
...
+
+static int
+storage_write(struct persist_data *mdata, void *data, size_t size)
+{
+    if (streq(mdata->storage, "fs"))
+        return fs_write(mdata->name, mdata->fs_dir_path, data, size);
+
+    if (streq(mdata->storage, "efivars"))
+        return efivars_write(mdata->name, data, size);
+
```
barbieri: nooooo...  You did a nice work by keeping API similar, then go one step further and make API exactly the same and here you call the pointer giving the same arguments. no `streq()`

src/modules/flow/persistence/persistence.json
```diff
@@ -0,0 +1,259 @@
...
+          {
+            "data_type": "string",
+            "description": "Storage where data will be persisted. It can be of the following: fs, efivars",
+            "name": "storage"
+          },
+          {
+            "data_type": "string",
+            "default": null,
+            "description": "If 'storage' is 'fs', directory path where data will be persisted. If 'storage' different of 'fs', this option will be ignored.",
+            "name": "fs_dir_path"
```
barbieri: you can remove this one to keep the API the same. What you do is take the path in the `name` option above. If `efivars` and `name` is not compatible, then you will try to open the file and it will fail anyway.

src/modules/flow/persistence/efivarfs-storage.c
edersondisouza: Would an `fseek(sizeof(uint32_t))` suffice? Not sure if it works, though,

barbieri: no idea, would need to test.

edersondisouza:
v2 on #646

PR #637, opened 2015-08-27
[v2] flow/random fixes

lpereira:
This is the V2 of #634; these changes were made:

* State is not cast to an `int` when reinitializing the PRNG.
* Random seed is initialized using platform-specific entropy pools (getrandom(), /dev/urandom, etc).
* GLIBC extension random_r(), if available, is used.

### DIFF
--- a/data/jsons/dependencies.json
+++ b/data/jsons/dependencies.json
@@ -274,6 +274,14 @@
       "fragment": "strndupa(0, 0);"
     },
     {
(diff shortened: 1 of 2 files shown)


lpereira:
(Repushed with minor stylistic fix to make Semaphore happy.)

lpereira:
Received ack by @barbieri on IRC, and merged.

PR #638, opened 2015-08-27
build: make sure we reset obj-*-[y,m] variables

dorileo:
After parsing a module's makefile we should make sure we reset
obj-*-[y,m] variables so we don't end up appending named clashing
objects variables.

An exposed problem was console.mod, we have flow/console.mod and
linux-micro/console.mod, after parsing flow/console.mod we left
obj-console-m set with console.o and then when parsing
linux-micro/console.mod (because it does obj-console-m += console.o) we
ended up with obj-console-m set as "console.o console.o", with that
Makefile.rules created 2 rules for linux-micro/console.o.


dorileo:
@barbieri could you test it?

barbieri:
+1 commit it

dorileo:
applied.

PR #639, opened 2015-08-27
Fix depending on packet buf [v2]

vcgomes:
Updated version of https://github.com/solettaproject/soletta/pull/248

The objective is to remove `sol_coap_packet_get_buf()` because it exposes the raw buffer of the packet.

On one discussion we had offline, the idea was to provide the function handles access to both the incoming packet (request) payload and the outgoing packet payload, so use cases where the response depends on the content of the packet payload can be addressed.

@lpereira Can you take a look and help test this?

### DIFF
--- a/src/lib/comms/sol-oic-server.c
+++ b/src/lib/comms/sol-oic-server.c
@@ -711,8 +711,11 @@ _sol_oic_resource_type_handle(
 {
     struct sol_coap_packet *response;
     sol_coap_responsecode_t code = SOL_COAP_RSPCODE_INTERNAL_ERROR;
-    uint8_t *payload;
-    uint16_t payload_len;
+    uint8_t *payload, *response_payload;
+    uint16_t payload_len, response_len;
+    const char response_prefix[] = "{\"oc\":[{\"rep\":";
+    const char response_suffix[] = "}]}";
+    size_t len;
 
     OIC_SERVER_CHECK(-ENOTCONN);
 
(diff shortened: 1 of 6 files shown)


src/lib/comms/sol-oic-server.c
```diff
@@ -722,9 +726,21 @@ _sol_oic_resource_type_handle(
         return -1;
     }
 
-    if (!handle_fn) {
-        code = SOL_COAP_RSPCODE_NOT_IMPLEMENTED;
+    len = strlen(response_prefix);
+
+    if (sol_coap_packet_get_payload(response, &response_payload, &response_len) < 0)
```
barbieri: change this function to return a `sol_buffer` with memory set to not-owned.

src/lib/comms/sol-oic-server.c
```diff
@@ -722,9 +726,21 @@ _sol_oic_resource_type_handle(
...
         goto done;
+
+    if (response_len <= (len + strlen(response_suffix)))
+        goto done;
+
+    memcpy(response_payload, response_prefix, len);
+    response_payload += len;
+    response_len -= len;
```
barbieri: you shouldn't need this. Use `sol_buffer` functions, they will append strings for you. If you set the buffer to not-owned/fixed capacity, then no resize is done and you can just check for errors if they happen.

src/lib/comms/sol-oic-server.c
```diff
@@ -722,9 +726,21 @@ _sol_oic_resource_type_handle(
...
+
+    if (!handle_fn) {
+	    code = SOL_COAP_RSPCODE_NOT_IMPLEMENTED;
+	    goto done;
```
barbieri: wrong indentation?

barbieri:
overall I think it should use more of `sol_buffer`, simplifying the internals and make it "safer". One option, for instance, is to use a sub-buffer for the response contents. We create a response buffer so we can use it to append the prefix/suffix easily. Then we create one starting right after the prefix, with capacity being response - prefix_len - suffix_len, give it to users to fill. Once you're done, you increment the bigger response with the amount used by the content and append the suffix.

vcgomes:
The source of this problem was solved by PR #835

PR #640, opened 2015-08-27
Regexp replace [v3]

glima:
Changes since v2:

- Fixed a missing overflow check.
- Fixed bad return values on a function
- Added moar tests


glima:
@edersondisouza can you re-check, please?

PR #641, opened 2015-08-27
Port sol-socket to RIOT so we can use sol-coap there

ibriano:
It's less RFCish, but there's still plenty that can be improved. It works well
enough for the time being though.


barbieri:
+1 overall, just try to make the socket class `const`

ibriano:
Merged with const fixes

PR #643, opened 2015-08-27
build: make sure each objects rule depend on its .dep rule

dorileo:
Every built object has its own .dep rule to generate the makefile
dependency rules. This patch makes sure to set it has explicit
dependency to the compilation rule.


dorileo:
I'll send another PR with additional stuffs.

PR #644, opened 2015-08-27
build: fix requires-private

dorileo:
We're not supporting the common "local context" idiom for
requires-private-y where each sub makefile may want to reset the
requires-private-y variable like it's done to all other variables i.e:

requires-private-y := $(DEP_REQUIRES_PRIVATE)

With that whenever we get a similar case the requires-private-y variable
will be reset.

This patch changes the code append a global requires-private variable
given a local requires-private-y.

The patch also fixes a makefile setting requires-private-y with LDFLAGS
instead of REQUIRES_PRIVATE.


dorileo:
applied, thanks.

PR #645, opened 2015-08-27
modules/gtk: set min width for labels

bdilly:
After scrolled window was added, labels width aren't change
window width anymore.
It leads to sad scenarios where a dev may think a node type
isn't working fine...

I tried to resize the window respecting changes on label
size, but I wasn't able to do it, so
I'm setting a minimum enforced width that I believe is good
enough to display most of our outputs.


cmarcelo:
+2

bdilly:
merged

PR #646, opened 2015-08-27
Persistence v2

edersondisouza:
v2:
Using more sol_buffer;


edersondisouza:
Checking what is wrong with `make allyesconfig`.

edersondisouza:
Found out bug with `make allyesconfig` (json `"name"` attribute was wrong) and fixed locally.

edersondisouza:
v3 on #648

PR #647, opened 2015-08-27
runner: Fix -I option

ceolin:
Do not include it's own path in the fbp path list, instead of this keep
a copy of this search path in its on struct. It avoids the memory leak
that was happening and not duplicate search paths in the list.

### DIFF
--- a/src/bin/sol-fbp-runner/runner.c
+++ b/src/bin/sol-fbp-runner/runner.c
@@ -57,6 +57,7 @@ struct runner {
 
     const char *filename;
     char *basename;
+    char *dir;
 
     struct sol_ptr_vector file_readers;
     struct sol_ptr_vector *fbp_paths;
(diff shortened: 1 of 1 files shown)


ceolin:
@lgywata could take a look if this patch is correct please ?

src/bin/sol-fbp-runner/runner.c
```diff
@@ -73,18 +74,22 @@ search_fbp_file(struct runner *r, const char *basename)
...
+    do { \
+        err = asprintf(&fullpath, "%s/%s", _path, basename); \
+        if (err < 0) { \
+            errno = ENOMEM;\
+            return NULL; \
+        } \
+        if (stat(fullpath, &s) == 0 && S_ISREG(s.st_mode)) \
+            return fullpath; \
+        free(fullpath); \
+    } while (0)
```
cmarcelo: -1 use a function instead of a macro.

ceolin: we are using macros like this in other places, is there a good advantage making this a funcion ?

I do not see any problem in change it, i'm just afraid i've missed something

ceolin:
closed, changes on https://github.com/solettaproject/soletta/pull/651

src/bin/sol-fbp-runner/runner.c
cmarcelo: Macros help in some cases, e.g. when you have a lot of shared context that would mean passing many arguments to a function. However they complicate in others, e.g. making it difficult to debug, giving less room for compiler to make decisions about inlining or not.

I prefer we stay with functions as a default form abstraction and use the macros when there's a need for them (since we pay a cost).

Sorry if my first message sounded harsh, the fix itself is good!

PR #648, opened 2015-08-28
Persistence v3

edersondisouza:
v3:
Fixed `make allyesconfig` failure;
Added option 'erron_on_no_value' - when true (default false) it'll send an error packet informing that no previous value was found upon node opening.

v2:
Using more sol_buffer;


edersondisouza:
v3 on #648

edersondisouza:
Ops, wrong tab.

edersondisouza:
Oh, it looks I screwed semaphore... sending a v4.

PR #650, opened 2015-08-28
Uuid

glima:
Add new UUID nodes and linux-micro service.

    The string/uuid node outputs fresh, new v4 UUIDs on a flow.

    The service will run after fstab one, and will guarantee that a
    machine-id exists (in known places) in the file system.

    The string/machine-id node gets said machine-id and outputs it on a
    flow.

    Tests for the nodes are included.


glima:
I left sol_util_get_machine_id() at sol-util-file.c. @barbieri, moving it to sol-platform-linux.c, is better, right?

glima:
The bot is failing because that old Ubuntu did not have systemd yet and neither /etc/machine-id nor /run/machine-id are available. The shell there forbids that one writes on /etc/ by hand on the build steps, so I'm looking for other solutions for a test involving the string/machine-id to pass there.

PR #651, opened 2015-08-28
Fix v2

ceolin:
- Using a functions instead of macro
- Added two commits (fixes)
 - Fix a segfault in the generator
 - Sanitizing the init() functions


ceolin:
merged with the fix. Thanks

cmarcelo:
Thanks!

PR #652, opened 2015-08-28
Build clearlinux fixes

dorileo:
This series brings some fixes that appeared while packaging soletta to clearlinux.


data/scripts/dependency-resolver.py
```diff
@@ -487,6 +487,14 @@ def handle_definitions(args, conf, context):
     context.debug("## ------------------- ##\n\n")
 
     variables = {"PREFIX": args.prefix}
+
+    # environment variables get priority, so set it beforehand
+    for k,v in os.environ.items():
```
lpereira: `definitions.update(os.environ)`.

dorileo: this is not the case, we'll only add the environ var if it's present in ```definitions```.

dorileo:
applied...

PR #654, opened 2015-08-31
oicgen: Use argparse instead of positional arguments

lpereira:
This is more consistent with the rest of the scripts.

### DIFF
--- a/data/oic/oicgen.py
+++ b/data/oic/oicgen.py
@@ -1585,16 +1585,29 @@ json_token_to_bool(struct sol_json_token *token, bool *out)
     return code.replace('\n\n\n', '\n')
 
 if __name__ == '__main__':
+    import argparse
+    parser = argparse.ArgumentParser()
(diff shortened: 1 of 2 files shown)


barbieri:
+1 overall, two comment unrelated to patch, but eventually oicgen should check those:

 * try to check JSON format using `jsonschema`. We're using a code that uses it and provides meaningful messages for files that don't respect the expected schema. The bad part is that we are copying the code around as we don't install any module in our python site-package. You can copy those and keep the comment, if we change to install shared files in site-package, then we will remember to change oicgen.py as well.
 * ensure files are read and write as `utf-8`. @anselmolsm found some problems running our python scripts in environments where locale wasn't set or was not using `utf-8`.

lpereira:
I'll merge as is, then -- and make the suggested changes alongside other changes that I'm making in the script.

lpereira:
Merged.

PR #655, opened 2015-08-31
flow: fs node transform into persistence node - with efivars - v5

edersondisouza:
v5:

v4:

v3:

v2:
Using more sol_buffer;


src/modules/flow/persistence/fs-storage.c
```diff
@@ -0,0 +1,100 @@
...
+{
+    FILE *file;
+
+    file = fopen(name, "w+e");
+    if (!file) {
+        SOL_WRN("Could not open persistence file [%s]", name);
+        return -errno;
+    }
+
+    fwrite(buffer->data, buffer->used, 1, file);
```
lpereira: Not checking the return valud of `fwrite()`: is this intentional?

src/modules/flow/persistence/persistence.c
```diff
@@ -0,0 +1,557 @@
...
+    char *storage;
+
+    struct sol_flow_packet *(*packet_new_fn)(const struct persist_data *data);
+    int (*packet_data_get_fn)(const struct sol_flow_packet *packet, void *value_ptr);
+    int (*packet_send_fn)(struct sol_flow_node *node);
+    void *(*node_get_default_fn)(struct sol_flow_node *node);
+
+    int (*storage_write)(const char *name, struct sol_buffer *buffer);
+    int (*storage_read)(const char *name, struct sol_buffer *buffer);
+    int (*storage_get_size)(const char *name, size_t *size);
```
lpereira: I'm not sure if I like this interface: it's prone to TOCTOU. Would it be possible to have only primitives for reading and writing, with the read primitive returning the size or an error?

src/modules/flow/persistence/persistence.c
```diff
@@ -0,0 +1,557 @@
...
+    struct persist_data *mdata = data;
+    size_t size;
+    int r;
+
+    if (!storage || *storage == '\0') {
+        SOL_WRN("Must define a storage type");
+        return -EINVAL;
+    }
+
+    if (streq(storage, "fs")) {
```
lpereira: sol_str_table?

src/modules/flow/persistence/fs-storage.c
edersondisouza: Maybe I'm kind of `fwrite()` won't fail, if something goes wrong it'll arise on `fclose()`, but I can check return here too - maybe string values can be big enough to fill file buffer.

src/modules/flow/persistence/persistence.c
edersondisouza: Hmmm... Fair. However, in this case reading will need to take care of allocating memory - but I think that `sol_buffer` can be of help.

src/modules/flow/persistence/persistence.c
edersondisouza: Didn't know about it, will check.

edersondisouza:
v6 on #666

PR #656, opened 2015-08-31
Uuid [v2]

glima:
Changes since #650 
 - Pass better flags to getrandom()
 - Use sol_buffer_append_printf() to generate uuid string
 - Accept an env var as machine id too (I'll work with Anselmo to test that on the bot)
 - sol_util_get_machine_id() moved to plaform headers, where other ID getters will land soon.


src/shared/sol-util-linux.c
```diff
@@ -45,3 +59,107 @@ sol_util_timespec_get_realtime(struct timespec *t)
...
+    int fd = -1;
+    int r;
+
+#ifdef HAVE_GETRANDOM
+    r = syscall(SYS_getrandom, p, n, 0);
+    if (r == n)
+        return 0;
+
+    if (r < 0) {
+        if (errno != EAGAIN) { /* fallback to /dev/urandom on EAGAIN
```
lpereira: getrandom() might return ENOSYS if the kernel doesn't support that. (The headers might, but the running kernel might not.)

src/shared/sol-util-linux.c
```diff
@@ -45,3 +59,107 @@ sol_util_timespec_get_realtime(struct timespec *t)
...
+
+    *ret = assert_uuid_v4(t);
+    return 0;
+}
+
+// 37 = 2 * 16 (chars) + 4 (hyphens) + 1 (\0)
+int
+sol_util_uuid_gen(bool upcase,
+    bool with_hyphens,
+    char id[37])
```
lpereira: Might be a good idea to pass `char id[static 37]` here.

src/shared/sol-util-linux.c
```diff
@@ -45,3 +59,107 @@ sol_util_timespec_get_realtime(struct timespec *t)
...
+{
+    struct sol_str_slice hyphen = SOL_STR_SLICE_LITERAL("-");
+    /* hyphens on positions 8, 13, 18, 23 (from 0) */
+    const int hyphens_pos[] = { 8, 13, 18, 23 };
```
lpereira: static const

src/shared/sol-util-linux.c
```diff
@@ -45,3 +59,107 @@ sol_util_timespec_get_realtime(struct timespec *t)
...
+    return 0;
+}
+
+// 37 = 2 * 16 (chars) + 4 (hyphens) + 1 (\0)
+int
+sol_util_uuid_gen(bool upcase,
+    bool with_hyphens,
+    char id[37])
+{
+    struct sol_str_slice hyphen = SOL_STR_SLICE_LITERAL("-");
```
lpereira: static const

src/shared/sol-util-linux.c
```diff
@@ -45,3 +59,107 @@ sol_util_timespec_get_realtime(struct timespec *t)
...
+    /* hyphens on positions 8, 13, 18, 23 (from 0) */
+    const int hyphens_pos[] = { 8, 13, 18, 23 };
+    struct sol_uuid uuid;
+    unsigned i;
+    int r;
+
+    struct sol_buffer buf = SOL_BUFFER_INIT_EMPTY;
```
lpereira: Can't this buffer be initialized passing `id` as the data pointer, to avoid the memcpy() below?

src/shared/sol-util-linux.c
```diff
@@ -45,3 +59,107 @@ sol_util_timespec_get_realtime(struct timespec *t)
...
+    if (r < 0) {
+        if (errno != EAGAIN) { /* fallback to /dev/urandom on EAGAIN
+                                * -- good enough for our uses */
+            return -errno;
+        }
+    } else
+        return -ENODATA;
+#endif
+
+    fd = open("/dev/urandom", O_RDONLY | O_CLOEXEC | O_NOCTTY);
```
lpereira: Any reason to pass O_NOCTTY?

src/shared/sol-util-linux.c
barbieri: it can, it should :-)

src/modules/flow/random/random.c
```diff
@@ -54,19 +55,17 @@ getrandom_shim(void *buf, size_t buflen, unsigned int flags)
...
 
-#ifdef SYS_getrandom
+#ifdef HAVE_GETRANDOM
     /* No wrappers are commonly available for this system call yet, so
      * use syscall(2) directly. */
     long gr_ret = syscall(SYS_getrandom, buf, buflen, flags);
-    if (gr_ret >= 0)
-        return gr_ret;
+    if (gr_ret > 0 && (size_t)(unsigned long)gr_ret == buflen)
+        return 0;
```
lpereira: This is wrong: it should return `gr_ret` (the number of bytes read). Or else the check in the call site will fail

src/shared/sol-util-linux.c
glima: Ok, so checking for ENOSYS instead (to continue) is fair enough? I'll remove the comment.

src/shared/sol-util-linux.c
glima: Ack.

src/modules/flow/random/random.c
glima: Ouch, ok.

PR #658, opened 2015-08-31
AIO setup was never called because of incorrect ifdef check

otaviobp:
ifdef was checking for an incorrect definition, so aio setup code was
never called. Using USE_PIN_MUX instead of HAVE_PIN_MUX


dorileo:
+1

PR #659, opened 2015-08-31
run-fbp-tests: UTF8 and verbose option

anselmolsm:
- Use enconding=UTF-8 to open files
 - Add a --verbose option, to help while debugging the script itself.


anselmolsm:
Applied, tks.

PR #660, opened 2015-08-31
Build templates everywhere

dorileo:
This series is another step in the direction of Makefile.rules simplification. The final result here is the *builtins*.h generation rules removal, with that we stick to the templates schema and let it happen in a uniform way.


data/scripts/template.py
```diff
@@ -119,18 +119,48 @@ def try_subst(verbatim):
 
     return verbatim.replace("@","")
 
-def run_template(f, context):
-    raw = f.read()
+def try_include(verbatim, globals, context):
+    p = re.compile("^.*include.*\"(?P<include_file>.*)\".*$")
```
lpereira: Is it a good idea to start the match with `.*`?

data/scripts/template.py
```diff
@@ -119,18 +119,48 @@ def try_subst(verbatim):
...
+    if not include_file:
+        print("Could not determine include_file to include: %s" % verbatim)
+        return None, True
+
+    dir = os.path.dirname(os.path.realpath(args.template.name));
+    path = "%s/%s" % (dir, include_file)
+
+    try:
+        f = open(path)
+    except FileNotFoundError:
```
lpereira: Maybe only capturing `FileNotFoundError` is not enough: one might not have permission to open a file, for instance.

data/scripts/template.py
```diff
@@ -119,18 +119,48 @@ def try_subst(verbatim):
...
+
+    if not m:
+        return None, False
+
+    include_file = m.group("include_file")
+    if not include_file:
+        print("Could not determine include_file to include: %s" % verbatim)
+        return None, True
+
+    dir = os.path.dirname(os.path.realpath(args.template.name));
```
lpereira: Please do not use `dir` as a variable name, as it's the name of a builtin function. Also, remove the semicolon.

data/scripts/template.py
```diff
@@ -119,18 +119,48 @@ def try_subst(verbatim):
...
+    path = "%s/%s" % (dir, include_file)
```
lpereira: os.path.join()?

lpereira:
While I like the idea, I'm not really sold on the implementation. Since things inside {{ .. }} blocks are Python code, and `include "foo"` isn't really Python, things can get quite confusing. Maybe a function to include a template could be exported instead?

dorileo:
Yep, the idea of exporting a include() function seems good. Thanks for reviewing.

dorileo:
v2 comming....

PR #662, opened 2015-08-31
Add timestamp packet and node types, and make changes on wallclock [v2]

bdilly:
I'm not confortable about
4a53df2d modules/converter: support format option on timestamp-to-string

I prefer to drop it, but since it was requested on pr #642
and some guys think it's safe, I'm pushing it.

I would like to hear more opnions about it.

Changelog:
  * use timespec instead of time_t to save info about nanoseconds
  * use node_type for comparison node types to set func per class,
    not node instances
  * make the same on other modules (int, float, trigonometry)
  * add general improvements for int and float modules regarding
    re-using structs and functions that could be shared between
    node types.
  * create macros to mark last ports and last array ports and use it
    on timestamp node types


src/bin/sol-fbp-runner/inspector.c
```diff
@@ -201,6 +203,20 @@ inspector_show_packet(const struct sol_flow_packet *packet)
             fprintf(stdout, "<lat=%g|lon=%g|alt=%g>", v.lat, v.lon, v.alt);
             return;
         }
+    } else if (type == SOL_FLOW_PACKET_TYPE_TIMESTAMP) {
+        struct timespec v;
+        if (sol_flow_packet_get_timestamp(packet, &v) == 0) {
+            struct tm cur_time;
+            char buf[32];
+            tzset();
+            if (localtime_r(&v.tv_sec, &cur_time)) {
```
vcgomes: I need review this more closely, but it looks that you should be using `gmtime_r()` instead of `localtime_r()`, as you are printing the value in the UTC timezone ('Z').

Am I missing something?

bdilly: Yeah, it's true.
I think it's a good idea to keep printed values in UTC timezone, so I'll use gmtime

bdilly:
OK, thanks @ibriano . I've talked to @lpereira and he agrees.
So I think we're good with this commit.

ibriano:
Seems fine

bdilly:
Merged with gmtime change.
Thanks

PR #666, opened 2015-09-01
Persistence v6

edersondisouza:
v6:
No more file size checking before use;
Previous topic enforced `sol_util_file` use and improvements/fixing:
                  `sol_util_load_file_fd_string()` to read a file descriptor - useful for efivars reading, that need to discard unused parts of file;
                   Fixed off-by-one reading;
Using sol_str_table to lookup which functions to use;
Error checking on `fwrite()`.

v5:

v4:

v3:

v2:
Using more sol_buffer;


src/modules/flow/persistence/persistence.c
```diff
@@ -0,0 +1,558 @@
...
+
+static int
+persist_reset(struct persist_data *mdata, struct sol_flow_node *node)
+{
+    void *value;
+
+    if (mdata->packet_data_size)
+        value = alloca(mdata->packet_data_size);
+
+    value = mdata->node_get_default_fn(node);
```
lpereira: Value is allocated in the previous if but overwritten here. Why?

src/modules/flow/persistence/persistence.c
```diff
@@ -0,0 +1,558 @@
...
+    char *fs_dir_path;
+    char *storage;
+
+    struct sol_flow_packet *(*packet_new_fn)(const struct persist_data *data);
+    int (*packet_data_get_fn)(const struct sol_flow_packet *packet, void *value_ptr);
+    int (*packet_send_fn)(struct sol_flow_node *node);
+    void *(*node_get_default_fn)(struct sol_flow_node *node);
+
+    int (*storage_write)(const char *name, struct sol_buffer *buffer);
+    int (*storage_read)(const char *name, struct sol_buffer *buffer);
```
lpereira: Put these two function pointers in a structure, have a pointer to the structure here, and make the str_table point to fs/efivars struct.

src/modules/flow/persistence/persistence.c
```diff
@@ -0,0 +1,558 @@
...
+
+    return persist_open(node, data, opts->storage, opts->name);
+}
+
+static void
+persist_string_close(struct sol_flow_node *node, void *data)
+{
+    struct persist_string_data *mdata = data;
+
+    free(mdata->default_value);
```
lpereira: Is default_value properly initialized when not passed down from the opts struct?

edersondisouza: Is it possible that opts struct does not have default_value?

src/modules/flow/persistence/persistence.c
edersondisouza: Drugs...

src/modules/flow/persistence/persistence.c
edersondisouza: ok

src/modules/flow/persistence/persistence.c
edersondisouza: I've tested it and found no issue.

edersondisouza:
v7 on #668

PR #667, opened 2015-09-01
Uuid [v3]

glima:
Changes since v2 (#656):
 - getrandom's return now checked correctly
 - char id[static <size>] used where possible
 - machine id does not deal with hyphens and case anymore
 - machine id is now a platform node
 - various other assorted fixes


src/lib/common/sol-platform-impl-linux-micro.c
```diff
@@ -784,6 +784,99 @@ sol_platform_impl_set_target(const char *target)
...
+    if (!sol_util_uuid_str_valid(id))
+        return -EINVAL;
+
+    return 0;
+}
+
+int
+sol_platform_impl_get_machine_id(char id[static 33])
+{
+    const char *etc_path, *run_path;
```
lpereira: Why not `static const char *etc_path = "/foo/bar"`?

glima: Fair, I'll fix it.

glima:
BTW, I'll leave this PR exposed a little bit more till I dive in fixes -- I'm working on something else in the meanwhile.

glima:
All fixed, new version now.

PR #668, opened 2015-09-01
Persistence v7

edersondisouza:
v7:
Refactor sol_str_table usage;
Fixed unnecessary code;

v6:
No more file size checking before use;
Previous topic enforced sol_util_file use and improvements/fixing:
sol_util_load_file_fd_string() to read a file descriptor - useful for efivars reading, that need to discard unused parts of file;
Fixed off-by-one reading;
Using sol_str_table to lookup which functions to use;
Error checking on fwrite().

v5:

v4:

v3:

v2:
Using more sol_buffer;


src/modules/flow/persistence/efivarfs-storage.c
```diff
@@ -0,0 +1,155 @@
...
+
+static const int EFIVARS_DEFAULT_ATTR = 0x7;
+
+static bool
+check_realpath(const char *path)
+{
+    char real_path[PATH_MAX];
+
+    if (realpath(path, real_path)) {
+        return strstartswith(real_path, EFIVARFS_VAR_DIR);
```
lpereira: Please add a trailing slash in `EFIVARFS_VAR_DIR` so this will actually check a path prefix.

lpereira:
LGTM with a comment

edersondisouza:
Fixed issue pointed by @lpereira and integrated.

PR #669, opened 2015-09-01
Build templates everywhere v2

dorileo:
## Changes
  + v2 (since #660):
    - Fixed issues pointed by @lpereira;
    - Removed more "duplicated" rules using template script;

## Rationale
This series is another step in the direction of Makefile.rules simplification. The final result here is the *builtins*.h generation rules removal, with that we stick to the templates schema and let it happen in a uniform way.


ibriano:
Seems fine

lpereira:
LGTM

dorileo:
applied.

PR #670, opened 2015-09-01
modules/flow: fix buffers on int and boolean regarding operations

bdilly:
Option 'operation' is a string and shouldn't fallback silently
to one of the supported operations in case of mismatch.
So it was added a warning. Also, int/buffer operation default
was invalid.

Supported operations were added to node type description.


cmarcelo:
+2

bdilly:
Thanks for reviewing it, @cmarcelo 
Merged

PR #672, opened 2015-09-02
doc: add documentation for gpio

ricardotk:
The important information here is that if the user call
sol_gpio_open() more than once for the same pin, the
behaviour will be undefined per platform.


ricardotk:
merged following the comments, thanks!

PR #673, opened 2015-09-02
build: cleanup static modules

dorileo:
We don't use or need static modules any more, this was useful when we
first had the libshared.a to share code between bin, tests and so
on. Now it becomes obsolete and shouldn't be maintained.


ibriano:
Go ahead

dorileo:
applied, thanks.

PR #674, opened 2015-09-02
pwm: Don't try to change pwm polarity if polarity is already correct

otaviobp:
Trying to change polarity in a chip that doesn't support polarity
changes will abort soletta and that is not needed when the current
polarity is equal the wanted polarity.

Soletta will still abort if the wanted polarity is different from
current polarity.

### DIFF
--- a/src/lib/io/sol-pwm-linux.c
+++ b/src/lib/io/sol-pwm-linux.c
@@ -193,6 +193,7 @@ _pwm_config(struct sol_pwm *pwm, const struct sol_pwm_config *config)
 {
     const char *pol_str;
     char path[PATH_MAX];
+    char pol_value[10];
 
     sol_pwm_set_enabled(pwm, false);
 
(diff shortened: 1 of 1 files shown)


bdilly:
I'm thinking a bit more about polarity on PWM. It isn't supported on RIOT and on Linux it depends on hardware support. Maybe it's the case of doing it in software (sol-pwm) to guarantee it always work (hopefully it's doable) or even dropping it completely.
Anyway, I'm ok with this patch while we discuss about it.

ibriano:
Polarity on RIOT is supported differently, and only if the hardware supports it. In fact, everything I saw about polarity of a PWM is on the Linux API only, because everyone else refers to it as RIOT does. RIOT, if the hardware supports it, will let you open a PWM as left-phase or right-phase, which is, I think, what the inverted polarity would do, but on Linux we have no way to set a center-phase, which RIOT allows, again, if the hardware works that way. It gets more complicated when you consider that these phases don't just change when the duty cycle will be active, but it can change the how the period is counted too.

PR #675, opened 2015-09-02
Improve int/buffer and boolean/buffer

bdilly:
Fixes, tests and more features


bdilly:
New version #691

PR #678, opened 2015-09-02
shared: Introduce sol-rng, a Random Number Generator API

lpereira:
This is a simple abstraction to generate random numbers. It is
loosely-based on the C++11 API supported by <random> include, sans the
distribution stuff.

Current implementation contains three engine implementations: a MT19937
(default), a random_r() generator (if available; currently only glibc
supports it), and a /dev/urandom reader.

Only raw bytes can be generated by the API at the moment.  The idea is
to implement custom type generation (with proper distribution) as the
need arises.

Currently, the implementation is completely opaque, but can be opened
up if the need to let Soletta API users implement their own engines
arise.

Usage is pretty simple at the moment:

	struct sol_rng_engine *engine;
	uint64_t value;

	/* Passing the default engine or an engine not supported by
	 * the platform will fallback to the MT19937 engine. */
	engine = sol_rng_engine_new(SOL_RNG_ENGINE_IMPL_DEFAULT);

	/* Passing 0 to the seed will try to use a good source of
	 * entropy available in the system.  This is /dev/urandom on
	 * Linux (using getrandom() if available). */
	sol_rng_engine_seed(engine, 0);

	assert(sizeof(value) == sol_rng_engine_generate_bytes(engine, &value, sizeof(value)));

	sol_rng_engine_destroy(engine);

As a bonus, move random node types to use this new API.

### DIFF
--- a/src/shared/sol-rng.h
+++ b/src/shared/sol-rng.h
@@ -1,0 +1,52 @@
+/*
+ * This file is part of the Soletta Project
+ *
+ * Copyright (C) 2015 Intel Corporation. All rights reserved.
+ *
+ * Redistribution and use in source and binary forms, with or without
+ * modification, are permitted provided that the following conditions
+ * are met:
+ *
+ *   * Redistributions of source code must retain the above copyright
+ *     notice, this list of conditions and the following disclaimer.
+ *   * Redistributions in binary form must reproduce the above copyright
+ *     notice, this list of conditions and the following disclaimer in
+ *     the documentation and/or other materials provided with the
+ *     distribution.
+ *   * Neither the name of Intel Corporation nor the names of its
+ *     contributors may be used to endorse or promote products derived
+ *     from this software without specific prior written permission.
+ *
+ * THIS SOFTWARE IS PROVIDED BY THE COPYRIGHT HOLDERS AND CONTRIBUTORS
+ * "AS IS" AND ANY EXPRESS OR IMPLIED WARRANTIES, INCLUDING, BUT NOT
+ * LIMITED TO, THE IMPLIED WARRANTIES OF MERCHANTABILITY AND FITNESS FOR
+ * A PARTICULAR PURPOSE ARE DISCLAIMED. IN NO EVENT SHALL THE COPYRIGHT
+ * OWNER OR CONTRIBUTORS BE LIABLE FOR ANY DIRECT, INDIRECT, INCIDENTAL,
+ * SPECIAL, EXEMPLARY, OR CONSEQUENTIAL DAMAGES (INCLUDING, BUT NOT
+ * LIMITED TO, PROCUREMENT OF SUBSTITUTE GOODS OR SERVICES; LOSS OF USE,
+ * DATA, OR PROFITS; OR BUSINESS INTERRUPTION) HOWEVER CAUSED AND ON ANY
+ * THEORY OF LIABILITY, WHETHER IN CONTRACT, STRICT LIABILITY, OR TORT
+ * (INCLUDING NEGLIGENCE OR OTHERWISE) ARISING IN ANY WAY OUT OF THE USE
+ * OF THIS SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
+ */
+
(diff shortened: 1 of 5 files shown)


src/shared/sol-rng.h
```diff
@@ -0,0 +1,52 @@
...
+struct sol_rng_engine;
+struct sol_rng_engine_impl;
+
+extern const struct sol_rng_engine_impl *SOL_RNG_ENGINE_IMPL_MT19937;
+extern const struct sol_rng_engine_impl *SOL_RNG_ENGINE_IMPL_URANDOM;
+extern const struct sol_rng_engine_impl *SOL_RNG_ENGINE_IMPL_RANDOMR;
+extern const struct sol_rng_engine_impl *SOL_RNG_ENGINE_IMPL_DEFAULT;
+
+struct sol_rng_engine *sol_rng_engine_new(const struct sol_rng_engine_impl *impl);
+struct sol_rng_engine *sol_rng_engine_seeded_new(const struct sol_rng_engine_impl *impl);
```
barbieri: make a single call that takes the seed, keep 0 as a good auto-default.

Then change your engine implementation, make `init()` take the seed and remove the explicit `seed()`. It's rare to change the seed after the generator is created. If one wants, just destroy and create a new generator.

src/shared/sol-rng.h
```diff
@@ -0,0 +1,52 @@
...
+void sol_rng_engine_destroy(struct sol_rng_engine *engine);
```
barbieri: we're using `_del()` not `_destroy()` in our exported headers.

src/shared/sol-rng.h
```diff
@@ -0,0 +1,52 @@
...
+
+void sol_rng_engine_seed(struct sol_rng_engine *engine, uint64_t value);
+
+size_t sol_rng_engine_generate_bytes(struct sol_rng_engine *engine,
```
barbieri: thinking if we should change the API to be similar to `sol_util_fill_buffer()`. What do you think?

src/shared/sol-rng.h
```diff
@@ -0,0 +1,52 @@
...
+ * THEORY OF LIABILITY, WHETHER IN CONTRACT, STRICT LIABILITY, OR TORT
+ * (INCLUDING NEGLIGENCE OR OTHERWISE) ARISING IN ANY WAY OUT OF THE USE
+ * OF THIS SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
+ */
+
+#pragma once
+
+#include <stdint.h>
+
+struct sol_rng_engine;
```
barbieri: I don't like the name.

`rng` is too hard to remember it's a random-number-generator. I'm thinking about users trying to find a random number generator if they would check `sol-rng.h`.

`engine` is not needed. I'd drop engine for sure.

lpereira:
Thanks for the review. Merged with the requested changes.

src/shared/sol-rng.h
lpereira: Didn't see this comment before merging. I'll rename tomorrow to `sol_random`. Not sure yet about dropping the `engine` part; my idea was to model this API roughly based off of C++11 <random>: http://en.cppreference.com/w/cpp/numeric/random, which has the concept of engines and distributions.

barbieri: the word `engine` doesn't add much here. You can keep the engine concept, just don`t need to make APIs named after that.

It's fine (and easier to read) to have:

```c
struct sol_random *sol_random_new(const struct sol_random_type *type);
ssize_t sol_random_fill_buffer(struct sol_random *engine, struct sol_buffer *buffer, size_t len);
static inline bool
sol_random_get_int32(struct sol_random *engine, int32_t *value)
{
   struct sol_buffer buf = SOL_BUFFER_INIT_FLAGS(value, sizeof(int32_t), SOL_BUFFER_FLAGS_MEMORY_NOT_OWNED);
   ssize_t r = sol_random_fill_buffer(engine, &buf);
   sol_buffer_fini(&buf);
   return r == (ssize_t)sizeof(int32_t);
}
```

PR #679, opened 2015-09-02
build: reintroduce the initial-services

dorileo:
After kconfig migration we lost initial-service generation and
instalattion, this patch brings it back.


barbieri:
+1

dorileo:
ok, applied.

PR #681, opened 2015-09-02
modules/grove: fix rotary-sensor packet range

bdilly:
It should properly set min, max and step to make it
possible to use packets directly with map if required.

@otaviobp could you please see if it fixes the issue you've
spotted?


otaviobp:
+1. Working fine now. Thanks

dorileo:
+1

bdilly:
thanks guys, merged

PR #682, opened 2015-09-03
pwm: Don't try to change pwm polarity if polarity is already correct - v2

otaviobp:
Trying to change polarity in a chip that doesn't support polarity
changes will abort soletta and that is not needed when the current
polarity is equal the wanted polarity.

Soletta will still abort if the wanted polarity is different from
current polarity.

Replaces #674


otaviobp:
I'll change it to:
pwm: Don't try to change polarity when not needed

PR #683, opened 2015-09-03
Uuid [v4]

glima:
Changes since #667 

 - goto typo fixed and code made an util function
 - vars that could made static
 - now using sol_rng (which has a fix commit also)


glima:
@edersondisouza, thanks a lot. All fixed and now merging.

PR #684, opened 2015-09-03
shared/rng: Rename sol_rng_engine to sol_random

lpereira:
This renames `sol_rnd_engine` to `sol_random` so it's easier to figure
out what it does from a cursory glance.

Also slightly change the API to use a `sol_buffer` instead of
writing bytes to an `unsigned char` array.  While at it, add a few
utility functions to generate integers, doubles, and booleans, without
requiring the user to actually instantiate a `sol_buffer`.

The internal API also changed slightly: each implementation now
generates an `uint32_t`, which is appended to the buffer given to
`sol_random_fill_buffer()` function.


lpereira:
Ping @barbieri, @glima.

lpereira:
(The force-push was to fix a style issue and keep Semaphore happy.)

lpereira:
(Another `push -f` to rebase/fix conflicts.)

src/lib/datatypes/sol-buffer.c
```diff
@@ -38,18 +38,25 @@
...
-    SOL_EXP_CHECK(buf->flags & SOL_BUFFER_FLAGS_FIXED_CAPACITY, -EPERM);
-    SOL_EXP_CHECK(buf->flags & SOL_BUFFER_FLAGS_NO_FREE, -EPERM);
 
     if (buf->capacity == new_size)
         return 0;
 
+    SOL_EXP_CHECK(buf->flags & SOL_BUFFER_FLAGS_FIXED_CAPACITY, -EPERM);
+    SOL_EXP_CHECK(buf->flags & SOL_BUFFER_FLAGS_NO_FREE, -EPERM);
+
     new_data = realloc(buf->data, new_size);
```
barbieri: why did you move these checks? I wanted to make sure nobody calls resize on not-owned buffers.

wanted to know what would break with that :-)

lpereira: These checks were moved, not removed. Shouldn't make a difference if old_size == new_size. (Maybe the move isn't necessary anymore, will check before pushing.)

barbieri: i said they were moved... that's why I'm wondering. Nobody should call resize on fixed buffers. If the capacity == new_size, it will not warn... and will make spotting call sites harder (ie: I test on my cases where this is equal and no warnings, then you use in your app and suddenly get some warnings to that same code that got in unoticed).

lpereira: Moving those checks is not required anymore, they were just artifacts.

lpereira:
If there are no other comments, I'll merge this soon.

lpereira:
Merged.

PR #686, opened 2015-09-03
Http Server - v2

ceolin:
From the first pr:
 - Api changed to be async
 - Code styles fix
 - Not bumping the api version
 - Create a request struct
 - Keeping reference of the listen's socket watch
 - Removed reference count for the server
 - Added enum for the most common/used status code


src/modules/flow/http-server/http-server.c
```diff
@@ -0,0 +1,626 @@
...
+    };
+
+    method = sol_http_request_get_method(request);
+    response.url = sol_http_request_get_url(request);
+
+#define STRTOD_(field_) \
+    do { \
+        errno = 0; \
+        modified = true; \
+        mdata->value.d.field_ = strtod(value->value.key_value.value, NULL); \
```
lpereira: Might be a good idea to use sol_util_strtodn() to not consider locale settings. (Also, the printf() to generate JSON might print out different formatting so you might want to check out how to fix that.)

src/modules/flow/http-server/http-server.c
```diff
@@ -0,0 +1,626 @@
...
+            return -errno; \
+    } while (0)
+
+    SOL_HTTP_PARAM_FOREACH_IDX (sol_http_request_get_params(request), value, idx) {
+        switch (value->type) {
+        case SOL_HTTP_PARAM_POST_FIELD:
+            if (streq(value->value.key_value.key, "value"))
+                STRTOD_(val);
+
+            if (streq(value->value.key_value.key, "min"))
```
lpereira: `else if`?

src/modules/flow/http-server/http-server.c
```diff
@@ -0,0 +1,626 @@
...
+    SOL_INT_CHECK_GOTO(r, < 0, end);
+
+    slice.data = str;
+    slice.len = strlen(str);
+
+    r = sol_http_param_add(&response.param, SOL_HTTP_REQUEST_PARAM_HEADER(
+        HTTP_HEADER_CONTENT_TYPE, (send_json) ? HTTP_HEADER_CONTENT_TYPE_JSON : HTTP_HEADER_CONTENT_TYPE_TEXT));
+    SOL_INT_CHECK_GOTO(r, != true, end);
+
+    r = sol_buffer_set_slice(&response.content, slice);
```
lpereira: Since slice is only initialized to be used here, might be better to use `sol_str_slice_from_str()` inside the call: sol_buffer_set_slice(&content,sol_str_slice_from_str(str));

src/modules/flow/http-server/http-server.c
```diff
@@ -0,0 +1,626 @@
...
+        .api_version = SOL_HTTP_RESPONSE_API_VERSION,
+        .content = SOL_BUFFER_INIT_EMPTY,
+        .param = SOL_HTTP_REQUEST_PARAM_INIT,
+        .response_code = SOL_HTTP_STATUS_OK
+    };
+
+    method = sol_http_request_get_method(request);
+    response.url = sol_http_request_get_url(request);
+
+    SOL_HTTP_PARAM_FOREACH_IDX (sol_http_request_get_params(request), value, idx) {
```
lpereira: Many of these loops are the same in all types; is there a way to reuse more code?

src/modules/flow/http-server/http-server.c
ceolin: nope, i'm considering that the request can contains any field of the irange/drange

src/modules/flow/http-server/http-server.c
ceolin: yep, i can do that

src/lib/comms/sol-http-server-impl-microhttpd.c
```diff
@@ -0,0 +1,585 @@
...
+
+        MHD_post_process(req->pp, upload_data, *upload_data_size);
+        if (*upload_data_size) {
+            *upload_data_size = 0;
+            return MHD_YES;
+        }
+    } else if (method && streq(method, "GET")) {
+        req->method = SOL_HTTP_METHOD_GET;
+    } else {
+        SOL_WRN("Method %s not implemented", method);
```
lpereira: What if method is NULL? Check it before these various ifs.

src/lib/comms/sol-http-server-impl-microhttpd.c
```diff
@@ -0,0 +1,585 @@
...
+
+    if (flags & (SOL_FD_FLAGS_ERR | SOL_FD_FLAGS_NVAL | SOL_FD_FLAGS_HUP)) {
+        SOL_VECTOR_FOREACH_IDX (&server->fds, connection, i) {
+            if (connection->fd == fd) {
+                sol_vector_del(&server->fds, i);
+                break;
+            }
+        }
+
+        return false;
```
lpereira: Should this callback return false even if a fd wasn't removed from the fds vector?

src/lib/comms/include/sol-http.h
```diff
@@ -0,0 +1,211 @@
...
+};
+
+enum sol_http_status_code {
+    SOL_HTTP_STATUS_OK = 200,
+    SOL_HTTP_STATUS_NOT_MODIFIED = 304,
+    SOL_HTTP_STATUS_BAD_REQUEST = 400,
+    SOL_HTTP_STATUS_FORBIDDEN = 403,
+    SOL_HTTP_STATUS_NOT_FOUND = 404,
+    SOL_HTTP_STATUS_METHOD_NOT_ALLOWED = 405,
+    SOL_HTTP_STATUS_METHOD_INTERNAL_ERROR = 500,
```
lpereira: Missing link to https://http.cat/ and, of course, HTTP 418: https://http.cat/418

src/lib/comms/sol-http-server-impl-microhttpd.c
```diff
@@ -0,0 +1,585 @@
...
+static time_t
+process_if_modified_since(const char *value)
+{
+    char *s;
+    struct tm t;
+
+    s = strptime(value, "%a, %d %b %Y %H:%M:%S GMT", &t);
+    if (!s)
+        return 0;
+
```
lpereira: Might be a good idea to see if there's anything left in that string (if *s == '\0', there's garbage there).

src/lib/comms/sol-http-server-impl-microhttpd.c
```diff
@@ -0,0 +1,585 @@
...
+    SOL_NULL_CHECK(request->connection, -EINVAL);
+    SOL_NULL_CHECK(response, -EINVAL);
+
+    MHD_resume_connection(request->connection);
+
+    mhd_response = build_mhd_response(response);
+    SOL_NULL_CHECK(mhd_response, -1);
+
+    ret = MHD_queue_response(request->connection, response->response_code, mhd_response);
+    MHD_destroy_response(mhd_response);
```
lpereira: Is it OK to destroy a response right after queuing it?

src/modules/flow/http-server/http-server.c
```diff
@@ -0,0 +1,626 @@
...
+        server = NULL;
+    }
+    return -1;
+}
+
+static void
+stop_server(struct http_data *http)
+{
+    sol_http_server_unregister_handler(server, http->path);
+    free(http->path);
```
lpereira: Shouldn't this happen only when init_count is zero?

src/modules/flow/http-server/http-server.c
lpereira: Right, but if it's "value", it's not "min"; if it's "min", it's not "max", and so on.

src/modules/flow/http-server/http-server.c
ceolin: nope, this is relative to the node and not the server.
The node will stop handle this path even if the server will keep alive.

src/lib/comms/sol-http-server-impl-microhttpd.c
ceolin: Yep, response is refcounted. The MDH just icrease the reference.

src/lib/comms/include/sol-http.h
ceolin: lol, i'm afraid you'll need to commit it after this pr

src/lib/comms/sol-http-server-impl-microhttpd.c
ceolin: yes, in this case, something wrong happened with the file descriptor and the MHD was already notified, so there is nothing to do anymore. It should never happened, but .... in the worst scenario this vector is cleared in the shutdown.

src/lib/comms/sol-http-server-impl-microhttpd.c
ceolin: If method is NULL an response with the status code 501 (not implemented) is sent.
I'm can change the warning to avoid %s with NULL.

src/lib/comms/sol-http-server-impl-microhttpd.c
ceolin: yep

lpereira:
Fixes on #696.

PR #687, opened 2015-09-03
Build header installation and fixes

dorileo:
this series changes the generated source headers names and directory of installation, also fixes an misspelled initial-service file name.


dorileo:
well, it seems I broke things on ubuntu again... checking it out. - I have a script to locally run the same tests semaphore runs, but locally it doesn't break anything... let's see...

PR #690, opened 2015-09-03
Fixes related to options parsing

bdilly:
Please review it with attention since it may impact stuff
depending on conf files and we don't have much tests on that
(if any).


ibriano:
Patches seem fine

glima:
Okay for me too. I almost exploded when I looked at the size of that macro now :(

lpereira:
A brave soul will eventually convert those macros to functions.

bdilly:
@lpereira it isn't a bad idea. These macros are hard to maintain.

bdilly:
Merged

PR #691, opened 2015-09-03
Improve int/buffer and boolean/buffer [v2]

bdilly:
Changelog:
 * stretch test timeouts a bit
 * use node->node_type->default_options instead of hardcoded values

@otaviobp please verify if now make check-fbp pass


otaviobp:
Tests are working fine now. +1

bdilly:
Merged

PR #692, opened 2015-09-04
Fix limited string size reads with sol_util_read_file() calls.

glima:
The rest of the calls are fine, it seems.


ibriano:
ok

glima:
merging, thanks.

PR #694, opened 2015-09-04
Build make check fbp bin

dorileo:


dorileo:
@otaviobp the idea is the to support what already works and then go fixing the issues, see there're also cases where we *really* want to skip - namely the parser error cases.

tools/run-fbp-tests
```diff
@@ -44,6 +44,104 @@ import re
...
+        self.skip_msg = None
+        self.skip = self.SKIP_NONE
+        self.output = False
+        self.output_regex = False
+        self.output_spec = ""
+        self.exit_code = 0
+
+        self.__parse_test__()
+
+    def __parse_skip__(self, cmd, splitted):
```
lpereira: Never use dunder for methods that are not part of the object interface or the Python data model.

tools/run-fbp-tests
```diff
@@ -44,6 +44,104 @@ import re
...
+            self.exit_code = 1
+        elif cmd.startswith("TEST-SKIP"):
+            self.__parse_skip__(cmd, splitted)
+        else:
+            print("Ignoring unknown test command '%s' in test %s" %
+                  (cmd, self.test_file))
+
+    def __parse_test__(self):
+        f = open(self.test_file, encoding="UTF-8")
+        for line in f:
```
lpereira: Nitpick, but this can be written like so:

     for line in open(...):
         do_stuff_with(line)

tools/run-fbp-tests
```diff
@@ -76,137 +174,130 @@ def collect_tests(tests):
...
+
+def get_skip_list(args):
+    result = []
+    for curr in args.skip:
+        skip = getattr(FbpTest, curr.upper(), None)
+        if isinstance(skip, int):
+            result.append(skip)
+        else:
+            print("Unknown skip spec: %s" % curr)
+    return result
```
lpereira: Return `set(result)`, to speed up `foo in get_skip_list()`.

tools/run-fbp-tests
```diff
@@ -44,6 +44,104 @@ import re
 import subprocess
 import sys
 
+valgrind_opts = [
```
lpereira: Use a tuple if these values aren't supposed to be changed. While there, rename to `VALGRIND_OPTS` so it screams it's a constant.

tools/run-fbp-tests
```diff
@@ -44,6 +44,104 @@ import re
...
+
+    def __parse_test__(self):
+        f = open(self.test_file, encoding="UTF-8")
+        for line in f:
+            if line.startswith("## "):
+                self.__parse_command__(line)
+            elif (self.output or self.output_regex) and line.startswith("#"):
+                self.output_spec += line[2:] or "\n"
+
+    def is_skip(self, skips):
```
lpereira: `is_skip()` is bad English. Rename to `should_skip()` or `is_skipped()`.

tools/run-fbp-tests
```diff
@@ -76,137 +174,130 @@ def collect_tests(tests):
...
+            result.append(skip)
+        else:
+            print("Unknown skip spec: %s" % curr)
+    return result
+
+def query_tests(args):
+    result = []
+    skips = get_skip_list(args)
+    tests = collect_tests(args.tests)
+    for t in sorted(set(tests)):
```
lpereira: Maybe `collect_tests()` can already return a set?

tools/run-fbp-tests
```diff
@@ -76,137 +174,130 @@ def collect_tests(tests):
...
-    print("%d tests passed." % passed)
-    if skipped > 0:
-        print("%d tests skipped." % skipped)
-    if failed > 0:
-        print("%d tests failed." % failed)
+        if not args.compiled:
+            cmd = cmd_prefix + [runner, path["file"]]
+            test_cwd = path["dir"]
+        else:
+            exe = "%s/%s" % (args.compiled_dir, path["file"].replace(".fbp", ""))
```
lpereira: Use `os.path.join()`.

tools/run-fbp-tests
```diff
@@ -44,6 +44,104 @@ import re
...
+                self.output_spec += line[2:] or "\n"
+
+    def is_skip(self, skips):
+        if self.skip == self.SKIP_NONE:
+            return False
+        elif self.skip in skips:
+            return True
+        return False
+
+    def check_result(self, out, code):
```
lpereira: This is not checking the result, it's printing values to the standard output. Either rename the function or use raise an exception with the error details so the call site can print them.

tools/run-fbp-tests
```diff
@@ -44,6 +44,104 @@ import re
...
+    "--leak-check=full",
+]
+
+class FbpTest:
+    SKIP_NONE = 0
+    SKIP_SIMPLE = 1
+    SKIP_VALGRIND = 2
+    SKIP_COMPILE = 3
+
+    supported_skip = {
```
lpereira: `SUPPORTED_SKIP`, as this is a constant.

tools/run-fbp-tests
```diff
@@ -76,137 +174,130 @@ def collect_tests(tests):
...
-                skip = ""
-        elif cmd in ["TEST-OUTPUT", "TEST-OUTPUT-REGEX"]:
-            out = ""
-            break
-        else:
-            print("Ignoring unknown test command '%s' in test %s" % (cmd, t))
-            continue
+    valgrind = os.path.abspath(args.valgrind)
+    if not os.path.isfile(valgrind):
+        sys.exit("Couldn't find valgrind program at", valgrind)
```
lpereira: Might be a good idea to see if the file is executable rather than just check if it exists.

tools/run-fbp-tests
```diff
@@ -76,137 +174,130 @@ def collect_tests(tests):
...
+if __name__ == "__main__":
+    parser = argparse.ArgumentParser()
+    parser.add_argument("--runner", help="Path to the sol-fbp-runner program",
+                        default="build/soletta_sysroot/usr/bin/sol-fbp-runner", type=str)
+    parser.add_argument("--valgrind", help="Enable valgrind, if path not passed, assume "
+                        "default location in the system", nargs='?', const="/usr/bin/valgrind",
+                        type=str)
+    parser.add_argument("--valgrind-supp", help="Path to valgrind's suppression file",
+                        type=str)
+    parser.add_argument("--verbose", help="Be verbose", dest="verbose", action="store_true")
```
lpereira: I like the fact that the help for the option that enables verbosity is quite terse. :)

tools/run-fbp-tests
```diff
@@ -76,137 +174,130 @@ def collect_tests(tests):
...
+        if v > 0:
+            print("%d tests %s." % (v, k))
+
+    if failures:
         sys.exit(1)
+
+def get_skip_list(args):
+    result = []
+    for curr in args.skip:
+        skip = getattr(FbpTest, curr.upper(), None)
```
lpereira: I'd also check `if curr.upper().startswith('FBP_')`, for good measure.

tools/run-fbp-tests
```diff
@@ -76,137 +174,130 @@ def collect_tests(tests):
...
+        if isinstance(skip, int):
```
lpereira: Then this `isinstance()` call can be removed.

dorileo:
ok guys, just sent a new version #698 - could you take a look at it again? thanks for reviewing.

PR #695, opened 2015-09-08
comms/http-client: destroy connection after calling cb

bdilly:
@lpereira could you please review it?


lpereira:
LGTM.

It's very unlikely there's going to be too many connections so the call to `sol_ptr_vector_remove()` doesn't bother me that much that we're exchanging a very cheap NULL pointer check with a loop.

In fact, I believe we could implement this function on top of `sol_vector` directly, by providing a `_remove()` method that takes a function pointer to compare items; I bet we can clean up a bunch of code this way (it always bothered me that, to find and and remove items in a vector, one has to traverse it in reverse).

bdilly:
ok, added to todo list.
merged.

PR #697, opened 2015-09-08
build: add -lasan ldflags

dorileo:
Since gcc < 5.0 will not adjust the ldflags for us - given an -fsanitize
we should do so ourselves to be backward compatible with older gcc.


anselmolsm:
:+1:

dorileo:
applied, thanks.

lpereira:
I love -lasan!

PR #698, opened 2015-09-08
Build make check fbp bin v2

dorileo:
## Changes
  + v2 (since #694):
    + run-fbp-tests:
      - fixed issues pointed by @lpereira in the previous review;
      - changed the output handling to use logging instead of simply printing out;
        - with that also introduced the ```--log``` argument, so the user may specify the logging level;
    + Makefile.target:
      - dropped the ```*.txt``` copy from ```check-fbp-bin``` rule - we're skipping the reader/writer tests, as result we're not using these files at all;

## Rationale
Add's ```check-fbp-bin``` rules, for that changes and adaptations were required on ```run-fbp-tests``` scripts so we can query the "wanted" tests to be compiled with that the build system can handle the compilation itself. Also in the process of changing the ```run-fbp-tests``` scripts we incorporated a bunch of improvements and design changes.

### DIFF
--- a/tools/run-fbp-tests
+++ b/tools/run-fbp-tests
@@ -39,10 +39,105 @@ from io import open
 
 import argparse
 import difflib
+import logging
 import os
 import re
 import subprocess
 import sys
+
+DEFAULT_VALGRIND_OPTS = (
+    "--show-reachable=no",
+    "--quiet",
+    "--error-exitcode=1",
+    "--leak-check=full",
+)
+
+class FbpTest:
+    SKIP_NONE = 0
+    SKIP_SIMPLE = 1
+    SKIP_VALGRIND = 2
+    SKIP_COMPILE = 3
+
+    SUPPORTED_SKIP = {
+        "TEST-SKIP": SKIP_SIMPLE,
+        "TEST-SKIP-VALGRIND": SKIP_VALGRIND,
+        "TEST-SKIP-COMPILE": SKIP_COMPILE,
+    }
+
+    def __init__(self, test_file):
+        self.test_file = test_file
+        self.skip_msg = None
+        self.skip = self.SKIP_NONE
+        self.output = False
+        self.output_regex = False
+        self.output_spec = ""
+        self.exit_code = 0
+
+        self._parse_test()
+
+    def _parse_skip(self, cmd, splitted):
+        self.skip = self.SUPPORTED_SKIP.get(cmd, self.SKIP_NONE)
+        if not self.skip:
+            return
+
+        if len(splitted) > 2:
+            self.skip_msg = splitted[2]
+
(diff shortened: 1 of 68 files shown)


tools/run-fbp-tests
```diff
@@ -39,11 +39,106 @@ from io import open
...
+        return False
+
+    def show_result_status(self, out, code):
+        if self.output and self.output_spec != out:
+            orig = out.splitlines(True)
+            dest = self.output_spec.splitlines(True)
+            logging.warning("FAILED running %s: wrong output" % os.path.basename(self.test_file))
+            for line in difflib.unified_diff(orig, dest):
+                sys.stdout.write(line)
+            return "failed"
```
lpereira: Why return a string, and not True/False? Symbols are less likely to be mistyped.

tools/run-fbp-tests
```diff
@@ -67,146 +162,162 @@ def collect_tests(tests):
...
+
+    for k,v in status.items():
+        if v > 0:
+            logging.warning("%d tests %s." % (v, k))
+
+    if failures:
         sys.exit(1)
+
+def get_skip_list(args):
+    result = []
```
lpereira: Maybe result can be a set already? (Use `.add()` instead of `.append()`.)

tools/run-fbp-tests
```diff
@@ -67,146 +162,162 @@ def collect_tests(tests):
...
+            continue
+
+        skip = getattr(FbpTest, curr.upper(), None)
+        if not skip:
+            logging.info("Unknown skip spec: %s" % curr)
+
+        result.append(skip)
+    return set(result)
+
+def query_tests(args):
```
lpereira: Maybe a generator expression could simplify this function (nitpick and bikeshedding in a single comment):

    skips = get_skip_list(args)
    return (FbpTest(t) for t in collect_tests(args.tests) if not should_skip(t))

And then just print the result outside:

   print(" ".join(query_tests(args))

lpereira:
LGTM with harmless comments.

tools/run-fbp-tests
dorileo: I use a dict to accumulate the status, so that's the why I returning the key part but we can change it, I just didn't want to have 3 if - else blocks to define the var to use and so on, latter on I do iterate over this dict items to print the status, just a matter of keeping the code a bit simpler.

tools/run-fbp-tests
dorileo: yep, sure. thanks.

tools/run-fbp-tests
dorileo: ok.

dorileo:
applied, thanks...

PR #699, opened 2015-09-08
Add new string/{starts,ends}-with nodes, with tests.

glima:


glima:
Thanks a lot. All fixed and merged.

PR #703, opened 2015-09-09
build: link to static libasan

dorileo:
## Changes
  + v2 (since #702)
    - fixed s/asan/ubsan/;
    - removed san intrumentation for no-san option;

## Rationale
In the commit 202bff12f21b114d39b818b887ff94771a034c4d we changed the
sanitze dependencies to also force linking against libasan - in the
commit message we state that this would be harmless to newer gcc
versions. I turns out not to be true, if we're linking against the
shared libasan we should make sure to preload the lib or make sure it is
the first DSO name[1].

To avoid this kind of runtime issues we're forcing libasan to be
statically linked.

[1] -
https://github.com/gcc-mirror/gcc/blob/master/libsanitizer/asan/asan_linux.cc#L112

### DIFF
--- a/data/jsons/dependencies.json
+++ b/data/jsons/dependencies.json
@@ -407,14 +407,14 @@
         "value": "-fsanitize=undefined"
       },
       "ldflags": {
-        "value": "-fsanitize=undefined -lasan"
+        "value": "-fsanitize=undefined -static-libubsan"
       }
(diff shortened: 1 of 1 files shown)


ceolin:
It does the job and seems a reasonable way to workaround the problem :+1:

PR #705, opened 2015-09-10
Set flags v2

rchiossi:


rchiossi:
Fixed missing new line (style-check) in local branch.

src/lib/common/include/sol-mainloop.h
```diff
@@ -143,6 +143,9 @@ enum sol_fd_flags {
 struct sol_fd;
 struct sol_fd *sol_fd_add(int fd, unsigned int flags, bool (*cb)(void *data, int fd, unsigned int active_flags), const void *data);
 bool sol_fd_del(struct sol_fd *handle);
+bool sol_fd_set_flags(struct sol_fd *handle, unsigned int flags);
+unsigned int sol_fd_get_flags(const struct sol_fd *handle);
+#define sol_fd_unset_flags(handle, flags) sol_fd_set_flags(handle, sol_fd_get_flags(handle) & ~flags)
```
lpereira: Make this a static inline function. It'll fix the bug if one passes more than one flag:

    sol_fd_unset_flags(handle, FLAG1 | FLAG2);

Will be expanded to:

    sol_fd_set_flags(handle, sol_fd_get_flags(handle) & ~FLAG1 | FLAG2)

Which is most likely not what you want (only FLAG1 is complemented). Making it a static inline function will avoid this gotcha while being the exact same thing. (In fact there's no need to even make it static inline, it could be a "normal" function instead.)

lpereira:
LGTM, but please fix `sol_fd_unset_flags()` before merging.

src/lib/common/include/sol-mainloop.h
ibriano: On Thu, 10 Sep 2015 10:53:34 -0700, Leandro A. F. Pereira wrote:
> > @@ -143,6 +143,9 @@ enum sol_fd_flags {
> >  struct sol_fd;
> >  struct sol_fd *sol_fd_add(int fd, unsigned int flags, bool (*cb)(void *data, int fd, unsigned int active_flags), const void *data);
> >  bool sol_fd_del(struct sol_fd *handle);
> > +bool sol_fd_set_flags(struct sol_fd *handle, unsigned int flags);
> > +unsigned int sol_fd_get_flags(const struct sol_fd *handle);
> > +#define sol_fd_unset_flags(handle, flags) sol_fd_set_flags(handle, sol_fd_get_flags(handle) & ~flags)
> 
> Make this a static inline function. It'll fix the bug if one passes more than one flag:
> 
>     sol_fd_unset_flags(handle, FLAG1 | FLAG2);
> 
> Will be expanded to:
> 
>     sol_fd_set_flags(handle, sol_fd_get_flags(handle) & ~FLAG1 | FLAG2)
> 

That's also why you should always wrap macro arguments with parens.

rchiossi:
fixed and merged. thanks

PR #706, opened 2015-09-10
Fixes

ceolin:
- netlink requests (using recvmsg instead of read) are not longer returning EINVAL
- freegeoip (allow build as module)
- Make the http nodes build as module (instead of builtin)

### DIFF
--- a/src/lib/comms/sol-network-linux.c
+++ b/src/lib/comms/sol-network-linux.c
@@ -275,7 +275,6 @@ _on_event(void *data, int nl_socket, unsigned int cond)
 static void
 _netlink_request(int event)
 {
-    size_t n;
     char buf[NLMSG_ALIGN(sizeof(struct nlmsghdr) + sizeof(struct rtgenmsg))] = { 0 };
     struct iovec iov = { buf, sizeof(buf) };
     struct sockaddr_nl snl = { .nl_family = AF_NETLINK };
@@ -290,9 +289,6 @@ _netlink_request(int event)
     };
     struct nlmsghdr *h;
     struct rtgenmsg *rt;
-    char buffer_receive_back[4096]; /* Backing storage for next sol_buffer */
-    struct sol_buffer buffer_receive = SOL_BUFFER_INIT_FLAGS(buffer_receive_back,
-        sizeof(buffer_receive_back), SOL_BUFFER_FLAGS_MEMORY_NOT_OWNED);
 
     h = (struct nlmsghdr *)buf;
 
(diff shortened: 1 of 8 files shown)


src/lib/comms/sol-network-linux.c
```diff
@@ -236,35 +236,36 @@ _on_event(void *data, int nl_socket, unsigned int cond)
         return false;
     }
 
-    status = recvmsg(nl_socket, &msg, MSG_WAITALL);
-    if (status <= 0) {
-        if (errno == EWOULDBLOCK || errno == EAGAIN)
-            return true;
+    while ((status = recvmsg(nl_socket, &msg, MSG_WAITALL))) {
+        if (status < 0) {
+            if (errno == EWOULDBLOCK || errno == EAGAIN)
```
lpereira: `EAGAIN` and `EWOULDBLOCK` are the same under Linux (and this is a Linux-specific part of the code), so just handle one of them. While you're at it, also handle `EINTR`.

src/modules/flow/freegeoip/freegeoip.json
```diff
@@ -14,7 +14,7 @@
         "open": "freegeoip_open",
         "close": "freegeoip_close"
       },
-      "name": "location/freegeoip",
+      "name": "freegeoip/location",
```
lpereira: This would ideally be `location/freegeoip`, as there might be other providers for the same interface (GPS, GPRS antenna triangulation, other IP->location databases, etc).

ceolin: ok, so we should change the filename to location.c, otherwise the .so will be freegeoip and the runner will no find the .so

I still prefer this, we can put the other modules that implement the same interface in the same category. I don't we are going to add other service under the freegeoip.json/c

src/lib/comms/sol-network-linux.c
ceolin: should have the same treatment of EAGAIN ?

lpereira: No; just restart the `recvmsg()` call (with a `continue`?).

ceolin: well, EAGAIN basically does it, but first return to mainloop.  I'll put explicitly an continue for this error. Thanks for pointing it.

ceolin:
fixes on https://github.com/solettaproject/soletta/pull/708

src/lib/comms/sol-network-linux.c
ibriano: On Fri, 11 Sep 2015 05:18:08 -0700, Flávio Vinicius Alvares Ceolin wrote:
> > @@ -236,35 +236,36 @@ _on_event(void *data, int nl_socket, unsigned int cond)
> >          return false;
> >      }
> >  
> > -    status = recvmsg(nl_socket, &msg, MSG_WAITALL);
> > -    if (status <= 0) {
> > -        if (errno == EWOULDBLOCK || errno == EAGAIN)
> > -            return true;
> > +    while ((status = recvmsg(nl_socket, &msg, MSG_WAITALL))) {
> > +        if (status < 0) {
> > +            if (errno == EWOULDBLOCK || errno == EAGAIN)
> 
> well, EAGAIN basically does it, but first return to mainloop.  I'll put explicitly an continue for this error. Thanks for pointing it.

The difference is that EAGAIN means "I have nothing more to give you
now, come back later", and EINTR is "There's stuff to read but we were
interrupted, you can try again right now"

PR #707, opened 2015-09-10
samples: fix build

bdilly:
@dorileo could you take a look on it please? Looks like
many samples never were built thanks to typos.


dorileo:
I've tested and it builds nicely, had just minor concerns I commented on 64ddfca5d83e6c1ee72d86cddc019ec6a0f2482b but the series looks nice. Thank you for the maintenance.

bdilly:
Thanks, guys.
Merged.

PR #710, opened 2015-09-11
modules/byte: add comparison node types

bdilly:
It has been a bit frequent to convert bytes to int to make
comparisons, so let's make this operation easier adding
node types to do it directly.

### DIFF
--- a/src/modules/flow/byte/byte.c
+++ b/src/modules/flow/byte/byte.c
@@ -227,7 +227,6 @@ byte_filter_open(struct sol_flow_node *node, void *data, const struct sol_flow_n
     return 0;
 }
 
-
 static int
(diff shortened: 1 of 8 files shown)


src/modules/flow/byte/byte.c
```diff
@@ -244,5 +243,76 @@ byte_filter_process(struct sol_flow_node *node, void *data, uint16_t port, uint1
...
+    struct byte_comparison_data *mdata = data;
+    bool output;
+    const struct byte_comparison_node_type *type;
+    int r;
+
+    r = sol_flow_packet_get_byte(packet, &mdata->val[port]);
+    SOL_INT_CHECK(r, < 0, r);
+
+    mdata->val_initialized[port] = true;
+    if (!(mdata->val_initialized[0] && mdata->val_initialized[1]))
```
lpereira: Unrelated to this patch (which LGTM, BTW), but maybe a bitmap would be better for the initialized vector?

PR #711, opened 2015-09-11
Http v4

ceolin:
from v3:
 - fix typos
 - not formatting the json output
 - added helpers to escape string and convert double to string (considering the locale)
 - removed dead code
 - clearing the requests vector when the server init fails
 - avoid possible hole in the server's structure
 - Add more debuggin information


src/modules/flow/http-server/http-server.c
```diff
@@ -0,0 +1,661 @@
...
+
+    method = sol_http_request_get_method(request);
+    response.url = sol_http_request_get_url(request);
+
+    SOL_HTTP_PARAM_FOREACH_IDX (sol_http_request_get_params(request), value, idx) {
+        switch (value->type) {
+        case SOL_HTTP_PARAM_POST_FIELD:
+            if (streq(value->value.key_value.key, "value")) {
+                char *escaped_str;
+                size_t escaped_len = sol_json_calculate_escaped_string_len(value->value.key_value.value);
```
lpereira: This will escape the string even if `send_json` is `false`. Is this OK?

src/modules/flow/http-server/http-server.c
```diff
@@ -0,0 +1,661 @@
...
+                    send_json = true;
+            }
+            break;
+        default:
+            break;
+        }
+    }
+
+    if (send_json) {
+        r = asprintf(&response_str, "{\"%s\": \"%s\"\n}", mdata->path,
```
lpereira: `\n` left in the string.

src/modules/flow/http-server/http-server.c
```diff
@@ -0,0 +1,661 @@
...
+                    send_json = true;
+            }
+            break;
+        default:
+            break;
+        }
+    }
+
+    if (send_json) {
+        r = snprintf(response_str, sizeof(response_str), "{\n\t\"%s\": %s\n}", mdata->path,
```
lpereira: `\n\t` and `\n` left in the string.

lpereira:
LGTM with comments.

src/modules/flow/http-server/http-server.c
ceolin: I did propositally, but i re-thinking it's better just escape when sending json.

src/modules/flow/http-server/http-server.c
ceolin: ack

src/modules/flow/http-server/http-server.c
ceolin: ack

ceolin:
merged with the fixes. Thanks a lot for reviewing this pr

PR #712, opened 2015-09-14
Basic implementantion of machine_id and serial number for RIOT

ibriano:
Makes use of the cpuid_get() function from RIOT if it's available.
For boards that don't implement it or don't provide a valid UUID there,
we need more infrastructure to create a valid and persistend UUID using
other characteristics of the board, like the MAC address if it has one.

### DIFF
--- a/src/lib/common/sol-platform-impl-riot.c
+++ b/src/lib/common/sol-platform-impl-riot.c
@@ -36,8 +36,10 @@
 
 #include "sol-platform.h"
 #include "sol-platform-impl.h"
+#include "sol-util.h"
 
 #include "lpm.h"
+#include "periph/cpuid.h"
 
 int
 sol_platform_impl_init(void)
@@ -119,23 +121,68 @@ sol_platform_impl_set_target(const char *target)
     return -ENOTSUP;
 }
 
+static inline char
+to_hex(int num)
+{
+    return num > 9 ? num - 10 + 'a' : num + '0';
+}
+
+static void
+serial_to_string(const char *buf, size_t len, char *dst)
+{
+    char *ptr;
+    int i;
+
+    for (ptr = dst, i = 0; i < len; i++) {
+        int h, l;
+
+        h = ((unsigned)buf[i] & 0xf0) >> 4;
+        l = buf[i] & 0x0f;
+        *(ptr++) = to_hex(h);
+        *(ptr++) = to_hex(l);
+    }
+    *ptr = 0;
+}
+
 int
 sol_platform_impl_get_machine_id(char id[static 33])
 {
-    SOL_WRN("Not implemented");
-    return -ENOTSUP;
+#ifdef CPUID_ID_LEN
+    char cpuid[CPUID_ID_LEN];
+
+    /* Assume, for now, the the cpuid we get is a valid UUID */
+    cpuid_get(cpuid);
+    serial_to_string(cpuid, CPUID_ID_LEN, id);
+
+    return 0;
+#else
+    return -ENOSYS;
+#endif
 }
 
 int
 sol_platform_impl_get_serial_number(char **number)
 {
-    SOL_WRN("Not implemented");
-    return -ENOTSUP;
+#ifdef CPUID_ID_LEN
+    char cpuid[CPUID_ID_LEN];
+
+    if (!number)
+        return -EINVAL;
+
+    *number = malloc(CPUID_ID_LEN * 2 + 1);
+    SOL_NULL_CHECK(*number, -ENOMEM);
+
+    cpuid_get(cpuid);
+    serial_to_string(cpuid, CPUID_ID_LEN, *number);
+
+    return 0;
+#else
+    return -ENOSYS;
+#endif
 }
 
 char *
 sol_platform_impl_get_os_version(void)
 {
-    SOL_WRN("Not implemented");
-    return NULL;
+    return strdup(RIOT_VERSION);
 }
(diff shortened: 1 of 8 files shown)


src/lib/common/sol-platform-impl-riot.c
```diff
@@ -119,23 +121,68 @@ sol_platform_impl_set_target(const char *target)
...
+    *ptr = 0;
+}
+
 int
 sol_platform_impl_get_machine_id(char id[static 33])
 {
-    SOL_WRN("Not implemented");
-    return -ENOTSUP;
+#ifdef CPUID_ID_LEN
+    char cpuid[CPUID_ID_LEN];
```
lpereira: Does CPUID_ID_LEN account for the terminating NUL?

src/lib/common/sol-platform-impl-riot.c
```diff
@@ -119,23 +121,68 @@ sol_platform_impl_set_target(const char *target)
...
+
+    if (!number)
+        return -EINVAL;
+
+    *number = malloc(CPUID_ID_LEN * 2 + 1);
+    SOL_NULL_CHECK(*number, -ENOMEM);
+
+    cpuid_get(cpuid);
+    serial_to_string(cpuid, CPUID_ID_LEN, *number);
```
lpereira: Why not implement this with `sol_platform_impl_get_machine_id()`?

src/lib/common/sol-platform-impl-riot.c
ibriano: On Mon, 14 Sep 2015 07:05:46 -0700, Leandro A. F. Pereira wrote:
> >  int
> >  sol_platform_impl_get_machine_id(char id[static 33])
> >  {
> > -    SOL_WRN("Not implemented");
> > -    return -ENOTSUP;
> > +#ifdef CPUID_ID_LEN
> > +    char cpuid[CPUID_ID_LEN];
> 
> Does CPUID_ID_LEN account for the terminating NUL?

It's an integer, not a string.

src/modules/linux-micro/machine-id/machine-id.c
```diff
@@ -94,7 +94,7 @@ run_do(void)
         SOL_INT_CHECK(r, == 0, 0);
     }
 
-    r = sol_util_uuid_gen(false, true, id);
+    r = sol_util_uuid_gen(false, false, id);
```
lpereira: Might be a good idea to have this function take an `enum sol_util_uuid_flags { SOL_UTIL_UUID_UPPER_CASE = 1<<0, SOL_UTIL_UUID_WITH_HYPHENS = 1<<1 }`.

src/lib/common/sol-platform-impl-riot.c
ibriano: On Mon, 14 Sep 2015 07:06:44 -0700, Leandro A. F. Pereira wrote:
> >  }
> >  
> >  int
> >  sol_platform_impl_get_serial_number(char **number)
> >  {
> > -    SOL_WRN("Not implemented");
> > -    return -ENOTSUP;
> > +#ifdef CPUID_ID_LEN
> > +    char cpuid[CPUID_ID_LEN];
> > +
> > +    if (!number)
> > +        return -EINVAL;
> > +
> > +    *number = malloc(CPUID_ID_LEN * 2 + 1);
> > +    SOL_NULL_CHECK(*number, -ENOMEM);
> > +
> > +    cpuid_get(cpuid);
> > +    serial_to_string(cpuid, CPUID_ID_LEN, *number);
> 
> Why not implement this with `sol_platform_impl_get_machine_id()`?

Because they won't necessarily be always the same, but we could make the
separation when it matters, I guess.

src/shared/sol-random.c
```diff
@@ -86,7 +86,9 @@ getrandom_shim(void *buf, size_t buflen, unsigned int flags)
 static uint64_t
 get_platform_seed(uint64_t seed)
```
lpereira: Doesn't RIOT have anything to generate some entropy? (Maybe we could test if `RDRAND` is available?)

ibriano: On Mon, 14 Sep 2015 07:09:51 -0700, Leandro A. F. Pereira wrote:
> > @@ -86,7 +86,9 @@ getrandom_shim(void *buf, size_t buflen, unsigned int flags)
> >  static uint64_t
> >  get_platform_seed(uint64_t seed)
> 
> Doesn't RIOT have anything to generate some entropy? (Maybe we could test if `RDRAND` is available?)

It has an implementation of the same algorithm we are using, and if the
cpu supports it and it's implemented, it has hardware random support,
but not on the board I'm testing with.
The Arduino Due seems to have one, I can get that board later to
implement that part.

lpereira: By the same algorithm, you mean Mersenne Twister? If so, it is only as good as the initial seed; if it's initialized to `time(NULL)`, it's not a really good starting point.

It might be a good idea later, then, to ensure that the RIOT implementation of `SOL_RANDOM_MT19937` uses the system implementation.

lpereira:
Patch series LGTM. Comments are unrelated to them, but might be added to the ever-growing TODO list.

ibriano:
On Mon, 14 Sep 2015 07:20:21 -0700, Leandro A. F. Pereira wrote:
> Patch series LGTM. Comments are unrelated to them, but might be added to the ever-growing TODO list.

I'll get a new SSD to make space to add those comments to the TODO list.

ibriano:
Merged

PR #714, opened 2015-09-14
pc: update description + add nodedescriptionpath

dorileo:
This patch updates the soletta's .pc description to some more acurate
text and introduces the nodedescriptionpath variable so the lib's users
may find out where the json descriptions are installed.


anselmolsm:
:+1:, tks

dorileo:
ok, applied.

PR #715, opened 2015-09-14
build: check-fbp-bin should know about modules

dorileo:
With alldefconfig or allmodconfig where tests depend on node types built
as modules the tests will not run due to lack of symbol definition -
since they were not linked to modules.

Since we don't configure the fbp tests but discover'em with directory
listing and skip codes we can't specify dependencies, this patch changes
the fbp-bin compilation to always depend on all modules - again, we
don't know exactly the ones we really depend on but for test suite
purpose it's ok to link against all modules.


dorileo:
actually we're already linking against the modules... we just have special requirements for the ubuntu's gcc version. I'm sending a new patch for that.

PR #716, opened 2015-09-14
Use device/machine ID instead of MAC address in OIC nodes

lpereira:
This enables the use of device/machine ID instead of MAC address in OIC nodes.

This is performed by making another query to each device that replies to a find_resource query, to obtain the device information, and comparing with the number provided in the node configuration. This will eventually be complemented by a device ID port, so that persistence nodes can be used as well.

As a bonus, a few odds and ends in related code were fixed.


lpereira:
Force-pushed to fix minor style issue.

ibriano:
Seems fine to me.

lpereira:
Merged, thank you.

PR #717, opened 2015-09-14
Mqtt v2

rchiossi:


src/samples/mqtt/mqtt-publish.c
```diff
@@ -0,0 +1,151 @@
...
+        SOL_INF("Usage: %s <ip> <port> <topic> <message>", argv[0]);
+        return 0;
+    }
+
+    port = atoi(argv[2]);
+    topic = argv[3];
+
+    payload = SOL_BUFFER_INIT_CONST(argv[4], strlen(argv[4]));
+
+    message = (struct sol_mqtt_message){
```
lpereira: Space missing before {

rchiossi: Not really. Thats a cast and uncrustify complains if you put a space there.

src/lib/datatypes/sol-buffer.c
```diff
@@ -301,3 +301,24 @@ sol_buffer_steal(struct sol_buffer *buf, size_t *size)
...
+    b_copy = sol_util_memdup(buf, sizeof(*buf));
+    if (!b_copy) return NULL;
+
+    b_copy->data = sol_util_memdup(buf->data, buf->used);
+    if (!b_copy->data) {
+        free(b_copy);
+        return NULL;
+    }
+
+    b_copy->flags = SOL_BUFFER_FLAGS_DEFAULT;
```
lpereira: There are other flags that should be copied over, if set on the original: SOL_BUFFER_FLAGS_FIXED_CAPACITY and SOL_BUFFER_FLAGS_NO_NUL_BYTE, to also copy the behavior of the original buffer.

src/lib/comms/sol-mqtt-impl-mosquitto.c
```diff
@@ -0,0 +1,518 @@
...
+    free(message->topic);
+    free(message);
+}
+
+static void
+sol_mqtt_init(void)
+{
+    SOL_LOG_INTERNAL_INIT_ONCE;
+    init_ref++;
+    mosquitto_lib_init();
```
lpereira: Shouldn't the lib be initialized only once?

src/samples/mqtt/mqtt-publish.c
lpereira: It is not just good, it is good enough

src/lib/datatypes/sol-buffer.c
rchiossi: But that's not true for the copy. The copy can be resized and should be freed even if coming from a MEMORY_NOT_OWNED or FIXED_CAPACITY buffer. I'm not sure about NUL_BYTE, I'll check it out.

lpereira:
LGTM with comments

src/lib/comms/sol-mqtt-impl-mosquitto.c
rchiossi: After the first initialization we only increase a "ref count" so we can track when it is safe to shutdown.

lpereira: That's not what's happening here: init_ref is incremented and mosquitto_lib_init() is called regardless of how many times init_ref has been incremented.

rchiossi: Ah yes, my bad. It shouldn't be a problem but there is no need to call it again anyway. I'll fix it. Thanks.

rchiossi:
fixed and merged.

PR #718, opened 2015-09-14
Memmap - yet another storage media

edersondisouza:
This series moves `persistence` node backing storage functions to `lib/io`, so they can be used anywhere, adding some helpers to write `double` our `sol_drange` and introduces a new backing storage: `memmap`.
This storage fits media where we only have a block to write stuff. To associate what is where on the media, a map should be provided. It can be created via C API or sent to sol-fbp-generator (or runner) via `-m` option. On the first, the output C file will contain the mapping, on the later it will be loaded on the fly.
There's doubt if we should keep the separated option for this map file, or if it should be part of `conffile`. Let's hope that with this PR, people can settle on a conclusion.

### DIFF
--- a/map.json
+++ b/map.json
@@ -1,0 +1,117 @@
+[
+    {
+        "version": 2,
+        "path": "nvram3.bin",
+        "entries": [
+            {
+                "name": "_version",
+                "offset": 53,
+                "size": 1
+            },
+            {
+                "name": "save3_bool",
+                "offset": 54,
+                "size": 1,
+                "bit_offset": 0,
+                "is_bool": true
+            },
+            {
+                "name": "save3_bool2",
+                "offset": 54,
+                "size": 1,
+                "bit_offset": 1,
+                "is_bool": true
+            },
+            {
+                "name": "save3_byte",
+                "offset": 55,
+                "size": 1
+            },
+            {
+                "name": "save3_int",
+                "offset": 56,
+                "size": 4
+            },
+            {
+                "name": "save3_double",
+                "offset": 60,
+                "size": 8
+            },
+            {
+                "name": "save3_string",
+                "offset": 68,
+                "size": 2
+            },
+            {
+                "name": "save3_irange",
+                "offset": 70,
+                "size": 16
+            },
+            {
+                "name": "save3_drange",
+                "offset": 86,
+                "size": 32
+            },
+            {
+                "name": "save_float",
+                "offset": 118,
+                "size": 8
+            }
+        ]
+    },
+    {
+        "version": 2,
+        "path": "nvram.bin",
+        "entries": [
+            {
+                "name": "_version",
+                "offset": 53,
+                "size": 1
+            },
+            {
+                "name": "save_bool",
+                "offset": 54,
+                "size": 1,
+                "bit_offset": 0,
+                "is_bool": true
+            },
+            {
+                "name": "save_bool2",
+                "offset": 54,
+                "size": 1,
+                "bit_offset": 1,
+                "is_bool": true
+            },
+            {
+                "name": "save_byte",
+                "offset": 55,
+                "size": 1
+            },
+            {
+                "name": "save_int",
+                "offset": 56,
+                "size": 4
+            },
+            {
+                "name": "save_double",
+                "offset": 60,
+                "size": 8
+            },
+            {
+                "name": "save_string",
+                "offset": 68,
+                "size": 2
+            },
+            {
+                "name": "save_irange",
+                "offset": 70,
+                "size": 16
+            },
+            {
+                "name": "save_drange",
+                "offset": 86,
+                "size": 32
+            }
+        ]
+    }
+]
--- a/src/bin/sol-fbp-generator/main.c
+++ b/src/bin/sol-fbp-generator/main.c
@@ -57,6 +57,7 @@ static struct {
     const char *conf_file;
     const char *output_file;
     const char *export_symbol;
+    const char *memory_map_file;
 
     struct sol_ptr_vector json_files;
     struct sol_ptr_vector fbp_search_paths;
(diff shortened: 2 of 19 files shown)


map.json
```diff
@@ -0,0 +1,117 @@
...
+            },
+            {
+                "name": "save3_irange",
+                "offset": 70,
+                "size": 16
+            },
+            {
+                "name": "save3_drange",
+                "offset": 86,
+                "size": 32
```
lpereira: I'm not sure if I like this. This is too error-prone to be considered practical: not only one has to know what's the size in bytes of a struct, it's possible to mess up offsets as well.

src/bin/sol-fbp-runner/main.c
```diff
@@ -149,6 +172,180 @@ parse_args(int argc, char *argv[])
...
+static void
+clear_memory_maps(void)
+{
+    const struct sol_str_table_ptr *iter;
+    struct sol_memmap_map *map;
+    int i;
+
+    SOL_PTR_VECTOR_FOREACH_IDX (&memory_maps, map, i) {
+        for (iter = map->entries; iter->key; iter++) {
+            free((void *)iter->key);
```
lpereira: Maybe use sol_str_arena instead?

map.json
barbieri: we can check if they are okay while loading. We can also add a script to draw SVG with it. Last but not least, one day someone can add a tool to graphically edit it.

That said, we can't always calculate it. Sometimes there is need for padding, etc. The best solution is to work low-level and leave the helpers to another layer (ie: scripts).

map.json
```diff
@@ -0,0 +1,117 @@
...
+                "name": "_version",
+                "offset": 53,
+                "size": 1
+            },
+            {
+                "name": "save3_bool",
+                "offset": 54,
+                "size": 1,
+                "bit_offset": 0,
+                "is_bool": true
```
barbieri: why do we need `is_bool`? Either we have types for all of them, or none of them.

src/bin/sol-fbp-generator/main.c
```diff
@@ -755,6 +756,116 @@ collect_context_info(struct generate_context *ctx, struct fbp_data *data)
...
+                SOL_JSON_SCANNER_ARRAY_LOOP (&entries, &token, SOL_JSON_TYPE_OBJECT_START, reason) {
+                    uint32_t size = 0, offset = 0, bit_offset = 0;
+                    bool is_bool = false;
+                    char *name;
+
+                    SOL_JSON_SCANNER_OBJECT_LOOP_NEST (&entries, &token, &key, &value, reason) {
+                        if (SOL_JSON_TOKEN_STR_LITERAL_EQ(&key, "name")) {
+                            sol_json_token_remove_quotes(&value);
+                            name = strndupa(value.start,
+                                sol_json_token_get_size(&value));
```
barbieri: you don't need to strdupa() just to print it. Use `%.*s`, that way you can provide 2 arguments and limit the string to a number of bytes. If you convert it to a `sol_str_slice` we even have macros to help with that.

map.json
lpereira: We can't check if they're OK while loading. If two variables overlap, how can they be checked? I'd just do something similar to what C does to structs, but specifying only the sizes; offsets would be calculated by the tool. Considering that this will be edited by hand and values will be added and removed, this is very likely to happen.

src/bin/sol-fbp-runner/main.c
edersondisouza: Will ckeck it, tks!

map.json
edersondisouza: I've put it just to avoiding define `bit_size: 1` and checking if `bit_size` is really `1`. Maybe another name, like `bit_field`? I'm open to suggestions. That said, maybe would be interesting have a `type` property, which would define `size` automatically, addressing @lpereira concern. I'm not sure about it because user, right now, has to know size to calc offsets.

src/bin/sol-fbp-generator/main.c
edersondisouza: It's not just to print it, I save it to print after this loop. Would be cool if we had instead a `sol_json_token_get_slice`, though.

barbieri: you can save the token (it's a struct, fixed size/cost).

I agree with the `sol_json_token_get_slice()`, please add it :-)

map.json
barbieri: you can calculate if the variables overlap while loading. "save3_irange" starts at 70 and have 16 bytes, then it ends at 86.  If some other variable intercepts this segment, they conflict.

map.json
barbieri: I'd say always consider bits, but default them to `bit_offset=0` and `bit_size=8`. This way there is no need to check for `is_bool` and it will work in cases where one is storing packed data such as:
```c
struct my_struct {
    uint8_t a_bit:1;
    uint8_t b_value:7;
};
```

Then you specify `bit_offset` is an offset to the left of the starting byte offset and that `bit_size` must be smaller than `size * 8`. This way you can generically encode numbers in any amount of bits. Of course take care to unpack/pack those (as you did in IIO's code).

map.json
lpereira: Why can't, then, the offset be calculated by the tool? If unspecified, it could be calculated; if specified, it can be used (and perform this check).

src/bin/sol-fbp-generator/main.c
lpereira: sol_json_token_get_slice() was added in another PR that's pending review.

map.json
edersondisouza: @lpereira To calculate the offset as you say, we can trust on `SOL_JSON_SCANNER_OBJECT_LOOP_NEST ` returning objects on the order they were written, or it would be better to have an `order` item on json?

lpereira: It's just a lexical scanner, so, yes, the order that things will be processed are the order that they're specified in the JSON file. Since every field is an object inside an array, the order shouldn't be changed by other tools (beautifiers, etc) as well.

src/bin/sol-fbp-generator/main.c
```diff
@@ -1057,6 +1183,14 @@ parse_args(int argc, char *argv[])
                 return false;
             }
             break;
+        case 'm':
+            if (access(optarg, R_OK) == -1) {
```
vcgomes: I would prefer if the check is made when the file is actually used.

As there shouldn't be any serious consequences, I will leave this suggestion as optional.

src/bin/sol-fbp-runner/main.c
```diff
@@ -149,6 +172,180 @@ parse_args(int argc, char *argv[])
...
+                }
+            }
+        }
+
+        /* Add ptr_table guard element */
+        ptr_table_entry = sol_vector_append(&entries_vector);
+        SOL_NULL_CHECK_GOTO(ptr_table_entry, error);
+
+        entries_vector_size = sizeof(struct sol_str_table_ptr) * entries_vector.len;
+        map = calloc(sizeof(struct sol_memmap_map) + entries_vector_size, 1);
```
vcgomes: This should be `map = calloc(1, sizeof(struct sol_memmap_map) + entries_vector_size);`

vcgomes: Suggestion: what about making `struct sol_memmap_map` abstract and using something like 
`struct sol_memmap_map *sol_memmap_map_new(const char path, struct sol_str_table_ptr table[])` to create a new instance, then the only place where you would check for `USE_MEMMAP` would be there.

Would this work?

edersondisouza: So this code would be compiled even if not using memmap? For the hate of `ifdef`? =D
(Or are there other reasons I'm not seeing?)

src/bin/sol-fbp-generator/main.c
edersondisouza: I've just followed other options handling nearby =D

src/bin/sol-fbp-runner/main.c
vcgomes: I would classify the second as the strongest one ;-) but those are the two main reasons. The idea is to decrease our compile/test surface.

edersondisouza: @vcgomes I see, but I'm not sure how `struct sol_memmap_entry` fits this plan. And the header would exist even when `USE_MEMMAP` is undefined? Why not just make `sol_memmap_add_map` return `-ENOSYS` or `-ENOTSUP`?

vcgomes: Hmm. Yeah, the header wouldn't even exist if `USE_MEMMAP` is undefined. This complicates stuff.

If I may change my mind, having these `ifdef`s seems the cleanest solution indeed.

I will only suggest that the `-m` option to not exist if memmap support is disabled.

edersondisouza:
v2 on #745

PR #720, opened 2015-09-15
generator: Use the right type name for float options

ibriano:
Since the option type is float, not double, the generator was not
properly producing the options initialization for each field, resulting
in a considerable amount of warnings when running make check-fbp-bin


lpereira:
:+1:

ibriano:
Added a few more commits to fix check-bin-fbp when running with allmodconfig

bdilly:
series lgtm

ibriano:
Merged

PR #721, opened 2015-09-15
comms:http:server: Set minimal microhttpd's version

ceolin:
The code is using symbols that are available only in versions greater than 0.9.40

### DIFF
--- a/data/jsons/dependencies.json
+++ b/data/jsons/dependencies.json
@@ -450,6 +450,7 @@
       "dependency": "libmicrohttpd",
       "type": "pkg-config",
       "pkgname": "libmicrohttpd"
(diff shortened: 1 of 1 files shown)


ceolin:
@glima could you check it please ?

data/jsons/dependencies.json
```diff
@@ -450,6 +450,7 @@
       "dependency": "libmicrohttpd",
       "type": "pkg-config",
       "pkgname": "libmicrohttpd"
```
lpereira: Missing comma at the end of this line

glima: Indeed. With the local fix, the build now passes.

ceolin:
thanks guys, commited with the fix

PR #723, opened 2015-09-15
build: check-fbp-bin rule should place the mods in the end

dorileo:
Some compilers will not like to have the list of external modules close
to the cflags but will expect it to be with the ldflags (namely ubuntu's
gcc). This patch changes that and make everybody happier.


ibriano:
Ack

dorileo:
ok, applied.

PR #728, opened 2015-09-15
Fix some more samples and add them to build

bdilly:
There is a warning fixed by ibriano on pr #720


src/samples/flow/unix-socket/byte-writer.fbp
```diff
@@ -32,6 +32,6 @@
 # On a timeout (every second) create a byte and send to the socket.
 # See byte-reader.fbp
 
-Timer(timer:interval=1000) OUT -> INC Acc(int/accumulator:setup_value=val:0|min:0|max:254:step:1)
+Timer(timer:interval=1000) OUT -> INC Acc(int/accumulator:setup_value=val:0|min:0|max:254|step:1)
```
lpereira: Was this a syntax error as far as the parser was concerned?

bdilly: Yeah, a syntax error.
This sample wasn't running or building.

bdilly:
It's building fine on my machine, on semaphore and tipbot.
We're just adding as dependencies of these samples node types that are built as modules by default. Maybe you're buliding console as module or something like that?
I'll see this keyboard dependency.
Thanks

bdilly:
@ceolin 

Ok, it's passing on all these machines because they don't have enough dependencies to build http-server, I've just checked the log.

I'll send a v2 fixing keyboard dependency but if something else goes wrong, please paste a log so I can fix all issues.

Thanks

bdilly:
pr #733

PR #734, opened 2015-09-16
get_libname(): resolve possible symlink returned by dladdr().

glima:
At least of Fedora systems, /lib64 -> /usr/lib64 symlink exists and, if
returned there, it messes with the rest of the logic without realpath().


barbieri:
+1

glima:
Thanks, merging.

PR #735, opened 2015-09-16
Ensure make check{,-fbp,-fbp-bin} work on non default configurations

ibriano:
Fix the checks for allmodconfig, basically. Tested also with
allyesconfig and alldefconfig, all work. User configurations still have
the potential to break due to assumptions from the tests that certain
node-types will be available always.


dorileo:
LGTM
tested, works fine.

PR #740, opened 2015-09-16
build: [samples] add all used dependencies

dorileo:
Since now we can successfully run check + check-fbp + check-fbp-bin with
allmodconfig it's reasonable to also run make samples with that
configuration, for that we should name all the required dependencies.

We plan to make it "automatic" in a near future but for now it's ok to
to namely declare the deps;


ibriano:
Fine by me

bdilly:
Closing since it was merged already

PR #744, opened 2015-09-16
flow: support receiving scaped suboptions keys

bdilly:
Since conffiles parts are passed to these options parser functions,
they need to be able to handle double quotes after keynames.

Example of possible value passed to rgb_parse:
  "rgb": {
      "red": 255
  }

It would fail because there is a char different from space
between keyname (rgb) and ':'.


ibriano:
Ack

bdilly:
merged

PR #745, opened 2015-09-16
Memmap v2

edersondisouza:
v2:

Some doc;
Using `sol_arena` to simplify allocations and deallocations;
Using `sol_token_to_slice`
Fixes on build;
`store_only_val` fixes;

### DIFF
--- a/src/bin/sol-fbp-generator/main.c
+++ b/src/bin/sol-fbp-generator/main.c
@@ -57,6 +57,7 @@ static struct {
     const char *conf_file;
     const char *output_file;
     const char *export_symbol;
+    const char *memory_map_file;
 
     struct sol_ptr_vector json_files;
     struct sol_ptr_vector fbp_search_paths;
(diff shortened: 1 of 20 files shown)


edersondisouza:
As it appears that no-one reviewed, I force pushed a new version fixing coverity-spotted bugs.

src/modules/flow/persistence/persistence.c
```diff
@@ -40,8 +40,293 @@
...
+
+#ifdef USE_MEMMAP
+#include "sol-memmap-storage.h"
+#endif
+
+static bool once = true;
+
+#ifdef USE_FILESYSTEM
+static void
+test_fs(void)
```
lpereira: How difficult would it be to have at least the filesystem tests being executed by `make check`?

edersondisouza: I'm working on tests right now. I believe that filesystem and memmap shall be possible (EFIVARS needs some luck). Now that you asked, I'm working on fbp tests, do you think it's important to have C API tests as well? Or only fbp shall be enough?

lpereira: FBP tests should be enough, if they cover most of the underlying C API.

src/bin/sol-fbp-generator/main.c
```diff
@@ -757,6 +758,119 @@ collect_context_info(struct generate_context *ctx, struct fbp_data *data)
...
+                out("    .entries = {\n");
+
+                SOL_JSON_SCANNER_ARRAY_LOOP (&entries, &token, SOL_JSON_TYPE_OBJECT_START, reason) {
+                    uint32_t size = 0, offset = 0, bit_offset = 0, bit_size = 0;
+                    struct sol_str_slice name = { };
+
+                    SOL_JSON_SCANNER_OBJECT_LOOP_NEST (&entries, &token, &key, &value, reason) {
+                        if (SOL_JSON_TOKEN_STR_LITERAL_EQ(&key, "name")) {
+                            sol_json_token_remove_quotes(&value);
+                            name = sol_json_token_to_slice(&value);
```
lpereira: Might be a good idea to see if `name.len > 0` here.

src/bin/sol-fbp-generator/main.c
```diff
@@ -757,6 +758,119 @@ collect_context_info(struct generate_context *ctx, struct fbp_data *data)
...
+                    if (!size) {
+                        SOL_WRN("Invalid size for entry");
+                        goto error;
+                    }
+
+                    out("       SOL_MEMMAP_ENTRY_BIT_SIZE(\"%.*s\", %u, %u, %u, %u),\n",
+                        SOL_STR_SLICE_PRINT(name), offset, size, bit_offset, bit_size);
+                }
+            }
+        }
```
lpereira: Don't forget to check `reason`, as the JSON file might be malformed.

src/bin/sol-fbp-runner/main.c
```diff
@@ -149,12 +181,194 @@ parse_args(int argc, char *argv[])
...
+                sol_json_scanner_init_from_token(&entries, &value);
+
+                SOL_JSON_SCANNER_ARRAY_LOOP (&entries, &token, SOL_JSON_TYPE_OBJECT_START, reason) {
+                    uint32_t size = 0, offset = 0, bit_offset = 0, bit_size = 0;
+                    char *name = NULL;
+
+                    SOL_JSON_SCANNER_OBJECT_LOOP_NEST (&entries, &token, &key, &value, reason) {
+                        if (SOL_JSON_TOKEN_STR_LITERAL_EQ(&key, "name")) {
+                            sol_json_token_remove_quotes(&value);
+                            name = strndupa(value.start,
```
lpereira: I don't think it'll be a problem here, but please avoid calling `alloca()` inside a loop. It's good practice, as the allocated memory is tied to the call frame and the stack might overflow. For this case, since the JSON data is still in RAM, you might want to use a string slice and use `sol_arena_strdup_slice()`.

src/bin/sol-fbp-runner/main.c
```diff
@@ -149,12 +181,194 @@ parse_args(int argc, char *argv[])
...
+
+                    memmap_entry = calloc(sizeof(struct sol_memmap_entry), 1);
+                    SOL_NULL_CHECK_GOTO(memmap_entry, error);
+
+                    memmap_entry->offset = offset;
+                    memmap_entry->size = size;
+                    memmap_entry->bit_offset = bit_offset;
+                    memmap_entry->bit_size = bit_size;
+
+                    ptr_table_entry->key = sol_arena_strdup(str_arena, name);
```
lpereira: Might be a good idea to assert that `name` is not empty here.

src/bin/sol-fbp-runner/main.c
```diff
@@ -149,12 +181,194 @@ parse_args(int argc, char *argv[])
...
+        }
+
+        free(map);
+    }
+
+    sol_ptr_vector_clear(&memory_maps);
+}
+
+static bool
+load_memory_map_file(const char *json_file)
```
lpereira: This funciton is quite large and has way too many variables. Is it possible to split it into smaller functions?

src/lib/io/include/sol-memmap-storage.h
```diff
@@ -0,0 +1,273 @@
...
+
+    return sol_memmap_read_raw(name, &buf);
+}
+
+static inline int
+sol_memmap_write_double(const char *name, double value)
+{
+    CREATE_BUFFER(&value, false);
+
+    return sol_memmap_write_raw(name, &buf);
```
lpereira: Since these implementations are the same for `memmap`, `fs`, and `efivars`, wouldn't it be possible to use something such as x-macros to implement these functions only once?

src/bin/sol-fbp-runner/main.c
edersondisouza: Fair, but I won't be addressing it for v3 - there's a possibility that this json will be merged with conffile - if that happens, splitting will be performed ther. Now that you are here, do you have any opinion on that?

src/lib/io/include/sol-memmap-storage.h
edersondisouza: Fair. But where to define it? A different header included by these headers?

edersondisouza:
v3 on #758

PR #752, opened 2015-09-18
check-license fixes, use spdx format for BSD-3-Clause

anselmolsm:
spdx identifiers: http://spdx.org/licenses/

Hopefully the check-license script will be replaced by a new one (python?) in the future, for the sake of the sanity.


bdilly:
seems ok

anselmolsm:
applied, tks

PR #754, opened 2015-09-18
Use uname(2) on get_os_version.

glima:
Also moving it to a sol_util function, that is now more suitable.

### DIFF
--- a/src/lib/common/sol-platform.c
+++ b/src/lib/common/sol-platform.c
@@ -394,7 +394,13 @@ sol_platform_get_sw_version(void)
 SOL_API char *
 sol_platform_get_os_version(void)
 {
-    return sol_platform_impl_get_os_version();
+    char *out = NULL;
+
+    int r = sol_util_get_os_version(&out);
+
+    SOL_INT_CHECK(r, < 0, NULL);
+
+    return out;
 }
 
 void
--- a/src/shared/sol-util-contiki.c
+++ b/src/shared/sol-util-contiki.c
@@ -54,3 +54,10 @@ sol_util_timespec_get_realtime(struct timespec *t)
     errno = ENOSYS;
     return -1;
 }
+
+int
+sol_util_get_os_version(char **out)
+{
+    SOL_WRN("Not implemented");
+    return -ENOTSUP;
+}
--- a/src/shared/sol-util-linux.c
+++ b/src/shared/sol-util-linux.c
@@ -38,6 +38,7 @@
 #include <stdio.h>
 #include <sys/stat.h>
 #include <sys/types.h>
+#include <sys/utsname.h>
 #include <unistd.h>
 
 struct timespec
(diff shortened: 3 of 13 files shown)


src/lib/common/sol-platform.c
```diff
@@ -394,7 +394,13 @@ sol_platform_get_sw_version(void)
 SOL_API char *
 sol_platform_get_os_version(void)
 {
-    return sol_platform_impl_get_os_version();
+    char *out = NULL;
+
+    int r = sol_util_get_os_version(&out);
```
barbieri: what about us storing this string as a global, returning a `const char *` to that string and free that on `sol_platform_shutdown()`? It's easier to use since people don't need to free() it. And also will cache the results:

```c
static char *os_version;

SOL_API const char *
sol_platform_get_os_version(void)
{
    int r;
    char *out;
    if (os_version) return os_version;
    r = sol_util_get_os_version(&out);
    SOL_INT_CHECK(r, < 0, NULL);
    os_version = out;
    return os_version;
}
```

src/shared/sol-util-contiki.c
```diff
@@ -54,3 +54,10 @@ sol_util_timespec_get_realtime(struct timespec *t)
     errno = ENOSYS;
     return -1;
 }
+
+int
+sol_util_get_os_version(char **out)
+{
+    SOL_WRN("Not implemented");
```
barbieri: really? @ibriano and @zehortigoza do you know of some way to get OS version there? Maybe some macro?

src/shared/sol-util-linux.c
```diff
@@ -53,3 +54,19 @@ sol_util_timespec_get_realtime(struct timespec *t)
...
+int
+sol_util_get_os_version(char **out)
+{
+    struct utsname hostinfo;
+    int r;
+
+    r = uname(&hostinfo);
+    SOL_INT_CHECK(r, == -1, -errno);
+
+    r = asprintf(out, "%s %s", hostinfo.sysname, hostinfo.release);
```
barbieri: why add `sysname` in there?

src/shared/sol-util-riot.c
```diff
@@ -68,3 +68,9 @@ sol_util_timespec_get_realtime(struct timespec *t)
     return -1;
 #endif
 }
+
+int
+sol_util_get_os_version(char **out)
+{
+    return strdup(RIOT_VERSION);
```
barbieri: bug! should be: `*out = strdup(RIOT_VERSION)`

barbieri:
I don't think we need to keep it in `sol_util` just keep the implementations in `common` as before... the internal call doesn't add much.

src/lib/common/sol-platform.c
glima: Fair to me.

src/shared/sol-util-linux.c
glima: drugs, I guess. removing, then.

src/shared/sol-util-riot.c
glima: sorry, ouch

glima:
It was in sol_util_file before. I still things it's better there in util.

barbieri:
I have no idea why people did put it in util, I guess because that one was for `file`... the platforms itself already have `-impl-` files and OS-specific code in there. There is no gain to keep an internal `util` version other than confusion -- which one should I use? why 2? etc.

glima:
Ok, no big deal for me, leaving as -impl- and not util.

glima:
2nd version coming, thanks.

src/shared/sol-util-contiki.c
zehortigoza: CONTIKI_VERSION_STRING

PR #755, opened 2015-09-18

ibriano:

### DIFF
--- a/src/bin/sol-fbp-generator/main.c
+++ b/src/bin/sol-fbp-generator/main.c
@@ -815,22 +815,23 @@ generate(struct sol_vector *fbp_data_vector)
 
     if (!args.export_symbol) {
         out(
+            "static const struct sol_flow_node_type *root_type;\n"
             "static struct sol_flow_node *flow;\n"
             "\n"
             "static void\n"
             "startup(void)\n"
             "{\n"
-            "    const struct sol_flow_node_type *type;\n\n"
             "    initialize_types();\n"
-            "    type = create_0_root_type();\n"
-            "    if (!type)\n"
+            "    root_type = create_0_root_type();\n"
+            "    if (!root_type)\n"
             "        return;\n\n"
-            "    flow = sol_flow_node_new(NULL, NULL, type, NULL);\n"
+            "    flow = sol_flow_node_new(NULL, NULL, root_type, NULL);\n"
             "}\n\n"
             "static void\n"
             "shutdown(void)\n"
             "{\n"
             "    sol_flow_node_del(flow);\n"
+            "    sol_flow_node_type_del((struct sol_flow_node_type *)root_type);\n"
             "}\n\n"
             "SOL_MAIN_DEFAULT(startup, shutdown);\n");
     } else {
--- a/src/bin/sol-fbp-generator/main.c
+++ b/src/bin/sol-fbp-generator/main.c
@@ -634,7 +634,15 @@ generate_node_type_assignments(const struct fbp_data *data)
 
     SOL_VECTOR_FOREACH_IDX (&data->graph.nodes, n, i) {
         struct type_description *desc = n->user_data;
-        out("    nodes[%d].type = %s;\n", i, desc->symbol);
+        struct sol_str_slice module;
+        char *sep;
+
+        module = sol_str_slice_from_str(desc->name);
+        sep = strstr(module.data, "/");
+        if (sep)
+            module.len = sep - module.data;
+        out("    nodes[%d].type = *(const struct sol_flow_node_type **)sol_symbol_get(\"flow\", \"%.*s\", %s);\n", i,
+            SOL_STR_SLICE_PRINT(module), desc->symbol);
     }
 
     SOL_VECTOR_FOREACH_IDX (&data->declared_fbp_types, dec_type, i) {
@@ -683,6 +691,11 @@ struct generate_context {
     struct sol_vector types_to_initialize;
 };
 
+struct type_to_init {
+    struct sol_str_slice symbol;
+    struct sol_str_slice module;
+};
+
 static bool
 is_declared_type(struct fbp_data *data, const struct sol_str_slice name)
 {
@@ -719,6 +732,7 @@ collect_context_info(struct generate_context *ctx, struct fbp_data *data)
         struct type_description *desc;
         const char *sep;
         struct sol_str_slice name, module, symbol;
+        struct type_to_init *t = NULL;
 
         /* Need to go via descriptions to get the real resolved name,
          * after conffile pass. */
@@ -732,11 +746,10 @@ collect_context_info(struct generate_context *ctx, struct fbp_data *data)
 
         symbol = sol_str_slice_from_str(desc->symbol);
         if (!contains_slice(&ctx->types_to_initialize, symbol)) {
-            struct sol_str_slice *t;
             t = sol_vector_append(&ctx->types_to_initialize);
             if (!t)
                 return false;
-            *t = symbol;
+            t->symbol = symbol;
         }
 
         module = name;
@@ -744,6 +757,9 @@ collect_context_info(struct generate_context *ctx, struct fbp_data *data)
         if (sep) {
             module.len = sep - module.data;
         }
+
+        if (t)
+            t->module = module;
 
         if (!contains_slice(&ctx->modules, module)) {
             struct sol_str_slice *m;
@@ -762,17 +778,19 @@ generate(struct sol_vector *fbp_data_vector)
 {
     struct generate_context _ctx = {
         .modules = SOL_VECTOR_INIT(struct sol_str_slice),
-        .types_to_initialize = SOL_VECTOR_INIT(struct sol_str_slice),
+        .types_to_initialize = SOL_VECTOR_INIT(struct type_to_init),
     }, *ctx = &_ctx;
 
     struct fbp_data *data;
-    struct sol_str_slice *module, *symbol;
+    struct sol_str_slice *module;
+    struct type_to_init *type;
     uint16_t i;
     int r;
 
     out(
         "#include \"sol-flow.h\"\n"
-        "#include \"sol-flow-static.h\"\n");
+        "#include \"sol-flow-static.h\"\n"
+        "#include \"sol-modules.h\"\n");
 
     if (!args.export_symbol) {
         out("#include \"sol-mainloop.h\"\n");
@@ -802,13 +820,15 @@ generate(struct sol_vector *fbp_data_vector)
     out(
         "static void\n"
         "initialize_types(void)\n"
-        "{\n");
-    SOL_VECTOR_FOREACH_IDX (&ctx->types_to_initialize, symbol, i) {
+        "{\n"
+        "    const struct sol_flow_node_type **t;\n\n");
+    SOL_VECTOR_FOREACH_IDX (&ctx->types_to_initialize, type, i) {
         out(
-            "    if (%.*s->init_type)\n"
-            "        %.*s->init_type();\n",
-            SOL_STR_SLICE_PRINT(*symbol),
-            SOL_STR_SLICE_PRINT(*symbol));
+            "    t = sol_symbol_get(\"flow\", \"%.*s\", %.*s);\n"
+            "    if ((*t)->init_type)\n"
+            "        (*t)->init_type();\n",
+            SOL_STR_SLICE_PRINT(type->module),
+            SOL_STR_SLICE_PRINT(type->symbol));
     }
     out(
         "}\n\n");
--- a/src/lib/common/include/sol-modules.h
+++ b/src/lib/common/include/sol-modules.h
@@ -32,21 +32,15 @@
 
 #pragma once
 
-{{
-st.on_value("PLATFORM_LINUX", "y", "#define SOL_PLATFORM_LINUX 1", "")
-st.on_value("PLATFORM_RIOTOS", "y", "#define SOL_PLATFORM_RIOT 1", "")
-st.on_value("PLATFORM_CONTIKI", "y", "#define SOL_PLATFORM_CONTIKI 1", "")
-}}
+#include "sol-common-buildopts.h"
 
-{{
-st.on_value("LOG", "y", "#define SOL_LOG_ENABLED 1", "")
-}}
+#ifdef SOL_DYNAMIC_MODULES
+#define sol_symbol_get(_ns, _mod, _sym) sol_modules_get_symbol(_ns, _mod, #_sym)
 
-{{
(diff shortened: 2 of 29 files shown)


src/bin/sol-fbp-generator/main.c
```diff
@@ -634,7 +634,15 @@ generate_node_type_assignments(const struct fbp_data *data)
...
     SOL_VECTOR_FOREACH_IDX (&data->graph.nodes, n, i) {
         struct type_description *desc = n->user_data;
-        out("    nodes[%d].type = %s;\n", i, desc->symbol);
+        struct sol_str_slice module;
+        char *sep;
+
+        module = sol_str_slice_from_str(desc->name);
+        sep = strstr(module.data, "/");
+        if (sep)
+            module.len = sep - module.data;
```
lpereira: Shouldn't module.data be updated as well?

ibriano: On Fri, 18 Sep 2015 14:39:48 -0700, Leandro A. F. Pereira wrote:
> > @@ -634,7 +634,15 @@ generate_node_type_assignments(const struct fbp_data *data)
> >  
> >      SOL_VECTOR_FOREACH_IDX (&data->graph.nodes, n, i) {
> >          struct type_description *desc = n->user_data;
> > -        out("    nodes[%d].type = %s;\n", i, desc->symbol);
> > +        struct sol_str_slice module;
> > +        char *sep;
> > +
> > +        module = sol_str_slice_from_str(desc->name);
> > +        sep = strstr(module.data, "/");
> > +        if (sep)
> > +            module.len = sep - module.data;
> 
> Shouldn't module.data be updated as well?

No, why? It's already pointing at the beginning of the string as set by
sol_str_slice_from_str()

barbieri: no, as this is the part before "/" (you reduce module.len from original "a/b" to strlen("a")).

src/bin/sol-fbp-generator/main.c
```diff
@@ -634,7 +634,15 @@ generate_node_type_assignments(const struct fbp_data *data)
...
+        out("    nodes[%d].type = *(const struct sol_flow_node_type **)sol_symbol_get(\"flow\", \"%.*s\", %s);\n", i,
```
barbieri: I don't get why we need to derreference `*(x **)`. Also, shouldn't it check `NULL` in the case of symbol is missing (no module, etc)? It would be the only reason to keep the derreference.

If we're going to fail badly -- such as abort, then remove the derreference and provide a nice error message inside `sol_symbol_get()`

src/bin/sol-fbp-generator/main.c
```diff
@@ -802,35 +820,38 @@ generate(struct sol_vector *fbp_data_vector)
...
-    SOL_VECTOR_FOREACH_IDX (&ctx->types_to_initialize, symbol, i) {
+        "{\n"
+        "    const struct sol_flow_node_type **t;\n\n");
+    SOL_VECTOR_FOREACH_IDX (&ctx->types_to_initialize, type, i) {
         out(
-            "    if (%.*s->init_type)\n"
-            "        %.*s->init_type();\n",
-            SOL_STR_SLICE_PRINT(*symbol),
-            SOL_STR_SLICE_PRINT(*symbol));
+            "    t = sol_symbol_get(\"flow\", \"%.*s\", %.*s);\n"
```
barbieri: `if (t)` to check missing symbols?

src/lib/common/include/sol-modules.h
```diff
@@ -0,0 +1,46 @@
...
+ * (INCLUDING NEGLIGENCE OR OTHERWISE) ARISING IN ANY WAY OUT OF THE USE
+ * OF THIS SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
+ */
+
+#pragma once
+
+#include "sol-common-buildopts.h"
+
+#ifdef SOL_DYNAMIC_MODULES
+#define sol_symbol_get(_ns, _mod, _sym) sol_modules_get_symbol(_ns, _mod, #_sym)
```
barbieri: please document it, we're going to do a docummentation effort soon -- but the more we do before, the better.

src/lib/common/include/sol-modules.h
```diff
@@ -0,0 +1,46 @@
...
+
+int sol_modules_init(void);
+void sol_modules_shutdown(void);
```
barbieri: these shouldn't be in the public header, rather go hidden somewhere private.

src/lib/common/sol-mainloop.c
```diff
@@ -104,6 +105,12 @@ sol_init(void)
 
     sol_log_domain_init_level(SOL_LOG_DOMAIN);
 
+#ifdef MODULES
```
barbieri: check `SOL_DYNAMIC_MODULES` to keep it consistent (eventually change it in other places as well).

src/bin/sol-fbp-generator/main.c
ibriano: On Fri, 18 Sep 2015 14:46:05 -0700, Gustavo Sverzut Barbieri wrote:
> > @@ -634,7 +634,15 @@ generate_node_type_assignments(const struct fbp_data *data)
> >  
> >      SOL_VECTOR_FOREACH_IDX (&data->graph.nodes, n, i) {
> >          struct type_description *desc = n->user_data;
> > -        out("    nodes[%d].type = %s;\n", i, desc->symbol);
> > +        struct sol_str_slice module;
> > +        char *sep;
> > +
> > +        module = sol_str_slice_from_str(desc->name);
> > +        sep = strstr(module.data, "/");
> > +        if (sep)
> > +            module.len = sep - module.data;
> > +        out("    nodes[%d].type = *(const struct sol_flow_node_type **)sol_symbol_get(\"flow\", \"%.*s\", %s);\n", i,
> 
> I don't get why we need to derreference `*(x **)`. Also, shouldn't it check `NULL` in the case of symbol is missing (no module, etc)? It would be the only reason to keep the derreference.

Because they symbol we are asking for is a pointer to something else,
and dlsym() will give you the address where the symbol is stored. If we
were asking for the struct directly, it would be a plain pointer.

> 
> If we're going to fail badly -- such as abort, then remove the derreference and provide a nice error message inside `sol_symbol_get()`

Nah, I should be less lazy and add the proper checks in the generator.

src/lib/common/sol-mainloop.c
barbieri: also, I've asked @ceolin to review the init/shutdowns... he will change it so we always have the calls, but in some cases they are empty. You may choose to take this approach right now and save him some work.

src/bin/sol-fbp-generator/main.c
ibriano: On Fri, 18 Sep 2015 14:47:44 -0700, Gustavo Sverzut Barbieri wrote:
> >          out(
> > -            "    if (%.*s->init_type)\n"
> > -            "        %.*s->init_type();\n",
> > -            SOL_STR_SLICE_PRINT(*symbol),
> > -            SOL_STR_SLICE_PRINT(*symbol));
> > +            "    t = sol_symbol_get(\"flow\", \"%.*s\", %.*s);\n"
> 
> `if (t)` to check missing symbols?

Yes, but I ended up wondering if that whole thing shouldn't be
re-adapted to consider that symbols might be missing, and do the
symbol_get thing only in one place. I'll look into that after.

src/lib/common/sol-modules.c
```diff
@@ -0,0 +1,222 @@
...
+        free(ns->name);
+    }
+    sol_vector_clear(cache);
+}
+
+static int
+get_internal_symbol(const char *symbol_name, void **symbol)
+{
+    *symbol = dlsym(RTLD_DEFAULT, symbol_name);
+    if (!*symbol)
```
barbieri: Please `SOL_WRN()` with `dlerror()`. Be verbose so we can easily spot problems.

If you think `WRN` is too much, at least `DBG` level.

src/lib/common/sol-modules.c
```diff
@@ -0,0 +1,222 @@
...
+        ret = sol_util_get_rootdir(rootdir, sizeof(rootdir));
+        SOL_INT_CHECK(ret, >= (int)sizeof(rootdir), NULL);
+    }
+
+    ret = snprintf(path, sizeof(path), "%s%s%s/%s.so", rootdir, MODULESDIR, nspace, modname);
+    SOL_INT_CHECK(ret, >= (int)sizeof(path), NULL);
+
+    SOL_INF("Loading module '%s'", path);
+
+    return dlopen(path, RTLD_LAZY | RTLD_LOCAL | RTLD_NODELETE);
```
barbieri: `SOL_WRN + path + dlerror()` if it failed.

src/lib/common/sol-modules.c
```diff
@@ -0,0 +1,222 @@
...
+    struct module_cache *mod;
+
+    ns = get_namespace(nspace);
+    SOL_NULL_CHECK(ns, -ENOMEM);
+
+    mod = get_module(ns, modname);
+    SOL_NULL_CHECK(mod, -ENOMEM);
+
+    *symbol = dlsym(mod->handle, symbol_name);
+    SOL_NULL_CHECK(*symbol, -ENOENT);
```
barbieri: `SOL_WRN + path + symbol_name + dlerror()`. You can recreate `path` from `nspace` and `modname`

src/lib/common/sol-modules.c
```diff
@@ -0,0 +1,222 @@
...
+
+    return 0;
+}
+
+SOL_API int
+sol_modules_init(void)
+{
+    modules_init_count++;
```
barbieri: we shouldn't need this counter. This `sol_modules_init()` should be internal and only called by `sol_init()` that already takes care of doing single-init.

src/samples/flow/c-api/lowlevel.c
```diff
@@ -142,7 +136,7 @@ startup(void)
     nodes[1 /* logic */].type = SOL_FLOW_NODE_TYPE_CUSTOM_NODE_TYPES_LOGIC;
     nodes[2 /* writer */].type = SOL_FLOW_NODE_TYPE_CUSTOM_NODE_TYPES_WRITER;
 #ifdef SOL_FLOW_NODE_TYPE_CONSOLE_DEFINED
```
barbieri: we don't need this `ifdef` here, right? Or this ifdef should also cover `#include "sol-flow/console.h"`

src/samples/flow/compass/Makefile
```diff
@@ -1,7 +1,2 @@
 sample-$(FLOW_COMPASS_LSM303_SAMPLE) += compass-lsm303
 sample-compass-lsm303-$(FLOW_COMPASS_LSM303_SAMPLE) := lsm303.fbp
-sample-compass-lsm303-$(FLOW_COMPASS_LSM303_SAMPLE)-deps := \
-	flow/accelerometer.mod \
-	flow/compass.mod \
-	flow/console.mod \
-	flow/magnetometer.mod
```
barbieri: hum... we still need the module, did you remove these dependencies just because we don't need them to be built before?

src/samples/flow/c-api/lowlevel.c
ibriano: On Fri, 18 Sep 2015 15:01:33 -0700, Gustavo Sverzut Barbieri wrote:
> > @@ -142,7 +136,7 @@ startup(void)
> >      nodes[1 /* logic */].type = SOL_FLOW_NODE_TYPE_CUSTOM_NODE_TYPES_LOGIC;
> >      nodes[2 /* writer */].type = SOL_FLOW_NODE_TYPE_CUSTOM_NODE_TYPES_WRITER;
> >  #ifdef SOL_FLOW_NODE_TYPE_CONSOLE_DEFINED
> 
> we don't need this `ifdef` here, right? Or this ifdef should also cover `#include "sol-flow/console.h"`

It's defined in console.h, and it's used to keep the sample working even
if the console module was disabled.

src/samples/flow/compass/Makefile
ibriano: On Fri, 18 Sep 2015 15:03:09 -0700, Gustavo Sverzut Barbieri wrote:
> > @@ -1,7 +1,2 @@
> >  sample-$(FLOW_COMPASS_LSM303_SAMPLE) += compass-lsm303
> >  sample-compass-lsm303-$(FLOW_COMPASS_LSM303_SAMPLE) := lsm303.fbp
> > -sample-compass-lsm303-$(FLOW_COMPASS_LSM303_SAMPLE)-deps := \
> > -	flow/accelerometer.mod \
> > -	flow/compass.mod \
> > -	flow/console.mod \
> > -	flow/magnetometer.mod
> 
> hum... we still need the module, did you remove these dependencies just because we don't need them to be built before?

Ah, yes, in order to keep build dependencies correct we need to keep all
these. Guess I'll just scrap those patches.

src/samples/flow/c-api/lowlevel.c
barbieri: in that case `#include "sol-flow/console.h"`  will fail since it doesn't exist.

Internally, we should have a way to check it. External apps should do something like `AC_HEADER()` to check for it.

src/samples/flow/compass/Makefile
barbieri: I don't think so, you don't need them to compile your code -- I think. Linking will happen at a later phase.

Try to make the sample/module before the module (explicit target), see if dependencies are pulled in.

ibriano: On Fri, 18 Sep 2015 15:08:53 -0700, Gustavo Sverzut Barbieri wrote:
> > @@ -1,7 +1,2 @@
> >  sample-$(FLOW_COMPASS_LSM303_SAMPLE) += compass-lsm303
> >  sample-compass-lsm303-$(FLOW_COMPASS_LSM303_SAMPLE) := lsm303.fbp
> > -sample-compass-lsm303-$(FLOW_COMPASS_LSM303_SAMPLE)-deps := \
> > -	flow/accelerometer.mod \
> > -	flow/compass.mod \
> > -	flow/console.mod \
> > -	flow/magnetometer.mod
> 
> I don't think so, you don't need them to compile your code -- I think. Linking will happen at a later phase.
> 
> Try to make the sample/module before the module (explicit target), see if dependencies are pulled in.

We don't need them linking, but we do need to fix up the build to make
the dependency clear. Right now I can build the samples and the modules
don't get built.

barbieri:
I like this patch overall and solves a long-standing problem... it's just what I imagined :-)

What I'm still unsure is the error handling an `sol_symbol_get()` signature. Maybe make it receive the pointer to store the symbol as parameter and return `int err`, then change all cases to `SOL_INT_CHECK()`... or since the call is actually a macro make it take a `goto label` and make the error handling always be a `goto`.

PR #756, opened 2015-09-18
http-client basic types

ceolin:
- The counterpart of the http-server's basic types
- Another double to str function (but this allocate the necessary memory)
- Small fix on keyboar node

### DIFF
--- a/src/lib/parsers/include/sol-json.h
+++ b/src/lib/parsers/include/sol-json.h
@@ -433,6 +433,8 @@ char *sol_json_escape_string(const char *str, char *buf, size_t len) SOL_ATTR_NO
 
 int sol_json_double_to_str(const double value, char *buf, size_t buf_len);
 
+int sol_json_double_to_str_alloc(const double value, char **buf);
+
 /**
  * @}
  */
--- a/src/lib/parsers/sol-json.c
+++ b/src/lib/parsers/sol-json.c
@@ -522,6 +522,8 @@ sol_json_double_to_str(const double value, char *buf, size_t len)
     struct lconv *lc = localeconv();
 #endif
 
+    SOL_NULL_CHECK(buf, -EINVAL);
+
     ret = snprintf(buf, len, "%f", value);
     if (ret < 0 || ret > (int)len)
         return -ENOMEM;
--- a/src/lib/parsers/sol-json.c
+++ b/src/lib/parsers/sol-json.c
@@ -512,25 +512,15 @@ sol_json_escape_string(const char *str, char *buf, size_t len)
     return out;
 }
 
-SOL_API int
-sol_json_double_to_str(const double value, char *buf, size_t len)
+static void
+double_to_str(char *buf, size_t len)
 {
-    int ret;
-    char *decimal_point;
-
 #ifdef HAVE_LOCALE
     struct lconv *lc = localeconv();
-#endif
+    char *decimal_point;
 
-    SOL_NULL_CHECK(buf, -EINVAL);
-
-    ret = snprintf(buf, len, "%f", value);
-    if (ret < 0 || ret > (int)len)
-        return -ENOMEM;
-
-#ifdef HAVE_LOCALE
     if (lc->decimal_point && streq(lc->decimal_point, "."))
-        return 0;
+        return;
 
 
     if ((decimal_point = strstr(buf, lc->decimal_point))) {
@@ -538,9 +528,37 @@ sol_json_double_to_str(const double value, char *buf, size_t len)
         char *fraction = decimal_point + decimal_len;
         *decimal_point = '.';
 
-        memmove(decimal_point + 1, fraction, (buf + ret + 1) - fraction);
+        memmove(decimal_point + 1, fraction, (buf + len + 1) - fraction);
     }
 #endif
+}
 
+SOL_API int
+sol_json_double_to_str(const double value, char *buf, size_t len)
+{
+    int ret;
+
+    SOL_NULL_CHECK(buf, -EINVAL);
+
+    ret = snprintf(buf, len, "%f", value);
+    if (ret < 0 || ret > (int)len)
+        return -ENOMEM;
+
+    double_to_str(buf, (size_t)ret);
+    return 0;
+}
+
+SOL_API int
+sol_json_double_to_str_alloc(const double value, char **buf)
+{
+    int ret;
+
+    SOL_NULL_CHECK(buf, -EINVAL);
+
+    ret = asprintf(buf, "%f", value);
+    if (ret < 0)
+        return -ENOMEM;
+
+    double_to_str(*buf, (size_t)ret);
     return 0;
 }
--- a/src/modules/flow/http-client/http-client.c
+++ b/src/modules/flow/http-client/http-client.c
@@ -354,4 +354,177 @@ string_open(struct sol_flow_node *node, void *data, const struct sol_flow_node_o
 
     return common_open(mdata, opts->url);
 }
+
+/*
+ * --------------------------------- irange node -----------------------------
+ */
+static void
+int_request_finished(void *data,
+    const struct sol_http_client_connection *connection,
+    struct sol_http_response *response)
+{
+    struct http_data *mdata = data;
+    struct sol_irange irange;
+
+    if (check_response(mdata, connection, response) < 0)
+        return;
+
+    if (streq(response->content_type, "application/json")) {
+        struct sol_json_scanner scanner;
+        struct sol_json_token token, key, value;
+        enum sol_json_loop_reason reason;
+
+        sol_json_scanner_init(&scanner, response->content.data, response->content.used);
+        SOL_JSON_SCANNER_OBJECT_LOOP (&scanner, &token, &key, &value, reason) {
+            struct sol_json_scanner sub_scanner;
+
+            sol_json_scanner_init(&sub_scanner, value.start, value.end - value.start);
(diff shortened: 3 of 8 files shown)


src/modules/flow/http-client/http-client.c
```diff
@@ -0,0 +1,702 @@
...
+        return;
+
+    if (streq(response->content_type, "application/json")) {
+        struct sol_json_scanner scanner;
+        struct sol_json_token token, key, value;
+        enum sol_json_loop_reason reason;
+
+        sol_json_scanner_init(&scanner, response->content.data, response->content.used);
+        SOL_JSON_SCANNER_OBJECT_LOOP (&scanner, &token, &key, &value, reason) {
+            if (!strncasecmp("true", value.start, value.end - value.start))
```
lpereira: JSON has boolean types, why not use them?

src/modules/flow/http-client/http-client.c
```diff
@@ -0,0 +1,702 @@
...
+        return;
+
+    if (streq(response->content_type, "application/json")) {
+        struct sol_json_scanner scanner;
+        struct sol_json_token token, key, value;
+        enum sol_json_loop_reason reason;
+
+        sol_json_scanner_init(&scanner, response->content.data, response->content.used);
+        SOL_JSON_SCANNER_OBJECT_LOOP (&scanner, &token, &key, &value, reason) {
+            result = strndup(value.start + 1, value.end - value.start - 2);
```
lpereira: This will leak memory.

src/lib/parsers/include/sol-json.h
```diff
@@ -433,6 +433,8 @@ char *sol_json_escape_string(const char *str, char *buf, size_t len) SOL_ATTR_NO
 
 int sol_json_double_to_str(const double value, char *buf, size_t buf_len);
 
+int sol_json_double_to_str_alloc(const double value, char **buf);
```
barbieri: -1 to this call. Even though float strings can be large, they have a relative small maximum size and by keeping those strings in the stack will avoid memory allocation, fragmentation etc.

src/lib/parsers/sol-json.c
```diff
@@ -512,23 +512,15 @@ sol_json_escape_string(const char *str, char *buf, size_t len)
     return out;
 }
 
-SOL_API int
-sol_json_double_to_str(const double value, char *buf, size_t len)
+static void
+double_to_str(char *buf, size_t len)
```
barbieri: bad API, it doesn't receive a double, so the name makes no sense. Make this one receive a double and do the `snprintf()` here.

src/lib/parsers/sol-json.c
```diff
@@ -536,9 +528,37 @@ sol_json_double_to_str(const double value, char *buf, size_t len)
...
+}
+
+SOL_API int
+sol_json_double_to_str(const double value, char *buf, size_t len)
+{
+    int ret;
+
+    SOL_NULL_CHECK(buf, -EINVAL);
+
+    ret = snprintf(buf, len, "%f", value);
```
barbieri: keep in mind `%f` may be a bad choice to encode doubles:

```c
printf("%f\n", 1e-3); // 0,001000 - OK
printf("%f\n", 1e-30); // 0,000000 - OUCH
printf("%g\n", 1e-3); // 0,001 - OK
printf("%g\n", 1e-30); // 1e-30 - OK
```

src/lib/parsers/sol-json.c
barbieri: as for the size, I recall it was 66 bytes for the number + two for exponent and sign encoding plus 3 for the actual exponent value, then one for the trailing null byte. That's: 66+2+3 = 71. Make it 100 and should be enough.

src/modules/flow/http-client/http-client.c
barbieri: agree, use JSON's native type.

I'd say you should also check for `key` being a known string.

src/modules/flow/http-client/http-client.c
```diff
@@ -0,0 +1,702 @@
...
+
+    r = sol_flow_packet_get_boolean(packet, &b);
+    SOL_INT_CHECK(r, < 0, r);
+
+    sol_http_param_init(&params);
+    if (!(sol_http_param_add(&params,
+        SOL_HTTP_REQUEST_PARAM_HEADER("Accept", "application/json"))) ||
+        !(sol_http_param_add(&params,
+        SOL_HTTP_REQUEST_PARAM_POST_FIELD("value", b ? "true" : "false")))) {
+        SOL_WRN("Failed to set query params");
```
barbieri: always try to provide more information such as the URL and so on. If you have multiple nodes doing requests, getting this message won't help you debug.

src/modules/flow/http-client/http-client.c
barbieri: indeed, if more keys were given, then it would leak. As for boolean, we should check for a specific key.

Also we shouldn't duplicate the memory and for JSON we do have a call to remove quotes and I've asked @bdilly to add one to also handle unescape of the string if needed (ie if it contains a `\'`).

src/modules/flow/http-client/http-client.c
```diff
@@ -0,0 +1,702 @@
...
+    return 0;
+}
+
+static int
+string_open(struct sol_flow_node *node, void *data, const struct sol_flow_node_options *options)
+{
+    struct http_data *mdata = data;
+    struct sol_flow_node_type_http_client_string_options *opts =
+        (struct sol_flow_node_type_http_client_string_options *)options;
+    mdata->node = node;
```
barbieri: you don't need to store the node. You can give the node pointer to the callbacks and later retrieve `mdata` from the node, that will save you a pointer for every node, and you can remove this call.

src/modules/flow/http-client/http-client.c
```diff
@@ -0,0 +1,702 @@
...
+    r = -ENOMEM;
+
+    if (asprintf(&val, "%d", value.val) == -1)
+        goto error;
+    if (asprintf(&min, "%d", value.min) == -1)
+        goto min_error;
+    if (asprintf(&max, "%d", value.max) == -1)
+        goto max_error;
+    if (asprintf(&step, "%d", value.step) == -1)
+        goto step_error;
```
barbieri: remove all `asprintf()`, do not allocate memory :-)

src/modules/flow/http-client/http-client.json
```diff
@@ -0,0 +1,195 @@
...
+          "name": "GET"
+        },
+	{
+	  "data_type": "boolean",
+	  "description": "Post into the url",
+          "methods": {
+            "process": "boolean_post_process"
+          },
+          "name": "POST"
+        }
```
barbieri: add an `URL` port for all nodes... it's simple and will allow you to change the URL during runtime (ie: from persistence nodes)

src/lib/parsers/include/sol-json.h
ceolin: i know that, that is the reason i add a new function instead of replace the other one

src/lib/parsers/sol-json.c
ceolin: this is an internal function, not exported and used internally by sol_json_double_str. If it's still a problem i don't see any problem in rename it.

For the string size, i did not know the necessary size and as the function is using %f in the snprintf, 71 bytes were not enough to convert DBL_MAX.

src/lib/parsers/sol-json.c
ceolin: ack

src/modules/flow/http-client/http-client.c
ceolin: ack

src/modules/flow/http-client/http-client.c
ceolin: Yep, it's being only used in a debug's message.

src/modules/flow/http-client/http-client.c
ceolin: i had done this in a previous commit, but someone had me asked to use asprintf() and I've done the same here. 

I agree with the change, i'll also review the usage of this in the other patch.

src/modules/flow/http-client/http-client.c
ceolin: ack

src/modules/flow/http-client/http-client.c
ceolin: i'll use. The key here is the server path, which i do not know (at least not easily)

src/lib/parsers/sol-json.c
ceolin: I found a problem with it. 

printf("%g", DBL_MIN); //2.22507e-308

if you try to use the returned string with strtod it will fail due an underflow :(
Well, i'm going to think how to make it works.

src/modules/flow/http-client/http-client.json
ceolin: ack

src/lib/parsers/sol-json.c
barbieri: even if it's internal, name matters.

As for 71... then check the required size, you can round up to make sure.

src/modules/flow/http-client/http-client.c
barbieri: should be in `mdata`, no?

ceolin:
closed, fixes on https://github.com/solettaproject/soletta/pull/773

PR #757, opened 2015-09-21
Add HTTP generic nodes.

cabelitos:
This patch introduces two new HTTP node types, one node
is responsible for making a HTTP GET request and output
its response as string and the another one will output
its response as a blob.

### DIFF
--- a/src/modules/flow/http/http.c
+++ b/src/modules/flow/http/http.c
@@ -1,0 +1,245 @@
+/*
+ * This file is part of the Soletta Project
+ *
+ * Copyright (C) 2015 Intel Corporation. All rights reserved.
+ *
+ * Redistribution and use in source and binary forms, with or without
+ * modification, are permitted provided that the following conditions
+ * are met:
+ *
+ *   * Redistributions of source code must retain the above copyright
+ *     notice, this list of conditions and the following disclaimer.
+ *   * Redistributions in binary form must reproduce the above copyright
+ *     notice, this list of conditions and the following disclaimer in
+ *     the documentation and/or other materials provided with the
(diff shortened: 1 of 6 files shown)


cabelitos:
Closing this one I will move this nodes to the http-client module.

src/modules/flow/http/http.c
```diff
@@ -0,0 +1,245 @@
...
+    SOL_NULL_CHECK(mdata->url, -ENOMEM);
+    return 0;
+}
+
+static void
+get_string_process(struct sol_http_response *response, struct http_data *mdata)
+{
+    char *result;
+
+    SOL_DBG("String process");
```
barbieri: make this debug more useful and print the `%.*s` with content.used and content.data

src/modules/flow/http/http.c
```diff
@@ -0,0 +1,245 @@
...
+    sol_flow_send_string_packet(mdata->node,
+        SOL_FLOW_NODE_TYPE_HTTP_GET_STRING__OUT__OUT, result);
+}
+
+static void
+get_blob_process(struct sol_http_response *response, struct http_data *mdata)
+{
+    struct sol_blob *blob;
+
+    SOL_DBG("Blob process");
```
barbieri: as above, it would be useful to either drop this debug or make it print something useful like the content.used (size)

src/modules/flow/http/http.c
cabelitos: Ok

src/modules/flow/http/http.c
cabelitos: Ack

PR #758, opened 2015-09-21
Memmap v3

edersondisouza:
v3:
Now with proper tests. Some other issues pointed by @lpereira were also addressed. Issues found by tests addressed as well =D
To have the tests, some modifications on `run-fbp-tests` were necessary: TEST-PRECONDITION and TEST-EXTRA-ARGS. Tests are disabled for compile, as they touch file system, and for valgrind, as double mappings fail on it (some NaNs appear).

v2:

Some doc;
Using sol_arena to simplify allocations and deallocations;
Using sol_token_to_slice
Fixes on build;
store_only_val fixes;

### DIFF
--- a/src/bin/sol-fbp-generator/main.c
+++ b/src/bin/sol-fbp-generator/main.c
@@ -57,6 +57,7 @@ static struct {
     const char *conf_file;
     const char *output_file;
     const char *export_symbol;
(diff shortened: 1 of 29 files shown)


edersondisouza:
Will reopen as soon as I get license issue solved.

PR #759, opened 2015-09-21
strings: add string/slice node, with tests.

glima:

### DIFF
--- a/src/modules/flow/string/string-ascii.c
+++ b/src/modules/flow/string/string-ascii.c
@@ -94,6 +94,8 @@ get_string_by_port(const struct sol_flow_packet *packet,
     free(mdata->string[port]);
 
     mdata->string[port] = strdup(in_value);
+    if (!mdata->string[port])
+        return false;
 
     if (!mdata->string[0] || !mdata->string[1])
         return false;
@@ -307,6 +309,8 @@ string_slice_input(struct sol_flow_node *node,
 
     free(mdata->str);
     mdata->str = strdup(in_value);
+    if (!mdata->str)
+        return -ENOMEM;
 
     return slice_do(mdata);
 }
--- a/src/modules/flow/string/string-ascii.c
+++ b/src/modules/flow/string/string-ascii.c
@@ -236,6 +236,130 @@ string_compare(struct sol_flow_node *node,
         SOL_FLOW_NODE_TYPE_STRING_COMPARE__OUT__OUT, result);
 }
 
+struct string_slice_data {
+    struct sol_flow_node *node;
+    char *str;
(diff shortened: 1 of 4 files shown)


src/modules/flow/string/string-ascii.c
```diff
@@ -94,6 +94,8 @@ get_string_by_port(const struct sol_flow_packet *packet,
     free(mdata->string[port]);
 
     mdata->string[port] = strdup(in_value);
+    if (!mdata->string[port])
```
barbieri: `SOL_NUL_CHECK()` to make it verbose on errors?

src/modules/flow/string/string-ascii.c
```diff
@@ -236,6 +238,132 @@ string_compare(struct sol_flow_node *node,
...
+    struct sol_str_slice slice;
+    int32_t start = mdata->idx[0], end = mdata->idx[1],
+        len = strlen(mdata->str);
+
+#define BOUND_CHECK(_val) \
+    do { \
+        if (_val >= 0) { \
+            if (!len && _val) \
+                _val = 0; \
+            else if (_val > len) \
```
barbieri: This is the generic version of the previous test. Keep only this one.

src/modules/flow/string/string-ascii.c
glima: Next commit.

src/modules/flow/string/string-ascii.c
glima: Fair, thanks.

src/modules/flow/string/string-ascii.c
```diff
@@ -236,6 +238,132 @@ string_compare(struct sol_flow_node *node,
...
+                _val = len; \
+        } else if (_val < -len) { \
+            _val = 0; \
+        } else { \
+            _val = len + _val; \
+        } \
+    } while (0)
```
barbieri: I'd say don't create a macro, but a `static inline` function in `sol-util.h` to clamp a number within 2 values.

Then you should compute the value and clamp it:
```c
static inline int32_t
sol_util_int32_clamp(int32_t start, int32_t end, int32_t value)
{
    if (value < start)
        return start;
    if (value > end)
        return end;
    return value;
}

static int
slice_do(...) {
    ...
    if (start < 0) start = len + start;
    if (end < 0) end = len + end;
    start = sol_util_int32_clamp(0, len, start);
    end = sol_util_int32_clamp(0, len, end);
```

glima: Okay, I'll test with that.

glima:
v2 coming.

PR #760, opened 2015-09-21
Memmap v3

edersondisouza:
v3:
Now with proper tests. Some other issues pointed by @lpereira were also addressed. Issues found by tests addressed as well =D
To have the tests, some modifications on run-fbp-tests were necessary: TEST-PRECONDITION and TEST-EXTRA-ARGS. Tests are disabled for compile, as they touch file system, and for valgrind, as double mappings fail on it (some NaNs appear).

v2:

Some doc;
Using sol_arena to simplify allocations and deallocations;
Using sol_token_to_slice
Fixes on build;
store_only_val fixes;

### DIFF
--- a/src/bin/sol-fbp-generator/main.c
+++ b/src/bin/sol-fbp-generator/main.c
@@ -57,6 +57,7 @@ static struct {
     const char *conf_file;
     const char *output_file;
     const char *export_symbol;
+    const char *memory_map_file;
 
     struct sol_ptr_vector json_files;
     struct sol_ptr_vector fbp_search_paths;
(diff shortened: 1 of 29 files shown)


src/bin/sol-fbp-generator/main.c
```diff
@@ -757,6 +758,135 @@ collect_context_info(struct generate_context *ctx, struct fbp_data *data)
...
+            out("static const struct sol_memmap_map _memmap%d = {\n", (*elements)++);
+
+            SOL_JSON_SCANNER_OBJECT_LOOP_NEST (&scanner, &token, &key, &value, reason) {
+
+                if (SOL_JSON_TOKEN_STR_LITERAL_EQ(&key, "path")) {
+                    sol_json_token_remove_quotes(&value);
+                    path = sol_json_token_to_slice(&value);
+                } else if (SOL_JSON_TOKEN_STR_LITERAL_EQ(&key, "version")) {
+                    if (sol_json_token_get_uint32(&value, &version) < 0) {
+                        SOL_WRN("Couldn't get memory map version");
```
barbieri: it would be nice to print more context in the error messages, making it easier to debug and fix errors. Here I'd say we should keep the json offset in mind as well as providing some "path" in the JSON hierarchy (not to be confused with `path` key above -- however if there is a `path` slice, you can print it to add even more context).

src/bin/sol-fbp-runner/main.c
```diff
@@ -149,12 +181,208 @@ parse_args(int argc, char *argv[])
...
+                        }
+                        if (!size) {
+                            SOL_WRN("Invalid size for entry");
+                            goto error;
+                        }
+
+                        ptr_table_entry = sol_vector_append(&entries_vector);
+                        SOL_NULL_CHECK_GOTO(ptr_table_entry, error);
+
+                        memmap_entry = calloc(sizeof(struct sol_memmap_entry), 1);
```
barbieri: change the order of parameters, to be consistent with other paths in your code.

and move `if (!name.len)` befre this `calloc()` and `sol_vector_append()` (fail as soon as possible).

src/bin/sol-fbp-runner/main.c
```diff
@@ -180,6 +408,17 @@ startup(void *data)
         }
     }
 
+    if (args.memory_map_file) {
+#ifdef USE_MEMMAP
+        if (!load_memory_map_file(args.memory_map_file)) {
+            fprintf(stderr, "Could not load memory map file\n");
```
barbieri: `%s` + args.memory_map_file... make messages more verbose. also prefix it with "error:" so tools that offer highlight will change the color of lines (ie: emacs).

src/lib/io/include/sol-memmap-storage.h
```diff
@@ -0,0 +1,273 @@
...
+ * Writes buffer contents to storage.
+ *
+ * @param name name of property. must be present in one of maps previoulsy
+ * added via @c sol_memmap_add_map (if present in more than one,
+ * behaviour is undefined)
+ * @param buffer buffer that will be written, according to its entry on map.
+ *
+ * return 0 on success, a negative number on failure
+ */
+int sol_memmap_write_raw(const char *name, struct sol_buffer *buffer);
```
barbieri: `const struct sol_buffer *buffer`. we shouldn't modify the buffer, right?

also, document that the "negative number" is the -errno on failure.

src/lib/io/include/sol-memmap-storage.h
```diff
@@ -0,0 +1,273 @@
...
+int sol_memmap_read_raw(const char *name, struct sol_buffer *buffer);
+
+/**
+ * Add a map to internal list of available maps.
+ *
+ * @param map map to be add.
+ *
+ * @return 0 on success, a negative number on failure.
+ */
+int sol_memmap_add_map(const struct sol_memmap_map *map);
```
barbieri: do we keep a reference to this map? yes. then document that, since it's not obvious :-)

Say the caller needs o keep the memory pointed by `map` alive during the memmap usage.

src/lib/io/sol-fs-storage.c
```diff
@@ -70,11 +73,14 @@ fs_write(const char *name, struct sol_buffer *buffer)
     return ret;
 }
 
-int
-fs_read(const char *name, struct sol_buffer *buffer)
+SOL_API int
+sol_fs_read_raw(const char *name, struct sol_buffer *buffer)
```
barbieri: modules shouldn't export `SOL_API`.  If multiple files of the same module use that function, just export it in the header, without the `SOL_API`.

src/lib/io/sol-memmap-storage.c
```diff
@@ -0,0 +1,366 @@
...
+    map = NULL;
+
+    return false;
+}
+
+static int
+sol_memmap_read_raw_do(const char *path, const struct sol_memmap_entry *entry, uint64_t mask, struct sol_buffer *buffer)
+{
+    int fd, ret = 0;
+    bool add_ending_char = buffer->capacity == 0;
```
barbieri: this should be the flag in the buffer... to append the trailing `\0`, no?

src/lib/io/sol-fs-storage.c
edersondisouza: Not sure if I understood. This is for everyone who wants to write a buffer...

src/lib/io/sol-memmap-storage.c
edersondisouza: Yeah, I forgot about that flag, tks!

src/lib/io/sol-fs-storage.c
barbieri: ouch, I did read the file name `src/modules/flow/persistence/fs-storage.c`, didn't notice it was moved to `src/lib/io/sol-fs-storage.c`. my bad

src/lib/io/sol-memmap-storage.c
edersondisouza: Hmmm... @barbieri but anyway it's memmap task to add the trailing char, right? For an instant I thought that `sol_util_fill_buffer` would do it for free.

edersondisouza:
v4 on #778

PR #761, opened 2015-09-21
Use uname(2) on get_os_version [v2]

glima:
Changes since #754:

- -impl- files kept as is
- common code moved to platform-linux
- os_version string persisted till shutdown
- base_os and os_version now printed for dgb

Also moving common linux code on that to a sol_platform_linux function,
that is now more suitable.

### DIFF
--- a/src/lib/common/sol-platform-impl-linux-micro.c
+++ b/src/lib/common/sol-platform-impl-linux-micro.c
@@ -846,16 +846,10 @@ sol_platform_impl_get_serial_number(char **number)
     return r;
 }
 
-char *
-sol_platform_impl_get_os_version(void)
+int
+sol_platform_impl_get_os_version(char **version)
 {
-    char *ret = NULL;
-    int r;
-
-    r = sol_util_get_os_version(&ret);
-    SOL_INT_CHECK(r, < 0, NULL);
-
-    return ret;
+    return sol_platform_linux_get_os_version(version);
 }
 
 SOL_API void
--- a/src/lib/common/sol-platform-impl-riot.c
+++ b/src/lib/common/sol-platform-impl-riot.c
@@ -181,8 +181,14 @@ sol_platform_impl_get_serial_number(char **number)
 #endif
 }
 
-char *
-sol_platform_impl_get_os_version(void)
+int
+sol_platform_impl_get_os_version(char **version)
 {
-    return strdup(RIOT_VERSION);
+    SOL_NULL_CHECK(version, -EINVAL);
+
+    *version = strdup(RIOT_VERSION);
+    if (*version)
+        return -ENOMEM;
+
+    return 0;
 }
--- a/src/lib/common/sol-platform-impl-systemd.c
+++ b/src/lib/common/sol-platform-impl-systemd.c
@@ -446,6 +446,7 @@ int
 sol_platform_impl_get_machine_id(char id[static 33])
 {
     int r;
+
     r = sol_util_read_file("/etc/machine-id", "%32s", id);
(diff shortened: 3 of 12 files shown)


src/lib/common/sol-platform-impl-riot.c
```diff
@@ -181,8 +181,14 @@ sol_platform_impl_get_serial_number(char **number)
...
-char *
-sol_platform_impl_get_os_version(void)
+int
+sol_platform_impl_get_os_version(char **version)
 {
-    return strdup(RIOT_VERSION);
+    SOL_NULL_CHECK(version, -EINVAL);
+
+    *version = strdup(RIOT_VERSION);
+    if (*version)
```
barbieri: incorrect check, this is success... not failure

src/lib/common/sol-platform-impl-systemd.c
```diff
@@ -446,6 +446,7 @@ int
 sol_platform_impl_get_machine_id(char id[static 33])
 {
     int r;
+
```
barbieri: whitespace changes...

barbieri: hum... i guess it's because of check-style. ok

src/lib/common/sol-platform-linux-common.c
```diff
@@ -243,3 +245,21 @@ sol_platform_linux_fork_run_exit(int status)
...
+{
+    int r;
+    struct utsname hostinfo;
+
+    SOL_NULL_CHECK(out, -EINVAL);
+
+    r = uname(&hostinfo);
+    SOL_INT_CHECK(r, == -1, -errno);
+
+    r = asprintf(out, "%s", hostinfo.release);
```
barbieri: what's wrong with strdup()?

src/lib/common/sol-platform-linux.h
```diff
@@ -159,6 +159,17 @@ uint64_t sol_platform_linux_fork_run_get_pid(const struct sol_platform_linux_for
...
+/**
+ * Retrieve Linux kernel version.
+ *
+ * @param out Where to store the version string, that conforms to
+ *            uname(2)'s return (release utsname struct field). Must
+ *            de freed after usage.
+ *
+ * @return 0 on success, negative error code otherwise.
+ */
+int sol_platform_linux_get_os_version(char **out);
```
barbieri: no, this is not the right place to put it. This is a public header that is installed... there shouldn't be another API to get OS version, just the `sol_platform_get_os_version()`.

src/lib/common/sol-platform-impl-linux-micro.c
```diff
@@ -846,16 +846,10 @@ sol_platform_impl_get_serial_number(char **number)
     return r;
 }
 
-char *
-sol_platform_impl_get_os_version(void)
+int
+sol_platform_impl_get_os_version(char **version)
```
barbieri: you can avoid this by moving `sol_platform_impl_get_os_version()` to `sol-platform-linux-common.c`, then it's automatically shared in both systemd and linux-micro and you don't need to replicate it.

glima: Okay.

src/lib/common/sol-platform-linux-common.c
glima: Indeed, better.

src/lib/common/sol-platform-impl-riot.c
glima: Damn it, thanks.

glima:
Thanks everyone, new version coming.

PR #763, opened 2015-09-21
sol-oic-gen: make it works for external projects

bdilly:
Required by PR
https://github.com/solettaproject/soletta-demos/pull/1

@lpereira could you please review this pr?


dorileo:
LGTM

data/scripts/sol-oic-gen.py
```diff
@@ -1488,18 +1493,30 @@ def = sol_oic_server_register_definition(defn_endpoint, resource_type,
...
+                        help="Relative path to header generated with "
+                        "sol-flow-node-type-gen for inclusion purposes.",
+                        required=True)
     args = parser.parse_args()
 
     def seems_schema(path):
         return path.endswith('.json') and (path.startswith('oic.r.') or path.startswith('core.'))
 
+    json_name = os.path.basename(args.node_type_json)
+    if (json_name[-5:] == ".json"):
```
barbieri: not a big deal, but in python  you should drop the parenthesis and you can use `json_name.endswith(".json")`

data/scripts/sol-oic-gen.py
```diff
@@ -1512,10 +1529,12 @@ def seems_schema(path):
             continue
 
     print('\nWriting master JSON: %s' % args.node_type_json)
-    open(args.node_type_json, 'w+').write(master_json_as_string(generated))
+    open(args.node_type_json, 'w+').write(master_json_as_string(generated,
```
barbieri: @anselmolsm spotted that on systems without locale set to some `UTF-8`, it would fail to open files with non-ASCII symbols. Then please take the time to specify the encoding so we're safe.

lpereira:
LGTM, with @barbieri's comments.

(I'm about to overhaul the script but haven't started yet, so this won't conflict with my current work.)

bdilly:
merged with corrections

PR #764, opened 2015-09-22
Add Key-value packet type && nodes.

cabelitos:


data/schemas/node-type-genspec.schema
```diff
@@ -39,7 +39,7 @@
     },
 
     "data_type": {
-      "enum": [ "boolean", "int", "byte", "string", "float", "rgb", "direction-vector", "error", "location", "timestamp" ]
+      "enum": [ "boolean", "int", "byte", "string", "float", "rgb", "direction-vector", "error", "location", "timestamp", "key-value" ]
```
barbieri: we can't use `key-value` as both strings if we plan to support other types of values in the future. Better call it `key-string-value` or `key-value-string`.

src/lib/common/sol-types.c
```diff
@@ -405,6 +405,87 @@ sol_rgb_set_max(struct sol_rgb *color, uint32_t max_value)
...
+SOL_API int
+sol_key_value_set_key(struct sol_key_value *kv, const char *key)
+{
+    return replace_string(&kv->key, key);
+}
+
+SOL_API int
+sol_key_value_set_string_value(struct sol_key_value *kv, const char *value)
+{
+    if (kv->type != SOL_KEY_VALUE_TYPE_STRING)
```
barbieri: `SOL_INT_CHECK(kv->type, != SOL_KEY_VALUE_TYPE_STRING, -EINVAL)`?

also, for all `SOL_API` please check for null pointers:
```c
   SOL_NULL_CHECK(kv, -EINVAL);
   SOL_NULL_CHECK(value, -EINVAL);
```

src/lib/common/sol-types.c
```diff
@@ -405,6 +405,87 @@ sol_rgb_set_max(struct sol_rgb *color, uint32_t max_value)
...
+        return -EINVAL;
+    return replace_string(&kv->value.s, value);
+}
+
+SOL_API int
+sol_key_value_set_int_value(struct sol_key_value *kv, int32_t value)
+{
+    if (kv->type != SOL_KEY_VALUE_TYPE_INT)
+        return -EINVAL;
+    kv->value.i = value;
```
barbieri: I'm unsure the key-value should only be an int, or an irange. :-/

src/lib/common/include/sol-types.h
```diff
@@ -249,6 +249,28 @@ struct sol_blob *sol_blob_ref(struct sol_blob *blob);
 void sol_blob_unref(struct sol_blob *blob);
 void sol_blob_set_parent(struct sol_blob *blob, struct sol_blob *parent);
 
+enum sol_key_value_type {
+    SOL_KEY_VALUE_TYPE_STRING,
+    SOL_KEY_VALUE_TYPE_INT,
+    SOL_KEY_VALUE_TYPE_INVALID
```
lpereira: Better move `INVALID` to be the first one, so checking if a `sol_key_value` is valid is cheaper (comparison with zero is cheaper than with any other number).

data/schemas/node-type-genspec.schema
cabelitos: The key-value type has an union in it, that can assume all the values that we need.
Still, should we make the distinction?

src/lib/common/sol-types.c
cabelitos: Oops, forgot about that. I will add these checks.

src/lib/common/sol-types.c
```diff
@@ -405,6 +405,87 @@ sol_rgb_set_max(struct sol_rgb *color, uint32_t max_value)
...
+    return 0;
+}
+
+SOL_API void
+sol_key_value_clear(struct sol_key_value *kv)
+{
+    free(kv->key);
+    if (kv->type == SOL_KEY_VALUE_TYPE_STRING)
+        free(kv->value.s);
+}
```
barbieri: you should set those pointers to null. While at it I'd set the type to unknown or something like that?

lpereira:
Just thinking out loud: Not sure what would be more useful in the long run: a key-value packet type, or a 2-tuple packet type.

src/lib/common/sol-types.c
cabelitos: Well, I had the same doubt, but when I looked at sol_location and sol_rgb I realized that they do not have irange/drange for its members. I tried to follow the same pattern.

src/lib/common/sol-types.c
cabelitos: Ack, we change it to SOL_KEY_VALUE_TYPE_INVALID

src/lib/common/include/sol-types.h
cabelitos: Ack.

cabelitos:
Please look at #771

PR #766, opened 2015-09-22

dorileo:
the following:

  1. sol-flow-$APP-$BOARD.json
  2. sol-flow-$APP.json
  3. sol-flow-$BOARD.json
  4. sol-flow.json

  1. $PWD
  2. app's dir
  3. $PKGSYSCONFIG

### DIFF
--- a/src/shared/sol-conffile.c
+++ b/src/shared/sol-conffile.c
@@ -31,17 +31,20 @@
  */
 
 #include <errno.h>
+#include <libgen.h>
+#include <limits.h>
 #include <stdio.h>
 #include <stdlib.h>
 #include <unistd.h>
-#include <limits.h>
 
-#include "sol-file-reader.h"
 #include "sol-conffile.h"
+#include "sol-file-reader.h"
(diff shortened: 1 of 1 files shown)


src/shared/sol-conffile.c
```diff
@@ -285,9 +288,18 @@ _get_json_include_paths(
             continue;
         }
         if (sol_json_token_str_eq(&key, include_fallback, strlen(include_fallback))) {
-            *include_fallbacks = strndup(value.start + 1, sol_json_token_get_size(&value) - 2);
-            if (!*include_fallbacks)
+            char *buff;
+            struct sol_str_slice *s;
+            buff = strndup(value.start + 1, sol_json_token_get_size(&value) - 2);
```
barbieri: is this to remove quotes? There is a function to do that and should also handle escaping. Ask @bdilly

dorileo: This is as it was before, I haven't even looked at it.

dorileo:
v2 comming

PR #767, opened 2015-09-22

ibriano:
Closes #755

Differences since v1:

### DIFF
--- a/src/modules/flow/calamari/calamari.c
+++ b/src/modules/flow/calamari/calamari.c
@@ -41,6 +41,7 @@
 #include "sol-flow.h"
 #include "sol-flow-static.h"
 #include "sol-mainloop.h"
+#include "sol-modules.h"
 #include "sol-pwm.h"
 #include "sol-spi.h"
 #include "sol-util.h"
@@ -274,7 +275,7 @@ calamari_7seg_new_type(const struct sol_flow_node_type **current)
     };
 
     nodes[SEG_CTL].type = SOL_FLOW_NODE_TYPE_CALAMARI_SEGMENTS_CTL;
-    nodes[SEG_CLEAR].type = nodes[SEG_LATCH].type = nodes[SEG_CLOCK].type = nodes[SEG_DATA].type = SOL_FLOW_NODE_TYPE_GPIO_WRITER;
+    nodes[SEG_CLEAR].type = nodes[SEG_LATCH].type = nodes[SEG_CLOCK].type = nodes[SEG_DATA].type = *(const struct sol_flow_node_type **)sol_symbol_get("flow", "gpio", SOL_FLOW_NODE_TYPE_GPIO_WRITER);
 
     type = sol_flow_static_new_type(&spec);
     SOL_NULL_CHECK(type);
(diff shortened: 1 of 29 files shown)


src/modules/flow/calamari/calamari.c
```diff
@@ -274,7 +275,7 @@ calamari_7seg_new_type(const struct sol_flow_node_type **current)
     };
 
     nodes[SEG_CTL].type = SOL_FLOW_NODE_TYPE_CALAMARI_SEGMENTS_CTL;
-    nodes[SEG_CLEAR].type = nodes[SEG_LATCH].type = nodes[SEG_CLOCK].type = nodes[SEG_DATA].type = SOL_FLOW_NODE_TYPE_GPIO_WRITER;
+    nodes[SEG_CLEAR].type = nodes[SEG_LATCH].type = nodes[SEG_CLOCK].type = nodes[SEG_DATA].type = *(const struct sol_flow_node_type **)sol_symbol_get("flow", "gpio", SOL_FLOW_NODE_TYPE_GPIO_WRITER);
```
barbieri: this is still not good  and will crash if gpio is a module and is missing at runtime :-(

ibriano: On Tue, 22 Sep 2015 12:06:53 -0700, Gustavo Sverzut Barbieri wrote:
> > @@ -274,7 +275,7 @@ calamari_7seg_new_type(const struct sol_flow_node_type **current)
> >      };
> >  
> >      nodes[SEG_CTL].type = SOL_FLOW_NODE_TYPE_CALAMARI_SEGMENTS_CTL;
> > -    nodes[SEG_CLEAR].type = nodes[SEG_LATCH].type = nodes[SEG_CLOCK].type = nodes[SEG_DATA].type = SOL_FLOW_NODE_TYPE_GPIO_WRITER;
> > +    nodes[SEG_CLEAR].type = nodes[SEG_LATCH].type = nodes[SEG_CLOCK].type = nodes[SEG_DATA].type = *(const struct sol_flow_node_type **)sol_symbol_get("flow", "gpio", SOL_FLOW_NODE_TYPE_GPIO_WRITER);
> 
> this is still not good  and will crash if gpio is a module and is missing at runtime :-(

Ah, I knew I was forgetting to change that somewhere.

src/test/test-flow.c
```diff
@@ -1151,12 +1152,14 @@ named_options_init_from_strv(void)
...
     struct sol_flow_node_named_options_member *m;
+    struct sol_flow_node_type **node_type;
     int r;
 
+    node_type = sol_symbol_get("flow", "timer", SOL_FLOW_NODE_TYPE_TIMER);
     {
         const char *strv[] = { "interval=1000", NULL };
 
-        r = sol_flow_node_named_options_init_from_strv(&named_opts, SOL_FLOW_NODE_TYPE_TIMER, strv);
+        r = sol_flow_node_named_options_init_from_strv(&named_opts, *node_type, strv);
```
barbieri: crashes if `node_type == NULL`.

What if we change to:
```c
struct sol_flow_node_type *node_type; /* pointer, not pointer to pointer */
if (sol_symbol_get("flow", "timer", SOL_FLOW_NODE_TYPE_TIMER, &node_type) < 0)
   goto error;

 r = sol_flow_node_named_options_init_from_strv(&named_opts, node_type, strv);
```

ibriano: On Tue, 22 Sep 2015 12:10:05 -0700, Gustavo Sverzut Barbieri wrote:
> >      {
> >          const char *strv[] = { "interval=1000", NULL };
> >  
> > -        r = sol_flow_node_named_options_init_from_strv(&named_opts, SOL_FLOW_NODE_TYPE_TIMER, strv);
> > +        r = sol_flow_node_named_options_init_from_strv(&named_opts, *node_type, strv);
> 
> crashes if `node_type == NULL`.
> 
> What if we change to:
> ```c
> struct sol_flow_node_type *node_type; /* pointer, not pointer to pointer */
> if (sol_symbol_get("flow", "timer", SOL_FLOW_NODE_TYPE_TIMER, &node_type) < 0)
>    goto error;
> 
>  r = sol_flow_node_named_options_init_from_strv(&named_opts, node_type, strv);
> ```

Those I just ignored, if the test crashes, someone fucked up anyway.

And passing the variable as a parameter, when modules are entirely
disabled would end up breaking, as we can't do:

&node_type = SOL_FLOW_NODE_TYPE_TIMER

There's the option of adding the & in the macro, but I dislike that.

barbieri: well, you can have an intermediate variable, but anyway the macro for disabled modules would be:

```c
#define sol_symbol_get(..., sym, var) ({var = sym; 0})
```

that is, assign symbol to variable and return 0 (success).

ibriano:
New  version in #782

PR #768, opened 2015-09-22
Is empty

glima:
strings: new node string/is-empty, with tests.


src/modules/flow/string/string-ascii.c
```diff
@@ -949,4 +949,23 @@ set_prefix_suffix_end(struct sol_flow_node *node,
...
+{
+    struct string_is_empty_data *mdata = data;
+    const char *in_value;
+    int r;
+
+    r = sol_flow_packet_get_string(packet, &in_value);
+    SOL_INT_CHECK(r, < 0, r);
+
+    return sol_flow_send_boolean_packet(node,
+        SOL_FLOW_NODE_TYPE_STRING_IS_EMPTY__OUT__OUT, strlen(in_value) == 0);
```
barbieri: likely to be optimized by the compiler, but writing `in_value[0] != '\0'` is guaranteed to be O(1)

src/modules/flow/string/string-icu.c
```diff
@@ -1195,4 +1195,21 @@ set_prefix_suffix_end(struct sol_flow_node *node,
...
+{
+    const char *in_value;
+    int r;
+
+    r = sol_flow_packet_get_string(packet, &in_value);
+    SOL_INT_CHECK(r, < 0, r);
+
+    return sol_flow_send_boolean_packet(node,
+        SOL_FLOW_NODE_TYPE_STRING_IS_EMPTY__OUT__OUT, strlen(in_value) == 0);
+}
```
barbieri: we should have a string-common.c with such methods that are the same in both ascii and icu implementations.

lpereira: Is it correct to call `strlen()` on UTF-8 strings anyway? Even if just to check if it's an empty string. For instance, a string with just a BOM or RTL markers would be empty, right?

glima: It seems we're ok to use strlen() on utf8 strings. Even in python, strings containing only these weird chars you mentioned were not empty. I was lazy about the common file, now doing it. I prefer to let strlen() explicit, instead of accessing [0].

glima:
sending v2

PR #769, opened 2015-09-22

dorileo:
## Changes ##

  + since v1 (#766):

## Rationale ##
the following:

  1. sol-flow-$APP-$BOARD.json
  2. sol-flow-$APP.json
  3. sol-flow-$BOARD.json
  4. sol-flow.json

  1. $PWD
  2. app's dir
  3. $PKGSYSCONFIG

### DIFF
--- a/src/shared/sol-conffile.c
+++ b/src/shared/sol-conffile.c
@@ -31,17 +31,20 @@
  */
 
 #include <errno.h>
+#include <libgen.h>
+#include <limits.h>
 #include <stdio.h>
 #include <stdlib.h>
 #include <unistd.h>
-#include <limits.h>
 
-#include "sol-file-reader.h"
 #include "sol-conffile.h"
+#include "sol-file-reader.h"
+#include "sol-json.h"
 #include "sol-log.h"
+#include "sol-mainloop.h"
+#include "sol-platform.h"
 #include "sol-util.h"
 #include "sol-vector.h"
-#include "sol-json.h"
 
 static struct sol_ptr_vector _conffile_entry_vector; /* entries created by conffiles */
 static struct sol_ptr_vector _conffiles_loaded; /* paths of the currently loaded conffiles */
@@ -250,7 +253,7 @@ static void
 _get_json_include_paths(
     struct sol_json_scanner json_scanner,
     char **include,
-    char **include_fallbacks)
+    struct sol_vector *include_fallbacks)
 {
     static const char *include_group = "config_includes";
     static const char *include_str = "include";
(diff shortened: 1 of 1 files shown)


src/shared/sol-conffile.c
```diff
@@ -285,9 +288,19 @@ _get_json_include_paths(
...
+            sol_json_token_remove_quotes(&value);
+            buff = strndup(value.start, sol_json_token_get_size(&value));
+            if (!buff)
                 SOL_DBG("Error: couldn't allocate memory for string.");
+            else {
+                s = sol_vector_append(include_fallbacks);
+                if (s) {
+                    s->data = buff;
+                    s->len = strlen(buff);
+                }
```
lpereira: `if (!s)`, `buff` leaks

src/shared/sol-conffile.c
```diff
@@ -432,28 +448,111 @@ _fill_vector(const char *path, const char *fallback_paths)
...
+        _fill_vector(include, &include_fallbacks);
+    }
 
 free_for_all:
     sol_file_reader_close(file_reader);
     free(full_path);
     free(include);
-    free(include_fallbacks);
+
+    SOL_VECTOR_FOREACH_IDX (&include_fallbacks, slice, i) {
```
lpereira: I'm seeing more and more `sol_vector` including `sol_str_slice`, and code like this, where `free()` is called with the needed cast to `char *`. Is this maybe a cry for a `sol_str_slice_vector`?

src/shared/sol-conffile.c
```diff
@@ -432,28 +448,111 @@ _fill_vector(const char *path, const char *fallback_paths)
...
+        int _r; \
+        char *_buff; \
+        struct sol_str_slice *_s; \
+        _r = asprintf(&_buff, _fmt, ## __VA_ARGS__); \
+        if (_r < 0 || !_buff) \
+            SOL_WRN("Couldn't allocate memory for config file formatting."); \
+        else { \
+            _s = sol_vector_append(_vector); \
+            if (_s) { \
+                _s->data = _buff; \
```
lpereira: `_buff` leaks `if (!_s)`

src/shared/sol-conffile.c
```diff
@@ -432,28 +448,111 @@ _fill_vector(const char *path, const char *fallback_paths)
...
+    sol_vector_clear(&include_fallbacks);
+}
+
+static void
+_add_lookup_path(struct sol_vector *vector, const char **search_dirs, char *appname, const char *board_name)
+{
+    size_t i;
+    struct sol_vector files = SOL_VECTOR_INIT(struct sol_str_slice);
+
+#define ADD_PATH_FMT(_vector, _fmt, ...) \
```
lpereira: Make this a function taking the vector and an allocated string, and use it in `_get_json_include_paths()` as well.

src/shared/sol-conffile.c
dorileo: true...

src/shared/sol-conffile.c
dorileo: ok

src/shared/sol-conffile.c
dorileo: I think the macro is good enough here, with that we do it all at once - no code duplication (even for the string allocation).

dorileo:
v3 coming...

dorileo:
v3 - #772

src/shared/sol-conffile.c
dorileo: but I agree with the previous comment about a possible ```sol_str_slice_vector```.

PR #770, opened 2015-09-22
strings: new node string/is-empty, with tests [v2]

glima:
Changes since #768:

- commits squashed
- common file for added function now present


glima:
Thanks, will merge.

PR #771, opened 2015-09-22
Add String tuple node/packet types.

cabelitos:


src/bin/sol-fbp-runner/inspector.c
```diff
@@ -217,6 +217,21 @@ inspector_show_packet(const struct sol_flow_packet *packet)
...
+            int r = sol_flow_packet_get(packet, &array);
+            if (r != 0)
+                return;
+            r = sol_flow_packet_get_string(array[0], &k);
+            if (r != 0)
+                return;
+            r = sol_flow_packet_get_string(array[1], &v);
+            if (r != 0)
+                return;
+            fprintf(stdout, "Key:%s Value:%s (String tuple)\n", k, v);
```
lpereira: Tuples don't have keys and values; maybe `first` and `second`.

barbieri: you can format that as `(%s, %s) (String Tuple)`. Also move the `const char *k, *v` inside the `if (sol_flow_packet_get(...))`.

lpereira:
Food for thought: If tuples worked similarly to Lisp lists, with primitives such as `cons` (construct), `car` (head), and `cdr` (tail), I'd be happier. The first and second elements could be packets themselves, and then no changes to the inspector to add new types would be required.

@vcgomes ^

barbieri:
we discussed in person the following idea:

 * new `sol_packet_type_composed_new(array-of-types)`. This would cache based on `array-of-types` and create new instances when required. This would keep all existing checks working with minimum runtime overhead (cache bookkeeping as an array of `struct {type*, array-of-types}`), plus the new type itself (dozen of bytes)).
 * new JSON packet type format, in addition to basic types and `custom:...`, we add `composed:type1,type2...`. Then we call `sol_packet_type_composed_new(type1, type2...)`. This would be used by `node-type-gen.py` and output more code in `init_type()` for composed.
 * temporary hack `sol_packet_type_is_composed(type)` that checks if a type is composed to be used in console and inspector. It can apply a trick to check if type could be composed by checking `type->data_size % sizeof(struct sol_packet_type *) == 0`. If yes, then loop cached types and see if it's in there -- not optimal, but temporary until we introduce some introspection information.
 * Two new `flow-meta` to be used in `DECLARE=` statements, they would be something like `composed-split` and `composed-new`. The payload would be the type description with ports names, such as `DECLARE=key-value-string-get:composed-split:KEY(string),VALUE(string)`, that creates a node type `key-value-string-get` with an input port `IN` of packet type `composed:string,string` and output ports `KEY` (type string) and `VALUE` (type string).  This way we reuse flow-meta infrastructure and don't need to manually create node types for all composed combinations.

@cmarcelo ^

lpereira:
[**@barbieri**](https://github.com/barbieri) commented on [Sep 22, 2015, 7:11 PM GMT-3](https://github.com/solettaproject/soletta/pull/771#issuecomment-142436714 "2015-09-22T22:11:43Z - Replied by Github Comment Enhancer"):
> we discussed in person the following idea:
>  (...)

Sounds like a good alternative.

src/modules/flow/tuple/tuple.json
```diff
@@ -0,0 +1,92 @@
...
+        {
+          "data_type": "string-tuple",
+          "description": "A tuple string tuple",
+          "name": "OUT"
+        }
+      ]
+    },
+    {
+      "category": "tuple",
+      "description": "Constructs a tuple packet",
```
edersondisouza: Description 'constructs' on something that 'splits' looks confusing...

cabelitos:
Closing in favor of #783

PR #772, opened 2015-09-22

dorileo:
## Changes ##

  + v2 (since #766):
  + v3 (since #769):
    - fixed leak on ```sol_vector_append()``` error - spotted by @lpereira ;

## Rationale ##
the following:

  1. sol-flow-$APP-$BOARD.json
  2. sol-flow-$APP.json
  3. sol-flow-$BOARD.json
  4. sol-flow.json

  1. $PWD
  2. app's dir
  3. $PKGSYSCONFIG

### DIFF
--- a/src/shared/sol-conffile.c
+++ b/src/shared/sol-conffile.c
@@ -31,17 +31,20 @@
  */
 
 #include <errno.h>
+#include <libgen.h>
+#include <limits.h>
 #include <stdio.h>
 #include <stdlib.h>
 #include <unistd.h>
-#include <limits.h>
 
-#include "sol-file-reader.h"
 #include "sol-conffile.h"
+#include "sol-file-reader.h"
+#include "sol-json.h"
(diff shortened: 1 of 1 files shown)


src/shared/sol-conffile.c
```diff
@@ -432,28 +450,113 @@ _fill_vector(const char *path, const char *fallback_paths)
...
+        else { \
+            _s = sol_vector_append(_vector); \
+            if (_s) { \
+                _s->data = _buff; \
+                _s->len = strlen(_buff); \
+            } else { \
+                free(_buff); \
+            } \
+        } \
+    } while (0) \
```
lpereira: Stray `\`. Please make this a function, there's no need to make it a macro. (Also, it can be reused above.)

dorileo: Ok

dorileo:
v4 coming...

PR #773, opened 2015-09-22
Client v2

ceolin:
- Added url port in the nodes
 - Fix problems when encoding/decoding doubles
 - Remove commit with double_to_str_alloc
 - Added commit with comments about how to test server's nodes
 - Re-using more code between the client nodes
 - Using json boolean types

### DIFF
--- a/src/modules/flow/http-client/http-client.c
+++ b/src/modules/flow/http-client/http-client.c
@@ -358,4 +358,121 @@ string_post_process(struct sol_flow_node *node, void *data, uint16_t port, uint1
     return 0;
 }
 
+/*
+ * --------------------------------- irange node -----------------------------
+ */
+static void
+int_request_finished(void *data,
+    const struct sol_http_client_connection *connection,
+    struct sol_http_response *response)
+{
+    struct sol_flow_node *node = data;
+    struct http_data *mdata = sol_flow_node_get_private_data(node);
+    struct sol_irange irange;
+
+    if (check_response(mdata, node, connection, response) < 0)
+        return;
+
+    if (streq(response->content_type, "application/json")) {
+        struct sol_json_scanner scanner;
+        struct sol_json_token token, key, value;
+        enum sol_json_loop_reason reason;
+
+        sol_json_scanner_init(&scanner, response->content.data, response->content.used);
+        SOL_JSON_SCANNER_OBJECT_LOOP (&scanner, &token, &key, &value, reason) {
+            struct sol_json_scanner sub_scanner;
+            if (check_key(mdata->url, &key))
+                continue;
+            sol_json_scanner_init(&sub_scanner, value.start, value.end - value.start);
+            SOL_JSON_SCANNER_OBJECT_LOOP (&sub_scanner, &token, &key, &value, reason) {
+                if (SOL_JSON_TOKEN_STR_LITERAL_EQ(&key, "value")) {
+                    if (sol_json_token_get_int32(&value, &irange.val) < 0)
+                        goto error;
+                } else if (SOL_JSON_TOKEN_STR_LITERAL_EQ(&key, "min")) {
+                    if (sol_json_token_get_int32(&value, &irange.min) < 0)
+                        goto error;
+                } else if (SOL_JSON_TOKEN_STR_LITERAL_EQ(&key, "max")) {
+                    if (sol_json_token_get_int32(&value, &irange.max) < 0)
+                        goto error;
+                } else if (SOL_JSON_TOKEN_STR_LITERAL_EQ(&key, "step")) {
+                    if (sol_json_token_get_int32(&value, &irange.step) < 0)
+                        goto error;
+                }
+            }
+            sol_flow_send_irange_packet(node,
+                SOL_FLOW_NODE_TYPE_HTTP_CLIENT_INT__OUT__OUT, &irange);
+            return;
+        }
+    } else {
+        errno = 0;
+        irange.val = strtol(response->content.data, NULL, 0);
+        if (errno)
+            goto error;
+        sol_flow_send_irange_packet(node,
+            SOL_FLOW_NODE_TYPE_HTTP_CLIENT_INT__OUT__OUT, &irange);
+        return;
+    }
+
+error:
+    sol_flow_send_error_packet(node, -EINVAL,
+        "%s Could not parser the url's contents ", mdata->url);
+}
+
+static int
+int_post_process(struct sol_flow_node *node, void *data, uint16_t port, uint16_t conn_id,
+    const struct sol_flow_packet *packet)
+{
+    int r;
+    struct sol_irange value;
+    struct sol_http_param params;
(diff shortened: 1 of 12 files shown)


src/modules/flow/http-client/http-client.c
```diff
@@ -0,0 +1,600 @@
...
+
+    return 0;
+}
+
+static int
+check_key(const char *url, struct sol_json_token *token)
+{
+    char *ptr;
+
+    if (strncmp(url, "http://", sizeof("http://") - 1) == 0)
```
lpereira: I believe we have a `strstartswith()` somewhere.

Also, what about `https://` URIs?

ceolin: so far, our server does not support https, but yes, i'll check for it before commit. Thanks

ceolin:
I've checked that common_url_process is not releasing the memory of mdata->url, i've already fixed it locally.

src/modules/flow/http-client/http-client.c
```diff
@@ -0,0 +1,600 @@
...
+    struct http_data *mdata = data;
+    struct sol_http_client_connection *connection;
+    const struct http_client_node_type *type;
+
+    SOL_NULL_CHECK(mdata->url, -EINVAL);
+
+    sol_http_param_init(&params);
+    if (!sol_http_param_add(&params,
+        SOL_HTTP_REQUEST_PARAM_HEADER("Accept", "application/json"))) {
+        SOL_WRN("Failed to set query params");
```
barbieri: add as much information as possible to error messages, here you could add `mdata->url` and also `Accept: application/json`, it would help debug to identify the node that caused it.

src/modules/flow/http-client/http-client.c
barbieri: you can cache this offset or pointer at open, then you only use `sol_json_token_str_eq(token, mdata->key.data, mdata->key.len)` (assuming `mdata->key` is `struct sol_str_slice`).

src/modules/flow/http-client/http-client.c
```diff
@@ -0,0 +1,600 @@
...
+    if (streq(response->content_type, "application/json")) {
+        struct sol_json_scanner scanner;
+        struct sol_json_token token, key, value;
+        enum sol_json_loop_reason reason;
+
+        sol_json_scanner_init(&scanner, response->content.data, response->content.used);
+        SOL_JSON_SCANNER_OBJECT_LOOP (&scanner, &token, &key, &value, reason) {
+            if (check_key(mdata->url, &key))
+                continue;
+            if (sol_json_token_get_type(&value) == SOL_JSON_TYPE_TRUE)
```
barbieri: up to this line this should be a common function. then here it would call a `node_type->process_token(node, &value)`. That would contain the block that converts the value to actual type and send a packet on output port.

src/modules/flow/http-client/http-client.c
```diff
@@ -0,0 +1,600 @@
...
+            else if (sol_json_token_get_type(&value) == SOL_JSON_TYPE_FALSE)
+                result = false;
+            else
+                goto err;
+            sol_flow_send_boolean_packet(node,
+                SOL_FLOW_NODE_TYPE_HTTP_CLIENT_BOOLEAN__OUT__OUT, result);
+            return;
+        }
+    } else {
+        if (!strncasecmp("true", response->content.data, response->content.used))
```
barbieri: similarly, this would be a `node_type->process_post(node, response->content)`

src/modules/flow/http-client/http-client.c
```diff
@@ -0,0 +1,600 @@
...
+    struct http_data *mdata = data;
+    struct sol_http_client_connection *connection;
+    const struct http_client_node_type *type;
+
+    SOL_NULL_CHECK(mdata->url, -EINVAL);
+
+    r = sol_flow_packet_get_boolean(packet, &b);
+    SOL_INT_CHECK(r, < 0, r);
+
+    sol_http_param_init(&params);
```
barbieri: the rest of this function should be a common one that receives the `value_as_string`. Then the only type-specific bit is the one that takes the packet and converts the value to string:

```c

   r = sol_flow_packet_get_boolean(packet, &b);
   SOL_INT_CHECK(r, < 0, r);
   return common_request(node, b ? "true" : "false");
}

```

src/modules/flow/http-client/http-client.c
```diff
@@ -0,0 +1,600 @@
...
+static int
+int_post_process(struct sol_flow_node *node, void *data, uint16_t port, uint16_t conn_id,
+    const struct sol_flow_packet *packet)
+{
+    int r;
+    struct sol_irange value;
+    struct sol_http_param params;
+    struct http_data *mdata = data;
+    struct sol_http_client_connection *connection;
+    char min[100], max[100], val[100], step[100];
```
lpereira: 400 bytes is a little bit too much for integer representation... `3 * sizeof(value.val)` should be a good approximation that doesn't waste too much too space.

src/modules/flow/http-client/http-client.c
```diff
@@ -0,0 +1,600 @@
...
+        else
+            goto err;
+        sol_flow_send_boolean_packet(node,
+            SOL_FLOW_NODE_TYPE_HTTP_CLIENT_BOOLEAN__OUT__OUT, result);
+        return;
+    }
+
+err:
+    sol_flow_send_error_packet(node, EINVAL,
+        "%s Could not parser the url's contents ", mdata->url);
```
edersondisouza: s/Could not parser/Could not parse/

src/modules/flow/http-client/http-client.c
```diff
@@ -0,0 +1,600 @@
...
+                        goto error;
+                }
+            }
+            sol_flow_send_irange_packet(node,
+                SOL_FLOW_NODE_TYPE_HTTP_CLIENT_INT__OUT__OUT, &irange);
+            return;
+        }
+    } else {
+        errno = 0;
+        irange.val = strtol(response->content.data, NULL, 0);
```
edersondisouza: Is it ok to send an `irange` with no `min`, `max` or `step` set? Shouldn't use some defaults?

src/modules/flow/http-client/http-client.c
```diff
@@ -0,0 +1,600 @@
...
+                    SOL_INT_CHECK_GOTO(r, != 0, error);
+                }
+            }
+            sol_flow_send_drange_packet(node,
+                SOL_FLOW_NODE_TYPE_HTTP_CLIENT_FLOAT__OUT__OUT, &drange);
+            return;
+        }
+    } else {
+        errno = 0;
+        drange.val = sol_util_strtodn(response->content.data, NULL, -1, false);
```
edersondisouza: Same as `irange` comment.

ceolin:
Fixes on https://github.com/solettaproject/soletta/pull/781

PR #774, opened 2015-09-23

dorileo:
## Changes ##

  + v2 (since #766):
  + v3 (since #769):
    - fixed leak on ```sol_vector_append()``` error - spotted by @lpereira ;
  + v4 (since #772):
    - got rid of ```ADD_FMT_PATH``` macro - as suggested by @lpereira ;

## Rationale ##

the following:

  1. sol-flow-$APP-$BOARD.json
  2. sol-flow-$APP.json
  3. sol-flow-$BOARD.json
  4. sol-flow.json

  1. $PWD
  2. app's dir
  3. $PKGSYSCONFIG

### DIFF
--- a/src/shared/sol-conffile.c
+++ b/src/shared/sol-conffile.c
@@ -31,17 +31,20 @@
  */
 
 #include <errno.h>
+#include <libgen.h>
+#include <limits.h>
 #include <stdio.h>
 #include <stdlib.h>
 #include <unistd.h>
-#include <limits.h>
 
-#include "sol-file-reader.h"
 #include "sol-conffile.h"
+#include "sol-file-reader.h"
+#include "sol-json.h"
 #include "sol-log.h"
+#include "sol-mainloop.h"
+#include "sol-platform.h"
 #include "sol-util.h"
 #include "sol-vector.h"
-#include "sol-json.h"
 
 static struct sol_ptr_vector _conffile_entry_vector; /* entries created by conffiles */
 static struct sol_ptr_vector _conffiles_loaded; /* paths of the currently loaded conffiles */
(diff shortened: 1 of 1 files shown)


src/shared/sol-conffile.c
```diff
@@ -285,9 +302,20 @@ _get_json_include_paths(
             continue;
         }
         if (sol_json_token_str_eq(&key, include_fallback, strlen(include_fallback))) {
-            *include_fallbacks = strndup(value.start + 1, sol_json_token_get_size(&value) - 2);
-            if (!*include_fallbacks)
+            char *buff;
+            struct sol_str_slice *s;
+            sol_json_token_remove_quotes(&value);
+            buff = strndup(value.start, sol_json_token_get_size(&value));
```
edersondisouza: Couldn't you get rid of this `strndup` and just append `value` to `include_fallbacks`, using`sol_json_token_to_slice`?

src/shared/sol-conffile.c
```diff
@@ -432,28 +463,112 @@ _fill_vector(const char *path, const char *fallback_paths)
...
+    int r;
+    struct sol_str_slice *slice;
+
+    va_start(ap, fmt);
+    r = vasprintf(&buff, fmt, ap);
+    va_end(ap);
+    if (r < 0 || !buff)
+        return;
+
+    slice = _vector_append_string_as_str_slice(vector, buff);
```
edersondisouza: Ouch, nevermind my first comment, you use other strings too.

src/shared/sol-conffile.c
```diff
@@ -432,28 +463,112 @@ _fill_vector(const char *path, const char *fallback_paths)
...
+        _add_formated_lookup_path(&files, "sol-flow-%s.json", appname);
+    }
+
+    if (board_name) {
+        _add_formated_lookup_path(&files, "sol-flow-%s.json", board_name);
+    }
+
+    _add_formated_lookup_path(&files, "sol-flow.json");
+
+    for (i = 0; i < ARRAY_SIZE(search_dirs); i++) {
```
edersondisouza: Does `ARRAY_SIZE` work here?

lpereira: Nope. In fact, it might be a good idea to change `ARRAY_SIZE` so that it checks if it's being applied to an array only. I believe kmod and the kernel have similar macros with these checks to serve as inspiration.

dorileo:
v5 coming...

PR #776, opened 2015-09-23
pin-mux: Keep track of opened gpios...

mbelluzzo:
...so struct sol_gpio won't leak and we can close them on shutdown.

We were avoiding to do this since introduces extra complexity on code
that should be straight and simple.

Primary reason is that we cannot close those gpios after usage
because they lose the state and multiplexer does not work. So at first
we were leaving them opened since there wasn't any major drawbacks.

But, by doing this, 'sol_gpio' structures started to leak what bring
us to this fix :(

Also took the opportunity to add some missing 'mux' namespace on
intel-common.{c,h}.


mbelluzzo:
pushing v2

PR #777, opened 2015-09-23
generator: irange values are divergent

dorileo:
If we run an fbp with sol-fbp-runner and the options has no sub-option
the values for .min, .max and step are set to INT32_MIN, INT32_MAX and 1
respectively, but the code generated by sol-fbp-generator doesn't set
those values by default.

The converter-string-int.fbp excercises the problem, i.e:

The following fbp fragment when generated + built + compiled should
produce the output string "003 / -2147483648.000 / 17777777777 / 0x1",
instead it produces "003 / 0.000 / 0 / 0x0":

_(constant/int:value=3) OUT -> IN
_(converter/int-to-string:format_spec="{val:03d} / {min:.3f} / {max:o} /
{step:#x}") OUT -> IN[0] fmt_test_cmp(string/compare)

### DIFF
--- a/src/bin/sol-fbp-generator/main.c
+++ b/src/bin/sol-fbp-generator/main.c
@@ -116,7 +116,7 @@ get_node_type_description(const struct fbp_data *data, uint16_t i)
 
 static void
 handle_suboptions(const struct sol_fbp_meta *meta,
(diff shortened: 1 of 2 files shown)


dorileo:
report purpose only, closing.

PR #781, opened 2015-09-24
Client v3

ceolin:
-v3
 - Fixed some debug messages
 - Set default value for irange/drange
 - Re-use all the common code
 - Cache the uri protocol (and startswith() for that)
 - Decreased the buffers used on irange convertion
 - Fixed the license

-v2
 - Added url port in the nodes
 - Fix problems when encoding/decoding doubles
 - Remove commit with double_to_str_alloc
 - Added commit with comments about how to test server's nodes
 - Re-using more code between the client nodes
 - Using json boolean types

### DIFF
--- a/src/modules/flow/http-client/http-client.json
+++ b/src/modules/flow/http-client/http-client.json
@@ -1,0 +1,65 @@
+{
+  "$schema": "http://solettaproject.github.io/soletta/schemas/node-type-genspec.schema",
+  "name": "http-client",
+  "meta": {
+    "author": "Intel Corporation",
+    "license": "BSD-3-Clause",
+    "version": "1"
+  },
+  "types": [
+    {
+      "name": "http-client/boolean",
+      "category": "output/network",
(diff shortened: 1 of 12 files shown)


src/modules/flow/http-client/http-client.json
```diff
@@ -0,0 +1,267 @@
...
+	    "description": "The url that will be used",
+	    "name": "url"
+	  }
+	],
+	"version": 1
+      },
+      "in_ports": [
+        {
+          "data_type": "string",
+          "description": "Url wich request will be done",
```
edersondisouza: s/wich/which/
Btw, there are places where url description is 'The url that will be used', maybe is good to use the same string on all places (I'd rather the second one, or something even more descriptive).

ceolin: thanks, i'll change it before commit

ceolin:
merged with the fixes. Thank you all involved.

PR #782, opened 2015-09-24

ibriano:
Differences since v2:
 * Moved from common to flow, since it's unlikely that it will be
   helpful for the generic case.

Differences since v1:

### DIFF
--- a/src/bin/sol-fbp-generator/main.c
+++ b/src/bin/sol-fbp-generator/main.c
@@ -815,22 +815,23 @@ generate(struct sol_vector *fbp_data_vector)
 
     if (!args.export_symbol) {
         out(
+            "static const struct sol_flow_node_type *root_type;\n"
             "static struct sol_flow_node *flow;\n"
             "\n"
             "static void\n"
             "startup(void)\n"
             "{\n"
-            "    const struct sol_flow_node_type *type;\n\n"
             "    initialize_types();\n"
-            "    type = create_0_root_type();\n"
-            "    if (!type)\n"
+            "    root_type = create_0_root_type();\n"
+            "    if (!root_type)\n"
             "        return;\n\n"
-            "    flow = sol_flow_node_new(NULL, NULL, type, NULL);\n"
+            "    flow = sol_flow_node_new(NULL, NULL, root_type, NULL);\n"
             "}\n\n"
             "static void\n"
             "shutdown(void)\n"
             "{\n"
             "    sol_flow_node_del(flow);\n"
+            "    sol_flow_node_type_del((struct sol_flow_node_type *)root_type);\n"
             "}\n\n"
             "SOL_MAIN_DEFAULT(startup, shutdown);\n");
     } else {
--- a/src/bin/sol-fbp-generator/main.c
+++ b/src/bin/sol-fbp-generator/main.c
@@ -83,6 +83,12 @@ struct declared_fbp_type {
     int id;
 };
 
+struct node_data {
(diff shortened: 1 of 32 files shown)


src/bin/sol-fbp-generator/main.c
```diff
@@ -633,8 +658,12 @@ generate_node_type_assignments(const struct fbp_data *data)
     out("\n");
 
     SOL_VECTOR_FOREACH_IDX (&data->graph.nodes, n, i) {
-        struct type_description *desc = n->user_data;
-        out("    nodes[%d].type = %s;\n", i, desc->symbol);
+        struct node_data *nd = n->user_data;
+
+        if (nd->is_declared)
```
barbieri: hum... maybe check `nd->desc` or `nd->desc->symbol` and remove `nd->is_declared`? is it possible?

ibriano: On Sep 24, 2015, at 18:14, Gustavo Sverzut Barbieri <notifications@github.com<mailto:notifications@github.com>> wrote:


In src/bin/sol-fbp-generator/main.c<https://github.com/solettaproject/soletta/pull/782#discussion_r40374136>:

> @@ -633,8 +658,12 @@ generate_node_type_assignments(const struct fbp_data *data)
>      out("\n");
>
>      SOL_VECTOR_FOREACH_IDX (&data->graph.nodes, n, i) {
> -        struct type_description *desc = n->user_data;
> -        out("    nodes[%d].type = %s;\n", i, desc->symbol);
> +        struct node_data *nd = n->user_data;
> +
> +        if (nd->is_declared)


hum... maybe check nd->desc or nd->desc->symbol and remove nd->is_declared? is it possible?

Both exist, I'd need to check for a given pattern or something. I don't see why removing is_declared would gain much.

�

src/lib/flow/sol-flow-internal.h
```diff
@@ -293,6 +293,8 @@ int sol_flow_builder_add_node_taking_options(
     const struct sol_flow_node_options *options);
 
 #ifdef ENABLE_DYNAMIC_MODULES
+void sol_flow_modules_cache_shutdown(void);
```
barbieri: even if you just use it inside flow, I'd keep all the machinery in the common/core. Then make the old function not-SOL_API and put it inside some private header, expose only the `sol-flow.h` one.

This way, if we need the logic for other parts (likely I'd say), we don't need to move back all of that.

src/lib/flow/include/sol-flow.h
```diff
@@ -457,6 +457,27 @@ struct sol_flow_port_type_in {
...
+ * Checks if the node type @a _type is built-in, if not, it loads the module
+ * @a _mod and fetches the type's symbol there. The result is stored in @a _var.
+ *
+ * @param _mod The name of the module to load if the symbol is not built-in.
+ * @param _type The node type's symbol.
+ * @param _var Variable where to store the type.
+ *
+ * @return 0 on success, < 0 on error.
+ */
+#define sol_flow_get_node_type(_mod, _type, _var) sol_flow_int_get_node_type(_mod, #_type, _var)
```
barbieri: `s/_int_/_internal_/g`, it may look like an integer...

src/bin/sol-fbp-generator/main.c
barbieri: ok, keep it then.

ibriano:
Hopefully the last version in #796

PR #785, opened 2015-09-24
modules/gtk: add more ports to gtk label

bdilly:
So it can be used to simulate led 7 segments.


bdilly:
new version on #790

PR #787, opened 2015-09-25
converter node: send error packet on bad boolean string.

glima:


mbelluzzo:
+2

PR #789, opened 2015-09-25
timestamp/time: have the option of sending an initial packet.

glima:

### DIFF
--- a/src/modules/flow/timestamp/timestamp.json
+++ b/src/modules/flow/timestamp/timestamp.json
@@ -21,7 +21,27 @@
          "required": true
        }
       ],
+      "methods": {
+        "open": "timestamp_time_open"
+      },
       "name": "timestamp/time",
+      "options": {
+        "members": [
+          {
+            "data_type": "boolean",
+            "default": true,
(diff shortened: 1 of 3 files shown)


src/modules/flow/timestamp/timestamp.json
```diff
@@ -21,7 +21,27 @@
...
+            "data_type": "boolean",
+            "default": true,
+            "description": "Whether to create an initial packet and send it upon connection to the port OUT or not (wait for packets in the TRIGGER port).",
+            "name": "initial_packet"
+          },
+          {
+            "data_type": "boolean",
+            "default": false,
+            "description": "Ignore case when comparing strings.",
+            "name": "ignore_case"
```
edersondisouza: What is this option about? It seems to not have any code associated to it...

glima: The famous wat. I'll remove that extra option coming from c&p, thanks.

bdilly:
Patch seems good removing "ignore_case" option.

glima:
Alright, fixing locally and merging. Thanks.

PR #790, opened 2015-09-25
Changes required for soletta-demos

bdilly:


bdilly:
Last commit was dropped from this series and sent in a v2.
Others were merged with proposed changes.

PR #791, opened 2015-09-27
conffile: Fix memory leak

ceolin:
filename was leaking in case of sol_file_reader_open() returns NULL.


mbelluzzo:
:+1:

PR #792, opened 2015-09-28
flow: fix issues when node type desc is not enabled

dorileo:
If node type description is not enabled the function
sol_flow_node_named_options_parse_member() is never run and its prototype
becomes broken since struct sol_flow_node_options_member_description is
never declared in that case.

Also the sol-fbp-runner and sol-fbp-generator both depend on description
if it's not present the targets check-fbp and check-fbp-bin should never
run due the lacking of the over mentioned binaries.

In the test/sol-flow.c we should also not build the description tests if
it's not enabled, otherwise the tests will fail at all.

### DIFF
--- a/src/lib/flow/include/sol-flow.h
+++ b/src/lib/flow/include/sol-flow.h
@@ -366,10 +366,12 @@ struct sol_flow_node_type {
 #endif
 };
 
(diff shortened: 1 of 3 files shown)


barbieri:
+1

glima:
LGTM as well

PR #793, opened 2015-09-28
modules/persistence: fix string leak

bdilly:
Default value is leaked if persist_open() fails on
persistence/string.

Also it was lacking proper handling of strdup error.


bdilly:
merged, thanks

PR #796, opened 2015-09-28

ibriano:
Differences since v3:
 * Kept sol-flow-modules as exposed API, but the core of the module
   loading moved to shared/. It will be used by other systems later.

Differences since v2:
 * Moved from common to flow, since it's unlikely that it will be
   helpful for the generic case.

Differences since v1:


barbieri:
+1 lgtm

dorileo:
+1

ibriano:
Merged

PR #798, opened 2015-09-28
flow/test: Fix crash in float-validator

lpereira:
If item couldn't be found in sequence, `op` would be NULL, causing
a NULL pointer dereference in the next line.


ibriano:
Ack

lpereira:
Merged, thanks!

PR #799, opened 2015-09-29
Fix conffile search

mbelluzzo:
It was corrupting "global" argv string and not validating basename output, what was causing a search for *-<appname>.fbp.json instead of *-appname.json


dorileo:
fair...
+2

mbelluzzo:
Merged thanks

PR #800, opened 2015-09-29
mqtt: Allow user and password to be set

rchiossi:
For brokers that implement MQTT spec 3.1, it may be required that a user
and a password to be provided for authentication. This patch adds
support for this feature.


rchiossi:
merged

PR #803, opened 2015-09-29
build: enable intree relative path

dorileo:
There're some cases where we want a module using a common object, i.e
flow/foo/foo.o and flow/bar/bar.o having common code to be shared as
flow/foobar/common.o

With this patch a module can declare its objects using relative path,
like:

obj-$(FLOW_NODE_TYPE_FOO) += foo.mod
obj-foo-$(FLOW_NODE_TYPE_FOO) := foo.o ../foobar/common.o

and with bar as:

obj-$(FLOW_NODE_TYPE_BAR) += bar.mod
obj-bar-$(FLOW_NODE_TYPE_BAR) := bar.o ../foobar/common.o

The build system will normalize the paths and make sure the objects are
built only once and decide where they should be.

### DIFF
--- a/tools/build/Makefile.rules
+++ b/tools/build/Makefile.rules
@@ -35,29 +35,27 @@ extra-headers = \
 	) \
 
 artifact-path = $(foreach curr,$(1), \
-			$(eval intree  := $(if $(filter $(abspath $(top_srcdir))%,$(abspath $(curr))),y,n)) \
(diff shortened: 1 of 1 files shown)


glima:
Mm, with both user places built-in, I get "multiple definition of"... :(

dorileo:
are you sure? I tested this case... double checking again...

dorileo:
ok, applied.

PR #804, opened 2015-09-29
http-client: Allow custom data to be sent on POST

rchiossi:
This patch allows custom data to be sent on POST operations, expanding
from the key=value restriction that we had up to this point. It relies
on content-type to be set via SOL_HTTP_PARAM_HEADER.

### DIFF
--- a/src/lib/comms/include/sol-http.h
+++ b/src/lib/comms/include/sol-http.h
@@ -61,7 +61,8 @@ enum sol_http_param_type {
     SOL_HTTP_PARAM_AUTH_BASIC,
     SOL_HTTP_PARAM_ALLOW_REDIR,
     SOL_HTTP_PARAM_TIMEOUT,
-    SOL_HTTP_PARAM_VERBOSE
+    SOL_HTTP_PARAM_VERBOSE,
+    SOL_HTTP_PARAM_POST_DATA
 };
 
 enum sol_http_status_code {
@@ -100,6 +101,9 @@ struct sol_http_param_value {
         struct {
             int value;
         } integer;
+        struct {
+            struct sol_buffer *value;
+        } data;
     } value;
 };
 
@@ -181,6 +185,12 @@ struct sol_http_response {
     (struct sol_http_param_value) { \
         .type = SOL_HTTP_PARAM_TIMEOUT, \
         .value.integer.value = (setting_) \
+    }
+
+#define SOL_HTTP_REQUEST_PARAM_POST_DATA(data_) \
+    (struct sol_http_param_value) { \
+        .type = SOL_HTTP_PARAM_POST_DATA, \
+        .value.data.value = (data_) \
     }
 
 #define SOL_HTTP_PARAM_FOREACH_IDX(param, itrvar, idx) \
--- a/src/lib/comms/sol-http-client-impl-curl.c
+++ b/src/lib/comms/sol-http-client-impl-curl.c
@@ -685,6 +685,53 @@ set_post_fields_from_params(CURL *curl, struct sol_arena *arena,
 }
 
 static bool
+set_post_data_from_params(CURL *curl, struct sol_arena *arena,
+    const struct sol_http_param *params)
+{
+    struct sol_buffer *data = NULL;
+    struct sol_http_param_value *iter;
+    uint16_t idx;
+    char *tmp;
+    bool type_set = false;
+
+    SOL_VECTOR_FOREACH_IDX (&params->params, iter, idx) {
+        const char *key = iter->value.key_value.key;
+        struct sol_buffer *value = iter->value.data.value;
+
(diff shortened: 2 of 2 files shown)


src/lib/comms/include/sol-http.h
```diff
@@ -100,6 +101,9 @@ struct sol_http_param_value {
         struct {
             int value;
         } integer;
+        struct {
+            struct sol_buffer *value;
```
barbieri: const? Or maybe `struct sol_str_slice`? (we can easily get a slice from a buffer)

src/lib/comms/sol-http-client-impl-curl.c
```diff
@@ -685,6 +685,53 @@ set_post_fields_from_params(CURL *curl, struct sol_arena *arena,
...
+    if (!type_set)
+        SOL_WRN("POST request has data but no content-type was set");
+
+    tmp = strndup((char *)data->data, (int)data->used);
+    if (!tmp)
+        goto cleanup;
+
+    return set_string_option(curl, CURLOPT_POSTFIELDS, arena, tmp);
+
+cleanup:
```
lpereira: No need to have this label, just `return false` on line 725. (Or just use `SOL_NULL_CHECK()`.)

src/lib/comms/sol-http-client-impl-curl.c
```diff
@@ -685,6 +685,53 @@ set_post_fields_from_params(CURL *curl, struct sol_arena *arena,
...
+        }
+    }
+
+    if (!data)
+        return false;
+
+    if (!type_set)
+        SOL_WRN("POST request has data but no content-type was set");
+
+    tmp = strndup((char *)data->data, (int)data->used);
```
barbieri: I wonder why these casts are required. They shouldn't exist.

src/lib/comms/sol-http-client-impl-curl.c
```diff
@@ -685,6 +685,53 @@ set_post_fields_from_params(CURL *curl, struct sol_arena *arena,
...
+    SOL_VECTOR_FOREACH_IDX (&params->params, iter, idx) {
+        const char *key = iter->value.key_value.key;
+        struct sol_buffer *value = iter->value.data.value;
+
+        if (iter->type == SOL_HTTP_PARAM_POST_FIELD) {
+            SOL_WRN("SOL_HTTP_PARAM_POST_FIELD and SOL_HTTP_PARAM_POST_DATA found in parameters."
+                   " Only one can be used at a time");
+            return false;
+        } else if (iter->type == SOL_HTTP_PARAM_HEADER) {
+            type_set = type_set || strcmp(key, "content-type");
```
lpereira: This should be a case-insensitive comparison. Headers are often passed with `Camel-Case-With-Dashes` (i.e. `Content-Type`) and this would fail.

src/lib/comms/sol-http-client-impl-curl.c
lpereira: Will this memory be cleaned when this `curl` handle dies?

src/lib/comms/sol-http-client-impl-curl.c
lpereira: Also, `strcmp()` returns 0 if the strings are equal.

src/lib/comms/include/sol-http.h
rchiossi: Changed it to use slice.

src/lib/comms/sol-http-client-impl-curl.c
rchiossi: encode_key_values() which is the equivalent when post_fields is set for key_values pair was also setting errno, so i just copied that. But indeed, there is no need for the label here. I'll remove it.

src/lib/comms/sol-http-client-impl-curl.c
rchiossi: It will die in set_string_option, just in the line below this.

rchiossi: Also, the cast is needed for sol_buffer, but not for sol_str_slice.

src/lib/comms/sol-http-client-impl-curl.c
rchiossi: Meh, my bad...

src/lib/comms/sol-http-client-impl-curl.c
ceolin: I prefer use SOL_NULL_CHECK. It gives at least some debug information

rchiossi: Changed it to SOL_NULL_CHECK already. ;)

rchiossi:
v2 #820

PR #807, opened 2015-09-29
Changes required for soletta-demos [v2]

bdilly:
v2 changelog:
 * switcher keep state support was rewritten to use
   generic code. Depends on pr #806.

v2 for pr #790


bdilly:
merged

PR #809, opened 2015-09-29
Buffer fixes

lpereira:
Fixes two issues found with the fuzzer.


glima:
Mm, both LGTM.

lpereira:
Merged, thank you.

PR #810, opened 2015-09-29
flow/string-format: Don't crash if precision is 0

lpereira:
Another one found with the fuzzer.

@glima, could you please take a look here, since you wrote this code?


lpereira:
Merged with slight difference after ack from@glima.

PR #811, opened 2015-09-30
flow/file: Fix segmentation fault while sending error packet

lpereira:
Functions taking a printf-style formatting string should never be
allowed to receive user-supplied strings.  In this case, a string has
been sent to `sol_flow_send_error_packet()` after being formatted by
`snprintf()` -- with `%s` and all -- causing the program to crash.

Modify the prototype for `sol_flow_send_error_packet()` so that it's
declared with `__attribute__((printf))`; this way, the compiler should
warn of other potential similar problems.

### DIFF
--- a/src/modules/flow/filter-repeated/filter-repeated.c
+++ b/src/modules/flow/filter-repeated/filter-repeated.c
@@ -144,12 +144,12 @@ error_filter(struct sol_flow_node *node, void *data, uint16_t port, uint16_t con
     free(mdata->msg);
     mdata->msg = strdup(in_value);
     if (!mdata->msg) {
-        sol_flow_send_error_packet(node, errno, sol_util_strerrora(errno));
+        sol_flow_send_error_packet(node, errno, "%s", sol_util_strerrora(errno));
         return -errno;
     }
     mdata->code = code_value;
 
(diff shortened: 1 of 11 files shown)


lpereira:
Please don't review this yet, I'm fixing a lot of compiler warnings due to the attribute change.

src/modules/flow/filter-repeated/filter-repeated.c
```diff
@@ -144,12 +144,12 @@ error_filter(struct sol_flow_node *node, void *data, uint16_t port, uint16_t con
     free(mdata->msg);
     mdata->msg = strdup(in_value);
     if (!mdata->msg) {
-        sol_flow_send_error_packet(node, errno, sol_util_strerrora(errno));
+        sol_flow_send_error_packet(node, errno, "%s", sol_util_strerrora(errno));
```
barbieri: this should be pretty common usage, what if we add a static inline to do this one, just give it the errno and get it formatted and send the packet.

lpereira: Sure, working on it.

rchiossi:
+1

lpereira:
Thanks. Merging with proposed changes.

PR #813, opened 2015-09-30
flow/string-format: Do not use alloca() with user-supplied size

lpereira:
It's undefined behavior to NULL-check the return of alloca(). Use malloc() instead, which will fail, and check that instead.

### DIFF
--- a/src/modules/flow/converter/string-format.c
+++ b/src/modules/flow/converter/string-format.c
@@ -494,17 +494,21 @@ fast_fill(struct sol_buffer *buffer,
 {
     char *str = NULL;
     struct sol_str_slice slice;
+    int r;
 
     assert(start >= 0);
 
-    str = alloca(length + 1);
+    str = malloc(length + 1);
     SOL_NULL_CHECK(str, -ENOMEM);
     memset(str, fill_char, length);
     str[length] = 0;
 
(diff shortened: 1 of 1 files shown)


src/modules/flow/converter/string-format.c
```diff
@@ -494,17 +494,21 @@ fast_fill(struct sol_buffer *buffer,
...
-    str = alloca(length + 1);
+    str = malloc(length + 1);
     SOL_NULL_CHECK(str, -ENOMEM);
     memset(str, fill_char, length);
     str[length] = 0;
 
     slice = SOL_STR_SLICE_STR(str, strlen(str));
 
-    return sol_buffer_set_slice_at(buffer, start, slice);
+    r = sol_buffer_set_slice_at(buffer, start, slice);
```
barbieri: well, given the meaning of this function (fast_fill), it's better to do it directly with `sol_buffer` by forcing it to grow to `used + length + 1` bytes, then `sol_buffer_get_at()` then `memset()`.

`malloc()`, to `memset()`, to `sol_buffer_set_slice_at()` is a bit of a waste.

glima:
The function overrides a certain portion of a buffer, not growing it. This is one of the portions where the original logic was kept.

lpereira:
Superseded by #814.

PR #814, opened 2015-09-30
Do not use alloca() with user-supplied sizes

lpereira:
This function does not have defined behavior when allocation fails, so
checking for NULL is not going to work.  Review all usages of alloca(),
strndupa() and strdupa() throughout the code base, reducing the amount
of damage that could happen if the allocation were to fail.

### DIFF
--- a/src/modules/flow/converter/string-format.c
+++ b/src/modules/flow/converter/string-format.c
@@ -492,19 +492,16 @@ fast_fill(struct sol_buffer *buffer,
     ssize_t length,
     unsigned char fill_char)
 {
-    char *str = NULL;
-    struct sol_str_slice slice;
+    int r;
 
     assert(start >= 0);
 
-    str = alloca(length + 1);
-    SOL_NULL_CHECK(str, -ENOMEM);
-    memset(str, fill_char, length);
-    str[length] = 0;
+    r = sol_buffer_ensure(buffer, buffer->used + length + 1);
+    SOL_INT_CHECK(r, < 0, r);
 
-    slice = SOL_STR_SLICE_STR(str, strlen(str));
+    memset(sol_buffer_at(buffer, start), fill_char, length);
 
-    return sol_buffer_set_slice_at(buffer, start, slice);
+    return sol_buffer_ensure_nul_byte(buffer);
 }
 
 static int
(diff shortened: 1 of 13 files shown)


lpereira:
@glima -- check-fbp is failing with this PR mostly under `string-format`. Could you please take a look if my changes are correct?

src/modules/flow/converter/string-format.c
```diff
@@ -594,14 +591,16 @@ insert_numbers(struct sol_buffer *out,
...
+        char *str;
 
-        SOL_NULL_CHECK(str, -ENOMEM);
-        r = snprintf(str, n_zeros + 1, "%0*d", (int)(long)n_zeros, 0);
-        SOL_INT_CHECK(r, < 0, r);
+        r = asprintf(&str, "%0*d", (int)(long)n_zeros, 0);
+        SOL_INT_CHECK(r, < 0, -ENOMEM);
+
+        slice = SOL_STR_SLICE_STR(str, r);
+        r = sol_buffer_insert_slice(out, pos, slice);
```
barbieri: no need to `asprintf()` and then `sol_buffer_insert_slice()`, we should have `sol_buffer_insert_printf()`. @glima, please review this whole file as you did the port, need to take care to use `sol_buffer` infra as much as possible.

src/modules/flow/flower-power/flower-power.c
```diff
@@ -402,16 +402,15 @@ get_timestamp(struct sol_json_token *value, struct timespec *timestamp)
...
-        return -EINVAL;
-    }
 
-    timestamp_str = strptime(timestamp_str, "%Y-%m-%dT%H:%M:%SZ", &time_tm);
-    if (!timestamp_str) {
+    timestamp_str = strndup(value->start, sol_json_token_get_size(value));
+    p = strptime(timestamp_str, "%Y-%m-%dT%H:%M:%SZ", &time_tm);
+    free(timestamp_str);
+
+    if (!p || !*p) {
```
lpereira: This should be `if (!p || *p)`. Fixed locally.

src/modules/flow/converter/string-format.c
glima: Agreed about ``sol_buffer_insert_printf()``, thanks for the tip. I pointed out what might be the error in the PR. The rest of the changes in the file should be fine, I guess.

lpereira:
I'm making the changes to make this pass and merge. The changes to add the convenience functions to sol_buffer won't be made this time around, as there are many other more important things to do. Thanks for the review.

PR #815, opened 2015-09-30
Eeprom

edersondisouza:
memmap-storage: `path` property now accepts `create,i2c,...` commands (the same from IIO) to create an I2C device - useful for Calamari EEPROM.

### DIFF
--- a/src/lib/io/include/sol-memmap-storage.h
+++ b/src/lib/io/include/sol-memmap-storage.h
@@ -79,7 +79,7 @@ extern "C" {
 
 struct sol_memmap_map {
     uint8_t version; /**< Version of map. Functions will refuse to read/write on storage if this version and the one storad differs */
-    char *path; /**< Where to find the storage. Under Linux, it is the file mapping the storage, like @c /dev/nvram */
+    const char *path; /**< Where to find the storage. Under Linux, it is the file mapping the storage, like @c /dev/nvram */
     struct sol_str_table_ptr entries[]; /**< Entries on map, containing name, offset and size */
 };
 
--- a/src/lib/io/include/sol-memmap-storage.h
+++ b/src/lib/io/include/sol-memmap-storage.h
@@ -79,7 +79,15 @@ extern "C" {
 
 struct sol_memmap_map {
     uint8_t version; /**< Version of map. Functions will refuse to read/write on storage if this version and the one storad differs */
-    const char *path; /**< Where to find the storage. Under Linux, it is the file mapping the storage, like @c /dev/nvram */
+    const char *path; /**< Where to find the storage. Under Linux, it is the file mapping the storage, like @c /dev/nvram.
+                       * Optionally, it can also be of form <tt> create,\<bus_type\>,\<rel_path\>,\<devnumber\>,\<devname\> </tt>, where:
+                       * @arg @a bus_type is the bus type, supported values are: i2c
+                       * @arg @a rel_path is the relative path for device on '/sys/devices',
+                       * like 'platform/80860F41:05'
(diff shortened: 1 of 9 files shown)


edersondisouza:
I'm checking the `make samples` issues.

edersondisouza:
Issue with `make sample` fixed. A new commit was added to fix `sol-fbp-generator`.

src/lib/io/include/sol-memmap-storage.h
```diff
@@ -79,7 +79,15 @@ extern "C" {
...
-    char *path; /**< Where to find the storage. Under Linux, it is the file mapping the storage, like @c /dev/nvram */
+    const char *path; /**< Where to find the storage. Under Linux, it is the file mapping the storage, like @c /dev/nvram.
+                       * Optionally, it can also be of form <tt> create,\<bus_type\>,\<rel_path\>,\<devnumber\>,\<devname\> </tt>, where:
+                       * @arg @a bus_type is the bus type, supported values are: i2c
+                       * @arg @a rel_path is the relative path for device on '/sys/devices',
+                       * like 'platform/80860F41:05'
+                       * @arg @a devnumber is device number on bus, like 0x50
+                       * @arg @a devname is device name, the one recognized by its driver
+                       */
+    char *resolved_path; /**< Reserved */
```
barbieri: can't you store this internally somewhere? It's always a PITA to have the user-structure to be modified... say the `sol_memmap_map` is declared as `static const `... OOoopps! You got that issue, just modified sol-fbp-generator/main.c ;-) bad decision.

src/lib/io/sol-memmap-storage.c
```diff
@@ -343,19 +355,123 @@ check_map(const struct sol_memmap_map *map)
...
+#endif
+}
+
 SOL_API int
-sol_memmap_add_map(const struct sol_memmap_map *map)
+sol_memmap_add_map(struct sol_memmap_map *map)
 {
+    SOL_NULL_CHECK(map, -EINVAL);
+
+    if (strstartswith(map->path, "create,i2c,")) {
```
barbieri: the best solution here is to duplicate `map` (just the struct, not the entries), setting its path to the resolved one and add this to a "to_free" vector, that is cleaned up later on `sol_memmap_remove_map()`.

you can `#ifdef USE_I2C` around this block.

src/modules/flow/int/int.json
```diff
@@ -123,6 +123,14 @@
             "process": "reset_process"
           },
           "name": "RESET"
+        },
+        {
+          "data_type": "any",
```
barbieri: not `"any"`, you require it to be an irange in the code.

src/lib/io/include/sol-memmap-storage.h
edersondisouza: Yeah, I was kinda lazy on it, probably a `sol_str_table` will suffice here.

src/modules/flow/int/int.json
edersondisouza: Ops!

src/lib/io/include/sol-memmap-storage.h
edersondisouza: Hmmm... `sol_str_table` is the other way round...

src/lib/io/sol-memmap-storage.c
edersondisouza: But `entries` is actually an array at the end of the struct, not a pointer to an array (so it's possible to use `sol_str_table`). So I guess it'd be necessary to duplicate the entries, or use something as a dictionary.

edersondisouza:
v2 on #816

PR #816, opened 2015-10-01
Eeprom v2

edersondisouza:
v2:
Keep generated memory mapped map `const`.


edersondisouza:
v3 on #831

Issue #746, opened 2015-09-16
JSON Packets & Node Types


barbieri:
It would be nice to have a JSON packet type, it's basically the same as the string, but will be used to differentiate ports and avoid feeding random text to JSON processing nodes. Of course one can add converters from `BLOB` and `STRING`, that will take care of validating using `sol-json.h` and if it passes will forward the packet.

Suggested Packet Types:
 * `JSON_OBJECT`: used for objects (`{...}`)
 * `JSON_ARRAY`:  used for arrays (`[...]`)

These packets should be implemented on top of `struct sol_blob` and keep references to the underlying memory, do not duplicate it to get slices. All other JSON types should be implemented on top of basic types such as `IRANGE`, `DRANGE`, `STRING`, `BOOLEAN`.

## Reading JSON

In order to make FBP easier to deal with JSON, we should offer nodes based on JSONPath http://goessner.net/articles/JsonPath/ to extract properties. This allows to reach a string or integer nested into deep structure with a single node.

Suggested Node Types:
 * `json/object-get-path`:
    * Options:
      * `path` (string): the JSON path to access the child element.
    * Input Ports:
      * `IN` (json object, required): the input json to get the child element
      * `PATH` (string, optional): allows changing the `path` given as option.
    * Output Ports:
      * `STRING` (string): the string at the given path.
      * `BOOLEAN` (boolean): the boolean at the given path.
      * `INT` (irange): the integer number at the given path (min=INT32_MIN, max=INT32_MAX, step=1)
      * `FLOAT` (drange): the floating point number at the given path (min=-DBL_MAX, max=DBL_MAX, step=DBL_MIN)
      * `OBJECT` (json object): the child JSON object at the given path.
      * `ARRAY` (json array): the child JSON array at the given path.
      * `NULL` (empty): empty packet if the object is `null``
 * `json/object-get-key`:
    * Options:
      * `key` (string): the key of the object child to access.
    * Input Ports:
      * `IN` (json object, required): the input json to get the child element
      * `KEY` (string, optional): allows changing the `key` given as option.
    * Output Ports:
      * `STRING` (string): the string at the given key.
      * `BOOLEAN` (boolean): the boolean at the given key.
      * `INT` (irange): the integer number at the given key (min=INT32_MIN, max=INT32_MAX, step=1)
      * `FLOAT` (drange): the floating point number at the given key (min=-DBL_MAX, max=DBL_MAX, step=DBL_MIN)
[...]

PR #818, opened 2015-10-01
Introduce json-object and json-array packet types

otaviobp:
Adding two packet types to handle json content as well as some nodes
to handle JSON objects and arrays.

This pull req is the beggining of the implementation proposed in task #746
Not all proposed nodes were implemented, but I am intending to merge them
as the development flows


src/bin/sol-fbp-runner/inspector.c
```diff
@@ -172,7 +172,9 @@ inspector_show_packet(const struct sol_flow_packet *packet)
             fprintf(stdout, "<\"%s\">", v);
             return;
         }
-    } else if (type == SOL_FLOW_PACKET_TYPE_BLOB) {
+    } else if (type == SOL_FLOW_PACKET_TYPE_BLOB ||
+        type == SOL_FLOW_PACKET_TYPE_JSON_OBJECT ||
+        type == SOL_FLOW_PACKET_TYPE_JSON_ARRAY) {
```
barbieri: meh, kinda bad to use the blob format for the inspector. You could do something like:

```c
if (type == SOL_FLOW_PACKET_TYPE_JSON_OBJECT || SOL_FLOW_PACKET_TYPE_JSON_ARRAY) {
    struct sol_blob *v;
    if (sol_flow_packet_get_blob(packet, &v) == 0) {
        fprintf(stdout, "%.*s", (int)v->size, (char *)v->data);
    }
}
```

src/lib/flow/include/sol-flow-packet.h
```diff
@@ -122,6 +124,11 @@ struct sol_flow_packet *sol_flow_packet_new_string_take(char *value);
 struct sol_flow_packet *sol_flow_packet_new_blob(const struct sol_blob *value);
 int sol_flow_packet_get_blob(const struct sol_flow_packet *packet, struct sol_blob **value);
 
+struct sol_flow_packet *sol_flow_packet_new_json_object(const struct sol_blob *value);
```
barbieri: would be nice to document in the header that we expect a valid JSON blob, that is, a string fragment with a JSON object (or array in the other packet), it may or not be NULL terminated.

src/lib/flow/sol-flow-packet.c
```diff
@@ -448,6 +448,54 @@ sol_flow_packet_get_blob(const struct sol_flow_packet *packet, struct sol_blob *
     return sol_flow_packet_get(packet, value);
 }
 
+static const struct sol_flow_packet_type _SOL_FLOW_PACKET_TYPE_JSON_OBJECT  = {
+    .api_version = SOL_FLOW_PACKET_TYPE_API_VERSION,
+    .name = "Json-Object",
```
barbieri: please use JSON

src/modules/flow/converter/converter.c
```diff
@@ -2066,4 +2067,81 @@ string_to_timestamp_convert(struct sol_flow_node *node, void *data, uint16_t por
...
+static bool
+json_validate(struct sol_blob *blob, enum sol_json_type type)
+{
+    struct sol_json_scanner scanner;
+    struct sol_json_token token;
+
+    sol_json_scanner_init(&scanner, blob->mem, blob->size);
+    return sol_json_scanner_next(&scanner, &token) &&
+           sol_json_token_get_type(&token) == type &&
+           sol_json_scanner_skip_over(&scanner, &token);
```
barbieri: would be too much to check if after skip-over token->end == ((char *)blob->mem + blob->size)? eventually `-1` if there is a trailing null byte in the blob.

src/modules/flow/json/Kconfig
```diff
@@ -0,0 +1,3 @@
+config FLOW_NODE_TYPE_JSON
+	tristate "Node type: json"
+	default m
```
barbieri: add some `help`:-)

src/modules/flow/json/json.c
```diff
@@ -0,0 +1,221 @@
...
+
+#include "sol-flow/json.h"
+#include "sol-flow-internal.h"
+
+#include <sol-json.h>
+#include <sol-util.h>
+#include <sol-types.h>
+#include <errno.h>
+
+static const struct sol_blob_type sub_json_blob_type = {
```
barbieri: when I saw this I thought we should use that one defined in `sol-blob.h` that does no-free... but we don't have one there. Please add one similar to `SOL_BLOB_TYPE_DEFAULT` called `SOL_BLOB_TYPE_NOFREE`, document it won't free data (and document the other saying it will free data).

src/modules/flow/json/json.c
```diff
@@ -0,0 +1,221 @@
...
+{
+    struct sol_json_scanner scanner;
+    const char *mem;
+
+    mem = token->start;
+    sol_json_scanner_init(&scanner, parent->mem, parent->size);
+    if (!sol_json_scanner_skip_over(&scanner, token))
+        return NULL;
+
+    if (sol_json_token_get_type(token) != SOL_JSON_TYPE_OBJECT_END) {
```
barbieri: this is similar to your validator macro in the converter. Please add a helper function in `sol-json.h` that validates and use it here and there.

src/modules/flow/json/json.c
```diff
@@ -0,0 +1,221 @@
...
+    int r;
+
+    if (!mdata->key[0] || !mdata->json_object)
+        return 0;
+
+    len = strlen(mdata->key);
+    sol_json_scanner_init(&scanner, mdata->json_object->mem,
+        mdata->json_object->size);
+    SOL_JSON_SCANNER_OBJECT_LOOP (&scanner, &token, &key, &value, reason) {
+        if (sol_json_token_str_eq(&key, mdata->key, len)) {
```
barbieri: you're not handling JSON Path here and it's okay, but you should prepare your code to.

To do so, make this whole loop a function and give it the path. Then when you want to add the JSON Path variant, what you can do is extract the pieces from the path (ie: skip `/`, get the key up to the next `/` or `\0`, then recurse with the remaining key slice).

You can also do an initial check if `len == 0`, there is no key and you can return as nothing will be found.

src/modules/flow/json/json.c
otaviobp: Actually they are slightly different. The only thing in common is the use of  sol_json_scanner_init and sol_json_scanner_skip_over and this is already exported in sol-json.h.
Anyway, I moved the validator function to sol-json.h to be used in converted, because it makes sense to have that helper over there.

PR #819, opened 2015-10-01
Http server

ceolin:
Re-factor of http-server node. Use class methods to avoid duplicated code. This is the very first step to create the http's group.

### DIFF
--- a/src/modules/flow/http-server/http-server.c
+++ b/src/modules/flow/http-server/http-server.c
@@ -55,7 +55,6 @@
 #define DOUBLE_STRING_LEN 64
 
 struct http_data {
-    struct sol_flow_node *node;
     union {
         struct sol_irange i;
         struct sol_drange d;
@@ -66,12 +65,85 @@ struct http_data {
     char *path;
 };
 
+struct http_server_node_type {
+    struct sol_flow_node_type base;
+    int (*post_cb)(struct http_data *mdata, struct sol_flow_node *node,
+        struct sol_http_param_value *value);
+    int (*response_cb)(struct http_data *mdata, struct sol_buffer *content, bool json);
+    int (*process_cb)(struct http_data *mdata, const struct sol_flow_packet *packet);
+    void (*send_packet)(struct http_data *mdata, struct sol_flow_node *node);
+};
+
 static struct sol_http_server *server = NULL;
 static int init_count = 0;
 
 static int
-start_server(struct http_data *http, const char *path,
-    int (*handler)(void *data, struct sol_http_request *request))
+common_response_cb(void *data, struct sol_http_request *request)
+{
+    int r = 0;
+    uint16_t idx;
+    bool send_json = false;
+    enum sol_http_method method;
+    struct sol_flow_node *node = data;
+    struct http_data *mdata = sol_flow_node_get_private_data(node);
+    const struct http_server_node_type *type;
+    struct sol_http_param_value *value;
+    struct sol_http_response response = {
+        .api_version = SOL_HTTP_RESPONSE_API_VERSION,
+        .content = SOL_BUFFER_INIT_EMPTY,
+        .param = SOL_HTTP_REQUEST_PARAM_INIT,
+        .response_code = SOL_HTTP_STATUS_OK
+    };
+
+    type = (const struct http_server_node_type *)
+        sol_flow_node_get_type(node);
+
+    method = sol_http_request_get_method(request);
+    response.url = sol_http_request_get_url(request);
+
+    SOL_HTTP_PARAM_FOREACH_IDX (sol_http_request_get_params(request), value, idx) {
+        switch (value->type) {
+        case SOL_HTTP_PARAM_POST_FIELD:
+            r = type->post_cb(mdata, node, value);
+            SOL_INT_CHECK_GOTO(r, < 0, end);
+
+            r = sol_http_server_set_last_modified(server, mdata->path, time(NULL));
+            SOL_INT_CHECK_GOTO(r, < 0, end);
+            break;
+        case SOL_HTTP_PARAM_HEADER:
+            if (streq(value->value.key_value.key, HTTP_HEADER_ACCEPT)) {
+                if (strstr(value->value.key_value.value, HTTP_HEADER_CONTENT_TYPE_JSON))
+                    send_json = true;
+            }
+            break;
+        default:
+            break;
+        }
+    }
+
+    r = type->response_cb(mdata, &response.content, send_json);
+    SOL_INT_CHECK_GOTO(r, < 0, end);
+
+    r = sol_http_param_add(&response.param, SOL_HTTP_REQUEST_PARAM_HEADER(
+        HTTP_HEADER_CONTENT_TYPE, (send_json) ? HTTP_HEADER_CONTENT_TYPE_JSON : HTTP_HEADER_CONTENT_TYPE_TEXT));
+    SOL_INT_CHECK_GOTO(r, != true, end);
+
+    r = sol_http_server_send_response(request, &response);
+    SOL_INT_CHECK_GOTO(r, < 0, end);
+
+    if ((method == SOL_HTTP_METHOD_POST) && type->send_packet)
+        type->send_packet(mdata, node);
+
+end:
+    sol_buffer_fini(&response.content);
+    sol_http_param_free(&response.param);
+
+    return 0;
+}
+
+static int
+start_server(struct http_data *http, struct sol_flow_node *node,
+    const char *path)
 {
     int r;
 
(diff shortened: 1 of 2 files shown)


src/modules/flow/http-server/http-server.c
```diff
@@ -66,62 +65,28 @@ struct http_data {
     char *path;
 };
 
+struct http_server_node_type {
+    struct sol_flow_node_type base;
+    int (*post_cb)(struct http_data *mdata, struct sol_flow_node *node,
+        struct sol_http_param_value *value);
+    int (*response_cb)(struct http_data *mdata, struct sol_buffer *content, bool json);
+    int (*process_cb)(struct http_data *mdata, const struct sol_flow_packet *packet);
+    void (*send_packet)(struct http_data *mdata, struct sol_flow_node *node);
```
barbieri: for consistency, add `_cb` to `send_packet` or remove that suffix from other callbacks.

src/modules/flow/http-server/http-server.c
```diff
@@ -157,30 +121,18 @@ boolean_response_cb(void *data, struct sol_http_request *request)
...
-        r = snprintf(response_str, sizeof(response_str), "{\"%s\": %s}", mdata->path,
-            mdata->value.b == true ? "true" : "false");
-    } else {
-        r = snprintf(response_str, sizeof(response_str), "%s", mdata->value.b == true ? "true" : "false");
-    }
-
+    r = type->response_cb(mdata, &response.content, send_json);
     SOL_INT_CHECK_GOTO(r, < 0, end);
 
     r = sol_http_param_add(&response.param, SOL_HTTP_REQUEST_PARAM_HEADER(
```
barbieri: I'd move this inside `response_cb()` as well, change to send the whole `&response` instead of just `&response.content`.  Then you don't need to use `send_json` in here. And you allow responses to add extra headers/parameters as they wish.

src/modules/flow/http-server/http-server.c
```diff
@@ -190,23 +142,50 @@ boolean_response_cb(void *data, struct sol_http_request *request)
...
+    http->path = strdup(path);
+    SOL_NULL_CHECK_GOTO(http->path, err);
+
+    r = sol_http_server_register_handler(server, http->path, common_response_cb, node);
+    SOL_INT_CHECK_GOTO(r, < 0, err_handler);
 
     return 0;
+
+err_handler:
+    free(http->path);
```
barbieri: since you don't free(http), it's recommended to `http->path = NULL` to avoid the removed pointer being used later.

src/modules/flow/http-server/http-server.c
ceolin: I was in doubt about this name, in my head it had a different meaning yesterday night. I'll change.

src/modules/flow/http-server/http-server.c
```diff
@@ -190,23 +142,50 @@ boolean_response_cb(void *data, struct sol_http_request *request)
...
+        server = NULL;
+    }
+    return -1;
+}
+
+static void
+stop_server(struct http_data *http)
+{
+    sol_http_server_unregister_handler(server, http->path);
+    free(http->path);
```
barbieri: also `http->path = NULL;`

src/modules/flow/http-server/http-server.c
barbieri: or you move the `strdup()` and `free()` to outside this function.

ceolin: this start_server is called only from the open(), when it fails open() returns something different from 0 and then the http_data is destroyied, so it's not necessary to set it NULL

src/modules/flow/http-server/http-server.c
ceolin: same here, it's called only from close(), the http_data will be destroyied

src/modules/flow/http-server/http-server.c
```diff
@@ -214,79 +193,100 @@ boolean_close(struct sol_flow_node *node, void *data)
...
+
+static int
+boolean_response_cb(struct http_data *mdata, struct sol_buffer *content, bool json)
+{
+    int r;
+    char response_str[128];
+
+    if (json) {
+        r = snprintf(response_str, sizeof(response_str), "{\"%s\": %s}", mdata->path,
+            mdata->value.b == true ? "true" : "false");
```
barbieri: this won't work later to convert to group, since if you call this function for each group element, the result will be `{"/path": {"/path": true}}` which is weird. Maybe have another virtual to just append the JSON value, then this function would call it:
```c
if (json) {
    sol_buffer_reset(content);
    r = sol_buffer_append_printf(content, "{\"%s\":", mdata->path);
   SOL_INT_CHECK(r, < 0, r);
   r = type->json_append_value(mdata, content);
   SOL_INT_CHECK(r, < 0, r);
   r = sol_buffer_append_slice(content, sol_str_slice_from_str("}"));
   SOL_INT_CHECK(r, < 0, r);
}   
```

this also shows that you should be using `sol_buffer_append_printf()` instead of `snprintf()` + `sol_buffer_set_slice()`.

src/modules/flow/http-server/http-server.c
```diff
@@ -214,79 +193,100 @@ boolean_close(struct sol_flow_node *node, void *data)
...
+static int
+string_response_cb(struct http_data *mdata, struct sol_buffer *content, bool json)
+{
+    int r = 0;
+    char *response_str;
+
+    if (json) {
         char *escaped_str;
         size_t escaped_len = sol_json_calculate_escaped_string_len(mdata->value.s);
         escaped_str = malloc(escaped_len);
```
barbieri: use the buffer potential! :-)

If you use the `type->json_append_value()` as I said before, this function would be:

```c
if (json) {
   sol_buffer_reset(buffer);
   r = sol_buffer_append_printf(buffer, "{\"%s\":", mdata->path); // see this is similar to the boolean? This could be in the caller function! :-)
   r = type->json_append_value(mdata, content);
   SOL_INT_CHECK(r, < 0, r);
   r = sol_buffer_append_slice(content, sol_str_slice_from_str("}"));
   SOL_INT_CHECK(r, < 0, r);
}
```

With that `json_append_value` being implemented as:
```c
escaped_len = sol_json_calculate_escaped_string_len(mdata->value.s)
r = sol_buffer_ensure(buffer, buffer->used + 2 + escaped_len);
SOL_INT_CHECK(r, < 0, r);
r = sol_buffer_append_slice(buffer, sol_str_slice_from_string("\"")); // we should have a sol_buffer_append_char() and sol_buffer_insert_char()!
SOL_INT_CHECK(r, < 0, r);
sol_json_escape_string(mdata->value.s, sol_buffer_at_end(buffer), escaped_len);
r = sol_buffer_append_slice(buffer, sol_str_slice_from_string("\"")); // we should have a sol_buffer_append_char() and sol_buffer_insert_char()!
SOL_INT_CHECK(r, < 0, r);
```

src/modules/flow/http-server/http-server.c
```diff
@@ -295,44 +295,45 @@ string_response_cb(void *data, struct sol_http_request *request)
...
+        free(mdata->value.s);
+        mdata->value.s = strdup(value->value.key_value.value);
+        if (!mdata->value.s) {
+            SOL_WRN("Could not duplicate the string");
+            mdata->value.s = NULL;
+            return -ENOMEM;
+        }
+    } else {
+        return -EINVAL;
     }
```
barbieri: here you could do a small but helpful optimization, create a:
```c
static int
string_replace_if_changed(char **storage, const char *input)
{
   if (*storage && input) {
      size_t storage_len = strlen(*storage);
      size_t input_len = strlen(input);
      if (storage_len == input_len && memcmp(*storage, input, storage_len) == 0) {
         return 0;
      } else if (storage_len < input_len) {
          char *tmp = realloc(*storage, input_len + 1);
          SOL_NULL_CHECK(tmp, -ENOMEM);
          *storage = tmp;
      }
     memcpy(*storage, input, input_len + 1);
   } else if (*storage) {
      free(*storage);
      *storage = NULL;
   } else {
      *storage = strdup(input);
      SOL_NULL_CHECK(*storage, -ENOMEM);
   }
  return 0;
}
```

barbieri: btw, this code could go in sol `sol-util.h` so more places could do this, like some string nodes (@glima).

ceolin: ack, i'm going to add it

src/modules/flow/http-server/http-server.c
ceolin: i think it's not necessary. We can set path befor call this function, so this function only set the value.

ceolin:
fixes on https://github.com/solettaproject/soletta/pull/839

PR #820, opened 2015-10-01
http-client: Allow custom data to be sent on POST

rchiossi:
This patch allows custom data to be sent on POST operations, expanding
from the key=value restriction that we had up to this point. It relies
on content-type to be set via SOL_HTTP_PARAM_HEADER.

### DIFF
--- a/src/lib/comms/include/sol-http.h
+++ b/src/lib/comms/include/sol-http.h
@@ -61,7 +61,8 @@ enum sol_http_param_type {
     SOL_HTTP_PARAM_AUTH_BASIC,
     SOL_HTTP_PARAM_ALLOW_REDIR,
     SOL_HTTP_PARAM_TIMEOUT,
-    SOL_HTTP_PARAM_VERBOSE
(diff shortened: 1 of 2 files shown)


src/lib/comms/include/sol-http.h
```diff
@@ -61,7 +61,8 @@ enum sol_http_param_type {
     SOL_HTTP_PARAM_AUTH_BASIC,
     SOL_HTTP_PARAM_ALLOW_REDIR,
     SOL_HTTP_PARAM_TIMEOUT,
-    SOL_HTTP_PARAM_VERBOSE
+    SOL_HTTP_PARAM_VERBOSE,
+    SOL_HTTP_PARAM_POST_DATA
```
barbieri: no big deal, but I'd move this close to `POST_FIELD`, it would be nice to also document they conflict in the header file.

lpereira:
LGTM, just fix @barbieri's nit before merging.

bdilly:
It was merged already

PR #831, opened 2015-10-01
Eeprom v3

edersondisouza:
v3:
To avoid simplify i2c paths, now keeping a 'shadow' map with resolved path. To avoid duplicating entries, `sol_memmap_map` structure was changed to have `entries` as a pointer to an array. So, generator and conffile were changed accordingly.

v2:


edersondisouza:
Initialising statics with compound literals, as `SOL_MEMMAP_ENTRY_BIT_SIZE` does, is not allowed on C99 mode of gcc 4.x. That's why build is failing on bots. I'll change this and send a v4.

edersondisouza:
v4 on #836

PR #832, opened 2015-10-02
Introduce json-object and json-array packet types - v2

otaviobp:
Version 2 of #818

Changelog from v1:
 - json nodes help added
 - inspector fixed and JSON packets are now handled properly
 - Json validator moved to sol-json.h and it is checking now if there
 is no trailing data after JSON object or array.
 - all other minor proposals in last review


src/lib/parsers/sol-json.c
```diff
@@ -546,3 +546,20 @@ sol_json_double_to_str(const double value, char *buf, size_t len)
 
     return 0;
 }
+
+SOL_API bool
+sol_json_validate(struct sol_json_scanner *scanner, enum sol_json_type type)
```
lpereira: What is this validating?

src/modules/flow/json/json.c
```diff
@@ -0,0 +1,226 @@
...
+        case SOL_JSON_TYPE_NUMBER:
+            r = sol_json_token_get_double(&value, &value_float.val);
+            SOL_INT_CHECK(r, < 0, r);
+            r = sol_flow_send_drange_packet(node,
+                SOL_FLOW_NODE_TYPE_JSON_OBJECT_GET_KEY__OUT__FLOAT,
+                &value_float);
+            SOL_INT_CHECK(r, < 0, r);
+
+            r = sol_json_token_get_int32(&value, &value_int.val);
+            if (r == -EINVAL) //Non an integer number
```
lpereira: `s/Non/Not/`

Please also use `/* ... */` comments.

src/lib/parsers/sol-json.c
otaviobp: I'll add this comment: Check if scanner is pointing to a valid string with a json element of informed type. May or may not be NULL terminated

otaviobp:
Iscaro, good point about adding support to print json objects/array in console.
I don't know if we should move it to data_type definitions. Why isn't blob in data_type definitions?

src/modules/flow/json/json.c
otaviobp: I don't like the idea of sending the integer value if we have a float. I think this is confusing. If the field is a float number and user wants the int value, he should get the float value and convert it to int.

src/lib/common/include/sol-types.h
```diff
@@ -242,6 +242,12 @@ struct sol_blob_type {
  * The default type uses free() to release the blob's memory
  */
 extern const struct sol_blob_type *SOL_BLOB_TYPE_DEFAULT;
+/*
+ * The no-free type doesn't free blob's memory. Used when pointing to inner
+ * position of a pre existing blob or any other case when blob's memory
+ * shouldn't be freed
```
barbieri: hum... this may cause a confusion. The blob `data` memory (internal pointed) is not freed. The `struct sol_blob` itself will be freed. Would try to clarify it.

PR #833, opened 2015-10-02
comms:http:server: Support to serve static files

ceolin:
Added support to serve static files. It has less precedence than the
registered handlers. So, when a request comes it first look for a
handler and if any is found it looks in the registered paths.

Signed-off-by: Flair's Ceolin <flavio.ceolin@intel.com>

### DIFF
--- a/src/lib/comms/sol-http-server-impl-microhttpd.c
+++ b/src/lib/comms/sol-http-server-impl-microhttpd.c
@@ -31,9 +31,13 @@
  */
 
 #include <errno.h>
+#include <fcntl.h>
 #include <microhttpd.h>
 #include <stdio.h>
 #include <stdlib.h>
+#include <sys/stat.h>
+#include <sys/types.h>
+#include <unistd.h>
 
 #include "sol-http-server.h"
 #include "sol-log.h"
@@ -63,6 +67,7 @@ struct sol_http_request {
 
 struct sol_http_server {
     struct MHD_Daemon *daemon;
+    struct sol_ptr_vector dirs;
     struct sol_vector handlers;
     struct sol_vector fds;
     struct sol_vector requests;
@@ -231,12 +236,33 @@ err:
 }
 
 static int
+get_static_file(const char *dir, const char *url)
+{
+    int ret;
+    char path[PATH_MAX];
+
+    /* url given by microhttpd starts from /. For instance,
+     * https://www.solettaproject.com => url == /
+     * https://www.solettaproject.com/thankyou => url == /thankyou
+     */
+    ret = snprintf(path, sizeof(path), "%s/%s", dir,
+        *(url + 1) ? (url + 1) : "index.html");
+    SOL_INT_CHECK(ret, < 0, NULL);
+
+    /*  According with microhttpd fd will be closed when response is
+     *  destroyed and fd should be in 'blocking' mode
+     */
+    return open(path, O_RDONLY | O_CLOEXEC);
+}
+
+static int
 http_server_handler(void *data, struct MHD_Connection *connection, const char *url, const char *method,
     const char *version, const char *upload_data, size_t *upload_data_size, void **ptr)
 {
-    int ret;
+    int ret, fd;
     uint16_t i;
-    struct MHD_Response *mhd_response;
+    char *dir;
+    struct MHD_Response *mhd_response = NULL;
     struct sol_http_server *server = data;
     struct http_handler *handler;
     struct sol_http_request *req = *ptr;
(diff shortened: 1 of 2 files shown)


src/lib/comms/sol-http-server-impl-microhttpd.c
```diff
@@ -231,12 +236,33 @@ headers_iterator(void *data, enum MHD_ValueKind kind, const char *key, const cha
...
+get_static_file(const char *dir, const char *url)
+{
+    int ret;
+    char path[PATH_MAX];
+
+    /* url given by microhttpd starts from /. For instance,
+     * https://www.solettaproject.com => url == /
+     * https://www.solettaproject.com/thankyou => url == /thankyou
+     */
+    ret = snprintf(path, sizeof(path), "%s/%s", dir,
```
lpereira: If `url` is `/../../../../../../../../../../../../../../../etc/passwd`, people will probably be able to get `/etc/passwd`. Use `realpath()`, see if the path starts with the root directory's `realpath()` up to and including the terminating `/`.

src/lib/comms/sol-http-server-impl-microhttpd.c
```diff
@@ -231,12 +236,33 @@ headers_iterator(void *data, enum MHD_ValueKind kind, const char *key, const cha
...
+        *(url + 1) ? (url + 1) : "index.html");
+    SOL_INT_CHECK(ret, < 0, NULL);
+
+    /*  According with microhttpd fd will be closed when response is
+     *  destroyed and fd should be in 'blocking' mode
+     */
+    return open(path, O_RDONLY | O_CLOEXEC);
```
lpereira: Call `fstat(fd, &st)` and bail out if `st.st_mode & (S_IRUSR | S_IRGRP | S_IROTH)` fails.

src/lib/comms/sol-http-server-impl-microhttpd.c
```diff
@@ -231,12 +236,33 @@ headers_iterator(void *data, enum MHD_ValueKind kind, const char *key, const cha
 }
 
 static int
+get_static_file(const char *dir, const char *url)
+{
+    int ret;
+    char path[PATH_MAX];
+
+    /* url given by microhttpd starts from /. For instance,
+     * https://www.solettaproject.com => url == /
```
lpereira: `while (startswith(url, '/')) url++`

src/lib/comms/sol-http-server-impl-microhttpd.c
```diff
@@ -293,10 +319,29 @@ http_server_handler(void *data, struct MHD_Connection *connection, const char *u
         return MHD_YES;
     }
 
-end:
+    SOL_PTR_VECTOR_FOREACH_IDX (&server->dirs, dir, i) {
+        struct stat st;
+        fd = get_static_file(dir, url);
```
lpereira: Might be a good idea to set `status` to "Access denied" on `EACCESS` and "Unavailable" on ENFILE/EMFILE.

src/lib/comms/sol-http-server-impl-microhttpd.c
```diff
@@ -231,12 +236,33 @@ headers_iterator(void *data, enum MHD_ValueKind kind, const char *key, const cha
...
+    int ret;
+    char path[PATH_MAX];
+
+    /* url given by microhttpd starts from /. For instance,
+     * https://www.solettaproject.com => url == /
+     * https://www.solettaproject.com/thankyou => url == /thankyou
+     */
+    ret = snprintf(path, sizeof(path), "%s/%s", dir,
+        *(url + 1) ? (url + 1) : "index.html");
+    SOL_INT_CHECK(ret, < 0, NULL);
```
lpereira: Truncation detection here is missing the `>= sizeof(path)` check.

src/lib/comms/sol-http-server-impl-microhttpd.c
lpereira: It's also important to return only a simple "Not found" error in the case that the prefixes don't match. This way, only one error will be returned, avoiding people trying to discover information about the system by testing if the server returns different errors (such as "Access denied") from different paths.

src/lib/comms/sol-http-server-impl-microhttpd.c
ceolin: What is the problem with the current approach of skipping the first char ?

src/lib/comms/sol-http-server-impl-microhttpd.c
ceolin: good catch with ../../ 
About the other observation i will address in a different commit since it's already happening.

src/lib/comms/sol-http-server-impl-microhttpd.c
ceolin: true

src/lib/comms/sol-http-server-impl-microhttpd.c
ceolin: ack

src/lib/comms/sol-http-server-impl-microhttpd.c
lpereira: URLs might begin with more than one `/`. Not much of a problem (`GET //////////////index.html` should serve `index.html`), but it's unusual and pretty cheap to sanitize.

ceolin:
fixes on https://github.com/solettaproject/soletta/pull/841

PR #835, opened 2015-10-02
Update OIC implementation to current version of the spec

lpereira:
CoAP does not handle OIC stuff anymore; its implementation of /oc/core is now gone.  This is handled by the OIC server like any other ordinary CoAP resource.    
    
OIC now requires TinyCBOR library (which this commit enables as a Git submodule).  All payloads have been converted from JSON to the current format as supported by IoTivity.    
    
All deprecated server resources have been removed; only /oic/res (for discovery) and /oic/d (for device information) remain.    
    
The server API has been also streamlined: now, one only pass a `sol_oic_resource_type` function to `sol_oic_server_add_resource()`, and the implementation does the rest (creating URIs, setting up for discovery, etc.).  This function returns a pointer to an opaque structure that can be used to unregister a resource and also to notify observers (which took a CoAP resource previously.)    
    
Since the encoding format changed from JSON to CBOR, the APIs to send and/or receive data changed.  This way, whoever consumes the OIC API does not need to encode/decode the payload, and TinyCBOR remains as an implementation detail.  Payload decoding is now performed within Soletta, and a vector of `sol_oic_repr_field` is passed to handlers as input, or received from them as output to be encoded and sent through the wire.  All formats supported by TinyCBOR are supported by this scheme.    

Deprecated functions and values required previously have been removed.    
    
The discovery process is now faster, since the device ID is now known without requiring another network hop to request from /oic/d.    
    
The samples and oicgen script have been updated as well.

This has been tested compatible with IoTivity as of ad4794090.

### DIFF
--- a/src/lib/comms/Makefile
+++ b/src/lib/comms/Makefile
@@ -16,8 +16,20 @@ obj-networking-$(COAP) += \
     sol-coap.o
 
 obj-networking-$(OIC) += \
+    sol-oic-cbor.o \
     sol-oic-client.o \
-    sol-oic-server.o
+    sol-oic-server.o \
+    $(TINYCBOR_SRC_PATH)/cborencoder.o \
+    $(TINYCBOR_SRC_PATH)/cborerrorstrings.o \
+    $(TINYCBOR_SRC_PATH)/cborparser.o \
+    $(TINYCBOR_SRC_PATH)/cborpretty.o
+
+obj-networking-$(OIC)-extra-cflags := \
+    -I$(TINYCBOR_SRC_PATH) \
+    -Wno-cpp \
+    -Wno-declaration-after-statement \
+    -Wno-float-equal \
+    -Wno-undef
 
 obj-networking-$(HTTP) += \
     sol-http-common.o
@@ -53,8 +65,10 @@ headers-networking-$(COAP) += \
     include/sol-coap.h
 
 headers-networking-$(OIC) += \
+    include/sol-oic-common.h \
     include/sol-oic-client.h \
-    include/sol-oic-server.h
+    include/sol-oic-server.h \
+    sol-oic-cbor.h
 
 headers-networking-$(HTTP) += \
     include/sol-http.h
--- a/src/lib/comms/include/sol-oic-common.h
+++ b/src/lib/comms/include/sol-oic-common.h
@@ -1,0 +1,116 @@
+/*
+ * This file is part of the Soletta Project
+ *
+ * Copyright (C) 2015 Intel Corporation. All rights reserved.
+ *
+ * Redistribution and use in source and binary forms, with or without
+ * modification, are permitted provided that the following conditions
+ * are met:
+ *
+ *   * Redistributions of source code must retain the above copyright
+ *     notice, this list of conditions and the following disclaimer.
+ *   * Redistributions in binary form must reproduce the above copyright
+ *     notice, this list of conditions and the following disclaimer in
+ *     the documentation and/or other materials provided with the
+ *     distribution.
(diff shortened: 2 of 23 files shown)


lpereira:
(Force-pushed to update TinyCBOR revision, which was local to my branch; that was the only change.)

lpereira:
(Force-pushed again to fix a style issue to make the bot happy. No other changes were made.)

src/lib/comms/include/sol-oic-common.h
```diff
@@ -0,0 +1,116 @@
...
+ * OF THIS SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
+ */
+
+#pragma once
+
+#include <sol-coap.h>
+#include <sol-vector.h>
+
+struct sol_oic_server_information {
+#define SOL_OIC_SERVER_INFORMATION_API_VERSION (2)
```
lpereira: This should be still version 1, as Soletta has not been released yet. Fixed locally.

src/lib/comms/Makefile
```diff
@@ -16,8 +16,20 @@ obj-networking-$(COAP) += \
...
+    $(TINYCBOR_SRC_PATH)/cborerrorstrings.o \
+    $(TINYCBOR_SRC_PATH)/cborparser.o \
+    $(TINYCBOR_SRC_PATH)/cborpretty.o
+
+obj-networking-$(OIC)-extra-cflags := \
+    -I$(TINYCBOR_SRC_PATH) \
+    -Wno-cpp \
+    -Wno-declaration-after-statement \
+    -Wno-float-equal \
+    -Wno-undef
```
barbieri: are these because of cbor? if so could we add a TODO FIXME so we remember to remove those once they are changed in that library? (most look like valid errors that should get warnings).

src/lib/comms/include/sol-oic-common.h
```diff
@@ -0,0 +1,116 @@
...
+    uint16_t api_version;
+    int : 0; /* save possible hole for a future field */
```
barbieri: this is not needed

src/lib/comms/Makefile
lpereira: Yes, because of TinyCBOR. I might create a `libtinycbor.a` and link the OIC stuff with it, segregating the `-Wno-foo` stuff to it.

I believe some of these warnings won't be fixed in TinyCBOR (specially `-Wcpp` and `-Wdeclaration-after-statement`), though.

src/lib/comms/include/sol-oic-common.h
lpereira: It's used almost everywhere else, though.

src/lib/comms/sol-oic-server.c
```diff
@@ -560,336 +551,129 @@ create_coap_resource(struct sol_str_slice endpoint)
...
+    char *buffer = NULL;
+    int r;
 
-    SOL_VECTOR_FOREACH_REVERSE_IDX (&def->resource_types, iter, idx) {
-        free(iter->resource_type);
-        /* FIXME: sol_coap_server_unregister_resource(iter->resource) */
-        free(iter->resource);
-    }
-    sol_vector_clear(&def->resource_types);
+    r = asprintf(&buffer, "/sol/%d", oic_server.resources.len);
```
barbieri: this is a problem if resources may be deleted and added, then they will clash. I'd use a `unsigned int32_t` as global and keep increasing it, let it overflow as it's less likely to create problems than the current approach.

lpereira:
Merged with ack from @barbieri in IRC. Thanks for the review, guys!

PR #836, opened 2015-10-02
Eeprom v4

edersondisouza:
v4
No more initialising statics with compound literals, so gcc < 5 can be happy.

v3:
To avoid simplify i2c paths, now keeping a 'shadow' map with resolved path. To avoid duplicating entries, sol_memmap_map structure was changed to have entries as a pointer to an array. So, generator and conffile were changed accordingly.

v2:

### DIFF
--- a/src/lib/io/sol-memmap-storage.c
+++ b/src/lib/io/sol-memmap-storage.c
@@ -40,15 +40,31 @@
 
 #include "sol-buffer.h"
 #include "sol-log.h"
+#include "sol-mainloop.h"
 #include "sol-str-slice.h"
 #include "sol-str-table.h"
 #include "sol-util.h"
 #include "sol-util-file.h"
+#include "sol-vector.h"
+
+#ifdef USE_I2C
+#include <sol-i2c.h>
+#endif
 
 #define MIN(a, b) (((a) < (b)) ? (a) : (b))
 
+#define REL_PATH_IDX 2
+#define DEV_NUMBER_IDX 3
+#define DEV_NAME_IDX 4
+
+struct map_resolved_path {
+    const struct sol_memmap_map *map;
+    char *resolved_path;
+};
+
 static struct sol_ptr_vector memory_maps = SOL_PTR_VECTOR_INIT;
 static struct sol_ptr_vector checked_maps = SOL_PTR_VECTOR_INIT;
+static struct sol_ptr_vector resolved_maps_to_free = SOL_PTR_VECTOR_INIT;
 
 static bool
 get_entry_metadata_on_map(const char *name, const struct sol_memmap_map *map, const struct sol_memmap_entry **entry, uint64_t *mask)
@@ -315,6 +331,8 @@ check_map(const struct sol_memmap_map *map)
     struct sol_memmap_entry *entry;
     uint32_t last_offset = 0;
 
+    SOL_DBG("Using memory file at [%s]", map->path);
+
     /* First, calculate any offset that was not set */
     for (iter = map->entries; iter->key; iter++) {
         entry = (void *)iter->val;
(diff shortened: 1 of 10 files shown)


src/lib/io/sol-memmap-storage.c
```diff
@@ -343,19 +361,160 @@ check_map(const struct sol_memmap_map *map)
     return true;
 }
 
+#ifdef USE_I2C
+static char *
+sol_slice_to_str(const struct sol_str_slice *slice)
```
barbieri: add this function as a helper in `sol-str-slice.h`, also fix the name to be `sol_str_slice_to_string()`.

src/lib/io/sol-memmap-storage.c
```diff
@@ -343,19 +361,160 @@ check_map(const struct sol_memmap_map *map)
...
+{
+    char *result = calloc(1, slice->len + 1);
+
+    SOL_NULL_CHECK(result, NULL);
+
+    memcpy(result, slice->data, slice->len);
```
barbieri: `strndup()` does it all :-)

src/lib/io/sol-memmap-storage.c
```diff
@@ -343,19 +361,160 @@ check_map(const struct sol_memmap_map *map)
...
+        start = sol_util_timespec_get_current();
+        while (stat(map->path, &st)) {
+            struct timespec elapsed, now = sol_util_timespec_get_current();
+
+            sol_util_timespec_sub(&now, &start, &elapsed);
+            /* Let's wait up to one second */
+            if (elapsed.tv_sec > 0) {
+                ret = -ENODEV;
+                goto end;
+            }
```
barbieri: would be helpful to `sched_yield();` before trying again, it basically says to the scheduler to yield the CPU to other processes before continuing.

src/lib/io/sol-memmap-storage.c
```diff
@@ -343,19 +361,160 @@ check_map(const struct sol_memmap_map *map)
...
+            goto error;
+        }
+
+        r = sol_ptr_vector_append(&resolved_maps_to_free, resolved_map);
+        SOL_INT_CHECK_GOTO(r, < 0, error);
+
+        return resolved_map;
+    }
+
+    return (void *)map;
```
barbieri: cast to `(void*)` is smaller to type, but potentially dangerous if we happen to change function signatures.

src/modules/flow/int/int.c
```diff
@@ -212,6 +215,32 @@ reset_process(struct sol_flow_node *node, void *data, uint16_t port, uint16_t co
...
+    const struct sol_flow_packet *packet)
+{
+    int r;
+    int32_t value;
+    struct accumulator_data *mdata = data;
+
+    r = sol_flow_packet_get_irange_value(packet, &value);
+    SOL_INT_CHECK(r, < 0, r);
+
+    if (value == mdata->val.val) return 0;
```
barbieri: what if we ignored the initial packet and value is the same as that? Should we send a packet or not?

src/lib/io/sol-memmap-storage.c
edersondisouza: Do you mean that this cast should be `(struct sol_memmap_map *)`?

src/modules/flow/int/int.c
edersondisouza: This verification is here to avoid 'infinite loops'. But thinking about the sample, `persistence` node is the one which knows `accumulator` value, and sending a packet after SET an accumulator is in fact useless. Maybe yet another option to define if SET should send a packet or not? Then get rid of this check?

edersondisouza:
v5 on #861

PR #837, opened 2015-10-02
Fix undefined behavior in align_power2()

lpereira:
If __builtin_clz() returns 1, the shift operation will try to shift by 31 bits, which is undefined behavior.  Guard against that by returning UINT_MAX, which should be the next power of two.

In addition, use C11 generic selection to pick the correct alignment function tailored to a specific type, which returns the correct maximum value, uses the correct leading zero counting builtin, and, in the case of unsigned shorts (often 16-bit), provide a function that clamps the results to USHRT_MAX.

### DIFF
--- a/src/lib/datatypes/sol-vector.c
+++ b/src/lib/datatypes/sol-vector.c
@@ -118,7 +118,7 @@ sol_vector_shrink(struct sol_vector *v)
         return;
     }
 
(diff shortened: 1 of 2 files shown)


lpereira:
(Force-pushed only to fix a style issue.)

lpereira:
_Generic() doesn't seem to be available in Semaphore. Closing for V2.

PR #839, opened 2015-10-02
Http server v2

ceolin:
from v1:
 - Added function to replace a string only if the content changed
 - Added helper to append a char to a buffer
 - Manage the response content directly (sol-buffer) instead of create strings and then set them.
 - The virtual response_cb only set the value. Before it was setting the key (path) making harder to use it further on groups.

### DIFF
--- a/src/lib/datatypes/sol-buffer.c
+++ b/src/lib/datatypes/sol-buffer.c
@@ -342,3 +342,33 @@ sol_buffer_ensure_nul_byte(struct sol_buffer *buf)
 
     return 0;
 }
+
+SOL_API int
+sol_buffer_append_char(struct sol_buffer *buf, const char c)
+{
+    const size_t nul_size = nul_byte_size(buf);
+    char *p;
+    size_t new_size;
+    int err;
+
+    SOL_NULL_CHECK(buf, -EINVAL);
+
+    err = sol_util_size_add(buf->used, 1, &new_size);
+    if (err < 0)
+        return err;
+
+    /* Extra room for the ending NUL-byte. */
+    if (new_size >= SIZE_MAX - nul_size)
+        return -EOVERFLOW;
+    err = sol_buffer_ensure(buf, new_size + nul_size);
+    if (err < 0)
+        return err;
+
+    p = sol_buffer_at_end(buf);
+    *p = c;
+
+    if (nul_size)
+        *(++p) = '\0';
+    buf->used++;
+    return 0;
+}
--- a/src/modules/flow/http-server/http-server.c
+++ b/src/modules/flow/http-server/http-server.c
@@ -55,7 +55,6 @@
 #define DOUBLE_STRING_LEN 64
 
 struct http_data {
-    struct sol_flow_node *node;
     union {
         struct sol_irange i;
         struct sol_drange d;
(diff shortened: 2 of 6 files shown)


ceolin:
fixes on https://github.com/solettaproject/soletta/pull/839

src/lib/datatypes/sol-buffer.c
```diff
@@ -342,3 +342,33 @@ sol_buffer_ensure_nul_byte(struct sol_buffer *buf)
...
+    size_t new_size;
+    int err;
+
+    SOL_NULL_CHECK(buf, -EINVAL);
+
+    err = sol_util_size_add(buf->used, 1, &new_size);
+    if (err < 0)
+        return err;
+
+    /* Extra room for the ending NUL-byte. */
```
barbieri: Ederson added `sol_buffer_ensure_nul_byte()` to do this part. Only append `c` and call that function.

Also, take the time to add the `insert_char` version.

src/modules/flow/http-server/http-server.c
```diff
@@ -214,125 +203,131 @@ boolean_close(struct sol_flow_node *node, void *data)
...
-        .content = SOL_BUFFER_INIT_EMPTY,
-        .param = SOL_HTTP_REQUEST_PARAM_INIT,
-        .response_code = SOL_HTTP_STATUS_OK
-    };
 
-    method = sol_http_request_get_method(request);
-    response.url = sol_http_request_get_url(request);
+    if (json) {
+        size_t escaped_len = sol_json_calculate_escaped_string_len(mdata->value.s);
+        r = sol_buffer_ensure(content, content->used + escaped_len);
```
barbieri: ooops! Since you're going to append `"` twice and these are not counted in `escaped_len`, you need to sum that amount here. Likely not crashing since `sol_buffer_ensure()` grows by power of 2 and you're being lucky.

src/modules/flow/http-server/http-server.c
```diff
@@ -214,125 +203,131 @@ boolean_close(struct sol_flow_node *node, void *data)
...
-            escaped_str);
-        free(escaped_str);
+        sol_json_escape_string(mdata->value.s, sol_buffer_at_end(content), escaped_len);
+        content->used += escaped_len - 1; /* remove \0 in the result */
+
+        r = sol_buffer_append_char(content, '\"');
+        SOL_INT_CHECK(r, < 0, r);
     } else {
-        r = asprintf(&response_str, "%s", mdata->value.s);
+        r = sol_buffer_append_printf(content, "%s", mdata->value.s);
```
barbieri: `sol_buffer_append_slice()` and `sol_str_slice_from_string()`?

src/modules/flow/http-server/http-server.c
```diff
@@ -375,287 +369,132 @@ string_open(struct sol_flow_node *node, void *data, const struct sol_flow_node_o
...
 
-    if (send_json) {
-        r = asprintf(&response_str, "{\"%s\":{\"value\":%s,\"min\":%s,\"max\":%s,\"step\":%s}}",
-            mdata->path, val, min, max, step);
+    if (json) {
+        r = sol_buffer_append_printf(content, "{\"value\":%s,\"min\":%s,\"max\":%s,\"step\":%s}",
+            val, min, max, step);
     } else {
-        r = asprintf(&response_str, "%s", val);
+        r = sol_buffer_append_printf(content, "%s", val);
```
barbieri: same here, maybe a helper to make it clear? like `sol_buffer_append_string()`?

src/shared/sol-util.c
```diff
@@ -329,3 +329,36 @@ sol_util_uuid_gen(bool upcase,
     sol_buffer_fini(&buf);
     return r;
 }
+
+SOL_API int
+sol_util_replace_str_if_changed(char **str, const char *new_str)
+{
+    if (!str)
```
barbieri: use `SOL_NULL_CHECK()` to be loud against errors.

src/shared/sol-util.c
```diff
@@ -329,3 +329,36 @@ sol_util_uuid_gen(bool upcase,
...
+        return -EINVAL;
+
+    if (!new_str) {
+        if (*str)
+            free(*str);
```
barbieri: no need to check for null before free.

src/modules/flow/http-server/http-server.c
ceolin: if i'm right sol_buffer_append_printf() checks the necessary size and increase the buffer if necessary

barbieri: it will, but you ensure `escaped_len`,  then you consume one char with `append_char()`, then you have `escaped_len - 1` available, then you:

```c
sol_json_escape_string(mdata->value.s, sol_buffer_at_end(content), escaped_len);
```
yet you don't have `escaped_len` anymore.

ceolin:
pushed with the fixes, thanks for reviewing it

PR #840, opened 2015-10-02
Fix undefined behavior in align_power2()

lpereira:
If __builtin_clz() returns 1, the shift operation will try to shift by 31 bits, which is undefined behavior. Guard against that by returning UINT_MAX, which should be the next power of two.

In addition, create a bunch of align_power2() functions to work with integers of different sizes.  C11's _Generic() would be a better candidate but for some reason it's not available in the CI server.

These functions will each return the correct maximum value and uses the correct leading zero counting builtin for the type, and, in the case of unsigned shorts (often 16-bit), provide a function that clamps the results to USHRT_MAX.

### DIFF
--- a/src/shared/sol-util.h
+++ b/src/shared/sol-util.h
@@ -193,18 +193,40 @@ char *sol_util_strerror(int errnum, char *buf, size_t buflen);
         sol_util_strerror((errnum), buf ## __COUNT__, sizeof(buf ## __COUNT__)); \
     })
 
-static inline unsigned int
-align_power2(unsigned int u)
-{
-    unsigned int left_zeros;
+/* Power of 2 alignment. */
+#define DEFINE_ALIGN_POWER2(name_, type_, max_, clz_fn_) \
+    static inline type_ \
+    name_(type_ u) \
+    { \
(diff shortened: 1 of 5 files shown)


src/shared/sol-util.h
```diff
@@ -193,16 +193,40 @@ char *sol_util_strerror(int errnum, char *buf, size_t buflen);
...
+    name_(type_ u) \
+    { \
+        unsigned int left_zeros; \
+        if (u == 1) \
+            return 1; \
+        if ((left_zeros = clz_fn_(u - 1)) < 1) \
+            return 0; \
+        if (unlikely(left_zeros == 1)) \
+            return max_; \
+        return 1 << ((sizeof(u) * 8) - left_zeros); \
```
lpereira: These should be `1U` and `8U`. Fixed locally.

lpereira:
Maybe this can be written using a macro that uses `__builtin_types_compatible_p()` to work around compilers without `_Generic()`, and still be easy to use...

lpereira:
Closing for version with `__builtin_types_compatible_p()`.

PR #841, opened 2015-10-02
Static files v2

ceolin:
from v1
 - Avoid bad requests (../../../) using realpath()
 - check the return of snprintf properly
 - check permission of the served file
 - Added commit with a small fixes on network node

### DIFF
--- a/src/lib/comms/sol-http-server-impl-microhttpd.c
+++ b/src/lib/comms/sol-http-server-impl-microhttpd.c
@@ -31,9 +31,13 @@
  */
 
 #include <errno.h>
+#include <fcntl.h>
 #include <microhttpd.h>
 #include <stdio.h>
 #include <stdlib.h>
+#include <sys/stat.h>
+#include <sys/types.h>
+#include <unistd.h>
 
 #include "sol-http-server.h"
 #include "sol-log.h"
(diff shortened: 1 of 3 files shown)


src/lib/comms/sol-http-server-impl-microhttpd.c
```diff
@@ -293,10 +333,36 @@ http_server_handler(void *data, struct MHD_Connection *connection, const char *u
...
+            ret = fstat(fd, &st);
+            if (ret < 0) {
+                close(fd);
+            } else {
+                if (!(st.st_mode & (S_IRUSR | S_IRGRP | S_IROTH))) {
+                    status = SOL_HTTP_STATUS_FORBIDDEN;
+                    close(fd);
+                }
+
+                mhd_response = MHD_create_response_from_fd(st.st_size, fd);
```
lpereira: `fd` is closed here if the above check happens.

src/lib/comms/sol-http-server-impl-microhttpd.c
```diff
@@ -590,3 +662,51 @@ sol_http_server_set_last_modified(struct sol_http_server *server, const char *pa
...
+
+    SOL_NULL_CHECK(server, -EINVAL);
+    SOL_NULL_CHECK(rootdir, -EINVAL);
+
+    SOL_PTR_VECTOR_FOREACH_IDX (&server->dirs, dir, i) {
+        if (streq(dir, rootdir))
+            return -EINVAL;
+    }
+
+    dir = strdup(rootdir);
```
lpereira: Instead of `strdup()`, `realpath()`, as that's what's being compared in the function that opens the file.

ceolin:
merged with fixes, thanks guys

PR #842, opened 2015-10-02
flow/string-format: Also ensure that precision == 0 is valid in another case

lpereira:
Do this not only in the case where there's an exponent.


lpereira:
@glima, could you please take a look on this one? It's in the other path.

lpereira:
Merged, thank you!

PR #844, opened 2015-10-03
Linux micro kmod

dorileo:


src/modules/linux-micro/kmod/kmod.c
```diff
@@ -0,0 +1,273 @@
...
+{
+    const int probe_flags = KMOD_PROBE_APPLY_BLACKLIST;
+    struct kmod_list *itr, *modlist = NULL;
+    int r = 0;
+
+    SOL_DBG("load: %s", mod_name);
+
+    r = kmod_module_new_from_lookup(ctx, mod_name, &modlist);
+    if (r < 0) {
+        SOL_ERR("Failed to lookup alias '%s': %m", mod_name);
```
lpereira: Please don't use `%m`. It's a GNU extension and we're not using it elsewhere anymore. Use `%s` and `sol_util_strerrora()`.

dorileo: Ok

src/modules/linux-micro/kmod/kmod.c
```diff
@@ -0,0 +1,273 @@
...
+        mod_name = remove_leading_whitespace(buf);
+        if (*mod_name == '#' || *mod_name == ';' || *mod_name == '\0')
+            continue;
+
+        remove_trailing_whitespace(mod_name);
+        ret = kmod_apply_value(ctx, mod_name);
+        if (ret < 0)
+            err = ret;
+
+        line_no++;
```
edersondisouza: Why this `line_no`? You don't appear to use it.

src/modules/linux-micro/kmod/kmod.c
```diff
@@ -0,0 +1,273 @@
...
+    return strcmp((const char *)data1, (const char *)data2);
+}
+
+static void
+read_directory_to_vector(DIR *dir, struct sol_ptr_vector *vec)
+{
+    struct dirent ent, *entbuf = NULL;
+    int r;
+
+    r = readdir_r(dir, &ent, &entbuf);
```
edersondisouza: This `ent` parameter must be allocated in a special fashion, according to `readdir_r(3)`.

lpereira: [**@edersondisouza**](https://github.com/edersondisouza) commented on [Oct 5, 2015, 10:04 AM GMT-3](https://github.com/solettaproject/soletta/pull/844#discussion_r41140575 "2015-10-05T13:04:28Z - Replied by Github Comment Enhancer"):
> This `ent` parameter must be allocated in a special fashion, according to `readdir_r(3)`.

The note is more for portable programs. This is Linux-specific, so it should be OK; unless alternative libcs do this differently than glibc, in which case this would need to be checked (specially MUSL, since that's the one we're supporting for `linux-micro` targets).

src/modules/linux-micro/kmod/kmod.c
dorileo: this peace of code came from other modules c&p, I'll adjust it and propagate wherever look appropriate. thanks.

src/modules/linux-micro/kmod/kmod.c
dorileo: I'll recheck it, also there're some pieces of this code that better deserve a common/util set of sharing. I'll also find a good place to share them across the linux-micro modules. Thanks for reviewing.

src/modules/linux-micro/kmod/Kconfig
```diff
@@ -0,0 +1,4 @@
+config LINUX_MICRO_KMOD
+	tristate "kmod"
+	depends on LINUX_MICRO && HAVE_LIBKMOD
+	default m if MODULES=y
```
barbieri: add help :-)

src/modules/linux-micro/kmod/kmod.c
```diff
@@ -0,0 +1,273 @@
...
+    kmod_list_foreach(itr, modlist) {
+        struct kmod_module *mod;
+        int state, err;
+
+        mod = kmod_module_get_module(itr);
+        state = kmod_module_get_initstate(mod);
+
+        switch (state) {
+        case KMOD_MODULE_BUILTIN:
+            SOL_INF("Module '%s' is builtin", kmod_module_get_name(mod));
```
barbieri: keep both log levels the same, so here you can use `SOL_DBG()`

src/modules/linux-micro/kmod/kmod.c
```diff
@@ -0,0 +1,273 @@
...
+static void
+remove_trailing_whitespace(char *input)
+{
+    char *ptr = input + strlen(input) - 1;
+
+    while (ptr > input && isspace(*ptr)) {
+        *ptr = '\0';
+        ptr--;
+    }
+}
```
barbieri: I'd move these two functions to `sol-util.h` and have them to work with `struct sol_str_slice`, see below where I say to change from `fgets()` to `mmap()`.

src/modules/linux-micro/kmod/kmod.c
```diff
@@ -0,0 +1,273 @@
...
+    char buf[1024];
+    int err = 0, line_no = 1;
+
+    conf = fdopen(fd, "re");
+    if (!conf) {
+        close(fd);
+        return -errno;
+    }
+
+    while (fgets(buf, sizeof(buf), conf)) {
```
barbieri: I'd avoid doing `fgets()` as lines bigger than `sizeof(buf)` will screw you. And it's slower, then use the `sol_file_reader` to get a slice and operate on that. If possible it will `mmap()`.

src/modules/linux-micro/kmod/kmod.c
```diff
@@ -0,0 +1,273 @@
...
+        "/usr/lib/modules-load.d",
+        "/run/modules-load.d",
+        "/etc/modules-load.d",
+        NULL
+    };
+
+    ctx = kmod_new(NULL, NULL);
+    if (!ctx) {
+        SOL_ERR("Failed to allocate memory for kmod.");
+        return EXIT_FAILURE;
```
barbieri: we're using `-errno`, if no `errno`, then use `-ENOMEM` here.

dorileo:
v2 sent...

PR #845, opened 2015-10-05
Introduce json-object and json-array packet types - v3

otaviobp:
Changelog from v2:
 - safety checks and documentation in sol_json_validate
 - fixing typo in comment s/Non/Not
 - checking strdup return
 - support json objects and arrays in console node


otaviobp:
this pr replaces #832

src/lib/parsers/sol-json.c
```diff
@@ -546,3 +546,25 @@ sol_json_double_to_str(const double value, char *buf, size_t len)
 
     return 0;
 }
+
+/* Check if scanner is pointing to a valid string with a json element of
+ * informed type. May or may not be NULL terminated */
+SOL_API bool
+sol_json_validate(struct sol_json_scanner *scanner, enum sol_json_type type)
```
lpereira: I still don't like this function name. It's fine for an internal name, but it's not OK for an exported API: there are way too many things one could validate in JSON.

Be more specific; something like `sol_json_pointing_at_type()`.

src/modules/flow/console/console.c
```diff
@@ -149,6 +149,22 @@ console_in_process(struct sol_flow_node *node, void *data, uint16_t port, uint16
         }
 
         fprintf(mdata->fp, "} (blob)%s\n", mdata->suffix);
+    } else if (packet_type == SOL_FLOW_PACKET_TYPE_JSON_OBJECT) {
+        struct sol_blob *val;
+
+        int r = sol_flow_packet_get_json_object(packet, &val);
+        SOL_INT_CHECK(r, < 0, r);
+
+        console_output(mdata, "%.*s (JSON object)", (int)val->size,
```
lpereira: There is a macro to print `sol_str_slice` structs; maybe you could make one for `sol_blob` as well?

barbieri: could be, but usually we don't print blobs like that, only when we know it's a string such as in JSON.

What we could have is a `sol_str_slice_from_blob()`  helper and then use that macro, one line more, but at least it's cleaner.

otaviobp: Ok. I agree. But as this method may be relevant to others, I'll implement it in another pr. Merging this pr without macro for blob printing

PR #846, opened 2015-10-05
Add safety checks to SOL_API methods in sol-json.c

otaviobp:


src/lib/parsers/sol-json.c
```diff
@@ -282,6 +282,9 @@ sol_json_token_get_double(const struct sol_json_token *token, double *value)
     char *endptr;
     int r;
 
+    SOL_NULL_CHECK(token, -EINVAL);
```
lpereira: Since these parameters are more likely to be allocated on the stack (and thus their addresses are guaranteed non-null), adding this overhead (as cheap as it is) is not the best course of action here. Use `SOL_ATTR_NONNULL` instead: this will generate a compile-time warning.

otaviobp: Actually this patch is invalid. Those parameters were already being checked with SOL_ATTR_NONNULL. I didn't realize that. thanks for pointing me that

PR #848, opened 2015-10-05
blob: Add helper to print string blobs

otaviobp:
Add a conversor from blob to str_slice to make it easier to print a string
blob using SOL_STR_SLICE_PRINT.

Note that creating a SOL_BLOB_PRINT macro is not a good approach because
not all blobs can be printed as strings. Making the conversion to
str_slice explicit may avoid errors


barbieri:
+1

PR #849, opened 2015-10-05
Add extra nodes to handle JSON object and JSON array

otaviobp:
Nodes added:
 - json/object-length: Get JSON object length
 - json/object-get-all-keys: Get all keys of a JSON object
 - json/array-get-at-index: Get a single element of a JSON array
 - json/array-length: Get the array length
 - json/array-get-all-elements: Send all elements from a JSON array to
   out PORTS

### DIFF
--- a/src/modules/flow/json/json.c
+++ b/src/modules/flow/json/json.c
@@ -72,17 +72,15 @@ json_object_get_key_close(struct sol_flow_node *node, void *data)
 }
 
 static struct sol_blob *
-create_sub_json_object(struct sol_blob *parent, struct sol_json_token *token)
+create_sub_json(struct sol_blob *parent, struct sol_json_scanner *scanner, struct sol_json_token *token, enum sol_json_type type)
 {
-    struct sol_json_scanner scanner;
     const char *mem;
 
     mem = token->start;
-    sol_json_scanner_init(&scanner, parent->mem, parent->size);
-    if (!sol_json_scanner_skip_over(&scanner, token))
+    if (!sol_json_scanner_skip_over(scanner, token))
         return NULL;
 
-    if (sol_json_token_get_type(token) != SOL_JSON_TYPE_OBJECT_END) {
+    if (sol_json_token_get_type(token) != type) {
         errno = EINVAL;
         return NULL;
     }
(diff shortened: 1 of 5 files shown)


src/modules/flow/json/json.c
```diff
@@ -224,4 +239,219 @@ json_object_get_key_in_process(struct sol_flow_node *node, void *data, uint16_t
...
+    struct sol_str_slice slice;
+    bool empty = true;
+
+    r = sol_flow_packet_get_json_object(packet, &in_value);
+    SOL_INT_CHECK(r, < 0, r);
+
+    sol_json_scanner_init(&scanner, in_value->mem, in_value->size);
+    SOL_JSON_SCANNER_OBJECT_LOOP (&scanner, &token, &key, &value, reason) {
+        sol_json_token_remove_quotes(&key);
+        slice = sol_json_token_to_slice(&key);
```
barbieri: actually I think we need to unescape it. We don't have a function to do that, but we should and it should be called where `sol_json_token_remove_quotes()` is currently in use (can be in a new/separate patch).

Ideally we'd be clever and only allocate memory if the string is escaped. We can do this with buffer trick:

```c
// buffer will be initialized in this function!
int
sol_json_token_get_string(const struct sol_json_token *token, struct sol_buffer *buffer)
{
   const char *start, *p;
   bool is_escaped = false;
   SOL_NULL_CHECK(token, -EINVAL);
   SOL_NULL_CHECK(buffer, -EINVAL);
   SOL_INT_CHECK(token->type, != SOL_JSON_TYPE_STRING, -EINVAL);
   sol_buffer_init_flags(buffer, SOL_BUFFER_FLAGS_NO_NUL_BYTE); // consistent in the case we don't need escaping.
   for (start = p = token->start + 1; p < token->end - 1; p++) {
       if (!is_escaped && *p == '\\') {
          struct sol_str_slice slice = {.data = start; .len = p - start; };
          is_escaped = true;
          sol_buffer_append_slice(buffer, &slice); // check return
       } else if (is_escaped) {
          start = p + 1;
          is_escaped = false;
          if (*p == '\\') sol_buffer_append_char(buffer, '\\');
          else if (*p == '/') sol_buffer_append_char(buffer, '/');
          else if (*p == '"') sol_buffer_append_char(buffer, '"');
          else if (*p == 'b') sol_buffer_append_char(buffer, '\b');
          else if (*p == 'r') sol_buffer_append_char(buffer, '\r');
          else if (*p == 'n') sol_buffer_append_char(buffer, '\n');
          else if (*p == 'f') sol_buffer_append_char(buffer, '\f');
          else if (*p == 't') sol_buffer_append_char(buffer, '\t');
          else if (*p == 'u' && p + 5 < token->end - 1) { // needs checking if this is correct
             char hexdigits[5], *endptr;
             struct sol_str_slice slice;
             uint16_t n;
             memcpy(hexdigits, p + 1, 4);
             hexdigits[4] = '\0';
             n = strtoul(hexdigits, &endptr, 16); // check error
             slice.data = &n; // needs checking if byte order is correct
             slice.len = sizeof(n);
             sol_buffer_append_slice(buffer, &slice); // check error
             start += 4; // extra bytes processed by this escape
          } else {
            SOL_WRN("invalid JSON string: %.*s", token...);
            return -EINVAL;
       }
   }
   if (is_escaped || *p != '"') {
      SOL_WRN("invalid JSON string..." , token ...);
      return -EINVAL;
   }
   if (start == token->start + 1) { // no escaping, re-init buffer to Not-Owned and point to token's memory
       sol_buffer_init_flags(buffer, SOL_BUFFER_FLAGS_MEMORY_NOT_OWNED);
       buffer->used = buffer->capacity = sol_json_token_get_size(token) - 2;
       buffer->data = token->start + 1;
   } else {
      struct sol_str_slice slice = {.data = start; .len = p - start; };
      sol_buffer_append_slice(buffer, &slice); // check return
   }
   return 0;
}
```

src/modules/flow/json/json.c
```diff
@@ -224,4 +239,219 @@ json_object_get_key_in_process(struct sol_flow_node *node, void *data, uint16_t
...
+
+    r = sol_flow_packet_get_json_array(packet, &in_value);
+    SOL_INT_CHECK(r, < 0, r);
+
+    sol_json_scanner_init(&scanner, in_value->mem, in_value->size);
+    SOL_JSON_SCANNER_ARRAY_LOOP_ALL(&scanner, &token, reason) {
+        if (!sol_json_scanner_skip_over(&scanner, &token))
+            return -EINVAL;
+
+        if (len.val == INT32_MAX)
```
barbieri: nice you did this here, but please add this check to the object length as well :-)

PR #850, opened 2015-10-05
oic: Various fixes when using flow based programming

lpereira:
This commit fixes some problems that appeared during the update to the
newer version of the protocol, including some segmentation faults (due
to NULL pointer dereferencing), incorrect return values, CoAP encoding
errors, among other things.

It also updates the light-client.fbp example, illustrating the use of
the `FOUND` port to only enable the toggle timer if the light has been
found on the network.


bdilly:
I've tested with soletta-demos and it worked fine.
Changes make sense. It seems good.

lpereira:
Merged, thanks.

PR #851, opened 2015-10-05
modules/int: fix division sigfpe

bdilly:
The most negative integer by -1 may generate SIGFPE,
so it must be properly tested by division node type
and should be avoided on multiplication underflow test.

@lpereira , @anselmolsm , @ibriano ^

### DIFF
--- a/src/lib/common/sol-types.c
+++ b/src/lib/common/sol-types.c
@@ -174,11 +174,12 @@ sol_drange_equal(const struct sol_drange *var0, const struct sol_drange *var1)
     ((var1 > 0) && (var0 < INT32_MIN + var1))
 
 #define SOL_IRANGE_MULTIPLICATION_OVERFLOW(var0, var1) \
+    ((var0 == INT32_MAX) && (var1 == -1)) || \
     ((var0 > 0) && (var1 > 0) && (var0 > INT32_MAX / var1)) || \
     ((var0 < 0) && (var1 < 0) && (var0 < INT32_MAX / var1))
 
 #define SOL_IRANGE_MULTIPLICATION_UNDERFLOW(var0, var1) \
-    ((var0 > 0) && (var1 < 0) && (var0 > INT32_MIN / var1)) || \
+    ((var0 > 0) && (var1 < -1) && (var0 > INT32_MIN / var1)) || \
     ((var0 < 0) && (var1 > 0) && (var0 < INT32_MIN / var1))
 
 SOL_API int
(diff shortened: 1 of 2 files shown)


anselmolsm:
* The new irange-arithmetic.fbp causes SIGFPE, when executed using a libsolletta without the patch to sol-types.c;
* The sol-types.c changes avoid the issue, it takes into account the overflow/underflow cases not covered before.

LGTM, but I'd like to hear @ibriano and @lpereira opinions too.

src/lib/common/sol-types.c
```diff
@@ -174,11 +174,12 @@ sol_drange_equal(const struct sol_drange *var0, const struct sol_drange *var1)
     ((var1 > 0) && (var0 < INT32_MIN + var1))
 
 #define SOL_IRANGE_MULTIPLICATION_OVERFLOW(var0, var1) \
+    ((var0 == INT32_MAX) && (var1 == -1)) || \
     ((var0 > 0) && (var1 > 0) && (var0 > INT32_MAX / var1)) || \
     ((var0 < 0) && (var1 < 0) && (var0 < INT32_MAX / var1))
 
 #define SOL_IRANGE_MULTIPLICATION_UNDERFLOW(var0, var1) \
-    ((var0 > 0) && (var1 < 0) && (var0 > INT32_MIN / var1)) || \
+    ((var0 > 0) && (var1 < -1) && (var0 > INT32_MIN / var1)) || \
```
lpereira: `var1` might be zero here.

src/lib/common/sol-types.c
```diff
@@ -174,11 +174,12 @@ sol_drange_equal(const struct sol_drange *var0, const struct sol_drange *var1)
     ((var1 > 0) && (var0 < INT32_MIN + var1))
 
 #define SOL_IRANGE_MULTIPLICATION_OVERFLOW(var0, var1) \
```
lpereira: Why not use `__builtin_mul_overflow()` here?

src/lib/common/sol-types.c
```diff
@@ -174,11 +174,12 @@ sol_drange_equal(const struct sol_drange *var0, const struct sol_drange *var1)
...
+    ((var0 == INT32_MAX) && (var1 == -1)) || \
```
lpereira: Another thing that should be fixed here while you're at it: either convert these overflow macros to `static inline` functions (which are [as fast as macros](https://gcc.gnu.org/onlinedocs/gcc/Inline.html)), or add parenthesis to these macro arguments.

src/lib/common/sol-types.c
ibriano: If var1 == 0, the second condition fails, and since they are all &&, the division won't take place. All praise lazy evaluation.

ibriano:
I concur with @lpereira's comments, but this is still a good patch in the mean time.

src/lib/common/sol-types.c
lpereira: @shenbot: %conclua

bdilly:
@lpereira it would be good to merge these checks in a single place. I saw overflow checks on sol-util.h using builtin when possible or with other implementations. When done, these macros could be changed to functions (or not).

bdilly:
merged, thanks

PR #852, opened 2015-10-06
Fix undefined behavior in align_power2() (V2)

lpereira:
If left_zeros is 0, then don't shift by 32 bits, which is undefined behavior.  Return UINT_MAX as the value can't get larger than that. If left_zeros is 1, return UINT_MAX, as that's the next power-of-two
aligned value.

Also ensure align_power2() will work with types of sizes different than `unsigned int` (at the moment, uint16_t and size_t). Do this automatically using __builtin_types_compatible_p() as to not patch every call of this function.

### DIFF
--- a/src/lib/datatypes/sol-vector.c
+++ b/src/lib/datatypes/sol-vector.c
@@ -118,7 +118,7 @@ sol_vector_shrink(struct sol_vector *v)
         return;
     }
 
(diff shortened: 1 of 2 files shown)


ibriano:
Looks ok

lpereira:
Merged, thanks

PR #854, opened 2015-10-06
oic-gen: don't include internal headers

bdilly:
Otherwise external modules won't compile.
Instead, just declare required macros.

@lpereira please review


lpereira:
:+1:

bdilly:
merged

PR #855, opened 2015-10-06
http:client: Avoid segfault

ceolin:
It gives to user the chance of steal the content buffer without happens
a double free. It's useful to avoid duplicate the buffer's content.


ceolin:
@otaviobp could you take a look on it please ?

ceolin:
merged, thanks !

PR #857, opened 2015-10-06
sol_buffer: Fix append_printf function bug

otaviobp:
Function sol_buffer_append_printf was not adding the last character to
the buffer if SOL_BUFFER_FLAGS_NO_NUL_BYTE was used and if this
character would be placed in the last available buffer position.

This was happening because printf function always place '\0' at the end
of the string being printed. So we need to allocate an extra byte even
if SOL_BUFFER_FLAGS_NO_NUL_BYTE is used.

### DIFF
--- a/src/lib/datatypes/sol-buffer.c
+++ b/src/lib/datatypes/sol-buffer.c
@@ -200,6 +200,8 @@ sol_buffer_insert_slice(struct sol_buffer *buf, size_t pos, const struct sol_str
     return 0;
 }
 
+/* Extra room for the ending NUL-byte, necessary even when the buffer has
+ * SOL_BUFFER_FLAGS_NO_NUL_BYTE because vsnprintf always write the '\0'. */
 SOL_API int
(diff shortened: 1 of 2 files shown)


src/lib/datatypes/sol-buffer.c
```diff
@@ -233,11 +236,11 @@ sol_buffer_append_vprintf(struct sol_buffer *buf, const char *fmt, va_list args)
             goto end;
 
         /* Extra room for the ending NUL-byte. */
-        if (new_size >= SIZE_MAX - nul_byte_size(buf)) {
+        if (new_size >= SIZE_MAX - NUL_SIZE) {
```
cmarcelo: Minor suggestions: I would change the comment above to say

   /* Extra room for the ending NUL-byte, necessary even when the buffer has
      SOL_BUFFER_FLAGS_NO_NUL_BYTE because vsnprintf always write the '\0'. */

And use the literal 1 or a local constant instead of a define macro.

PR #860, opened 2015-10-06
mqtt-samples: Add IBM Bluemix phonemotion sample

rchiossi:
This sample is a fully functional replacement of the client side of the
phonemotion example for IBM Bluemix. It is a demo of usage of Soletta
native MQTT and HTTP APIs.

### DIFF
--- a/src/samples/mqtt/bluemix.c
+++ b/src/samples/mqtt/bluemix.c
@@ -1,0 +1,355 @@
+/*
+ * This file is part of the Soletta Project
+ *
+ * Copyright (C) 2015 Intel Corporation. All rights reserved.
+ *
+ * Redistribution and use in source and binary forms, with or without
+ * modification, are permitted provided that the following conditions
+ * are met:
+ *
+ *   * Redistributions of source code must retain the above copyright
+ *     notice, this list of conditions and the following disclaimer.
+ *   * Redistributions in binary form must reproduce the above copyright
+ *     notice, this list of conditions and the following disclaimer in
+ *     the documentation and/or other materials provided with the
+ *     distribution.
+ *   * Neither the name of Intel Corporation nor the names of its
+ *     contributors may be used to endorse or promote products derived
+ *     from this software without specific prior written permission.
+ *
+ * THIS SOFTWARE IS PROVIDED BY THE COPYRIGHT HOLDERS AND CONTRIBUTORS
+ * "AS IS" AND ANY EXPRESS OR IMPLIED WARRANTIES, INCLUDING, BUT NOT
+ * LIMITED TO, THE IMPLIED WARRANTIES OF MERCHANTABILITY AND FITNESS FOR
+ * A PARTICULAR PURPOSE ARE DISCLAIMED. IN NO EVENT SHALL THE COPYRIGHT
+ * OWNER OR CONTRIBUTORS BE LIABLE FOR ANY DIRECT, INDIRECT, INCIDENTAL,
+ * SPECIAL, EXEMPLARY, OR CONSEQUENTIAL DAMAGES (INCLUDING, BUT NOT
+ * LIMITED TO, PROCUREMENT OF SUBSTITUTE GOODS OR SERVICES; LOSS OF USE,
+ * DATA, OR PROFITS; OR BUSINESS INTERRUPTION) HOWEVER CAUSED AND ON ANY
+ * THEORY OF LIABILITY, WHETHER IN CONTRACT, STRICT LIABILITY, OR TORT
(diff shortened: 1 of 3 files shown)


src/samples/mqtt/bluemix.c
```diff
@@ -0,0 +1,355 @@
...
+
+    SOL_INT_CHECK(r, != 0, false);
+
+    return true;
+}
+
+static bool
+try_reconnect(void *data)
+{
+    return sol_mqtt_reconnect((struct sol_mqtt *)data) != 0;
```
barbieri: no need to cast.

src/samples/mqtt/bluemix.c
```diff
@@ -0,0 +1,355 @@
...
+ * Authenticate to HTTP Server
+ */
+
+static bool
+json_token_to_string(struct sol_json_token *token, char **out)
+{
+    if (sol_json_token_get_type(token) != SOL_JSON_TYPE_STRING)
+        return false;
+    free(*out);
+    *out = strndup(token->start + 1, token->end - token->start - 2);
```
barbieri: so far, so good. But we're adding API to unescape the string, this will be the recommended way to extract the string. Sync with @otaviobp so either you or him update this function to use it.

src/samples/mqtt/bluemix.c
```diff
@@ -0,0 +1,355 @@
...
+    struct sol_mqtt_message message;
+    char topic[] = "iot-2/evt/sensorData/fmt/json-iotf";
+    int r;
+    double v;
+    static int pulse = 0;
+
+    pulse = (pulse + 1) % 360;
+    v = sin(pulse * PI / 180) * 10;
+
+    r = sol_buffer_append_printf(&payload, SENSOR_FMT, v * 1.25, v * 1, v * 0.75, v * 20, v * 15, v * 10);
```
barbieri: so far, so good. But @ceolin is working on json serializer so we handle it properly. If users copy&paste this fragment on code that also calls `setlocale()`, then we may get float-points formatted with `,` instead of `.` for decimal and it will break the format.

sync with him so this is updated once those functions are merged.

src/samples/mqtt/bluemix.c
```diff
@@ -0,0 +1,355 @@
...
+    int r;
+
+    SOL_NULL_CHECK_GOTO(data, error);
+    SOL_NULL_CHECK_GOTO(data->user, error);
+    SOL_NULL_CHECK_GOTO(data->pin, error);
+    SOL_NULL_CHECK_GOTO(data->url, error);
+
+    data->connect_timeout = NULL;
+
+    r = sol_buffer_append_printf(&buffer, "{\"email\":\"%s\", \"pin\":\"%s\"}", data->user, data->pin);
```
barbieri: similar to above, sync with @ceolin to use the string serialize function as the proper way to do this is to escape the string.

src/samples/mqtt/bluemix.c
```diff
@@ -0,0 +1,355 @@
...
+static bool get_user_token(void *vdata);
+
+static void
+request_callback(void *vdata, const struct sol_http_client_connection *connection,
+    struct sol_http_response *response)
+{
+    struct bluemix_data *data = (struct bluemix_data *)vdata;
+
+    if (!response->content.used || response->response_code != SOL_HTTP_STATUS_OK) {
+        SOL_DBG("Message: %.*s", (int)response->content.used, (char *)response->content.data);
```
barbieri: you may use `sol_buffer_get_slice()` and then `SOL_STR_SLICE_PRINT()`.

src/samples/mqtt/bluemix.c
```diff
@@ -0,0 +1,355 @@
...
+#include <sol-mainloop.h>
+#include <sol-mqtt.h>
+
+struct bluemix_data {
+    /**
+     * HTML POST authentication
+     */
+    char *user;
+    char *pin;
+    char *url;
```
barbieri: const char those to make sure they are not modified.

src/samples/mqtt/bluemix.c
rchiossi: ok

src/samples/mqtt/bluemix.c
rchiossi: ack

src/samples/mqtt/bluemix.c
rchiossi: I don't think really think SOL_DBG("Message: %.*s", SOL_STR_SLICE_PRINT(sol_buffer_get_slice(response->content))); is that much simpler... but anyway, I will update.

src/samples/mqtt/bluemix.c
ceolin: i'm workgin on it, if this patch goes before mine i replace it.

PR #861, opened 2015-10-06
Eeprom v5

edersondisouza:
v5
Added `sol_str_slice_to_string` and use it.
`sched_yield();` on sysfs busy waiting.
Fix some casts.

v4
No more initialising statics with compound literals, so gcc < 5 can be happy.

v3:
To avoid simplify i2c paths, now keeping a 'shadow' map with resolved path. To avoid duplicating entries, sol_memmap_map structure was changed to have entries as a pointer to an array. So, generator and conffile were changed accordingly.

v2:


barbieri:
+1

glima:
Other than those, +1.

edersondisouza:
Integrated with fixes to issues pointed by @glima. Tks!

PR #862, opened 2015-10-06
test, shared, modules: Fix compiler warnings when building with -O3

lpereira:


lpereira:
There are still problems in test.h regarding ASSERT_INT_EQ and overflows, but I'll leave those for another day.

edersondisouza:
lgtm

lpereira:
Merged, thanks

PR #863, opened 2015-10-06
Selector

glima:
Quoting the Kconfig help verbatim:

    "It is common to have IoT products with LCD display and buttons (arrows)
    and the need to ask for user input. This is a collection of high-level
    node types to deal with that, the *selector* nodes.

    These node types format a multi-line string to be used by other node
    types doing visual output, such as LCD display (grove/lcd-string), in
    conjuntion with others providing input."

    I did my best to have broad tests, but much more than that was tested by
    hand while trying to make the code robust. It's a PITA to write
    validation to this node, so let's grow that with the time.


glima:
The ``SELECTED`` input and output ports from the node *spec* are missing, doing that now.

barbieri:
didn't read the PR yet, but the module shouldn't be `selector`, rather a broader scope name that will cover all the input-form elements (boolean, integer...). Need to think of a name, maybe something like `textual-input-form`

glima:
* Gustavo Sverzut Barbieri <notifications@github.com> [2015-10-06 16:15:07 -0700]:

> didn't read the PR yet, but the module shouldn't be `selector`, rather a broader scope name that will cover all the input-form elements (boolean, integer...). Need to think of a name, maybe something like `textual-input-form`
>

I agree on a single module for the selector and input forms, but I
could not think of a reasonable name yet, either. Remember that the
selector itself is not a form, just the others. Maybe ``textual-io``
or something like that?

> ---
> https://github.com/solettaproject/soletta/pull/863#issuecomment-146030768

-- 
Gustavo Lima Chaves

src/modules/flow/selector/selector.c
```diff
@@ -0,0 +1,688 @@
...
+#include "sol-flow/selector.h"
+#include "sol-flow-internal.h"
+
+#include "sol-util.h"
+#include "sol-buffer.h"
+
+#include <errno.h>
+
+#define MIN_SIZE(a, b) (((a) < (b)) ? (a) : (b))
+#define MAX_SIZE(a, b) (((a) > (b)) ? (a) : (b))
```
edersondisouza: These 'min' and 'max' macro are becoming more widespread throughout the code. Maybe unify them on utils someday?

src/modules/flow/selector/selector.c
```diff
@@ -0,0 +1,688 @@
...
+    mdata->value_tag = strstr(mdata->format, VALUE_TAG);
+
+    if (!mdata->value_tag) {
+        SOL_WRN("Bad format, no {value} tag: %s", mdata->format);
+        r = -EINVAL;
+        goto err_format;
+    }
+
+    if (mdata->title_tag > mdata->value_tag) {
+        SOL_WRN("Bad format, {title} tag placet after {value} tag: %s",
```
edersondisouza: s/placet/placed

src/modules/flow/selector/selector.c
glima: Yeah, it bothers me too. I'll do that.

src/modules/flow/selector/selector.c
glima: Thanks.

src/modules/flow/selector/selector.json
```diff
@@ -0,0 +1,124 @@
...
+          "data_type": "string",
+          "description": "Defines the sequency of items to be selected by the user. Each incoming packet is added to an internal list. Use the CLEAR port to start over. The strings here will fill in the '{value}' field of the format option (e. g. '{title}:\\n{value} px'), when sent to output. Any line breaking characters is the input strings for this port will be translated to a space instead.",
+          "methods": {
+            "process": "add_item"
+          },
+          "name": "ADD_ITEM"
+        },
+        {
+          "data_type": "any",
+          "description": "Change selection to the next one in the sequence",
```
edersondisouza: I think that this description is not related to the port CLEAR.

src/modules/flow/selector/selector.json
```diff
@@ -0,0 +1,124 @@
...
+          {
+            "data_type": "string",
+            "default": "* ",
+            "description": "If set, selected items will be laid out with this string pre-fixed, separated by a space",
+            "name": "selection_marker"
+          },
+          {
+            "data_type": "string",
+            "default": "> ",
+            "description": "If set, selected items will be laid out with this string pre-fixed, separated by a space",
```
edersondisouza: Do options `cursor_marker` and `selection_marker` have the same descriptions?

src/modules/flow/selector/selector.json
```diff
@@ -0,0 +1,124 @@
...
+        ],
+        "version": 1
+      },
+      "out_ports": [
+        {
+          "data_type": "string",
+          "description": "The formatted, selected entry. This is meant to feed real display nodes (like grove/lcd-string).",
+          "name": "STRING"
+        }
+      ],
```
edersondisouza: Not sure if I understood node purpose, but shouldn't it output raw value too? To use on the flow?

src/modules/flow/selector/selector.c
ibriano: On Wed, 07 Oct 2015 07:23:44 -0700, Gustavo Lima Chaves wrote:
> > + * DATA, OR PROFITS; OR BUSINESS INTERRUPTION) HOWEVER CAUSED AND ON ANY
> > + * THEORY OF LIABILITY, WHETHER IN CONTRACT, STRICT LIABILITY, OR TORT
> > + * (INCLUDING NEGLIGENCE OR OTHERWISE) ARISING IN ANY WAY OUT OF THE USE
> > + * OF THIS SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
> > + */
> > +
> > +#include "sol-flow/selector.h"
> > +#include "sol-flow-internal.h"
> > +
> > +#include "sol-util.h"
> > +#include "sol-buffer.h"
> > +
> > +#include <errno.h>
> > +
> > +#define MIN_SIZE(a, b) (((a) < (b)) ? (a) : (b))
> > +#define MAX_SIZE(a, b) (((a) > (b)) ? (a) : (b))
> 
> Yeah, it bothers me too. I'll do that.

And since we have compound statements and other things like it all over
the place, might be good to use those to ensure we are comparing the
same type.

src/modules/flow/selector/selector.json
glima: Argh, these json docs really need love. Thanks for spotting.

src/modules/flow/selector/selector.json
glima: There'll be another output with the raw value of the selection, yes.

src/modules/flow/selector/selector.c
```diff
@@ -0,0 +1,688 @@
...
+        SOL_WRN("Selector columns number must be a positive integer, "
+            "but %d was given.", opts->columns.val);
+        return -EINVAL;
+    } else
+        mdata->columns = opts->columns.val;
+
+    if (opts->rows.val <= 0) {
+        SOL_WRN("Selector rows number must be a positive integer, "
+            "but %d was given.", opts->rows.val);
+        return -EINVAL;
```
bdilly: Just set rows to "1" and continue. Sure, warning should say that.
We're trying to follow this approach in our node types to be less frustrating as possible to users.

src/modules/flow/selector/selector.c
bdilly: It doesn't seem to be explained on "format" description.
Also, instead of just aborting here I would suggest to use default format.

src/modules/flow/selector/selector.c
```diff
@@ -0,0 +1,688 @@
...
+        mdata->rows = opts->rows.val;
+
+    mdata->format = strdup(opts->format);
+    SOL_NULL_CHECK(mdata->format, -ENOMEM);
+
+    mdata->title_tag = strstr(mdata->format, TITLE_TAG);
+    mdata->value_tag = strstr(mdata->format, VALUE_TAG);
+
+    if (!mdata->value_tag) {
+        SOL_WRN("Bad format, no {value} tag: %s", mdata->format);
```
bdilly: Can't you just use default format in this case (warning the user about it)?

src/modules/flow/selector/selector.c
```diff
@@ -0,0 +1,688 @@
...
+        goto err_format;
+    }
+
+    mdata->text_mem = calloc(1, get_buf_size(mdata));
+    if (!mdata->text_mem) {
+        r = -ENOMEM;
+        goto err_format;
+    }
+
+    if (opts->title) {
```
bdilly: Shouldn't you check if mdata->title_tag is set here?
Does it makes sense to have a title and don't use it in "format"?
Also, if title_tag is set and no title was defined, will you display "Bla (null): 10" for a format "Bla {title}: {value}"?
I would suggest to warn users.

src/modules/flow/selector/selector.c
```diff
@@ -0,0 +1,688 @@
...
+static void
+selector_close(struct sol_flow_node *node, void *data)
+{
+    struct selector_data *mdata = data;
+    uint16_t i;
+    char *it;
+
+    sol_buffer_fini(&mdata->text_grid);
+
+    free(mdata->format);
```
bdilly: Didn't you forget to free mdata->sel_mark, mdata->title and mdata->text_mem here?
Have you tried to run your samples with valgriind?

bdilly:
@glima  I'm missing tests and samples here, even for string generator.
I remember to see a test in your desk some days ago. Did you forget to commit it?

src/modules/flow/selector/selector.c
glima: I thought about that, but is a 1x1 display any more useful than no display at all (error)?

src/modules/flow/selector/selector.c
glima: A bit weird to *tratorar* the user format, but yes, I could.

src/modules/flow/selector/selector.c
glima: Mmm, I guess it's better to skip this if ``mdata->title_tag`` is NULL, indeed.

src/modules/flow/selector/selector.c
glima: Thanks. Weird thing is that valgrind said nothing on ctrl+c.

glima:
Sorry, I forgot to add ``src/test-fbp/selector.fbp``. Doing it now on v2.

glima:
v2 coming, thanks to all.

PR #864, opened 2015-10-07
Builders v2

lgywata:
Changes from V1:
- Remove unused platforms from test dir
- Add test-builders.sh script


cmarcelo:
> each builder currently have two different clones, a HOST clone (compile the host generator executable)

One improvement for the future is to share the HOST clone.

bdilly:
@lgywata not only script are missing proper licensing, all the files are.
You need to add them to fbps, json files, etc

Also I would suggest to merge README  on "Go client/server for builders" commit. But even doing this, c&ping a couple lines on commit msg won't hurt.

I intend to test and do a more granular review later

lgywata:
v3: #865

PR #865, opened 2015-10-07
Builders v3

lgywata:
Changes from V2:

- Add license to files
- Fix style and 'echo' invocations in scripts
- Remove duplicate code and change test-builders


lgywata:
v4: #873

PR #866, opened 2015-10-07
Sol-fbp-generator: Add meta-type code generation support.

cabelitos:


cabelitos:
Merged with ederson suggestions.

PR #867, opened 2015-10-07
Selector [v2]

glima:
Change since #863 :

- we don't quit on bad input, we "fix" it and warn about that
- all json mistakes fixed
- selected ports now there
- missed tests for the node there
- tests for new buffer function there
- all the minor fixes contemplated


glima:
@edersondisouza, any other complaint?

glima:
The extra col for the ``\n`` can't be *extra*, my bad. I'll fix that. Also, the ``grove/lcd-string`` node seems to be b0rked. I'll address that too.

glima:
#898 is there to support this one. After that is merged, we'll be ready to move on with this too.

glima:
It turned out the ``\n`` was indeed right, the support was missing in the lcd node (now done). I won't bother issuing a new version here, I'll fix and merge once that other gets in. Thanks to all.

glima:
Merged with due fixes.

PR #868, opened 2015-10-07
board detection: Add rules for minnowboard max

mbelluzzo:
And have a distinction between minnowboards with kernel newer than 3.18,
when gpio numbering changed, so we can find a proper conffile
'automagically'.

btw I tryed to escape the '.' in an old fashion way but didn't work. That was the only way I found that passed my tests. Suggestions are welcome.

### DIFF
--- a/data/jsons/board_detect.json
+++ b/data/jsons/board_detect.json
@@ -36,6 +36,28 @@
        "match": [ "SALT BAY" ]
       }
      ]
+    },
+    {
+     "name": "intel-minnow-max-linux_gt_3.19",
+     "validation": [
+      {
+       "file_path": "/sys/devices/virtual/dmi/id/bios_version",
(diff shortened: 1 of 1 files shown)


data/jsons/board_detect.json
```diff
@@ -36,6 +36,28 @@
        "match": [ "SALT BAY" ]
       }
      ]
+    },
+    {
+     "name": "intel-minnow-max-new-gpio-numbering",
```
barbieri: awful name, try to add a version or something we can refer. What happens if another new schema appears? -new-new? :-D

Add something like `intel-minnow-max-linux_gt_3.19`.

mbelluzzo: I didn't like the name as well, but nothing better came to my mind. Kind
like yours

mbelluzzo:
Merged with suggested changes

PR #869, opened 2015-10-08
Base64

barbieri:
Base64 allows encoding binary formats in printable form with "safe" chars (if a safe set of 65 are specified). It is often used in web contents.

### DIFF
--- a/src/modules/flow/string/string-common.c
+++ b/src/modules/flow/string/string-common.c
@@ -51,3 +51,226 @@ string_is_empty(struct sol_flow_node *node,
     return sol_flow_send_boolean_packet(node,
         SOL_FLOW_NODE_TYPE_STRING_IS_EMPTY__OUT__OUT, strlen(in_value) == 0);
 }
+
+int
+string_b64encode_string(struct sol_flow_node *node,
+    void *data,
+    uint16_t port,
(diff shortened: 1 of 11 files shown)


src/modules/flow/string/string-common.c
```diff
@@ -51,3 +51,226 @@ string_is_empty(struct sol_flow_node *node,
...
+    /* both b64encode and b64decode have the same options structure, however
+     * the generator emits different symbols. Let's pick one.
+     */
+    opts = (const struct sol_flow_node_type_string_b64encode_options *)options;
+
+    if (!opts->base64_map &&
+        (opts->base64_map[0] == '\0' ||
+        streq(opts->base64_map, SOL_BASE64_MAP)))
+        mdata->base64_map = SOL_BASE64_MAP;
+    else if (opts->base64_map) {
```
lpereira: Might be a good idea to see if all characters are distinct in the map as well

barbieri: fair, but will not validate as it's rare to provide those and it's a PITA to check (would need to walk all 65 entries and call `memchr()` to see if it's in the prior or next segment).

barbieri:
merged with the doc change spotted by @ibriano.

PR #870, opened 2015-10-08
Http static node

ceolin:
- Add new node to serve static files
- Fix the remove_dir function on impl-microhttpdc

### DIFF
--- a/src/modules/flow/http-server/http-server.c
+++ b/src/modules/flow/http-server/http-server.c
@@ -497,4 +497,77 @@ float_process_cb(struct http_data *mdata, const struct sol_flow_packet *packet)
     return sol_flow_packet_get_drange(packet, &mdata->value.d);
 }
 
+/* ---------------------------  static files ----------------------------------------- */
+
+static int
+static_open(struct sol_flow_node *node, void *data, const struct sol_flow_node_options *options)
+{
+    struct http_data *mdata = data;
+    struct sol_flow_node_type_http_server_static_options *opts =
+        (struct sol_flow_node_type_http_server_static_options *)options;
+
+    if (!server) {
+        server = sol_http_server_new(HTTP_SERVER_PORT);
+        SOL_NULL_CHECK(server, -1);
+    }
+    init_count++;
+
+    mdata->path = strdup(opts->path);
+    SOL_NULL_CHECK_GOTO(mdata->path, err);
+
+    mdata->value.b = opts->start;
+
+    if (opts->start)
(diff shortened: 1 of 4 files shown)


src/modules/flow/http-server/http-server.c
```diff
@@ -497,4 +497,77 @@ float_process_cb(struct http_data *mdata, const struct sol_flow_packet *packet)
...
+static int
+static_open(struct sol_flow_node *node, void *data, const struct sol_flow_node_options *options)
+{
+    struct http_data *mdata = data;
+    struct sol_flow_node_type_http_server_static_options *opts =
+        (struct sol_flow_node_type_http_server_static_options *)options;
+
+    if (!server) {
+        server = sol_http_server_new(HTTP_SERVER_PORT);
+        SOL_NULL_CHECK(server, -1);
```
barbieri: should use some -errno, maybe `-ENOMEM` or something like that.

also, would be nice to have this and the init_count as a function that is common to all node types.

src/modules/flow/http-server/http-server.c
```diff
@@ -497,4 +497,77 @@ float_process_cb(struct http_data *mdata, const struct sol_flow_packet *packet)
...
+    }
+    init_count++;
+
+    mdata->path = strdup(opts->path);
+    SOL_NULL_CHECK_GOTO(mdata->path, err);
+
+    mdata->value.b = opts->start;
+
+    if (opts->start)
+        return sol_http_server_add_dir(server, mdata->path);
```
barbieri: there is a missing argument here. This directory should be exposed in the server at some other directory. Like the directory `/srv/http` being exposed as `/static` in the server.

barbieri: also, similar to request handlers that return a handler, we could return a handle from `sol_http_server_add_dir()`, using that to remove/destroy.

src/modules/flow/http-server/http-server.c
```diff
@@ -497,4 +497,77 @@ float_process_cb(struct http_data *mdata, const struct sol_flow_packet *packet)
...
+    struct http_data *mdata = data;
+
+    if (mdata->value.b)
+        sol_http_server_remove_dir(server, mdata->path);
+
+    init_count--;
+    if (!init_count) {
+        sol_http_server_del(server);
+        server = NULL;
+    }
```
barbieri: this block should be a common function as well.

ceolin:
merged with the suggestion, api change will come in another pullreq.Thanks

PR #871, opened 2015-10-08
Http uuid v2

ceolin:
- from v1
  - Use machine_id instead of just using uuid
  - The header is only set on the cliend nodes and it's optional

### DIFF
--- a/src/modules/flow/http-client/http-client.c
+++ b/src/modules/flow/http-client/http-client.c
@@ -42,6 +42,7 @@
 #include "sol-http-client.h"
 #include "sol-json.h"
 #include "sol-mainloop.h"
+#include "sol-platform.h"
 #include "sol-util.h"
 #include "sol-vector.h"
 #include "sol_config.h"
@@ -51,6 +52,7 @@ struct http_data {
     struct sol_str_slice key;
     char *url;
     char *content_type;
+    bool machine_id;
     bool strict;
 };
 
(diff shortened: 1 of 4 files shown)


src/modules/flow/http-client/http-client.c
```diff
@@ -60,6 +62,21 @@ struct http_client_node_type {
     int (*process_data)(struct sol_flow_node *node, struct sol_http_response *response);
 };
 
+static int
+machine_id_header_add(struct sol_http_param *params, char *id)
```
barbieri: Well, the `sol_platform_get_machine_id()` call is strange, it should cache the value and return a pointer to the internal buffer.

would you change that as well, then your call will be simplified as you can keep the reference to that buffer.

barbieri: while at that, also fix the `sol_platform_get_serial_number()` to be the same. These should all be consistent!

src/modules/flow/http-client/http-client.json
```diff
@@ -21,19 +21,25 @@
         ],
         "data_type": "struct http_client_node_type",
         "extra_methods": {
-	  "process_token": "boolean_process_token",
-	  "process_data": "boolean_process_data"
+          "process_token": "boolean_process_token",
+          "process_data": "boolean_process_data"
```
barbieri: avoid these whitespace noise.

ceolin: it was using tab, i fixed it to use spaces

ceolin:
fixes on https://github.com/solettaproject/soletta/pull/886

PR #873, opened 2015-10-08
Builders v4

lgywata:
Changes from V3:
- small fix in README file.
- Change test-builders script so it can be called from any dir.


cmarcelo:
LGTM.

builders/sol/server.go
```diff
@@ -0,0 +1,176 @@
...
+
+	defer os.RemoveAll(dir)
+
+	tr := tar.NewReader(r.Body)
+
+	header, _ := tr.Next()
+	for ; header != nil; header, _ = tr.Next() {
+		switch header.Typeflag {
+		case tar.TypeDir:
+			os.MkdirAll(dir+"/"+header.Name, 0755)
```
lpereira: Is this controlled by the user?

builders/sol/server.go
```diff
@@ -0,0 +1,176 @@
...
+
+		case tar.TypeReg:
+			f, err := os.Create(dir + "/" + header.Name)
```
lpereira: Same here?

builders/sol/server.go
```diff
@@ -0,0 +1,176 @@
...
+	fmt.Fprintln(w, "Internal error")
+}
+
+func compile(dir string, platform string) (*bytes.Buffer, error) {
+	var out bytes.Buffer
+	mw := io.MultiWriter(&out, os.Stdout)
+
+	// TODO: check permission denied error
+
+	cmd := exec.Command(initialDir + "/" + platform + "/compile")
```
lpereira: Are `initialDir` or `platform` controlled by the HTTP request?

builders/sol/server.go
```diff
@@ -0,0 +1,176 @@
...
+			internalError(w, "couldn't call compile: %s", err)
+		}
+	} else {
+		bin, err := os.Open(dir + "/output.zip")
+		if err != nil {
+			// TODO: set HTTP header.
+			fmt.Fprintf(w, "Internal error: %s\n", err)
+			return
+		}
+		w.Header().Set("Content-Type", "application/octet-stream")
```
lpereira: Why not `application/zip`?

builders/sol/server.go
lpereira: Is there any way the environment can be controlled by whoever sends the request? Or, better, yet, will `exec.Command()` give the command a completely new, empty environment, in order to avoid possible contamination?

builders/sol/client.go
```diff
@@ -0,0 +1,253 @@
...
+	}
+	defer resp.Body.Close()
+
+	contentType := resp.Header["Content-Type"][0]
+
+	if strings.HasPrefix(contentType, "text/plain") {
+		io.Copy(os.Stdout, resp.Body)
+	} else if contentType == "application/octet-stream" {
+		os.MkdirAll("out", 0755)
+		bin, err := os.Create("out/" + targetPlatform + ".zip")
```
lpereira: Since `targetPlatform` comes from the server, it might be a good idea to sanitize it before blindly using it to create files (maybe only accept `[a-zA-Z0-9]+`).

builders/sol/server.go
cmarcelo: At the moment, you can provide more isolation between each build by bootstrapping it in the `compile` command (currently a script) for the particular target.

Note the idea is that the sol server itself is expected to run isolated from users environment (we happen to run it directly for testing because it is convenient).

cmarcelo:
@lpereira thanks for looking. I was sure you would since it's a piece of Go code :-)

I'll add the handling of the inputs. For the purposes of this PR we don't expect sol server/client pair to be in the shape of being used in a network (it sends data in plain, it doesn't verify the server identity). We opted to start by figuring out the workflow we wanted for building apps. And then iterate on top of that.

builders/sol/server.go
cmarcelo: No particular reason. But note that could evolve to return more data than the zip file -- so we would fall back to octet-stream.

Specifically for the sol client/server pair I don't see the difference this could make.

cmarcelo:
Replaced by #904.

builders/sol/client.go
cmarcelo: Fixed, thanks.

builders/sol/server.go
cmarcelo: Fixed here and the case below. Thanks.

In this case, the server will consider the input invalid and issues a client error, since renaming the file might make internal references between files to break.

builders/sol/server.go
lpereira: The thing is: if this spawns a shell, this is passing environment variables from the server to the compile script. This might be a problem. See how Shellshock was abused with CGI scripts. Ideally you'd pass a clean environment, containing only the necessary to run the build environment.

cmarcelo: In the CGI case, user input was added as contents of environment variables, that's not the case for this code. The only user input is `targetPlatform` which is validated.

As for the clean environment, I was worrying more about file system and such. Leandro explained to me in IRC. I'll give it a shot (`exec` has API for that).

PR #874, opened 2015-10-08
comms: Send multicast packets to all interfaces

lpereira:
On multi-homed systems, OIC discovery messages was being sent to only
one interface, often not the desired one.  Because of this, it would
often fail to find devices.

Modify sol_socket::sendmsg() so that, whenever sending data to a
multicast address, it loops through all interfaces that are running and
multicast capable, and sets a `struct ip_pktinfo` (and its IPv6
counterpart, `struct ip6_pktinfo`) as ancillary data.  This allows
setting the network interface index that `sol_network` is already aware
of.

### DIFF
--- a/src/lib/comms/sol-socket-impl-linux.c
+++ b/src/lib/comms/sol-socket-impl-linux.c
@@ -251,6 +251,95 @@ sol_socket_recvmsg(struct sol_socket *s, void *buf, size_t len, struct sol_netwo
     return r;
 }
 
+static int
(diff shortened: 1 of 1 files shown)


lpereira:
I can't test here to see if discovery on multi-homed systems will actually work (as I'm from home today), but from a quick glance with `strace`, things seem to be working as expected.

@vcgomes, @bdilly, could you please take a look?

lpereira:
Closing for V2 in #883.

PR #876, opened 2015-10-08
Add extra nodes to handle JSON object and JSON array - v2

otaviobp:
Changelog from v1:
 - Adding function to unescape JSON strings
 - Improve comments
 - some minnor fixes

### DIFF
--- a/src/lib/parsers/include/sol-json.h
+++ b/src/lib/parsers/include/sol-json.h
@@ -41,6 +41,7 @@
 #include <stdint.h>
 #include <sol-macros.h>
 #include <sol-str-slice.h>
+#include <sol-buffer.h>
 
 #ifdef __cplusplus
 extern "C" {
@@ -193,17 +194,6 @@ sol_json_token_str_eq(const struct sol_json_token *token, const char *str, unsig
 
 #define SOL_JSON_TOKEN_STR_LITERAL_EQ(token_, str_) \
     sol_json_token_str_eq(token_, str_, sizeof(str_) - 1)
-
-static inline void
-sol_json_token_remove_quotes(struct sol_json_token *token)
-{
-    if (sol_json_token_get_size(token) < 3)
-        return;
-    if (*token->start == '"')
-        token->start++;
-    if (*(token->end - 1) == '"')
-        token->end--;
-}
 
 /**
  * Get the numeric value of the given token as an 64 bits unsigned integer.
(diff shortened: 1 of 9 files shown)


otaviobp:
Replaces #849

src/lib/parsers/include/sol-json.h
```diff
@@ -436,6 +434,34 @@ int sol_json_double_to_str(const double value, char *buf, size_t buf_len);
...
+ *  Helper function to get a copy of the unescaped and no-quotes string
+ *  produced by sol_json_token_get_unescaped_string.
+ *
+ *  Caller is responsable to free string memory.
+ *
+ *  @param token A string type token
+ *
+ *  @return A a copy of the unescaped string on sucess, NULL on failure.
+ */
+char *sol_json_token_get_unescaped_string_copy(struct sol_json_token *value);
```
barbieri: minor, but this can be `const struct sol_json_token...`, just fix it locally before pushing (or directly pus this small fix to master if it's already committed).

src/lib/parsers/sol-json.c
```diff
@@ -567,3 +567,144 @@ sol_json_is_valid_type(struct sol_json_scanner *scanner, enum sol_json_type type
...
+
+    SOL_NULL_CHECK(token, -EINVAL);
+    SOL_NULL_CHECK(token->start, -EINVAL);
+    SOL_NULL_CHECK(token->end, -EINVAL);
+
+    if (sol_json_token_get_type(token) != SOL_JSON_TYPE_STRING) {
+        sol_buffer_init_flags(buffer, (char *)token->start,
+            sol_json_token_get_size(token), SOL_BUFFER_FLAGS_MEMORY_NOT_OWNED);
+        buffer->used = buffer->capacity;
+        return 0;
```
barbieri: not sure this is required... we're already returning errors if the string is invalid, so I'd say requesting an string with a different token is error as well.

src/lib/parsers/sol-json.c
```diff
@@ -567,3 +567,144 @@ sol_json_is_valid_type(struct sol_json_scanner *scanner, enum sol_json_type type
...
+            case 'u':
+                if (p + 4 < token->end - 1) {
+                    char hexdigits[3], *endptr;
+                    uint16_t n;
+                    hexdigits[2] = '\0';
+
+                    //Add first unicode byte
+                    memcpy(hexdigits, p + 3, 2);
+                    n = strtoul(hexdigits, &endptr, 16);
+                    SOL_INT_CHECK(errno, < 0, -errno);
```
barbieri: you must manually set `errno = 0` before calling `strtoul()`, the man page says:

```
       Since strtoul() can legitimately return 0 or ULONG_MAX (ULLONG_MAX  for
       strtoull()) on both success and failure, the calling program should set
       errno to 0 before the call, and then determine if an error occurred  by
       checking whether errno has a nonzero value after the call.
```

barbieri: also, `errno` is usually `>0` (positive).

src/modules/flow/json/json.c
```diff
@@ -106,83 +105,103 @@ json_object_search_for_token(struct sol_json_scanner *scanner, const char *key,
...
+        return sol_flow_send_boolean_packet(node,
+            SOL_FLOW_NODE_TYPE_JSON_OBJECT_GET_KEY__OUT__BOOLEAN,
+            false);
+    case SOL_JSON_TYPE_NULL:
+        return sol_flow_send_empty_packet(node,
+            SOL_FLOW_NODE_TYPE_JSON_OBJECT_GET_KEY__OUT__NULL);
+    case SOL_JSON_TYPE_STRING:
+        r = sol_json_token_get_unescaped_string(token, &buffer);
+        SOL_INT_CHECK(r, < 0, r);
+        r = sol_flow_send_string_slice_packet(node,
```
barbieri: you could use `get_unescaped_string_copy()` and then use `sol_flow_send_string_take_packet()`, so you reduce allocations if it was escaped (the `take` variant will handle the pointer ownership to the packet).

src/modules/flow/json/json.c
```diff
@@ -224,4 +243,226 @@ json_object_get_key_in_process(struct sol_flow_node *node, void *data, uint16_t
...
+};
+
+
+static bool
+json_array_get_at_index(struct sol_json_scanner *scanner, struct sol_json_token *token, int32_t index, enum sol_json_loop_reason *reason)
+{
+    int32_t cur_index = 0;
+
+    SOL_JSON_SCANNER_ARRAY_LOOP_ALL(scanner, token, *reason) {
+        if (index == cur_index)
```
barbieri: avoid using `index` as it shadows a `<strings.h>` function of that name. New compilers are savvy enough to avoid such errors, but old tools will trigger warnings due shadow declaration.

src/lib/parsers/sol-json.c
otaviobp: This was implemented because conffile uses this to get the string value of other JSON types. I could create another function for that, or checking if the token is a JSON String, but I preferred this way.

src/lib/parsers/sol-json.c
otaviobp: my bad. Thanks

src/lib/parsers/sol-json.c
barbieri: i think it's too error prone, add a conffile function (in its file) and wrap this call with such handling.

otaviobp: ok. I'll do it.

PR #877, opened 2015-10-08
Linux micro kmod v2

dorileo:
## Changes
  + v2 (since #844)
    - impemented coldplug;
    - implemented hotplug;
    - introduced patches:

## Rationale

via uevent netlink channel.


src/lib/datatypes/include/sol-str-slice.h
```diff
@@ -123,6 +124,29 @@ sol_str_slice_to_string(const struct sol_str_slice slice)
     return strndup(slice.data, slice.len);
 }
 
+static inline void
+sol_str_slice_remove_leading_whitespace(struct sol_str_slice *slice)
```
lpereira: I'd rather have a funciton that takes a `struct sol_str_slice` (non-pointer) and returns another `struct sol_str_slice` (non-pointer).

src/lib/datatypes/include/sol-str-slice.h
```diff
@@ -123,6 +124,29 @@ sol_str_slice_to_string(const struct sol_str_slice slice)
...
+{
+    while (isspace(*slice->data)) {
+        slice->data++;
+        slice->len--;
+    }
+}
+
+static inline void
+sol_str_slice_remove_trailing_whitespace(struct sol_str_slice *slice)
```
lpereira: Same here.

This is more consistent with many other usages of the string slice in APIs, where the structure is copied rather than passing a pointer to that structure. It's simple struct so it's pretty cheap.

src/modules/linux-micro/kmod/kmod.c
```diff
@@ -0,0 +1,474 @@
...
+        if (fstatat(dfd, ent.d_name, &st, AT_SYMLINK_NOFOLLOW) == -1) {
+            SOL_ERR("Could not stat %s", ent.d_name);
+            goto next;
+        }
+
+        if (S_ISDIR(st.st_mode)) {
+            int nfd;
+            DIR *ndir;
+
+            nfd = openat(dfd, ent.d_name, O_RDONLY);
```
lpereira: `O_CLOEXEC` missing

src/modules/linux-micro/kmod/kmod.c
```diff
@@ -0,0 +1,474 @@
...
+                goto next;
+            }
+
+            kmod_coldplug_find_devices(kdata, ndir);
+            closedir(ndir);
+        } else if (!strcmp(ent.d_name, "modalias")) {
+            int fd;
+            char buff[PATH_MAX];
+
+            fd = openat(dfd, ent.d_name, O_RDONLY);
```
lpereira: `O_CLOEXEC` missing

src/modules/linux-micro/kmod/kmod.c
```diff
@@ -0,0 +1,474 @@
...
+static void
+kmod_coldplug_find_devices(struct kmod_data *kdata, DIR *dir)
+{
+    struct dirent ent, *entbuf = NULL;
+    int r, dfd;
+
+    dfd = dirfd(dir);
+    if (dfd < 0) {
+        SOL_DBG("Could not obtain directory file descriptor");
+        goto next;
```
lpereira: Why `goto next`?

src/modules/linux-micro/kmod/kmod.c
```diff
@@ -0,0 +1,474 @@
...
+        r = readdir_r(dir, &ent, &entbuf);
+    }
+}
+
+static int
+kmod_start_coldplug(struct kmod_data *kdata)
+{
+    DIR *dir;
+
+    dir = opendir("/sys/devices");
```
lpereira: Is there any way to assert that the filesystem mounting code has executed before this one?

src/modules/linux-micro/kmod/kmod.c
```diff
@@ -0,0 +1,474 @@
...
+    if (dfd < 0) {
+        SOL_DBG("Could not obtain directory file descriptor");
+        goto next;
+    }
+
+    r = readdir_r(dir, &ent, &entbuf);
+    while (r == 0 && entbuf) {
+        struct stat st;
+
+        if (!strcmp(ent.d_name, ".") || !strcmp(ent.d_name, "..")) {
```
lpereira: Do we want to process hidden files as well?

src/modules/linux-micro/kmod/kmod.c
```diff
@@ -0,0 +1,474 @@
...
+kmod_uevent_start(struct kmod_data *kdata)
+{
+    struct sockaddr_nl nls;
+
+    memset(&nls, 0, sizeof(struct sockaddr_nl));
+    nls.nl_family = AF_NETLINK;
+    nls.nl_pid = getpid();
+    nls.nl_groups = -1;
+
+    kdata->uevent.fd = socket(PF_NETLINK, SOCK_DGRAM, NETLINK_KOBJECT_UEVENT);
```
lpereira: `SOCK_DGRAM | SOCK_CLOEXEC`

src/modules/linux-micro/kmod/kmod.c
```diff
@@ -0,0 +1,474 @@
...
+    while (i < len) {
+        char *curr, *del;
+        size_t value_len;
+
+        curr = msg + i;
+        del = strrchr(curr, '=');
+        value_len = del - curr;
+
+        if (del && !strncmp(curr, "MODALIAS", value_len))
+            modalias = strdupa(++del);
```
lpereira: This seems like a dangerous use of `strdupa()` with a user-controllable string.

src/modules/linux-micro/kmod/kmod.c
```diff
@@ -0,0 +1,474 @@
...
+    lines = sol_util_str_split(content, "\n", 0);
+    SOL_VECTOR_FOREACH_IDX (&lines, itr, idx) {
+        char *alias = NULL;
+
+        sol_str_slice_remove_leading_whitespace(itr);
+        if (*itr->data == '#' || *itr->data == ';' || *itr->data == '\0')
+            continue;
+
+        sol_str_slice_remove_trailing_whitespace(itr);
+        alias = strndupa(itr->data, itr->len);
```
lpereira: Possibly dangerous use of strndupa().

src/modules/linux-micro/kmod/kmod.c
barbieri: we do mount `/sys` before proceeding to services.

src/lib/datatypes/include/sol-str-slice.h
barbieri: +1, this makes it easier to use.

src/modules/linux-micro/kmod/kmod.c
barbieri: indeed, using `alloca()` and variations will explode stack if called inside the loop... the stack just goes back once the loop is finished :-/

I do recommend you to use `struct sol_buffer` here, initialize it before the loop and then call `sol_buffer_set_slice()`, it will reuse the allocated memory if possible, otherwise will grow as needed. Remember to call `sol_buffer_fini()` afterwards to release allocated memory.

by default, the buffer adds null-terminator, so safe to use as a string.

src/modules/linux-micro/kmod/kmod.c
barbieri: same here, use the `sol_buffer` trick

src/modules/linux-micro/kmod/kmod.c
dorileo: skipping them not processing them.

src/modules/linux-micro/kmod/kmod.c
dorileo: ok

src/modules/linux-micro/kmod/kmod.c
dorileo: code shifting/left over.

src/modules/linux-micro/kmod/kmod.c
dorileo: ok

src/modules/linux-micro/kmod/kmod.c
dorileo: ok

src/modules/linux-micro/kmod/kmod.c
lpereira: Yes, but this is skipping only `.` and `..`. Hidden files will still be processed.

dorileo: if sysfs is exporting devices data as hidden files then yes... this is a very restrict scenario, we're dealing only with ```/sys/devices``` here, so ok to not be so pedantic with that. we're mostly interested on ```modalias``` files, if we find one then try to coldplug it, even if it's inside a hidden dir or not.

dorileo:
v3 coming

PR #878, opened 2015-10-08
json: Add helpers to serialize basic types

ceolin:

### DIFF
--- a/src/lib/parsers/include/sol-json.h
+++ b/src/lib/parsers/include/sol-json.h
@@ -39,6 +39,7 @@
 #include <string.h>
 #include <errno.h>
 #include <stdint.h>
+#include <sol-buffer.h>
 #include <sol-macros.h>
 #include <sol-str-slice.h>
 
@@ -434,6 +435,11 @@ char *sol_json_escape_string(const char *str, char *buf, size_t len) SOL_ATTR_NO
 int sol_json_double_to_str(const double value, char *buf, size_t buf_len);
 
 bool sol_json_is_valid_type(struct sol_json_scanner *scanner, enum sol_json_type start_type) SOL_ATTR_NONNULL(1);
+
+int sol_json_serialize_string(struct sol_buffer *buffer, const char *str) SOL_ATTR_NONNULL(1, 2);
+int sol_json_serialize_double(struct sol_buffer *buffer, double val) SOL_ATTR_NONNULL(1);
+int sol_json_serialize_int(struct sol_buffer *buffer, int val) SOL_ATTR_NONNULL(1);
+int sol_json_serialize_boolean(struct sol_buffer *buffer, bool val) SOL_ATTR_NONNULL(1);
 
 /**
  * @}
--- a/src/lib/parsers/sol-json.c
+++ b/src/lib/parsers/sol-json.c
@@ -567,3 +567,75 @@ sol_json_is_valid_type(struct sol_json_scanner *scanner, enum sol_json_type type
            sol_json_scanner_skip_over(scanner, &token) &&
            token.end == last_position;
 }
+
+SOL_API int
+sol_json_serialize_string(struct sol_buffer *buffer, const char *str)
+{
+    int r;
+    size_t escaped_len, new_size;
+
+    SOL_NULL_CHECK(buffer, -EINVAL);
+    SOL_NULL_CHECK(str, -EINVAL);
+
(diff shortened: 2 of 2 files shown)


src/lib/parsers/include/sol-json.h
```diff
@@ -435,6 +436,11 @@ int sol_json_double_to_str(const double value, char *buf, size_t buf_len);
 
 bool sol_json_is_valid_type(struct sol_json_scanner *scanner, enum sol_json_type start_type) SOL_ATTR_NONNULL(1);
 
+int sol_json_serialize_string(struct sol_buffer *buffer, const char *str) SOL_ATTR_NONNULL(1, 2);
+int sol_json_serialize_double(struct sol_buffer *buffer, double val) SOL_ATTR_NONNULL(1);
+int sol_json_serialize_int(struct sol_buffer *buffer, int val) SOL_ATTR_NONNULL(1);
```
barbieri: call it int32 and use that specific type, otherwise this will cause problems in systems where int is not 4 bytes.

src/lib/parsers/sol-json.c
```diff
@@ -567,3 +567,75 @@ sol_json_is_valid_type(struct sol_json_scanner *scanner, enum sol_json_type type
...
+    size_t escaped_len, new_size;
+
+    SOL_NULL_CHECK(buffer, -EINVAL);
+    SOL_NULL_CHECK(str, -EINVAL);
+
+    escaped_len = sol_json_calculate_escaped_string_len(str);
+    r = sol_util_size_add(buffer->used, escaped_len + 2, &new_size);
+    SOL_INT_CHECK(r, < 0, r);
+
+    r = sol_buffer_ensure(buffer, buffer->used + escaped_len + 2);
```
barbieri: this should be `new_size` as it was computed to be `buffer->used + escaped_len + 2`

src/lib/parsers/sol-json.c
```diff
@@ -567,3 +567,75 @@ sol_json_is_valid_type(struct sol_json_scanner *scanner, enum sol_json_type type
...
+{
+    int r;
+    char val_str[64];
+
+    SOL_NULL_CHECK(buffer, -EINVAL);
+
+    r = sol_json_double_to_str(val, val_str, sizeof(val_str));
+    SOL_INT_CHECK(r, < 0, r);
+
+    r = sol_buffer_append_slice(buffer, SOL_STR_SLICE_STR(val_str, strlen(val_str)));
```
barbieri: minor, but here you could ensure the buffer is `buffer->used + 64` (with proper overflow checks) and then use `p = sol_buffer_at_end(buffer)` to get the pointer and handle it to `sol_json_double_to_str()`. As this function does not return the number of bytes written (maybe it should), you can then `buffer->used += strlen(p)`

src/lib/parsers/sol-json.c
```diff
@@ -567,3 +567,75 @@ sol_json_is_valid_type(struct sol_json_scanner *scanner, enum sol_json_type type
...
+}
+
+SOL_API int
+sol_json_serialize_int(struct sol_buffer *buffer, int val)
+{
+    int r;
+
+    SOL_NULL_CHECK(buffer, -EINVAL);
+
+    r = sol_buffer_append_printf(buffer, "%d", val);
```
barbieri: change all to `int32_t` and `%" PRId32`

src/lib/parsers/sol-json.c
```diff
@@ -567,3 +567,75 @@ sol_json_is_valid_type(struct sol_json_scanner *scanner, enum sol_json_type type
...
+    SOL_INT_CHECK(r, < 0, r);
+
+    return 0;
+}
+
+SOL_API int
+sol_json_serialize_boolean(struct sol_buffer *buffer, bool val)
+{
+    int r;
+    const char *val_str = (val) ? "true" : "false";
```
barbieri: this could be easier:

```c
    static const struct sol_str_slice t_str = SOL_STR_SLICE_LITERAL("true");
    static const struct sol_str_slice f_str = SOL_STR_SLICE_LITERAL("false");
    r = sol_buffer_append_slice(buffer, val ? t_str : f_str);
    SOL_INT_CHECK(r, < 0, r);
   return 0;
```

ceolin:
fixes on https://github.com/solettaproject/soletta/pull/879

PR #879, opened 2015-10-08
Json v2

ceolin:
from v1:
 - use int32
 - added commit fix a leak on http client


barbieri:
+1, commit it.

To make it complete, please add a int64, uint64 and uint32 versions as well. no need to make a new PR.

ceolin:
merged with the suggested functions. Thanks

PR #881, opened 2015-10-09
Base16

barbieri:
Similar to base64, this is the hexadecimal encoder -- often called base16.

It's mostly a copy&paste from base64, that's why I fixed some base64 issues while editing the code.


barbieri:
rebased and merged

PR #882, opened 2015-10-09
modules/flower-power: Initialize variable

anselmolsm:
Coverity (130998) pointed out that a SOL_INT_CHECK_GOTO could
happen with r uninitialized.


anselmolsm:
Applied, tks.

PR #883, opened 2015-10-09
comms: Send multicast packets to all interfaces

lpereira:
Changes from V1:
   - Better error handling

On multi-homed systems, OIC discovery messages were being sent to only
one interface, often not the desired one.  Because of this, it would
fail to find devices.

Modify sol_socket::sendmsg() so that, whenever sending data to a
multicast address, it loops through all interfaces that are running and
multicast capable, calling sendmsg(2) with a `struct ip_pktinfo` (and
its IPv6 counterpart, `struct ip6_pktinfo`) as ancillary data with the
network interface index.  Consider a success if at least one of these
sendmsg(2) calls succeed.

This change is only required in the Linux implementation at the moment,
and is completely transparent on an API level.

### DIFF
--- a/src/lib/comms/sol-socket-impl-linux.c
+++ b/src/lib/comms/sol-socket-impl-linux.c
@@ -251,6 +251,97 @@ sol_socket_recvmsg(struct sol_socket *s, void *buf, size_t len, struct sol_netwo
     return r;
 }
 
+static bool
+sendmsg_multicast_addrs(int fd, const struct sol_network_link *link,
+    struct msghdr *msg)
+{
+    union {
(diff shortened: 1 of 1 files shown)


bdilly:
I've tested the following cases (without success) - dev machine is client and board is server:
 1) on dev machine (usbethernet + ethernet) and on board (usbethernet) : didn't work
 2) on dev machine (usbethernet + ethenert + wifi) and on board (usbethernet + wifi) didn't work

We've tried some variations disabling multicast on interfaces that shouldn't be used but that didn't work as well.

The only way communication worked was only leaving wifi on dev machine

lpereira:
I've found an alternative method and will try that next week.

lpereira:
Closing for V3 at #891.

PR #884, opened 2015-10-09
test-str-slice: fix memory leak

dorileo:
sol_str_slice_to_string() will allocate a new string, this string must
be freed.


dorileo:
applied

PR #885, opened 2015-10-09
datatypes: Make room for the nul char at the end of the string

ibriano:


dorileo:
+1

ibriano:
Merged

barbieri:
doesn't hurt, but I wonder why... It's a map, not a string :-D

ibriano:
On Fri, 09 Oct 2015 12:17:03 -0700, Gustavo Sverzut Barbieri wrote:
> doesn't hurt, but I wonder why... It's a map, not a string :-D

The node type is using `streq()` on the map it gets from the options
with this one to see if they are equal, which then tries to read the
final \0 that's not part of the array declared size. Probably should
have changed that call to use strneq instead, but I saw that afterwards.

barbieri:
then my code, my bad... autospank :-D

PR #886, opened 2015-10-09
Http uuid v3

ceolin:
from v2:

 - Cache serial number
 - Cache machine id


glima:
Ack on the rest.

ceolin:

PR #888, opened 2015-10-09
Some more GPIO fixing

ibriano:
Better handling of interrupts not being supported on Linux, and don't
open a gpio/reader node if it won't generate any events.


ibriano:
Merged

PR #889, opened 2015-10-12
Crypto message digest

barbieri:
Add crypto subsystem with message-digest support using openssl or linux kernel api.

unfortunately the linux kernel API uses socket() that is not poll() friendly, thus a worker thread is used (or a timer if it's not enabled).

The non-threaded approach doesn't try to segment load, thus if a big blob is used it will take some time and may block the main loop. We can improve this later by limiting the feed size if its required.

Please check @lpereira @ceolin @ibriano and @ulissesf.


barbieri:
did a push with some fixes and md5sum.c, I guess nobody did the review during the holiday :-D

src/lib/crypto/Kconfig
```diff
@@ -0,0 +1,41 @@
...
+	depends on FEATURE_CRYPTO_MESSAGE_DIGEST
+	default CRYPTO_MESSAGE_DIGEST_LINUX_KCAPI if LINUX
+	help
+            Message Digest algorithms will take a byte stream and
+            compute a hash that may be used to later validate the
+            identity. Even the smallest variation of the input data
+            will have an avalanche effect that drastically change the
+            output data.
+
+            Common Message Digest algorithms are CRC32, MD5, SHA1,
```
lpereira: Associating `CRC32` with crypto is quite funny. Please make a note that `MD5` and `SHA1` should not be used unless required for interoperability. They've been broken for a while. Even `SHA256` is beginning to show its age.

src/lib/crypto/Kconfig
```diff
@@ -0,0 +1,41 @@
...
+            This implementation will use the Kernel-Userspace interface
+            provided by Linux. Linux implementation are often very optimized
+            and in some case are hardware-accelerated.
+
+config CRYPTO_MESSAGE_DIGEST_OPENSSL
+	bool "OpenSSL"
+	depends on HAVE_OPENSSL && FEATURE_CRYPTO_MESSAGE_DIGEST
+	select CRYPTO_MESSAGE_DIGEST
+	help
+            This implementation will use OpenSSL.
```
lpereira: Maybe only require `libcrypto` here instead of the full OpenSSL stack?

src/lib/crypto/sol-message-digest-impl-linux-kcapi.c
```diff
@@ -0,0 +1,957 @@
...
+    } req;
+    struct sockaddr_nl snl = {
+        .nl_family = AF_NETLINK,
+    };
+    struct iovec iov;
+    struct msghdr msg;
+    bool ret = false;
+    int fd;
+
+    fd =  socket(AF_NETLINK, SOCK_RAW, NETLINK_CRYPTO);
```
lpereira: `SOCK_RAW | SOCK_CLOEXEC` ?

src/lib/crypto/sol-message-digest-impl-linux-kcapi.c
```diff
@@ -0,0 +1,957 @@
...
+                    SOL_DBG("message digest algorithm \"%s\" digest_size=%zd",
+                        info->name, info->digest_size);
+                    ret = true;
+                    break;
+                }
+            }
+        }
+    }
+
+error_sendmsg:
```
lpereira: If all error paths are the same, why not just a single cleanup label?

src/lib/crypto/sol-message-digest-impl-linux-kcapi.c
```diff
@@ -0,0 +1,957 @@
...
+    pd->blob = input;
+    pd->is_digest = false;
+
+    _sol_message_digest_unlock(handle);
+    sol_worker_thread_feedback(handle->thread);
+    return;
+
+error:
+    _sol_message_digest_unlock(handle);
+    sol_blob_unref(input); /* this may cause problems if main thread changes blob refcnt */
```
lpereira: Maybe it is the case we `assert(pthread_self() is the correct thread)`?

src/lib/crypto/sol-message-digest-impl-openssl.c
```diff
@@ -0,0 +1,780 @@
...
+            config->api_version, SOL_MESSAGE_DIGEST_CONFIG_API_VERSION);
+        return NULL;
+    }
+
+    if (!did_openssl_load_digests) {
+        OpenSSL_add_all_digests();
+        did_openssl_load_digests = true;
+    }
+
+    md = EVP_get_digestbyname(config->algorithm);
```
lpereira: Wouldn't it be a good idea to standardize the digest names in order to keep the uses portable? Or are the strings exported by `kcapi` the same supported by `libcrypto`?

Maybe a `sol_str_table` with a fallback to NULL, where it would then be sent to the implementation to decide?

src/samples/crypto/message-digest.c
```diff
@@ -0,0 +1,297 @@
...
+    struct timespec elapsed;
+    double size, rate, seconds;
+    const char *s_unit, *r_unit;
+
+    sol_util_timespec_sub(&now, &ctx->start, &elapsed);
+    seconds = elapsed.tv_sec + (double)elapsed.tv_nsec / NSEC_PER_SEC;
+
+    size = amount;
+    if (size >= 1.0e9) {
+        s_unit = "Gb";
```
lpereira: Gigabit?

lpereira:
It's a large code dump and I can't possibly review every nook and cranny without spending a lot of time. I'd say that, from a cursory glance, it looks good, sans a few nits.

ibriano:
Nothing to add to what @lpereira said, looks good overall.

src/lib/crypto/sol-message-digest-impl-linux-kcapi.c
barbieri: oops, did forget about it.

src/lib/crypto/sol-message-digest-impl-linux-kcapi.c
barbieri: could be, but I like this way to keep it explicit when I was writing.

src/lib/crypto/sol-message-digest-impl-linux-kcapi.c
barbieri: it won't be the correct thread, we're from a worker thread trying to unref the input blob... this should happen from the main thread, but we can't allocate memory to append this request for the main thread -- it's a cumbersome situation.

The other way is to leak the input blob.

src/lib/crypto/sol-message-digest-impl-openssl.c
barbieri: most match and I'm documenting the standard names are those for the kcapi.

we can't simply use a translation table as openssl uses another set of functions for HMAC, thus we need to "parse" the string `"hmac(sha1)"` and `EVP_get_digestbyname("sha1")` then use the HMAC function to create the digest handle.

barbieri:
granted it grew bigger than I wondered... my initial implementation based solely on `sol_fd` was smaller, but the fscking crypto api doesn't behave like I wished and thus I had to do the thread version... to later realize it wasn't available everywhere then I did the timer, then the mess was big enough.

that said, all the possibly nasty bits are not implemented in this code, it's only a shim to abstract the actual algorithms that were written elsewhere. What matters is that it works and doesn't leak memory -- the md5sum.c test can be quite helpful to test various cases, particularly by handling a large blob at once (files) or to handle smaller chunks via stdin -- this caught one of the bugs

barbieri:
fixed @lpereira's comments and pushed to master.

later on I'll send a patch creating a "-common" that will do the thread and timer logic, leaving each implementation to be much smaller.

PR #891, opened 2015-10-13
comms: Send multicast packets to all interfaces [v3]

lpereira:
Changes from v2:
*  Use a different approach with `setsockopt()` and `getsockopt()` instead of ancillary messages

On multi-homed systems, OIC discovery messages were being sent to only one interface, often not the desired one.  Because of this, it would fail to find devices.

Modify sol_socket::sendmsg() so that, whenever sending data to a multicast address, it loops through all interfaces that are running and multicast capable, setting `IP_MULTICAST_IF` or `IPV6_MULTICAST_IF` (as appropriate per the address family), sending the packet, and then
restoring the original option.

This change is only required in the Linux implementation at the moment, and is completely transparent on an API level.

### DIFF
--- a/src/lib/comms/sol-socket-impl-linux.c
+++ b/src/lib/comms/sol-socket-impl-linux.c
@@ -251,6 +251,113 @@ sol_socket_recvmsg(struct sol_socket *s, void *buf, size_t len, struct sol_netwo
     return r;
 }
 
+static bool
+sendmsg_multicast_addrs(int fd, struct sol_network_link *net_link,
+    struct msghdr *msg)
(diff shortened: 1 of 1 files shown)


bdilly:
lgtm

I've tested it our regular development configuration (using a dev board connected via usbethernet), while client had other interfaces, and it worked just fine.

lpereira:
Awesome. I'll address @vcgomes concerns and merge the patch ASAP. Thanks for the test and review!

lpereira:
Merged, thanks!

PR #893, opened 2015-10-13
http:server: Document the api

ceolin:


lpereira:
Rubberstamp+.

src/lib/comms/include/sol-http-server.h
```diff
@@ -56,25 +56,126 @@ extern "C" {
 struct sol_http_server;
 struct sol_http_request;
 
+/**
+ * It creates a http's server on that will bind on all interfaces in the given port.
```
lpereira: HTTP. It's an acronym. No need to start with `It`:

   Creates an HTTP server, binding on all interfaces at the specified @port.

src/lib/comms/include/sol-http-server.h
```diff
@@ -56,25 +56,126 @@ extern "C" {
...
+ * same port will result in fail.
+ *
+ * @param port The port where the server will bind.
+ *
+ * @return a handle to the server on success, otherwise @c NULL is returned.
+ */
 struct sol_http_server *sol_http_server_new(uint16_t port);
+
+/**
+ * Deletes the server and consequently all handlers and dir added.
```
lpereira: Will it also destroy the directories I've added, removing all the files contained therein? That sounds dangerous.

src/lib/comms/include/sol-http-server.h
```diff
@@ -56,25 +56,126 @@ extern "C" {
...
+ * return automatically a response with the code 304 (not modified) when the request contains the header
+ * If-Since-Modified greater than the value given in this function.
+ *
+ * @param server The value got with @c sol_http_server_new
+ * @param path The same path given on @c sol_http_server_register_handler
+ * @param modified The time it was modified
+ *
+ * @return '0' on success, error code (always negative) otherwise.
+ */
 int sol_http_server_set_last_modified(struct sol_http_server *server, const char *path, time_t modified);
```
lpereira: Is that global per instance of the server? I don't get how this API would be useful...

src/lib/comms/include/sol-http-server.h
ceolin: nope, i want to say that is not need remove the added dirs

lpereira: I know, but the documentation is dubious and scary-sounding. I'd just say that this destroys the server instance, leaving the removal of handlers and directory references implicit.

src/lib/comms/include/sol-http-server.h
ceolin: nope, it's path's specific

src/lib/comms/include/sol-http-server.h
ceolin: ack

ceolin:
merged with the fixes. Thanks

src/lib/comms/include/sol-http-server.h
barbieri: then you should always get from `struct stat::st_mtime`

lpereira: If `path` is a file, then,  yeah, it's possible. Otherwise, it doesn't make much sense.

barbieri: are we able to server something that is not a file?

if we're generating index.html, then we can use the newest file as the index.html mtime.

PR #894, opened 2015-10-13
crypto/message-digest: add common infrastructure, improve non-threaded.

barbieri:
this moves the common bits from linux-kcapi and openssl to a single
file, allows that cumbersome core to be shared with other
implementations.

it also introduces a maximum limit per feed if not using threads, this
way we avoid locking the main loop for too long.

### DIFF
--- a/src/test/test-message-digest.c
+++ b/src/test/test-message-digest.c
@@ -1,0 +1,405 @@
+/*
+ * This file is part of the Soletta Project
+ *
+ * Copyright (C) 2015 Intel Corporation. All rights reserved.
+ *
+ * Redistribution and use in source and binary forms, with or without
+ * modification, are permitted provided that the following conditions
+ * are met:
+ *
+ *   * Redistributions of source code must retain the above copyright
+ *     notice, this list of conditions and the following disclaimer.
+ *   * Redistributions in binary form must reproduce the above copyright
+ *     notice, this list of conditions and the following disclaimer in
+ *     the documentation and/or other materials provided with the
+ *     distribution.
+ *   * Neither the name of Intel Corporation nor the names of its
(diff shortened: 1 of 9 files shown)


barbieri:
updated to include tests and a fix to coverity issue CID 131380

src/test/test-message-digest.c
```diff
@@ -0,0 +1,405 @@
...
+
+static void
+test_md5_single(void)
+{
+    /* note: mem includes string trailing null byte */
+    static const struct md_test md5_single_test[] = {
+        MD_TEST("md5", SOL_STR_SLICE_EMPTY, "test", "e2a3e68d23ce348b8f68b3079de3d4c9"),
+        MD_TEST("md5", SOL_STR_SLICE_EMPTY, "long line of text bla bla bla more text here yada yada", "18511ce4f220de4744390ca3ae72873f"),
+        MD_TEST("md5", SOL_STR_SLICE_EMPTY, big_blob_of_zeros, "ab893875d697a3145af5eed5309bee26"),
+        MD_TEST("md5", SOL_STR_SLICE_EMPTY, big_blob_of_chars, "9a36eacb09f8e98e103e9ee897f8e31c"),
```
lpereira: I"d also test for the empty string (which should evaluate to `d41d8cd98f00b204e9800998ecf8427e`).

ibriano:
Seems fine

barbieri:
pushed to master a rebased and reordered patch (moving the kcapi fix earlier).

It adds the hash for zero-sized memory as requested by @lpereira.

It fixes a bug for the chunked content test that would leave a timer to be called after the context was deleted, causing a bug in the test code (not the tested code). Now keep a reference to the timer and delete it alongside with chunked_ctx.

PR #896, opened 2015-10-14
Board detect

mbelluzzo:
Fix minnow name because I f**k up and applied k-s suggested name without test <because had lost the minnow to someone else already>.

Add detection to "raspberry pi", actually it's processor. I couldn't find a way to detect that the processor was actually on a raspberry board.

### DIFF
--- a/data/jsons/board_detect.json
+++ b/data/jsons/board_detect.json
@@ -38,7 +38,7 @@
      ]
     },
     {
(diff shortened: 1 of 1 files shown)


brunobottazzini:
:+1:  for invalid char in minnowboard name.

I have tested it here and the warning has disappeared.

Thanks!

PR #897, opened 2015-10-14
Add nodes to extract data from JSON using JSON path

otaviobp:
To extract data from complex JSON objects and JSON arrays, we need to
create multiple json/object-get-key and json/array-get-index to find the
desired elements.

Nodes json/object-get-path and json/array-get-path were created to
extract inner data from JSON elements using a single path that points to
the specific desired element.

JSON path follows specification from http://goessner.net/articles/JsonPath/,
but multiple elements extraction was not implemented

### DIFF
--- a/src/modules/flow/converter/converter.c
+++ b/src/modules/flow/converter/converter.c
@@ -2127,13 +2127,25 @@ timestamp_error:
     return 0;
 }
 
-static bool
+static struct sol_blob *
 json_validate(struct sol_blob *blob, enum sol_json_type type)
 {
     struct sol_json_scanner scanner;
+    char *p;
+    size_t size;
+
+    p = (char *)blob->mem + blob->size - 1;
+    while (isspace(*p))
+        p--;
+    size = p - (char *)blob->mem + 1;
+
+    if (size < blob->size)
+        blob = sol_blob_new(SOL_BLOB_TYPE_NOFREE, blob, blob->mem, size);
 
     sol_json_scanner_init(&scanner, blob->mem, blob->size);
-    return sol_json_is_valid_type(&scanner, type);
+    if (sol_json_is_valid_type(&scanner, type))
+        return blob;
+    return NULL;
 }
 
 static int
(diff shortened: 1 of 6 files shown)


otaviobp:
I just did a push -f (as no one has commented yet) to add extra tests and fix a minor mistake I did

src/modules/flow/converter/converter.c
```diff
@@ -2127,13 +2127,25 @@ string_to_timestamp_convert(struct sol_flow_node *node, void *data, uint16_t por
...
+static struct sol_blob *
 json_validate(struct sol_blob *blob, enum sol_json_type type)
 {
     struct sol_json_scanner scanner;
+    char *p;
+    size_t size;
+
+    p = (char *)blob->mem + blob->size - 1;
+    while (isspace(*p))
+        p--;
```
lpereira: Buffer underrun here: if string is comprised solely of spaces, `p--` may go past it. Ensure `p >= blob->mem` in the `while` condition.

src/modules/flow/json/json.c
```diff
@@ -90,15 +120,72 @@ create_sub_json(struct sol_blob *parent, struct sol_json_scanner *scanner, struc
 }
 
 static bool
-json_object_search_for_token(struct sol_json_scanner *scanner, const char *key, struct sol_json_token *value)
+_get_next_slice(const struct sol_str_slice path, struct sol_str_slice *slice)
```
lpereira: This function is pretty tough to review. Maybe you could use the FBP lexer/parser as an inspiration?

src/modules/flow/json/json.c
```diff
@@ -186,28 +273,163 @@ send_token_packet(struct sol_flow_node *node, struct sol_json_scanner *scanner,
...
+    const char *end, *p, *p2;
+    char *dest = NULL, *last;
+
+    end = slice.data + slice.len;
+    for (p = slice.data + 2; p < end; p = p2 + 1) {
+        p2 = memchr(p, '\'', end - p);
+        if (!p2 || p2 == end - 2)
+            break;
+        if (!dest) {
+            dest = malloc(slice.len - 5);
```
lpereira: Too many magic numbers and variables with meaningless names. I'm not sure what they mean.

src/modules/flow/json/json.c
```diff
@@ -186,28 +273,163 @@ send_token_packet(struct sol_flow_node *node, struct sol_json_scanner *scanner,
...
+                scanner.current = value.start;
+                continue;
+            }
+            break;
+        case SOL_JSON_TYPE_ARRAY_START:
+            if (_sol_json_loop_helper_init(&scanner, &token,
+                SOL_JSON_TYPE_ARRAY_START) != SOL_JSON_LOOP_REASON_OK)
+                goto error;
+
+            if (*key_slice.data != '[' || *(key_slice.data + 1) == '\'')
```
lpereira: Why not just `if (!strncmp(key_slice.data, "['", 2))`?

PR #898, opened 2015-10-14
Lcd

glima:
While testing my #867 PR, I found out that the lcd nodes were rotting bad. It took me a while to fix the concurrency/bad access issues, but here it goes.

@zehortigoza, could you help reviewing?

### DIFF
--- a/src/lib/io/sol-i2c-impl-linux.c
+++ b/src/lib/io/sol-i2c-impl-linux.c
@@ -903,11 +903,11 @@ sol_i2c_write_register(struct sol_i2c *i2c, uint8_t reg, const uint8_t *values, 
 
 #ifdef WORKER_THREAD
     i2c->async.worker = sol_worker_thread_new(&spec);
-    SOL_NULL_CHECK(i2c->async.worker, false);
+    SOL_NULL_CHECK(i2c->async.worker, NULL);
     return (struct sol_i2c_pending *)i2c->async.worker;
 #else
     i2c->async.timeout = sol_timeout_add(0, i2c_write_reg_timeout_cb, i2c);
-    SOL_NULL_CHECK(i2c->async.timeout, false);
+    SOL_NULL_CHECK(i2c->async.timeout, NULL);
     return (struct sol_i2c_pending *)i2c->async.timeout;
 #endif
 }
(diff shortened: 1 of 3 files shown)


src/lib/io/sol-i2c-impl-linux.c
```diff
@@ -903,11 +903,11 @@ sol_i2c_write_register(struct sol_i2c *i2c, uint8_t reg, const uint8_t *values,
 
 #ifdef WORKER_THREAD
     i2c->async.worker = sol_worker_thread_new(&spec);
-    SOL_NULL_CHECK(i2c->async.worker, false);
+    SOL_NULL_CHECK(i2c->async.worker, NULL);
```
lpereira: I suspect there are ohter places in the code where `NULL` is returned instead of `false`, or an `int` is returned instead of a boolean (with the implicit casting and all). I wonder if there's a way to find all occurrences of this.

src/modules/flow/grove/grove.c
```diff
@@ -660,101 +675,150 @@ write_string(struct lcd_data *mdata, char *str)
...
+        return -EIO;
     }
 
-    SOL_DBG("ChipAddr: 0x%02x, DataAddr: 0x%02x, Value 0x%02x - %c",
-        cmd->chip_addr, cmd->data_addr, cmd->value, cmd->value);
+    /* Unfortunately we can't afford not replicating this memory. At
+     * least &cmd->value must be alive during the whole callback time,
+     * and we can't guarantee that on our side with so many list
+     * operations going on. */
+    cpy = calloc(1, sizeof(*cpy));
```
lpereira: Use `malloc()` here, since the whole block is going to be overwritten anyway, there's no need to zero it out.

mbelluzzo:
+2, works for me. Just s/calloc/malloc as Leandro pointed.

glima:
Thanks, everybody. Merging soon with the due fixes.

src/lib/io/sol-i2c-impl-linux.c
glima: Maybe coccinelle again? :P

src/modules/flow/grove/grove.c
glima: Ack.

PR #899, opened 2015-10-15
generator: Fix memory leak

ceolin:


ceolin:
@cabelitos the others return don't have problem because it's already keeping the reference to the allocated memory.

ceolin:
merged, thanks !

PR #900, opened 2015-10-15
JSON unescape of unicode chars to utf-8 was incorrect

otaviobp:
The implementation of sol_json_token_get_unescaped_string was converting
unicode number to utf-8 in an incorrect way. The conversion is a bit
more complex than what was implemented.

This patch fixes this issue and also adds a test to check unicode
conversions

### DIFF
--- a/src/lib/parsers/sol-json.c
+++ b/src/lib/parsers/sol-json.c
@@ -690,6 +690,24 @@ sol_json_serialize_boolean(struct sol_buffer *buffer, bool val)
     return 0;
 }
 
+static int
+_hex_decode(const char *p, uint8_t *val)
+{
+    int16_t n;
+    char hexdigits[3], *endptr;
+
+    hexdigits[2] = '\0';
+    memcpy(hexdigits, p, 2);
+    errno = 0;
+    n = strtoul(hexdigits, &endptr, 16);
+    SOL_INT_CHECK(errno, > 0, -errno);
+    if (n < 0 || n > UINT8_MAX || endptr != hexdigits + 2)
+        return -EINVAL;
+
+    *val = n;
+    return 0;
+}
+
 SOL_API int
 sol_json_token_get_unescaped_string(const struct sol_json_token *token, struct sol_buffer *buffer)
 {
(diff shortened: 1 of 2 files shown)


src/lib/parsers/sol-json.c
```diff
@@ -690,6 +690,24 @@ sol_json_serialize_boolean(struct sol_buffer *buffer, bool val)
...
+static int
+_hex_decode(const char *p, uint8_t *val)
+{
+    int16_t n;
+    char hexdigits[3], *endptr;
+
+    hexdigits[2] = '\0';
+    memcpy(hexdigits, p, 2);
+    errno = 0;
+    n = strtoul(hexdigits, &endptr, 16);
```
lpereira: `strtoul()` returns an `unsigned long`, and the return value is being truncated to an `int16_t`.

lpereira: Maybe this is faster/clearer?

      bool
      parse_hex_digit(uint8_t digit, uint8_t *nibble)
      {
        if (digit >= '0' && digit <= '9') {
          *nibble = digit - '0';
          return true;
        }
        if (digit >= 'a' && digit <= 'f') {
          *nibble = digit - 'a';
          return true;
        }
        if (digit >= 'A' && digit <= 'F') {
          *nibble = digit - 'A';
          return true;
        }

        return false;
      }
   
      int
      parse_hex(char *in, uint8_t *out)
      {
        uint8_t nibble;

        if (!parse_hex_digit(in[0], &nibble))
          return -EINVAL;

        *out = nibble;

        if (!parse_hex_digit(in[1], &nibble))
          return -EINVAL;

        *out = (*out << 4) | nibble;

        return 0;
      }

otaviobp: not a bad idea

otaviobp: And strtoul returns negative numbers casted as unsined int. I am getting it as a signed int to check for negative values. But anyway I prefer your other solution

lpereira: The thing is: they're being truncated to a 16-bit int from a 64-bit unsigned int. That's not going to work well. :)

PR #902, opened 2015-10-15
Async

edersondisouza:
Tests were also added, memap only now - although `make check-valgrind` failed here in a way I couldn't understand, no leaks, but an assert failure o.O.
@barbieri there's no `delayed`. All operations are async, so I couldn't find use for it. Reasoning is that we will only have async option when writing via i2c, so I kept things uniform. Comments/objections?

### DIFF
--- a/src/lib/io/include/sol-efivarfs-storage.h
+++ b/src/lib/io/include/sol-efivarfs-storage.h
@@ -36,114 +36,161 @@
 #include <stdint.h>
 
 #include "sol-buffer.h"
+#include "sol-log.h"
 #include "sol-types.h"
 
 #ifdef __cplusplus
 extern "C" {
 #endif
 
-int sol_efivars_write_raw(const char *name, const struct sol_buffer *buffer);
+int sol_efivars_write_raw(const char *name, struct sol_blob *blob,
+    void (*cb)(void *data, const char *name, struct sol_blob *blob, int status),
+    const void *data);
 int sol_efivars_read_raw(const char *name, struct sol_buffer *buffer);
 
-#define CREATE_BUFFER(_val, _empty) \
+#define CREATE_BUFFER(_val) \
     struct sol_buffer buf = SOL_BUFFER_INIT_FLAGS(_val, \
-    sizeof(*(_val)), SOL_BUFFER_FLAGS_MEMORY_NOT_OWNED); \
-    buf.used = (_empty) ? 0 : sizeof(*(_val));
+    sizeof(*(_val)), SOL_BUFFER_FLAGS_MEMORY_NOT_OWNED | SOL_BUFFER_FLAGS_NO_NUL_BYTE);
+
+#define CREATE_BLOB(_val) \
+    struct sol_blob *blob; \
+    size_t _s = sizeof(*_val); \
+    void *v = malloc(_s); \
+    SOL_NULL_CHECK(v, -EINVAL); \
+    memcpy(v, _val, _s); \
+    blob = sol_blob_new(SOL_BLOB_TYPE_DEFAULT, NULL, v, _s); \
+    SOL_NULL_CHECK(blob, -EINVAL);
 
 static inline int
 sol_efivars_read_uint8(const char *name, uint8_t *value)
 {
-    CREATE_BUFFER(value, true);
+    CREATE_BUFFER(value);
 
     return sol_efivars_read_raw(name, &buf);
 }
 
 static inline int
-sol_efivars_write_uint8(const char *name, uint8_t value)
+sol_efivars_write_uint8(const char *name, uint8_t value,
+    void (*cb)(void *data, const char *name, struct sol_blob *blob, int status),
+    const void *data)
 {
-    CREATE_BUFFER(&value, false);
+    int r;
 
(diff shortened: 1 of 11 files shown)


src/lib/io/include/sol-efivarfs-storage.h
```diff
@@ -36,114 +36,161 @@
...
+    sizeof(*(_val)), SOL_BUFFER_FLAGS_MEMORY_NOT_OWNED | SOL_BUFFER_FLAGS_NO_NUL_BYTE);
+
+#define CREATE_BLOB(_val) \
+    struct sol_blob *blob; \
+    size_t _s = sizeof(*_val); \
+    void *v = malloc(_s); \
+    SOL_NULL_CHECK(v, -EINVAL); \
+    memcpy(v, _val, _s); \
+    blob = sol_blob_new(SOL_BLOB_TYPE_DEFAULT, NULL, v, _s); \
+    SOL_NULL_CHECK(blob, -EINVAL);
```
lpereira: `v` leaks

src/lib/io/include/sol-efivarfs-storage.h
```diff
@@ -36,114 +36,161 @@
...
-    sizeof(*(_val)), SOL_BUFFER_FLAGS_MEMORY_NOT_OWNED); \
-    buf.used = (_empty) ? 0 : sizeof(*(_val));
+    sizeof(*(_val)), SOL_BUFFER_FLAGS_MEMORY_NOT_OWNED | SOL_BUFFER_FLAGS_NO_NUL_BYTE);
+
+#define CREATE_BLOB(_val) \
+    struct sol_blob *blob; \
+    size_t _s = sizeof(*_val); \
+    void *v = malloc(_s); \
+    SOL_NULL_CHECK(v, -EINVAL); \
+    memcpy(v, _val, _s); \
```
lpereira: Use `sol_util_memdup()`

src/lib/io/include/sol-efivarfs-storage.h
edersondisouza: I don't think so. `sol_blob_new` takes care of it, when using `SOL_BLOB_TYPE_DEFAULT`.

src/lib/io/include/sol-efivarfs-storage.h
edersondisouza: Good to know.

src/lib/io/include/sol-efivarfs-storage.h
lpereira: Even if `sol_blob_new()` fails?

edersondisouza: Ah, you mean if `sol_blob_new` fails. True that.

ibriano: On Thu, 15 Oct 2015 13:41:18 -0700, Ederson de Souza wrote:
> >      struct sol_buffer buf = SOL_BUFFER_INIT_FLAGS(_val, \
> > -    sizeof(*(_val)), SOL_BUFFER_FLAGS_MEMORY_NOT_OWNED); \
> > -    buf.used = (_empty) ? 0 : sizeof(*(_val));
> > +    sizeof(*(_val)), SOL_BUFFER_FLAGS_MEMORY_NOT_OWNED | SOL_BUFFER_FLAGS_NO_NUL_BYTE);
> > +
> > +#define CREATE_BLOB(_val) \
> > +    struct sol_blob *blob; \
> > +    size_t _s = sizeof(*_val); \
> > +    void *v = malloc(_s); \
> > +    SOL_NULL_CHECK(v, -EINVAL); \
> > +    memcpy(v, _val, _s); \
> > +    blob = sol_blob_new(SOL_BLOB_TYPE_DEFAULT, NULL, v, _s); \
> > +    SOL_NULL_CHECK(blob, -EINVAL);
> 
> I don't think so. `sol_blob_new` takes care of it, when using `SOL_BLOB_TYPE_DEFAULT`.
> 

But if sol_blob_new() fails, it leaks.

barbieri:
Didn't read the code yet, but regarding delayed its just a way to when to schedule the write. Doesn't have to do with the i2c working. For instance, when delayed you could add a 50ms timer, then you queue all write requests and only dcrcutr them once the timer expires, the actual writes may be in parallel or in sequence

edersondisouza:
@barbieri Could that be only a global option? Like, everything is async, with timer(0), and value of timer defined globally, using env var for instance? Reason: not sure if mixing writing to be done 'asap' and delayed ones is a good thing. Besides one more parameter in API, confusion may arise to one writing `X` to property `a` on t0->t50, then 'Y' on t20->t20 (thus discarding first), then 'Z' on t30->t50 (which would replace second).

edersondisouza:
v2 on #919

barbieri:
I didn't think about that confusion, but you're right.

OTOH there are different EEPROM or memory in the system, making it global shouldn't work nice. 

Can we get a middle ground? Like per device? Devices (by path) could define their best initial value and one could query/set programmatically using C API (hidden at node type level)

edersondisouza:
Sounds like a plan. I'll give a try on v3.

PR #903, opened 2015-10-16
JSON unescape of unicode chars to utf-8 was incorrect - v2

otaviobp:
Replaces #900

Changelog from v1:
 - using method proposed by acidx to parse hex digits


barbieri:
See sol-util.h base16 stuff. Let's keep it uniform and a single implementation

If this allows both upper and lower case, we may change that function from a Boolean to an enum with 3 states, adding "auto" (or "both")

PR #904, opened 2015-10-16
Builders [v5]

cmarcelo:
Differences from v4:

- On the server side, added check for valid names before
  compiling files, invalid names lead to failure. On the client side,
  added a sanitize step to the filename used to store the output.

- Added some context to the third patch (sol tool) commit message.

- Improved error functions in the server, now for internal
  errors, the contents is sent to the user. At least for
  development this is helpful.

- Some sol TODO entries were edited.


cmarcelo:
Replaces #873.

cmarcelo:
Replaced by #908.

PR #906, opened 2015-10-16
Make composed types API similar to API for other packet types

otaviobp:
API for sending or receiving a composed type was different than other
packet types. User needed to allocate memory manually to get composed
members data and do some checks that are not necessary for other
packets. This patch tries to make a bit easier to work with packet
types.

Receiving packets changed from:

r = sol_flow_packet_get_composed_members_len(
        sol_flow_packet_get_type(packet), &len);
SOL_INT_CHECK(r, < 0, r);
packets = malloc(len * sizeof(struct sol_flow_packet *));
SOL_NULL_CHECK(packets, -ENOMEM);
r = sol_flow_packet_get(packet, &packets);
...
err_exit:
free(packets);
return r;

to:

r = sol_flow_packet_get_composed_members(packet, &children, &len);
SOL_INT_CHECK(r, < 0, r);

Sending packets changed from:

struct sol_flow_packet *composed = sol_flow_packet_new(type, children);
SOL_NULL_CHECK(composed, -ENOMEM);
return sol_flow_send_packet(node, 0, composed);

to:

return sol_flow_send_composed_packet(node, 0, type, children);

### DIFF
--- a/data/scripts/sol-flow-node-type-stub-gen.py
+++ b/data/scripts/sol-flow-node-type-stub-gen.py
@@ -337,7 +337,7 @@ send_%(name)s_packet(struct sol_flow_node *src, uint16_t src_port
 static int
 send_%s_packet(struct sol_flow_node *src, uint16_t src_port, %s)
 {
-   struct sol_flow_packet **children, *composed_packet;
+   struct sol_flow_packet **children;
    const struct sol_flow_packet_type *p_type;
    uint16_t len, i;
    int r;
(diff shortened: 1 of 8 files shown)


otaviobp:
@cabelitos take a look

barbieri:
Why not a sol-vector?

otaviobp:
It was already a regular array when I touched the code. If there is no issue in changing it to sol_vector, I will

otaviobp:
Why not sol-vector: As this array has a fixed size it won't get bigger or smaller. Using a sol-vector would allocate unnecessary extra space.

glima:
This belongs in another PR now, no?

otaviobp:
I put this is a separate pr because I don't want the discussions about json create nodes to block this pr.

barbieri:
I wondered this would be an answer. Then I thought if we should have a way to flag this somehow. Or use another struct to do so.

My point is that it will common to have this pattern and if we should try to cover it or not.

So far leave it like you proposed.

PR #907, opened 2015-10-16
fbp-generator: Don't include memmap-storage header if not needed

ibriano:
It was being included every time Soletta has support for
memmap-storage, regardless of whether the FBP uses it or not, which
makes generating programs for cross-building problematic.

Also, include errno.h always, as the composed node types generate code
using ENOMEM. Previously it was working because memmap-storage includes
sol-buffer.h, which includes errno.h.


cmarcelo:
+1

PR #908, opened 2015-10-16
Builders [v6]

cmarcelo:
Differences from v5:

- Added notes indicating the code here is experimental. The goal of
  it was to figure out the workflow we wanted. If/Once this is
  settled, then worth focusing on hardening it (i.e. how can we
  communicate using crypto, and so on).

- Cleaned environment variables before calling compile in the
  server. Starting with PATH, HOME and LANG=C. Should be tweaked
  later as necessary.

- New commit: try to fail early when running as server from wrong
  location.

- New commit: Improve test-builders.sh to actually use `sol` to
  build things.

- `sol` returns 1 when the build fails, and don't always print
  the list of available targets (unnecessary).

- Replace `-connect` options with an `-addr` option, that's used
  both for connecting to and serving.

- Added explanation of why platform-test builds always fail: they
  are used to show the environment and debug the generic part of
  compile scripts (generate C from FBP, pick the right files, etc.).

- Added builders/.gitignore.


cmarcelo:
Replaces #904.

cmarcelo:
@lpereira @ibriano could you take another look?

ibriano:
Seems fine to me. Tested here and works as expected.

builders/sol/util.go
```diff
@@ -0,0 +1,47 @@
...
+ * THEORY OF LIABILITY, WHETHER IN CONTRACT, STRICT LIABILITY, OR TORT
+ * (INCLUDING NEGLIGENCE OR OTHERWISE) ARISING IN ANY WAY OUT OF THE USE
+ * OF THIS SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
+ */
+
+package main
+
+import "regexp"
+
+var validNamePattern = "0-9A-Za-z-_\\.\\/"
```
lpereira: Will this pattern allow `../../../../../`? Then it might be a problem.

lpereira:
LGTM since it's clearly marked as experimental, although I haven't tried it.

bdilly:
@cmarcelo it seems to be ok, but listing dependencies to make it work may save people some time.
For RIOT I had to install arm-none-eabi-gcc-cs.x86_64 arm-none-eabi-binutils-cs.x86_64 arm-none-eabi-newlib.noarch (after installing 32 bits packages yesterday).

Regarding OS detection, I've talked to @ibriano and we're going to change dependency resolver to abort things if no OS is detected, avoid this issue of not having dependencies to build it for any platform and continue.

cmarcelo:
> but listing dependencies to make it work may save people some time.

One idea is to let the `prepare` for each platform should do verification if all it needs is there.

I'll push a v7 now.

cmarcelo:
Replaced by #929.

PR #912, opened 2015-10-19
JSON unescape of unicode chars to utf-8 was incorrect - v3

otaviobp:
Using sol_util_base16_decode to parse unicode codes when unescaping


otaviobp:
Replaces #903

barbieri:
Well, I'd not do the toupper(), rather change the decode to accept an enum. This way we cover this case of unknown more easily withou writing to memory. For speed sake maybe we keep another implementation for that less common special case.

Also, it seems that the conversion of hex to Unicode may be split into a sol-util function. Eventually other systems need that, such as some rich text or xml. Analougly add the encode version so we can get the hex numbers. Don't add prefixes as those change from one system to another

PR #915, opened 2015-10-19
parser: fix parsing of large integers on i686 systems.

glima:
Our integer packets are set in int32_t for size, but parsing them
involved an intermediary strtol(), which could overflow on our inputs on
32 bit systems. By moving the strtoll() we stay safe on both 64/32-bit
systems.


glima:
By printing the returned values with the target cast done exactly as in this code, we checked that the values are being converted right we solely this code. All other attempts at checking the validity were tricky and spoiling valid results.

Any other argument on not keeping this (direct) approach?

glima:
Merging.

PR #916, opened 2015-10-19
Linux micro kmod v3

dorileo:
## Changes
  + v2 (since #844)
    - impemented coldplug;
    - implemented hotplug;
    - introduced patches:

  + v3 (since #877)

## Rationale

via uevent netlink channel.


src/lib/datatypes/include/sol-str-slice.h
```diff
@@ -123,6 +124,34 @@ sol_str_slice_to_string(const struct sol_str_slice slice)
...
+    size_t len = slice.len;
+
+    while (ptr > (slice.data + slice.len - 1) && isspace(*ptr)) {
+        ptr--;
+        len--;
+    }
+
+    return SOL_STR_SLICE_STR(ptr, len);
+}
+
```
lpereira: Maybe create a `sol_str_slice_strip(slice)`?

    struct sol_str_slice
    sol_str_slice_strip(struct sol_str_slice slice)
    {
        return sol_str_slice_remove_leading_whitespace(sol_str_slice_remove_trailing_whitespace(slice));
    }

src/lib/datatypes/include/sol-str-slice.h
```diff
@@ -123,6 +124,34 @@ sol_str_slice_to_string(const struct sol_str_slice slice)
     return strndup(slice.data, slice.len);
 }
 
+static inline struct sol_str_slice
+sol_str_slice_remove_leading_whitespace(struct sol_str_slice slice)
+{
+    char *data = (char *)slice.data;
```
lpereira: You can just do `struct sol_str_slice copy = slice;`, and then operate directly on `copy`.

src/lib/datatypes/include/sol-str-slice.h
```diff
@@ -123,6 +124,34 @@ sol_str_slice_to_string(const struct sol_str_slice slice)
...
+    size_t len = slice.len;
+
+    while (isspace(*data)) {
```
lpereira: You must check for `len` here, in case the slice isn't zero-terminated.

src/lib/datatypes/include/sol-str-slice.h
```diff
@@ -123,6 +124,34 @@ sol_str_slice_to_string(const struct sol_str_slice slice)
...
+    return SOL_STR_SLICE_STR(data, len);
+}
+
+static inline struct sol_str_slice
+sol_str_slice_remove_trailing_whitespace(struct sol_str_slice slice)
+{
+    char *ptr = (char *)slice.data + slice.len;
+    size_t len = slice.len;
+
+    while (ptr > (slice.data + slice.len - 1) && isspace(*ptr)) {
```
lpereira: You can simplify this: `while (copy.data >= slice.data && isspace(*copy.data)) { ... }`.

src/modules/linux-micro/kmod/kmod.c
```diff
@@ -0,0 +1,475 @@
...
+
+static int
+kmod_apply_value(struct kmod_ctx *kmod, struct sol_str_slice modalias)
+{
+    char *alias;
+    const int probe_flags = KMOD_PROBE_APPLY_BLACKLIST;
+    struct kmod_list *itr, *modlist = NULL;
+    int r = 0;
+
+    alias = strndupa(modalias.data, modalias.len);
```
lpereira: Maybe `SOL_INT_CHECK(modalias.len, < 512, -ENOMEM);`, to avoid stack overflows with malformed files?

src/modules/linux-micro/kmod/kmod.c
```diff
@@ -0,0 +1,475 @@
...
+        if (fstatat(dfd, ent.d_name, &st, AT_SYMLINK_NOFOLLOW) == -1) {
+            SOL_ERR("Could not stat %s", ent.d_name);
+            goto next;
+        }
+
+        if (S_ISDIR(st.st_mode)) {
+            int nfd;
+            DIR *ndir;
+
+            nfd = openat(dfd, ent.d_name, O_RDONLY | O_CLOEXEC);
```
lpereira: Maybe also `O_PATH`, to ensure `dfd` isn't used for anything else than path-related operations?

src/modules/linux-micro/kmod/kmod.c
```diff
@@ -0,0 +1,475 @@
...
+            int fd;
+            char buff[PATH_MAX];
+
+            fd = openat(dfd, ent.d_name, O_RDONLY | O_CLOEXEC);
+            if (fd < 0) {
+                SOL_ERR("Could not obtain uevent file descriptor");
+                goto next;
+            }
+
+            if (read(fd, buff, sizeof(buff)) > 0) {
```
lpereira: Why not use `sol_file_reader` here, that will handle `EINTR`, `EAGAIN` and whatnot?

src/modules/linux-micro/kmod/kmod.c
```diff
@@ -0,0 +1,475 @@
...
+}
+
+static int
+kmod_start_coldplug(struct kmod_data *kdata)
+{
+    DIR *dir;
+
+    dir = opendir("/sys/devices");
+    if (!dir) {
+        SOL_ERR("Could not run dir");
```
lpereira: Maybe "Could not open /sys/devices directory: coldplug unavailable"?

src/modules/linux-micro/kmod/kmod.c
```diff
@@ -0,0 +1,475 @@
...
+}
+
+static int
+kmod_conf_cmp(const void *data1, const void *data2)
+{
+    return strcmp((const char *)data1, (const char *)data2);
+}
+
+static void
+read_directory_to_vector(DIR *dir, struct sol_ptr_vector *vec)
```
lpereira: Not for this patch, but maybe create an utility function to do this, filtering by file extension? Or, better yet, a function that walks a directory path and calls a callback function for each name a `fnmatch()` pattern match? This would simplify the `kmod`, `sysctl`, and possibly other modules as well.

src/modules/linux-micro/kmod/kmod.c
```diff
@@ -0,0 +1,475 @@
...
+            ndir = fdopendir(nfd);
+            if (!ndir) {
+                SOL_ERR("Could not open sys directory %s", ent.d_name);
+                close(nfd);
+                goto next;
+            }
+
+            kmod_coldplug_find_devices(kdata, ndir);
+            closedir(ndir);
+        } else if (!strcmp(ent.d_name, "modalias")) {
```
lpereira: `streq()`?

src/modules/linux-micro/kmod/kmod.c
dorileo: I've considered this, for now we would have at least 2 users for that... I'll handle the other linux-micro modules I still have to write and come back to this.

src/modules/linux-micro/kmod/kmod.c
dorileo: ok

src/lib/datatypes/include/sol-str-slice.h
dorileo: yep, maybe, but I think it's not high priority and I'll postpone this to the end of my todo list (namely my linux-micro modules to be written).

src/lib/datatypes/include/sol-str-slice.h
dorileo: nice the copy function wasn't on the top of my mind while writing this. thanks.

src/lib/datatypes/include/sol-str-slice.h
dorileo: yep, true.

src/lib/datatypes/include/sol-str-slice.h
dorileo: yep, the copy would indeed simplify things out.

src/modules/linux-micro/kmod/kmod.c
dorileo: yep, true, the strings from uevent are all null terminated, if kmod handled that we would not even need to dupa anything here - I'll check that out, and maybe get rid of it.

src/modules/linux-micro/kmod/kmod.c
dorileo: true.

src/modules/linux-micro/kmod/kmod.c
dorileo: just a matter of laziness. I'll save some time to possibly change that. thanks.

src/modules/linux-micro/kmod/kmod.c
dorileo: nicer, thanks.

src/lib/datatypes/include/sol-str-slice.h
dorileo: ok, I first misunderstood the copy thing, dismiss my comments on that regard.

dorileo:
v4 coming

PR #917, opened 2015-10-20
Add nodes to extract data from JSON using JSON path - v2

otaviobp:
Changelog from v1:
 - refactor in parser to make it human readable

### DIFF
--- a/src/modules/flow/converter/converter.c
+++ b/src/modules/flow/converter/converter.c
@@ -2127,13 +2127,27 @@ timestamp_error:
     return 0;
 }
 
-static bool
+static struct sol_blob *
 json_validate(struct sol_blob *blob, enum sol_json_type type)
 {
     struct sol_json_scanner scanner;
+    char *p;
+    size_t size;
+
+    p = (char *)blob->mem + blob->size - 1;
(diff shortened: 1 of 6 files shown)


otaviobp:
Replaces #897

bdilly:
I've got a design question. Why are we adopting packet types like "json-array" and "json-object", and consequently node types that handle only with arrays or object (like get_length() or get_path()) instead of having a unique packet type and nodes that handle both kinds?

barbieri:
@bdilly the packet types exist (instead of a generic  "array" or "object") because they are just slices of another one. Making them pretty cheap as we don't need to decode/parse all elements if they are not used.

Last, we could emulate the behavior you want with 2input ports for get path or get length, it may be confusing but easy to document. 

That said, the two types and nodes exist because they are different in properties and access model. We're providing "length" for object just to make it complete, but that's not to be widely used, instead people should use the get keys. Then IMO it's better to keep different and not try to overload a single type

Depending on the usages we may even add more node types that will take the blob and actually parse it all, making recurrent access faster (better if keys are used more than once)

bdilly:
@otaviobp I've fixed this "file support" issue already (commit as merged).
So no need to change it to use strings (just remove the skip).

PR #918, opened 2015-10-20
Add nodes to create JSON objects and JSON arrays - v2

otaviobp:
Use str tables when possible and simplifying json_array_*_process functions

Replaces #909

### DIFF
--- a/data/schemas/node-type-genspec.schema
+++ b/data/schemas/node-type-genspec.schema
@@ -56,7 +56,7 @@
         },
         {
           "type": "string",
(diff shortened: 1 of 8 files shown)


cabelitos:
+1 for the composed types changes.

otaviobp:
Packet is sent only when create is received. But it is a good idea to state that.
I'll s/flush/clear

bdilly:
@otaviobp just declaring empty constant after the others should be enough

Also, I've pushed a PR #930 to make conversions like str -> blob -> json a bit more direct

PR #919, opened 2015-10-20
Async v2

edersondisouza:
v2:
Fixed pointed issues;
Added a commit that has a `delayed` option, so some requests can be delayed by up to SOL_PERSISTENCE_WRITE_TIMEOUT envvar. It's separated from main commit of async, as I'm not sure about this approach. If it goes in, it should (probably) be squashed onto that commit - or it maybe dropped altogether.


edersondisouza:
Check-sytle issues fixed locally.

edersondisouza:
v3 on #934

PR #920, opened 2015-10-20
Get rid of legacy tool sol-flow-node-types.

glima:
We don't use that anymore in the build system.


anselmolsm:
:fire:

glima:
I'll interpret that as a +2. Merging.

PR #922, opened 2015-10-20
run-fbp-tests: use same path for fbps on runner and compiled tests

bdilly:
Otherwise tests trying to read files will work on runner but
fail in compiled form.

Required to make test on pr #917 work without skipping


ibriano:
Looks ok to me

bdilly:
merged

PR #924, opened 2015-10-20
Add RIOT back-end for message-digest and default to OpenSSL on Linux

ibriano:


barbieri:
Looks good, but are you sure if final is not called there is no need to cleanup? If so we may need to add a placeholder call to final on cleanup to avoid leaks. 

That reminds me we should add a test to call del without reading the digest. Could you do it?

ibriano:
On Wed, 21 Oct 2015 06:15:54 -0700, Gustavo Sverzut Barbieri wrote:
> Looks good, but are you sure if final is not called there is no need to cleanup? If so we may need to add a placeholder call to final on cleanup to avoid leaks. 
> 
> That reminds me we should add a test to call del without reading the digest. Could you do it?

Right, forgot to check that. I'll add the test too.

PR #925, opened 2015-10-21
Change string/concatenate to support mulitple inputs

otaviobp:
String concatenate was only concatenating 2 strings. Now we are able to
concatenate up to 32 strings using a single node. The approach taken is
the same approach from int opeartions that have multiple inputs

### DIFF
--- a/src/modules/flow/string/string-ascii.c
+++ b/src/modules/flow/string/string-ascii.c
@@ -51,8 +51,11 @@ struct string_data {
 };
 
 struct string_concatenate_data {
-    struct string_data base;
+    uint32_t n;
+    char *string[32];
     char *separator;
+    uint32_t var_initialized;
+    uint32_t var_connected;
 };
 
 struct string_compare_data {
@@ -73,15 +76,17 @@ static void
 string_concatenate_close(struct sol_flow_node *node, void *data)
 {
     struct string_concatenate_data *mdata = data;
+    uint8_t i;
 
-    string_close(node, data);
+    for (i = 0; i < 32; i++)
+        free(mdata->string[i]);
     free(mdata->separator);
 }
 
-static bool
+static int
 get_string_by_port(const struct sol_flow_packet *packet,
     uint16_t port,
-    struct string_data *mdata)
+    char **string)
 {
     const char *in_value;
     int r;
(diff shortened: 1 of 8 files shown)


src/modules/flow/string/string-ascii.c
```diff
@@ -146,27 +181,45 @@ string_concat(struct sol_flow_node *node,
...
 
-    dest = strcpy(dest, mdata->base.string[0]);
+    for (i = 0; i < 32; i++) {
+        if (!(mdata->var_initialized & (1u << i)))
+            continue;
 
-    if (mdata->separator)
-        dest = strcat(dest, mdata->separator);
+        if (count && mdata->separator) {
+            dest = strncat(dest, mdata->separator, len);
```
lpereira: Why not use `sol_buffer`?

otaviobp: I was planning on using buffer, but then I just followed the approach that was being used before. But I think that it really better use buffer for that.

otaviobp:
lol. the idea in the title of the commit message is the oposite. Correct:

strings: Rename bytes option to chars in concatenate and compare

Unicode strings are counting the number of characters and not the number
of bytes.

glima:
``FAILED running string-concat.fbp: wrong exit code, got 1 but expected 0
Output:
==14433== WARNING: AddressSanitizer failed to allocate 0xc00800018078 bytes``

check the bot's output. Enable the sanitizers on your machine.

bdilly:
@otaviobp wasn't the conclusion of our discussion on IRC yesterday with @glima to completely remove this counting option?

So we would get rid of this logic inside this node. It's specially true imo if you're going to fix the size only of the final string. It can be easily achieved using string/slice later, right? Something like:

concat(string/concat) OUT -> IN _(string/slice:end=CHARS) OUT -> IN _(console)

instead of

concat(string/concat:chars=CHARS) OUT -> IN _(console)

@glima, what do you think about it?

otaviobp:
For me it is ok to remove it. I had understood that I was suppose to keep it, but limiting the size of the full string.

glima:
I was ok with the semantics changed to the full size, but removing it altogether is cool as well. I vote for removing code :)

PR #926, opened 2015-10-21
flow: only consider named descriptions

bdilly:
It was a condition to stop but when checking if it failed
to find descriptions it was checking only for a not null
mdesc pointer.


bdilly:
ty, merged

PR #927, opened 2015-10-21
fbp meta types on conf files

bdilly:
This is more like a WIP since it's leaking parsers
and I intent to re-write it following an approach
discussed with @ibriano yesterday.

Parsers would have a way to call back its users
to handle with unknown types (or metatypes).

So runner would handle node types declared on conf files
just like it would with ones declared on fbp files.

I'm pushing it here to see if anybody else would like
to comment on that before doing such change.

====

flow: support fbp meta types on conffiles

It support to declare them on conffiles like:

{
   "name": "NodeName",
   "options": {
       ...
   },
   "type": "fbp:FILE_NAME.fbp"
}

Support was added only on runner (no generator support yet).
It leaks parsers (and consequently node types).

Paths used to look for fbp files are passed by runner.

### DIFF
--- a/src/bin/sol-fbp-runner/runner.c
+++ b/src/bin/sol-fbp-runner/runner.c
@@ -92,10 +92,6 @@ search_fbp_file(struct runner *r, const char *basename)
     char *fullpath;
     uint16_t i;
 
-    fullpath = stat_fullpath(r->dir, basename);
-    if (fullpath)
-        return fullpath;
-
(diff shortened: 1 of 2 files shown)


barbieri:
Just don't abuse "type" and then an encoded string. JSON is structured, so add a new key with an object instructing the mets type and all

cabelitos:
Should we support composed as well?

barbieri:
Composed is a meta, so we're good if we support them

PR #928, opened 2015-10-21
Change string/concatenate to support mulitple inputs - v2

otaviobp:
Changelog:
 - using sol_buffer
 - chars option removed from string/concatenate
 - fix sanity check reported issue


bdilly:
Wouldn't we remove "chars / bytes" from concatenate?

otaviobp:
I removed chars options from concatenate. I just kept the commit renaming them because this rename is necessary for compare. We are not counting bytes in compare. We are counting chars.

otaviobp:
merged

PR #929, opened 2015-10-21
Builders [v7]

cmarcelo:
Differences from v6:

- Improve verification of the valid filenames for the files received from the server: only certain characters are accepted, and the path must not "leave" the directory used for the base
path.

- Add error message for the case input contains unsupported filetypes (e.g. links). our client doesn't support it, but when it does, we will now where to handle in the server.

- `test-builders.sh`: add function die() to do echo and leaving, this avoids missing an exit 1 when writing an error message.

- `test-builders.sh`: run 'go test' for the sol tool tests. The exisiting test is small but serves as example.


cmarcelo:
Replaces #908.

cmarcelo:
@bdilly @lpereira could you take another look?

cmarcelo:
Applied in 45b2520c2637eb914b6332d6fe4872ccc38c1383 and parents.

PR #930, opened 2015-10-21
modules/converter: add string-to-json-[array|object]

bdilly:
It seems to be useful since all our tests need to
convert from string to blob and then blob to json.

I've let some usages of string-to-blob -> blob-to-json-*
just to make sure this converters are tested as well.

@otaviobp could you please take a look on this PR?


otaviobp:
other than that, +1

bdilly:
@otaviobp I just used a null terminator because all your tests were using it (it's default on conversions between strings and blobs). I'll test them without it and if it works, it's even better, so I can remove the open() method.

bdilly:
@otaviobp it worked fine without null terminators.
merged, thanks

Issue #748, opened 2015-09-17
HTTP Client generic node types


barbieri:
We should provide some generic node types to deal with HTTP/REST APIs that we lack native/optimized components. Ideally we should be as simple as `wget | sed` to solve simple problems, linking with `string/regexp-search`, `string/regexp-replace` and the node types from #746.

Suggested node types:
 * `http/get-string`: does a `GET` and outputs a string packet.
    * Options:
      * `url` (string): the URL to execute `GET`.
      * `strict` (boolean, default=false): validate `Content-type` is same as requested.
      * `content_type` (string, default="text/plain"): provided as `Accept: ` and then validated with `Content-type:` HTTP headers.
    * Input Ports:
      * `URL` (string, optional): allows changing the `url` given as option.
      * `TRIGGER` (any): execute the HTTP method.
    * Output Ports:
      * `OUT` (string): the string returned as payload of the HTTP request.
 * `http/get-json`: does a `GET` and outputs a json object packet.
    * Options:
      * `url` (string): the URL to execute `GET`.
      * `strict` (boolean, default=false): validate `Content-type` is same as requested.
      * `content_type` (string, default="application/json"): provided as `Accept: ` and then validated with `Content-type:` HTTP headers.
    * Input Ports:
      * `URL` (string, optional): allows changing the `url` given as option.
      * `TRIGGER` (any): execute the HTTP method.
    * Output Ports:
      * `OUT` (json object): the json object returned as payload of the HTTP request.
 * `http/get-blob`: does a `GET` and outputs a blob packet.
    * Options:
      * `url` (string): the URL to execute `GET`.
      * `strict` (boolean, default=false): validate `Content-type` is same as requested.
      * `content_type` (string, default=null): if non-null, provided as `Accept: ` and then validated with `Content-type:` HTTP headers.
    * Input Ports:
      * `URL` (string, optional): allows changing the `url` given as option.
      * `TRIGGER` (any): execute the HTTP method.
    * Output Ports:
      * `OUT` (blob): the blob returned as payload of the HTTP request.
 * `http/request`: creates a HTTP request to the server (complex version that covers all cases):
    * Options:
      * `url` (string): the URL to execute method.
      * `method` (string): one of `GET`, `POST`, `HEAD`, `PUT`...
      * `timeout` (int, default=0): the number of milliseconds to wait for this request to be completed.
      * `allow_redir` (boolean, default=true): allows redirections to be followed.
    * Input Ports:
[...]

PR #931, opened 2015-10-21
Add HTTP generic node.

cabelitos:
Please look at https://github.com/solettaproject/soletta/issues/748


barbieri:
This needs a careful review. Of both details and some key aspects.

In the details front, things like keeping http types in the sol-http header, not sol-types. Also using lookup tables for method to enum. Inspector formatting seems wrong and missing a "return" should be printing "<?>"

On the key aspects, "filter" is not good, I'd read it as "filter out", like remove. Use "get-" so we match JSON names. Also, headers (and maybe cookies?) are case Insensitive. And we need to start thinking how to make all of theses types extensible...we need to add too much boilerplate to introduce a new type! Maybe a new JSON in another directory to specify packets?

barbieri:
And talk to @rchiossi, he implemented something like post with data. Maybe change his function to allow your needs?

cabelitos:
@barbieri I agree with you, adding new packet types is always a pain. Maybe a json specification can help us indeed. 

Details: 
* I will talke a look at the inspector formation + return issue.
* I did not understand the part that you said 
* Should we keep the http types in sol-types.h even if sol is compiled without HTTP? I did not move those types there because, soletta can be compiled with support.

Keys -
* Maybe we should stay safe and compare all keys with strcasecmp() then.
* I will talk to @rchiossi about the post with data function

cabelitos:
Closing in favor of #945

PR #932, opened 2015-10-21
Add nodes to extract data from JSON using JSON path - v3

otaviobp:
Chagelog from v2:
  skip removed from code generation 
  INT ports receives truncated float values
  path documented in json.json
  extra_methods used to reduce duplicated methods

Replaces #917


otaviobp:
merged

PR #933, opened 2015-10-21
Changes to modules int and boolean related to multiple ports and not filtering packets with the same value on these node types.

bdilly:
These modifications are consequence of discussions related to
pr #928


otaviobp:
+1

glima:
If the tests *and samples* are OK after that, go for it.

bdilly:
@glima tests are OK for sure. I've tested a few samples using int arithmetic and they looked to be working fine.

bdilly:
merged

PR #935, opened 2015-10-21
Linux micro kmod v4

dorileo:
## Changes
  + v2 (since #844)
    - impemented coldplug;
    - implemented hotplug;
    - introduced patches:

  + v3 (since #877)

  + v4 (since #916)

## Rationale

via uevent netlink channel.


src/lib/datatypes/include/sol-str-slice.h
```diff
@@ -123,6 +124,36 @@ sol_str_slice_to_string(const struct sol_str_slice slice)
...
+static inline struct sol_str_slice
+sol_str_slice_remove_trailing_whitespace(struct sol_str_slice slice)
+{
+    struct sol_str_slice copy = slice;
+    int len = copy.len;
+
+    if (copy.data[len] == '\0')
+        len--;
+
+    while (isspace(copy.data[len])) {
```
lpereira: Check if `len >= 0` before checking for `isspace(...)`.

src/lib/datatypes/include/sol-str-slice.h
```diff
@@ -123,6 +124,36 @@ sol_str_slice_to_string(const struct sol_str_slice slice)
     return strndup(slice.data, slice.len);
 }
 
+static inline struct sol_str_slice
+sol_str_slice_remove_leading_whitespace(struct sol_str_slice slice)
+{
+    struct sol_str_slice copy = slice;
+
+    while (isspace(*copy.data) && copy.len) {
```
lpereira: Check `copy.len` before dereferencing `copy.data`.

src/modules/linux-micro/automount/automount.c
```diff
@@ -0,0 +1,339 @@
...
+
+#include "sol-file-reader.h"
+#include "sol-mainloop.h"
+#include "sol-platform-linux-micro.h"
+#include "sol-uevent.h"
+#include "sol-util.h"
+
+#define EXTMAGIC "\123\357\001"
+#define EXT_SB_OFFSET 1024
+#define EXT_MAGIC_OFFSET EXT_SB_OFFSET + 0x38
```
lpereira: This macro doesn't seem to be used.

src/test/test-str-slice.c
```diff
@@ -108,6 +108,68 @@ test_str_slice_str_eq(void)
...
+
+#define TEST_EQUAL(X) { SOL_STR_SLICE_LITERAL(X), true }
+#define TEST_NOT_EQUAL(X) { SOL_STR_SLICE_LITERAL(X), false }
+
+    static const struct {
+        struct sol_str_slice input;
+        bool equal;
+    } table[] = {
+        TEST_NOT_EQUAL("with one trailing whitespace "),
+        TEST_NOT_EQUAL("with two trailing whitespace  "),
```
lpereira: Might be interesting to test other kinds of whitespace that `isspace()` tests for, such as `\n` or `\t`.

barbieri:
Mark some global structs/table as static.

Use sizeof() to declare magic whenever possible

Use memcmp() to check magic, if one day magic happens to include \0 the current code will fail.

I'm not sure we mount /mnt as tmpfs (see Linux micro main setup), if not add it there or use another mount point base. We shouldn't be touching the actual FS to create the mount point directories.

We need to find a way to request unmount. It shouldn't be Linux-micro specific, but we do need a way to explicitly get mount points and "eject" or unmount them. This way UI (or LED or webUI) can offer a way to save/load some stuff (i.e. Backup) and immediately unmount. Linux micro can hook to that request and implement it.

src/lib/datatypes/include/sol-str-slice.h
```diff
@@ -123,6 +124,36 @@ sol_str_slice_to_string(const struct sol_str_slice slice)
...
+    }
+
+    return copy;
+}
+
+static inline struct sol_str_slice
+sol_str_slice_remove_trailing_whitespace(struct sol_str_slice slice)
+{
+    struct sol_str_slice copy = slice;
+    int len = copy.len;
```
lpereira: Why is `len` an `int`?

src/lib/datatypes/include/sol-str-slice.h
```diff
@@ -123,6 +124,36 @@ sol_str_slice_to_string(const struct sol_str_slice slice)
...
+
+    if (copy.data[len] == '\0')
+        len--;
+
+    while (isspace(copy.data[len])) {
+        copy.len--;
```
lpereira: No need to update `copy.len` inside the loop, update it once right after the loop's over.

src/modules/linux-micro/automount/automount.c
```diff
@@ -0,0 +1,339 @@
...
+
+    sb = (struct ext_super_block *)buff;
+    if (!(sb->s_feature_compat & EXT3_FEATURE_COMPAT_HAS_JOURNAL))
+        return false;
+
+    if ((sb->s_feature_ro_compat & EXT3_FEATURE_RO_COMPAT_UNSUPPORTED) ||
+        (sb->s_feature_incompat & EXT3_FEATURE_INCOMPAT_UNSUPPORTED))
+        return false;
+
+    return false;
```
lpereira: All paths of this function returns `false`. Is it correct?

src/modules/linux-micro/automount/automount.c
dorileo: ok

src/lib/datatypes/include/sol-str-slice.h
dorileo: cause the programmer is too stupid to not use ```size_t``` :) thanks...

src/lib/datatypes/include/sol-str-slice.h
dorileo: actually it must be, ```len``` and ```copy.len``` may differ if dealing with null terminated string... probably we could change the whole approach...

src/modules/linux-micro/automount/automount.c
dorileo: yep, absolutely wrong, the last one should return ```true```.

src/lib/datatypes/include/sol-str-slice.h
dorileo: ok

src/lib/datatypes/include/sol-str-slice.h
dorileo: ouch, sure... thanks.

src/test/test-str-slice.c
lpereira: Also nice to test: declare the string `s = "  foo"`, and remove the leading whitespace from `s + 1`, to ensure that no buffer underflow will happen.

src/lib/datatypes/include/sol-str-slice.h
dorileo: I changed the approach... to it doesn't apply anymore and don't face the address violation we saw with the sanitizer.

dorileo:
v5 coming....

PR #936, opened 2015-10-22
Add nodes to create JSON objects and JSON arrays - v3

otaviobp:
Changelog from v2:
 - Add commit changing type names of packets and using it in the commit that fixes composed bug
 - s/FLUSH/CLEAR port name
 - Improve CREATE port documentation
 - Use single pulse instead of timer in test. Also using new converter/string-to-json-*
 - Check strdup returns
 - Reorganizing functions and using extra_methods to reduce the number of duplicated functions

### DIFF
--- a/data/schemas/node-type-genspec.schema
+++ b/data/schemas/node-type-genspec.schema
@@ -56,7 +56,7 @@
         },
         {
           "type": "string",
-          "pattern": "^composed:([a-zA-Z]*?)((,[a-zA-Z]+)+)$"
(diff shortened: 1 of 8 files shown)


otaviobp:
Replaces #918

edersondisouza:
+1 No outstanding issues for me, but someone else looking over this series would be nice

glima:
Looking ok for me too.

otaviobp:
merged

PR #937, opened 2015-10-22
Async v4

edersondisouza:
v4:

v3:

v2:
Fixed pointed issues;
Added a commit that has a delayed option, so some requests can be delayed by up to SOL_PERSISTENCE_WRITE_TIMEOUT envvar. It's separated from main commit of async, as I'm not sure about this approach. If it goes in, it should (probably) be squashed onto that commit - or it maybe dropped altogether.

### DIFF
--- a/src/bin/sol-fbp-generator/main.c
+++ b/src/bin/sol-fbp-generator/main.c
@@ -1320,9 +1320,10 @@ generate_memory_map_struct(const struct sol_ptr_vector *maps, int *elements)
         out("\nstatic const struct sol_memmap_map _memmap%d = {\n"
             "   .version = %d,\n"
             "   .path = \"%s\",\n"
+            "   .timeout = %u,\n"
             "   .entries = _memmap%d_entries\n"
             "};\n",
-            i, map->version, map->path, i);
+            i, map->version, map->path, map->timeout, i);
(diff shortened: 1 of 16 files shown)


barbieri:
I didn't read the commit details, but please check:

On shutdown (or memap_del), if any pending writes then we should execute them if there was a timeout.

Document that order of writes is guaranteed for a given key, but may be reordered for different keys. I saw that you replace an existing request in place, it does make sense, but may cause issues if people are not aware. Actually we may even reorder the vector based on memory address so we keep it linear/sequential (future work)

Review some allocations. Like the version of the map, you can use a blob with a NOFREE type instead of duplicating the memory.

edersondisouza:
I've documented about the 'replacing' request and how callback will be called.
Fair about the memmap_del.
I'll review doc about order and NOFREE candidates.

edersondisouza:
@barbieri I've failed to understand how a NOFREE blob could be helpful...

barbieri:
You have the map->version alive while the map is used, then you can use NOFREE with &(map->version), sizeof(map->version)

edersondisouza:
Ah, got it.

edersondisouza:
v5 on #966

PR #939, opened 2015-10-22
Persistence test

edersondisouza:
It should help debugging persistence. More information on logs and a way to establish dependency between fbp tests.


otaviobp:
lgtm

barbieri:
+1, however persistence is critical and I think we should add forced tests in plain C to control all failure cases, like exit (sol_quit) before timeouts fire, replace of writes, interleaved writes and reads, particularly reads before write commit to disc.

edersondisouza:
Async PR does have C API testing that can be improved on future.

edersondisouza:
Integrated.

PR #940, opened 2015-10-22
form: new form/boolean node, for bool state selection (plus lcd output)

glima:
Tests are included, naturally. The form/selector node got a missing
test, too -- the input SELECTED port.


glima:
v2 coming.

PR #941, opened 2015-10-22
shared: Don't set buffer flags before NULL-checking it

lpereira:


ricardotk:
:+1:

lpereira:
Merged, thanks!

PR #942, opened 2015-10-22
modules/test: use "int" for time interval option

bdilly:
Instead of "float", since it converts it to int later
to create timers anyway.


cabelitos:
+1

bdilly:
merged

PR #943, opened 2015-10-22
Pin map

mbelluzzo:


mbelluzzo:
Forgot to tell that this is like a beta PR. I will change the nodes to actually use this, so the idea is to use it through the nodes and not directly.

src/lib/common/include/sol-pin-mux-modules.h
```diff
@@ -32,6 +32,11 @@
```
barbieri: Do not increase API version until we release a new version. During development numbered should be the same

mbelluzzo: But that is intended for the next release, or we should bump it in a separated commit? I think this way will be very easy to forget to bump it.

barbieri:
Minor, but the log initializer defines a local context static Boolean. Then it's better to use a function log_init() to do it instead of calling the macro multiple times. It will reduce number of static variables.

About the name, I like more "_open_by_label()" instead of "label_open()"

mbelluzzo:
I forgot we dont use yoda style <what a shame> ... by_label it is

mbelluzzo:
V2: #958

PR #945, opened 2015-10-23
Add HTTP generic node v2.

cabelitos:
Changes since v1.

* Rename filter nodes to get; Example: filter-response-code is now get-response-code.
* Fixed how cookies are parsed.
* Cookies and Headers are now compared in a case insensitive way.
* Removed sol_http_client_request_with_data() function; It now uses the HTTP_POST_DATA param.
* The HTTP Packet data is now private, meaning that there is a sol_type assosiated to it.
* Various suggestions made by @edersondisouza and @ceolin


cabelitos:
@bdilly That's a great Idea I will work on it.

cabelitos:
Merged. I will add the proposed @bdilly example as a separated commit.

PR #946, opened 2015-10-23
Fix warning when building message-digest sample

otaviobp:
Printf format for signed value was used with an unsigned variable


otaviobp:
merged

PR #947, opened 2015-10-23
Http fixes

ceolin:
- Fix wrong vector usage on server (it was crashing when receiving multiple requests)
 - Notify curl wherever their file descriptors have something to do
 - Add function to escape strings (url encoding)
 - Add common status code related with redirect
 - Handle cookies, postdata and get arguments on server


cabelitos:
Btw, It would be nice to merge this commits ASAP. I have a url contructor node that can make good use of the escape string function.

cabelitos:
lgtm

ceolin:
merged with the suggestion

Issue #724, opened 2015-09-15
review node type options to use basic types


barbieri:
Currently we map "int" to "irange", "float" to "drange" and sometimes these are too-much. Take GPIO pins, for instance, we're interested in a single integer to represent that as the other 3 (min, max, step) are meaningless in such context.

Then we should introduce new mappings for "irange_spec"/"irange" and "drange_spec"/"drange" and use them where a range specification is needed (but not the value) or the whole set of members are needed (ie: constants).

bdilly:
all changes merged. we're done.

PR #948, opened 2015-10-23
flow: use options 'float' + 'drange_spec'

bdilly:
It's the float part of task #724.
The same will be done for integers next.

Since it's not a single float, it's a full drange data,
the name was misleading.

Actually, after some discussion, we get to conclusion this
type will be dropped, and instead another two should
be used:
    * float for single floats
    * float_spec (for min / max / step)

If drange is wanted, two options will be required (float +
        drange_spec).

So, for now, a "deprecated" warning will be displayed when
a type using it is generated.

### DIFF
--- a/data/schemas/node-type-genspec.schema
+++ b/data/schemas/node-type-genspec.schema
@@ -62,6 +62,7 @@
         "byte",
         "direction-vector",
         "drange",
+        "float",
         "int",
         "rgb",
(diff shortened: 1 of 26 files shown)


bdilly:
I'm merging this after a quick review from @glima too.
I'm going to use the same approach for "int" option.

PR #949, opened 2015-10-23
Add crypto stuff for RIOT

ibriano:
Also worker threads, but disabled by default because malloc() in newlib
is not thread safe.

### DIFF
--- a/src/lib/common/Kconfig
+++ b/src/lib/common/Kconfig
@@ -240,11 +240,15 @@ config MAINLOOP_CONTIKI
 endchoice
 
 config PTHREAD
-	bool "POSIX threads (pthread)"
-	depends on HAVE_PTHREAD_H && LINUX
+	bool
+
+config THREADS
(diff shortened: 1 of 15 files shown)


ibriano:
Turns out I'm stupid and I don't know how to thread.

barbieri:
What/why?

ibriano:
On Oct 28, 2015, at 20:25, Gustavo Sverzut Barbieri <notifications@github.com<mailto:notifications@github.com>> wrote:


What/why?

It's not about `malloc()` not being thread safe, it doesn't even need to be because threads are not preemptible. But when the main thread sleeps waiting to join a thread, the respective thread needs to set its state to active again, without yielding to it, or the main thread runs immediately and frees the stack before RIOT actually notices the thread finished.

All that got me thinking about whether all he other stuff I put there to handle threads is really necessary, when I decide on all that I'll send a new version of this patch set.

�

PR #950, opened 2015-10-23
Fix memory leak in json convert node

otaviobp:
A new blob was created and the memory was leaked. Stop creating this
node, because it was not necessary.


ibriano:
+1

otaviobp:
merged with fixes

PR #951, opened 2015-10-23
tests: change filter tests names and fix execution mode

bdilly:
So they can see listed together with others of the
same module.


bdilly:
merged

PR #952, opened 2015-10-26
float: New way to map irange values - no need for long double

edersondisouza:
As long double is not always bigger than double, we now do the
calculations without need for a bigger temporary storage.
The idea is to find the distance to interval midpoint on input, and
replicate this distance on output interval. This restric calculations to
normal double variables.


glima:
:ok:

edersondisouza:
Thanks, merged.

PR #954, opened 2015-10-26
persistence: Avoid bad mem access when store_only_val opt is set.

glima:


edersondisouza:
lgtm

glima:
Thanks, merging.

PR #956, opened 2015-10-26
Mqtt flow

rchiossi:


barbieri:
JSON description is not that helpful, like "QOS" or Keepalive", they should explain the purpose and value (Keepalive is a timeout? In seconds? What happens after the expire?)

Also, what's the need to publish and subscribe? Can't they be automatic? Is there a way to remove publish (if so change from any to Boolean)

rchiossi:
QOS is described in the MQTT documentation. Not sure where/how i should duplicate that info in the module.

The keepalive is the number of seconds after which the broker should send a PING message to the client if no other messages have been exchanged in that time. Indeed that needs to be documented.

How can you automate publish/subscribe? They are dependent on the other parameters which can arrive in any order.

barbieri:
Please replicate the documentation, it's better to find it right away instead to look for another doc.

As for publish, if the only problem is the ports, you can flag which one was received (starts = true if an option parameter) and do automatically when all of them is known? I don't know the MQTT behavior, just asking from the node type user PoV. If not, document when PUBLISH needs to be triggered

rchiossi:
The problem is that I don't know how many of the ports will be needed. In one case I may need to wait until something arrives in the 'user' port, while in another 'user' may not even be required. That's true for most ports.

barbieri:
ok, then just write something describing when to call them in their doc.

PR #957, opened 2015-10-26
Move function to convert unicode code to utf8 to sol-utils

otaviobp:
Also adds a function to get the unicode code from a utf8 character

### DIFF
--- a/src/lib/datatypes/include/sol-buffer.h
+++ b/src/lib/datatypes/include/sol-buffer.h
@@ -37,6 +37,7 @@
 #include <errno.h>
 #include <sol-str-slice.h>
 #include <stdarg.h>
+#include <sol-util.h>
 
 #ifdef __cplusplus
 extern "C" {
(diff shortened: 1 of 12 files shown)


otaviobp:
sol-util-file.h was including sol-buffer.h, but it was not using it. When I added sol-util include to sol-buffer I had issues with circular dependencies. So I removed the include from sol-util-file and add it to sol-uart-impl-linux.c  that was really using buffer.
A similar problem occurred in message-digest test. I needed to change a define name because it start to clash with a define from sol-util-file.h when I changed the include files.

otaviobp:
We are accepting case for performance issues, afaik. We don't need to check for both cases everytime and for the cases we know the case base16_decode_digit will be faster.
To be honest I don't know if this is really affecting the performance. @barbieri, is this really the reason you suggested adding a enum to this methods?

barbieri:
Not only performance, in some cases one just accept one case and the other is to be considered error.

As for sol-util.h, it's not installed so can't be included in the sol-buffer. You can translate two enums or take the time to make utils stuff to be installed.

otaviobp:
Ow, I haven't realize that. I'll see what I can do here

PR #958, opened 2015-10-26
Pin map v2

mbelluzzo:
Changes from v1:
* Changed 'label_open' to 'open_by_label' on API
* 'Un-bump' the version of pin-mux struct
* Fix log_init
* Use the new API on hw io nodes (and fix the nodes that used this as sub-node)
 * Fix conf files, samples and tests


mbelluzzo:
v3: #978

PR #960, opened 2015-10-26
Linux micro kmod v5

dorileo:
## Changes
  + v2 (since #844)
    - impemented coldplug;
    - implemented hotplug;
    - introduced patches:

  + v3 (since #877)

  + v4 (since #916)

  + v5 (since #935)
    - fixed issues pointed by @lpereira in the last review (#935);
    - moved uevent api to ```sol_platform_linux```;
    - added ```mount/umount/list mount points``` api's - with async options;
    - migrated the fstab linux-micro service to use the ```sol_platform_mount()``` api;
    - migrated the automount linux-micro service to use the ```sol_platform_mount()``` api;
    - changed the general approach for the ```str-slice: implement trailing + leading whitespace helpers``` and fixed the addr sanitizer issue;

## Rationale

via uevent netlink channel.


src/lib/common/sol-platform-linux-common.c
```diff
@@ -263,3 +307,581 @@ sol_platform_impl_get_os_version(char **version)
...
+        bool noauto = false;
+
+        mpoint = calloc(1, sizeof(struct sol_mount_point));
+        if (!mpoint) {
+            SOL_ERR("Could not allocate mount point memory");
+            goto err;
+        }
+
+        if (mbuf.mnt_fsname)
+            mpoint->dev = strdup(mbuf.mnt_fsname);
```
lpereira: Null-check missing from these `strdup()` calls

src/lib/common/sol-platform-linux-common.c
```diff
@@ -263,3 +307,581 @@ sol_platform_impl_get_os_version(char **version)
...
+    }
+    endmntent(tab);
+    return 0;
+
+err:
+    SOL_PTR_VECTOR_FOREACH_IDX (vector, mp, i) {
+        sol_mount_point_free(mp);
+    }
+    sol_ptr_vector_clear(vector);
+    return -ENOMEM;
```
lpereira: `endmntent()` not called.

src/lib/common/sol-platform-linux-common.c
```diff
@@ -263,3 +307,581 @@ sol_platform_impl_get_os_version(char **version)
...
+        if (mountopt_str_to_mask(*itr, &flags)) {
+            continue;
+        } else if (sol_str_slice_str_eq(*itr, "noauto")) {
+            *noauto = true;
+            continue;
+        }
+
+        if (!first_custom)
+            sol_buffer_append_char(custom_opts, ',');
+        sol_buffer_append_slice(custom_opts, *itr);
```
lpereira: Rellocation might fail here.

src/lib/common/sol-platform-linux-common.c
```diff
@@ -263,3 +307,581 @@ sol_platform_impl_get_os_version(char **version)
...
+                sol_ptr_vector_remove(&mtab, mp2);
+            } else {
+                sol_ptr_vector_append(&delete, mp2);
+            }
+        }
+        sol_mount_point_free(mp);
+    }
+
+    SOL_PTR_VECTOR_FOREACH_IDX (&delete, mp, i) {
+        sol_mount_point_free(mp);
```
lpereira: From the loop above, it looks like a mount point can get in the `delete` vector twice.

src/lib/common/sol-platform-linux-common.c
```diff
@@ -263,3 +307,581 @@ sol_platform_impl_get_os_version(char **version)
...
+{
+    int fd = 0;
+    struct sol_vector lines;
+    struct sol_str_slice content = SOL_STR_SLICE_EMPTY;
+    struct sol_str_slice *itr;
+    struct sol_file_reader *reader;
+    uint16_t idx;
+
+    fd = open(SOL_MTAB_FILE, O_RDONLY | O_CLOEXEC);
+    if (!fd) {
```
lpereira: `fd` 0 is a valid file descriptor. `open()` returns `-1` on error.

src/lib/common/sol-platform-linux-common.c
```diff
@@ -263,3 +307,581 @@ sol_platform_impl_get_os_version(char **version)
...
+            sol_util_strerrora(errno));
+        return -errno;
+    }
+
+    content = sol_file_reader_get_all(reader);
+    lines = sol_util_str_split(content, "\n", 0);
+    SOL_VECTOR_FOREACH_IDX (&lines, itr, idx) {
+        char buff[512];
+        snprintf(buff, itr->len, "%s", itr->data);
+        if (!strstr(buff, mpoint))
```
lpereira: `strstr()` will search the whole string. If there's a comment with the mount point name, it's going to be a hit. Is that the desired behavior?

src/lib/common/sol-platform-linux-common.c
```diff
@@ -263,3 +307,581 @@ sol_platform_impl_get_os_version(char **version)
...
+        SOL_ERR("Could not read "SOL_MTAB_FILE " file - %s",
+            sol_util_strerrora(errno));
+        return -errno;
+    }
+
+    content = sol_file_reader_get_all(reader);
+    lines = sol_util_str_split(content, "\n", 0);
+    SOL_VECTOR_FOREACH_IDX (&lines, itr, idx) {
+        char buff[512];
+        snprintf(buff, itr->len, "%s", itr->data);
```
lpereira: Is the copy here necessary? If so, why not use `memcpy()` or `strncpy()`?

src/lib/common/sol-platform-linux-common.c
```diff
@@ -263,3 +307,581 @@ sol_platform_impl_get_os_version(char **version)
...
+    uint16_t idx;
+
+    fd = open(SOL_MTAB_FILE, O_RDONLY | O_CLOEXEC);
+    if (!fd) {
+        SOL_ERR("Could not open "SOL_MTAB_FILE " - %s",
+            sol_util_strerrora(errno));
+        return -errno;
+    }
+
+    reader = sol_file_reader_fd(fd);
```
lpereira: Is there the necessity to `open()` the file and pass to this function? Can't you just use the function that creates a reader from a file path?

src/lib/common/sol-platform-linux-common.c
```diff
@@ -263,3 +307,581 @@ sol_platform_impl_get_os_version(char **version)
...
+    struct sol_buffer output;
+
+    res = fd = 0;
+    sol_buffer_init(&output);
+
+    err = sol_mtab_remove_entry(mpoint, &output);
+    SOL_INT_CHECK(err, < 0, err);
+
+    fd = open(SOL_MTAB_FILE, O_RDWR | O_CLOEXEC | O_TRUNC);
+    if (!fd) {
```
lpereira: `0` is a valid fd; see comment above.

src/lib/common/sol-platform-linux-common.c
```diff
@@ -263,3 +307,581 @@ sol_platform_impl_get_os_version(char **version)
...
+        SOL_ERR("Could not open " SOL_MTAB_FILE);
+        return -errno;
+    }
+
+    err = write(fd, output.data, sizeof(output.data));
+    if (err < -1) {
```
lpereira: `write()` returns `-1` on error, this will never handle it.

src/lib/common/sol-platform-linux-common.c
```diff
@@ -263,3 +307,581 @@ sol_platform_impl_get_os_version(char **version)
...
+}
+
+// since we don't have a mtab entry remove function we must implement it ourselves
+static int
+sol_mtab_cleanup(const char *mpoint)
+{
+    int res, err, fd;
+    struct sol_buffer output;
+
+    res = fd = 0;
```
lpereira: If you must initialize `fd` for some reason, initialize it with an invalid value, such as `-1`. `0` is `stdin` usually.

src/lib/common/sol-platform-linux-common.c
```diff
@@ -263,3 +307,581 @@ sol_platform_impl_get_os_version(char **version)
...
+    err = write(fd, output.data, sizeof(output.data));
+    if (err < -1) {
+        SOL_ERR("Could not write "SOL_MTAB_FILE " file - %s",
+            sol_util_strerrora(errno));
+        res = -errno;
+    }
+
+    err = close(fd);
+    if (err < -1) {
+        SOL_ERR("Could not close "SOL_MTAB_FILE " file - %s",
```
lpereira: Same comment for the `write()` call applies here.

src/lib/common/sol-platform-linux-common.c
```diff
@@ -263,3 +307,581 @@ sol_platform_impl_get_os_version(char **version)
...
+    SOL_NULL_CHECK(mpoint, -EINVAL);
+
+    if (!async_cb)
+        return run_umount(mpoint);
+
+    udata = calloc(1, sizeof(struct umount_data));
+    SOL_NULL_CHECK(udata, -ENOMEM);
+
+    udata->data = data;
+    udata->mpoint = strdup(mpoint);
```
lpereira: Null-check missing

src/lib/common/sol-platform-linux-common.c
```diff
@@ -263,3 +307,581 @@ sol_platform_impl_get_os_version(char **version)
...
+    if (!async_cb)
+        return run_mount(dev, mpoint, fstype, flags, opts);
+
+    mdata = calloc(1, sizeof(struct mount_data));
+    SOL_NULL_CHECK(mdata, -ENOMEM);
+
+    mdata->data = data;
+    mdata->dev = strdup(dev);
+    mdata->mpoint = strdup(mpoint);
+    mdata->fstype = strdup(fstype);
```
lpereira: Null checks missing.

src/lib/common/sol-platform-linux-common.c
```diff
@@ -263,3 +307,581 @@ sol_platform_impl_get_os_version(char **version)
...
+        SOL_INF("We're running side-by-side with udevd, skipping udevd generated event");
+        return;
+    }
+
+    while (i < len) {
+        char *curr, *del;
+        size_t value_len;
+
+        curr = msg + i;
+        del = strrchr(curr, '=');
```
lpereira: Might be a good idea to NULL-check this to guard against a possible bug in udev.

src/lib/common/sol-platform-linux-common.c
```diff
@@ -263,3 +307,581 @@ sol_platform_impl_get_os_version(char **version)
...
+        value_len = del - curr;
+
+        if (!strncmp(curr, "MODALIAS", value_len))
+            uevent.modalias = sol_str_slice_from_str(++del);
```
lpereira: You can `del++` only once before these comparisons; will probably save a few bytes from this routine.

src/lib/common/sol-platform-linux-common.c
```diff
@@ -263,3 +307,581 @@ sol_platform_impl_get_os_version(char **version)
...
+static bool
+sol_uevent_handler(void *data, int fd, unsigned int cond)
+{
+    int len;
+    struct uevent_context *ctx = data;
+    char buffer[512];
+
+    len = recv(ctx->uevent.fd, buffer, sizeof(buffer), MSG_WAITALL);
+    if (len == -1) {
+        SOL_ERR("Could not read netlink socket. %s", sol_util_strerrora(errno));
```
lpereira: Retry if `errno == EINTR`?

src/lib/datatypes/include/sol-str-slice.h
```diff
@@ -123,6 +124,43 @@ sol_str_slice_to_string(const struct sol_str_slice slice)
...
+    }
+
+    return copy;
+}
+
+static inline struct sol_str_slice
+sol_str_slice_remove_trailing_whitespace(struct sol_str_slice slice)
+{
+    struct sol_str_slice copy = slice;
+    long idx = copy.len;
```
lpereira: `size_t`?

src/lib/datatypes/include/sol-str-slice.h
```diff
@@ -123,6 +124,43 @@ sol_str_slice_to_string(const struct sol_str_slice slice)
...
+    bool nterm = false;
+
+    while (idx >= 0) {
```
lpereira: `while (idx != 0) {`?

src/lib/datatypes/include/sol-str-slice.h
```diff
@@ -123,6 +124,43 @@ sol_str_slice_to_string(const struct sol_str_slice slice)
...
+        char c = copy.data[idx];
+        if (isspace(c))
+            copy.len--;
+        else if (c == '\0')
```
lpereira: A slice might contain multiple `\0` characters, not necessarily it's going to be the terminating `NUL`.

src/modules/linux-micro/automount/automount.c
```diff
@@ -0,0 +1,347 @@
...
+};
+
+struct filesystem {
+    const char *id;
+    const struct filesystem_magic *magics;
+    bool (*fs_cb)(int fd);
+};
+
+static const struct filesystem_magic vfat_magic[] = {
+    { .magic = "MSWIN",    .length = 5, .offset = 0x52 },
```
lpereira: Perhaps a macro here will help maintain the table?

    #define MAGIC(str_, offset_)  { .magic = (str_), .length = sizeof(str_) - 1,  .offset = (offset_) }
        MAGIC("MSWIN", 0x52),
        MAGIC("FAT32   ", 0x52),
        MAGIC("MSDOS", 0x36),
        ...
    #undef MAGIC

src/modules/linux-micro/automount/automount.c
```diff
@@ -0,0 +1,347 @@
...
+    { .magic = "FAT16   ", .length = 8, .offset = 0x36 },
+    { .magic = "FAT12   ", .length = 8, .offset = 0x36 },
+    { .magic = "FAT     ", .length = 8, .offset = 0x36 },
+    { .magic = "\353",     .length = 1, },
+    { .magic = "\351",     .length = 1, },
+    { .magic = "\125\252", .length = 2, .offset = 0x1fe },
+    { NULL }
+};
+
+static const struct filesystem_magic hfsplus_magic[] = {
```
lpereira: Do we need to support HFS+ and HFS? I'm guessing only FAT and EXT filesystems are of interest for us.

src/modules/linux-micro/automount/automount.c
```diff
@@ -0,0 +1,347 @@
...
+        goto err;
+    }
+
+    err = read(fd, buff, magic->length);
+    if (err != magic->length) {
+        SOL_ERR("Could not read dev: %s", dev);
+        goto err;
+    }
+
+    res = streq(buff, magic->magic);
```
lpereira: Better to use `memcmp()` here, as the magic is possibly some arbitrary binary value.

src/modules/linux-micro/automount/automount.c
```diff
@@ -0,0 +1,347 @@
...
+        return;
+    }
+
+    err = snprintf(dev, sizeof(dev), "/dev/%s", uevent->devname.data);
+    if (err < 0 || err >= (int)sizeof(dev)) {
+        SOL_ERR("Could not format nodedev path");
+        return;
+    }
+
+    err = snprintf(mpoint, sizeof(mpoint), "/mnt/%s", uevent->devname.data);
```
lpereira: Do we want to mount everything under `/mnt`, or `/media` or other mount points are also valid?

src/modules/linux-micro/automount/automount.c
```diff
@@ -0,0 +1,347 @@
...
+    }
+
+    fstype = automount_get_fstype(dev);
+    if (!fstype) {
+        SOL_ERR("Could not determine the fstype for %s, not mounting", dev);
+        return;
+    }
+
+    err = mkdir(mpoint, 0755);
+    if (err < 0 && errno != EEXIST) {
```
lpereira: What if the mount point exists but is already mounted?

src/modules/linux-micro/fstab/fstab.c
```diff
@@ -42,6 +42,7 @@
 SOL_LOG_INTERNAL_DECLARE_STATIC(_log_domain, "linux-micro-fstab");
```
lpereira: IIRC there are many things in `fstab.c` that is very similar to the filesystem mounting changes in the platform implementation file. Can't them be used here?

src/modules/linux-micro/kmod/kmod.c
```diff
@@ -0,0 +1,398 @@
...
+    if (dfd < 0) {
+        SOL_DBG("Could not obtain directory file descriptor");
+        return;
+    }
+
+    r = readdir_r(dir, &ent, &entbuf);
+    while (r == 0 && entbuf) {
+        struct stat st;
+
+        if (!strcmp(ent.d_name, ".") || !strcmp(ent.d_name, "..")) {
```
lpereira: This will process hidden files as well, is that on purpose?

src/test/test-str-slice.c
```diff
@@ -108,6 +108,80 @@ test_str_slice_str_eq(void)
...
+        TEST_NOT_EQUAL("  with two leading whitespace"),
+        TEST_NOT_EQUAL(" "),
+        TEST_NOT_EQUAL("\twith one leading whitespace"),
+        TEST_NOT_EQUAL("\t\twith two leading whitespace"),
+        TEST_NOT_EQUAL("\t"),
+        TEST_NOT_EQUAL("\nwith one leading whitespace"),
+        TEST_NOT_EQUAL("\n\nwith two leading whitespace"),
+        TEST_NOT_EQUAL("\n"),
+        TEST_EQUAL(""),
+        TEST_EQUAL("without leading whitespace"),
```
lpereira: Please add the test I've asked for in the previous PR. It'll ensure that a buffer underflow won't happen when executed with address sanitizer or Valgrind (which the bots do by default):

     const char test_underflow_buf[] = "         some random string";
     const char test_underflow_no_leading_space[] = "some random string";
     sol_str_slice test_underflow = sol_str_slice_from_str(test_underflow_buf + 3);

     ASSERT(sol_str_slice_eq(test_underflow, test_underflow_no_leading_space));

src/lib/common/sol-platform-linux-common.c
dorileo: This file is only manipulated by us,  destroyed and recreated each reboot,  I'm no considering a user change in this file a valid thing.

src/lib/common/sol-platform-linux-common.c
dorileo: Yep,  true.  Thanks.

src/lib/common/sol-platform-linux-common.c
dorileo: Yep,  true...  Typo here.

src/lib/common/sol-platform-linux-common.c
dorileo: We are skipping all udev events which by the way is valid only while testing cause it's meant to run without any udev running.

src/lib/datatypes/include/sol-str-slice.h
dorileo: Intentional. A negative index here is "valid"  if we reach -1 we leave the loop,  if we use size_t when 0 - 1 will produce INT_MAX and end up in infinite loop.

src/modules/linux-micro/automount/automount.c
dorileo: Just used to test the overall thing since I had an Ipod in hand...  May drop it if we really don't care.

src/modules/linux-micro/automount/automount.c
dorileo: Yes, forgot to change,  it really makes sense.

src/modules/linux-micro/automount/automount.c
dorileo: We'll always mount it in the same base dir,  I must check the one we are using in the images.

src/modules/linux-micro/automount/automount.c
dorileo: The mount will fail since we are no providing the remount flag

src/modules/linux-micro/fstab/fstab.c
dorileo: Yes,  I see one extra api for that,  I can add that and reuse the whole bits.

src/modules/linux-micro/kmod/kmod.c
dorileo: Sysfs has no hidden files as far as I remember,  but it wouldn't be a problem if so.

src/test/test-str-slice.c
dorileo: Ok,  I thought I had added all the cases you've asked...  Will add these ones as well,  thanks.

src/lib/common/sol-platform-linux-common.c
lpereira: It's cheap to do `if (!del) continue;`. It's likely that Coverity will point this out and you'll have to fix this anyway. :)

src/lib/datatypes/include/sol-str-slice.h
lpereira: `ssize_t`, then, to be consistent and platform-independent. However, if you find a way to make this without `-1` being the loop finishing condition, then it's better.

src/lib/common/include/sol-platform.h
```diff
@@ -156,6 +160,75 @@ int sol_platform_restart_service(const char *service);
...
 int sol_platform_set_target(const char *target);
 
+struct sol_mount_point {
+    char *dev;
+    char *mpoint;
+    char *fstype;
+    unsigned long flags;
+    char *opts;
+    bool noauto;
+};
```
barbieri: this shouldn't be exposed in a general header file like this, at least not the way it is.

flags, opts and dev are all linux-specific. One could argue that dev could be virtual and on other systems could be encoded as required (ie: "usb:xyz" for a device "xyz" in an usb bus in RIOT/Contiki, if these are ever supported). But should we bother to list where the device came from? Or the flags?

Right now I'd keep it simple and just do not expose anything but the mount point.

src/lib/common/include/sol-platform.h
```diff
@@ -156,6 +160,75 @@ int sol_platform_restart_service(const char *service);
...
+    free(mp->fstype);
+    free(mp->opts);
+    free(mp);
+}
+
+/**
+ * List mounted mount points, if @c hplug_only is provided this function will list
+ * only mount points mounted by us, all mounted mount points are listed otherwise.
+ *
+ * @param hplug_only List only mount points only mounted by us?
```
barbieri: what's the point to list other mount points? Here we care about non-system mount points, so only hot plug (auto mounted). Those that are system-specific, such as linux-micro setup (/proc, /sys, /dev...) shouldn't show here. I'd consider /etc/fstab the same.

src/lib/common/include/sol-platform.h
```diff
@@ -156,6 +160,75 @@ int sol_platform_restart_service(const char *service);
...
+ * If the umount operation is desired to be run synchronously just provide @c async_cb
+ * as NULL.
+ *
+ * @param mpoint The mount point to be unmounted.
+ * @param async_cb The async callback to be called when async operation is required.
+ * @param data Context data to be provided to @c async_cb function.
+ *
+ * @return 0 on success, negative errno otherwise
+ */
+int sol_platform_umount(const char *mpoint, void (*async_cb)(const char *mpoint, void *data, uint64_t pid, int status), void *data);
```
barbieri: always take `const void *data`, to allow user given const pointers without cast.

and our convention is to always pass `void *data` (casted at callback time) as the first parameter to the callback function.

`pid` and `status` are not something that should be exposed in the documentation, they are implementation-dependent and generally as no interest to the user.

the documentation of this function should not mention anything about processes, it's an implementation detail. `async_cb` should always be given and it should receive the errno.

Proposed signature:

```c
int sol_platform_unmount(const char *mpoint, void (*cb)(void *data, const char *mpoint, int error), const void *data);
```

In linux-micro this will use fork-run and call `umount(2)` from the process, convert the result into exit code that is then returned by fork-run's `on_dead`, that is given to the user.

Not sure about systemd, but likely there is a way to request a mount operation via dbus, then you proceed like that and convert back the resulting error code, calling the user callback.

The `sol_platform_unmount()` error return is only if the implementation fail to start the async operation, like fork-run failing. The actual unmount operation error is given as callback's `error` parameter.

src/lib/common/include/sol-platform.h
```diff
@@ -156,6 +160,75 @@ int sol_platform_restart_service(const char *service);
...
+ * @param mpoint The mount point where this block device is to be mounted on
+ * @param fstype The filesystem type (i.e vfat, ext2, ext3 etc)
+ * @param flags The mount flags
+ * @param opts File system's custom options
+ * @param async_cb The async callback to be called when async operation is required.
+ * @param data Context data to be provided to @c async_cb function
+ *
+ * @return 0 on success, negative errno otherwise
+ */
+int sol_platform_mount(const char *dev, const char *mpoint, const char *fstype, unsigned long flags, const char *opts, void (*async_cb)(const char *dev, const char *mpoint, void *data, uint64_t pid, int status), void *data);
```
barbieri: like above, however I don't think all of these parameters should be there.

I'm particularly concerned with `flags` and `opts`. If we want to keep something like that, then we should have our own enums that will be translated to current OS values.

If we keep `fstab` linux-micro module like it is currently, then I think we're good to keep all of this internal to a platform. IOW, do not export `sol_platform_mount()` and only define the mount call inside linux-micro. The mount calls will be executed from platform's automount, not from the user.

src/lib/common/sol-platform-linux-common.c
```diff
@@ -263,3 +307,581 @@ sol_platform_impl_get_os_version(char **version)
...
+        { "relatime", MS_RELATIME },
+        { "remount", MS_REMOUNT },
+        { "silent", MS_SILENT },
+        { "strictatime", MS_STRICTATIME },
+        { "sync", MS_SYNCHRONOUS },
+        { "defaults", ~MS_RDONLY | ~MS_NOSUID | ~MS_NODEV | ~MS_NOEXEC | ~MS_SYNCHRONOUS },
+    };
+
+    for (i = 0; i < ARRAY_SIZE(table); i++) {
+        if (sol_str_slice_str_eq(slice, table[i].opt)) {
```
barbieri: why not use the str table?

src/lib/common/sol-platform-linux-common.c
```diff
@@ -263,3 +307,581 @@ sol_platform_impl_get_os_version(char **version)
...
+    }
+
+    sol_ptr_vector_init(&mtab);
+    err = parse_mount_point_file(SYS_MTAB_FILE, &mtab);
+    SOL_INT_CHECK_GOTO(err, < 0, err_mtab);
+
+    sol_ptr_vector_init(&sol_mtab);
+    err = parse_mount_point_file(SOL_MTAB_FILE, &sol_mtab);
+    SOL_INT_CHECK_GOTO(err, < 0, err_sol_mtab);
+
```
barbieri: I really don't get why you're doing all of this here. It would be much simpler to only record the auto-mounted devices and return it.

Or at least for linux-specific code you parse /proc/mounts and exclude the system mounts, but it's cumbersome as you'd have to check /etc/fstab and with systemd it support ".mount" units (actually /etc/fstab is converted to a series of ".mount" by a generator).

all in all i think this over complicates things.

src/lib/common/sol-platform-linux-common.c
```diff
@@ -263,3 +307,581 @@ sol_platform_impl_get_os_version(char **version)
...
+int
+sol_platform_impl_umount(const char *mpoint, void (*async_cb)(const char *mpoint, void *data, uint64_t pid, int status), void *data)
+{
+    struct umount_data *udata;
+
+    SOL_NULL_CHECK(mpoint, -EINVAL);
+
+    if (!async_cb)
+        return run_umount(mpoint);
```
barbieri: avoid doing this kind of widely different behavior. If you find another case like this in the future, keep the fork-run (or thread, or idler/timer) flow and just skip the callback at the end, it will keep the execution flow the same, avoiding problems.

src/lib/common/sol-platform-linux-common.c
```diff
@@ -263,3 +307,581 @@ sol_platform_impl_get_os_version(char **version)
...
+    err = run_umount(udata->mpoint);
+    sol_platform_linux_fork_run_exit(err);
+}
+
+static void
+on_umount_fork_exit(void *data, uint64_t pid, int status)
+{
+    struct umount_data *udata = data;
+
+    udata->async_cb(udata->mpoint, udata->data, pid, status);
```
barbieri: if you want to keep the list of mount points, here you'd be checking the `status`, if 0, then you append to the list before calling user's callback. Simple as that.

src/lib/common/sol-platform-linux-common.c
```diff
@@ -263,3 +307,581 @@ sol_platform_impl_get_os_version(char **version)
...
+static void
+sol_uevent_event_dispatch(struct uevent_context *ctx, struct sol_uevent *uevent)
+{
+    uint16_t idx;
+    struct uevent_callback *cb;
+
+    SOL_VECTOR_FOREACH_IDX (&ctx->callbacks, cb, idx) {
+        if ((!cb->action || (cb->action && sol_str_slice_str_eq(uevent->action, cb->action))) &&
+            (!cb->subsystem || (cb->subsystem && sol_str_slice_str_eq(uevent->subsystem, cb->subsystem)))) {
+            cb->uevent_cb((void *)cb->cb_data, uevent);
```
barbieri: you need to be careful here, if `uevent_cb` changes `ctx->callbacks` (ie: remove a callback handler), it will crash.

See other parts of our code where we keep a pending list + walking counter.

src/lib/common/sol-platform-linux-common.c
```diff
@@ -263,3 +307,581 @@ sol_platform_impl_get_os_version(char **version)
...
+        .modalias = SOL_STR_SLICE_EMPTY,
+        .action = SOL_STR_SLICE_EMPTY,
+        .subsystem = SOL_STR_SLICE_EMPTY,
+        .devtype = SOL_STR_SLICE_EMPTY,
+        .devname = SOL_STR_SLICE_EMPTY,
+    };
+    int i = 0;
+
+    /** lets avoid it to misbehave when people trying to test it with systemd running */
+    if (!strncmp(msg, LIBUDEV_ID, sizeof(LIBUDEV_ID))) {
```
barbieri: `streq()`

src/lib/common/sol-platform-linux-common.c
```diff
@@ -263,3 +307,581 @@ sol_platform_impl_get_os_version(char **version)
...
+    struct uevent_callback *cb;
+
+    if (!uevent_ctx.running)
+        sol_vector_init(&uevent_ctx.callbacks, sizeof(struct uevent_callback));
+
+    cb = sol_vector_append(&uevent_ctx.callbacks);
+    SOL_NULL_CHECK(cb, -1);
+
+    cb->action = action;
+    cb->subsystem = subsystem;
```
barbieri: don't keep pointers to user provided strings or document the expected lifecycle. If user calls you with strings on its stack (ie: snprintf()), then you'll be using garbage later.

src/lib/common/sol-platform-linux-common.c
```diff
@@ -263,3 +307,581 @@ sol_platform_impl_get_os_version(char **version)
...
+    if (!uevent_ctx.running) {
+        sol_uevent_register(&uevent_ctx);
+    }
+
+    uevent_ctx.running = true;
+    return 0;
+}
+
+static void
+sol_uevent_cleaup(struct uevent_context *ctx)
```
barbieri: typo... `cleaNup`, missing that N.

src/lib/common/sol-platform-linux-common.c
```diff
@@ -263,3 +307,581 @@ sol_platform_impl_get_os_version(char **version)
...
+    if (ctx->uevent.fd)
+        close(ctx->uevent.fd);
+
+    sol_vector_clear(&ctx->callbacks);
+    ctx->running = false;
+}
+
+SOL_API int
+sol_platform_linux_uevent_unsubscribe(const char *action, const char *subsystem,
+    void (*uevent_cb)(void *cb_data, struct sol_uevent *uevent))
```
barbieri: you're not using `action` and `subsystem`. And always provide the `data` to verify the callback you're using.

When people program using wrappers, like doing a Language binding (Python/JS) or just helpers, he will give you the same callback function for all cases, what changes is the context (void *data).

src/lib/common/sol-platform-linux-common.c
```diff
@@ -263,3 +307,581 @@ sol_platform_impl_get_os_version(char **version)
...
+{
+    struct uevent_callback *callback;
+    uint16_t idx;
+    int res = -1;
+
+    SOL_NULL_CHECK(uevent_cb, -1);
```
barbieri: -1 is bad... use -ENOENT.

src/lib/common/sol-platform.c
```diff
@@ -442,3 +442,21 @@ sol_platform_inform_service_monitors(const char *service,
         ((void (*)(void *data, const char *service, enum sol_platform_service_state state))m->base.cb)((void *)m->base.data, service, state);
     }
 }
+
+SOL_API int
+sol_platform_get_mount_points(bool hplug_only, struct sol_ptr_vector *vector)
+{
```
barbieri: do the common verification here, such as checking if `vector` is valid.

src/lib/datatypes/include/sol-str-slice.h
barbieri: then use `ssize_t` (signed) and check for overflows. I'd still follow @lpereira's advice and use `size_t`, adapting checks to not go to underflow case.

src/modules/linux-micro/automount/automount.c
barbieri: keep it, it's low cost.

src/modules/linux-micro/automount/automount.c
```diff
@@ -0,0 +1,347 @@
...
+        SOL_ERR("Coud not lseek to ext superblock");
+        return false;
+    }
+
+    err = read(fd, buff, sizeof(struct ext_super_block));
+    if (err != sizeof(struct ext_super_block)) {
+        SOL_ERR("Could not read ext superblock features");
+        return false;
+    }
+
```
barbieri: all of those should be retried when `errno == EINTR`

barbieri:
@dorileo this is getting too big, let's improve the workflow and focus on delivering smaller patchsets, then get them merged and just then follow with new features.

The file reader addition should be merged.

Then please get your string whitespace stuff in, since @cabelitos will use it as well. It's simple and almost there (just one comment about `long` x `size_t`).

After that follow up with cleanup to sysctl. AFAIR it is good and can be merged soon (or right now).

Proceed with kmod without automount stuff. It's good to go.

The automount stuff will likely need more review rounds and it's bad to keep the patchset growing like this -- for us it's always a pain to re-review, for you it's always a pain to keep rebasing and not getting your code in the master branch.

Use this workflow for next patchsets. If the infrastructure parts are good, merge them before the rest of the code. Just keep infrastructure bits with your patchset for too long if they are very tied to your other patches and may change depending on your approach -- like the current bits for mount/unmount APIs.

src/lib/datatypes/include/sol-str-slice.h
cabelitos: Also not that you should  set idx = copy.len - 1; and not copy.len

cabelitos:
@dorileo @barbieri It would be since to merge trim whitespaces patch ASAP.

src/lib/datatypes/include/sol-str-slice.h
barbieri: so likely change the loop to:

```c
size_t idx = copy.len;

while (idx > 0) {
   idx--;
   ... 
}
```

dorileo:
@barbieri I don't intend to put more features (at least big ones) in this series - the gdbserver bits for example are in a different PR #972.

src/lib/common/include/sol-platform.h
dorileo: ok

src/lib/common/include/sol-platform.h
dorileo: ok.

src/lib/common/include/sol-platform.h
dorileo: ok

src/lib/common/sol-platform-linux-common.c
dorileo: Just realized that a generic api to get all the mount points in the system would be fair enough. But, sure, it makes things way much simpler to not consider the system ones and go on only with the hplug mpoints.

src/lib/common/sol-platform-linux-common.c
dorileo: ok

src/lib/common/sol-platform-linux-common.c
dorileo: ok

src/lib/common/sol-platform-linux-common.c
dorileo: ok

src/lib/common/sol-platform-linux-common.c
dorileo: ok

src/lib/common/sol-platform-linux-common.c
dorileo: ok

src/lib/common/sol-platform-linux-common.c
dorileo: ok

src/lib/common/sol-platform-linux-common.c
dorileo: nope, got rid of it.

src/lib/common/sol-platform-linux-common.c
dorileo: not at all.

src/lib/common/sol-platform-linux-common.c
dorileo: but it's not.

dorileo:
ok guys, v6 coming...

PR #961, opened 2015-10-26
Create a node to build URLs.

cabelitos:
Some applications may need to construct an URL before a request may be performed.
This new node will make it easier to construct an URL.


cabelitos:
Closing in favor of #967

PR #966, opened 2015-10-27
Async v5

edersondisouza:
v5:
Less memory allocation (and leaks)

v4:

v3:

v2:
Fixed pointed issues;
Added a commit that has a delayed option, so some requests can be delayed by up to SOL_PERSISTENCE_WRITE_TIMEOUT envvar. It's separated from main commit of async, as I'm not sure about this approach. If it goes in, it should (probably) be squashed onto that commit - or it maybe dropped altogether.

### DIFF
--- a/src/lib/io/include/sol-memmap-storage.h
+++ b/src/lib/io/include/sol-memmap-storage.h
@@ -64,6 +64,15 @@ extern "C" {
  * is a @c uint8_t and that versions should start on 1, so Soletta will know
  * if dealing with a totally new storage. It also considers 255 (0xff) as a
  * non-value, so fit new EEPROMs.
+ * Note that a map may define a timeout value to perform writes. This way,
+ * writings can be grouped together, as all writing operations will be
+ * performed at timeout end. That said, it's important to note that when
+ * writing is performed, no order is guaranteed for different keys, i.e.,
+ * writing to 'a' and 'b' can be performed, at timeout end, as 'b' and 'a'
+ * - but for a given key, only the last write will be performed at timeout end.
+ * Multiple writes for the same key before timeout end will result in previous
+ * writes being replaced, and their callbacks will be informed with status
+ * -ECANCELED.
  *
  * @{
  */
@@ -89,7 +98,9 @@ struct sol_memmap_map {
                        * @arg @a devnumber is device number on bus, like 0x50
                        * @arg @a devname is device name, the one recognized by its driver
                        */
-    const struct sol_str_table_ptr *entries; /**< Entries on map, containing name, offset and size */
+    unsigned int timeout; /**< Timeout, in milliseconds, of writing operations. After a write is requested, a timer will run and group all
+                           * writing operations until it expires, when real writing will be performed */
+    const struct sol_str_table_ptr *entries; /**< Entries on map, containing name, offset and size */ /* Memory trick in place, must be last on struct*/
 };
 
 struct sol_memmap_entry {
@@ -155,6 +166,21 @@ int sol_memmap_add_map(const struct sol_memmap_map *map);
  * @return 0 on success, a negative number on failure.
  */
 int sol_memmap_remove_map(const struct sol_memmap_map *map);
+
+/**
+ * Defines map timeout to actually perform write.
+ *
+ * @param map map to have its timeout changed
+ * @param timeout new timeout, in milliseconds.
+ *
+ * @return true if successfuly set map timeout.
+ *
+ * @note This change will take effect after current active timer expires.
+ * Active ones will remain unchanged
+ */
+bool sol_memmap_set_timeout(struct sol_memmap_map *map, unsigned int timeout);
+
+unsigned int sol_memmap_get_timeout(struct sol_memmap_map *map);
 
 #define CREATE_BUFFER(_val) \
     struct sol_buffer buf = SOL_BUFFER_INIT_FLAGS(_val, \
--- a/src/lib/io/include/sol-memmap-storage.h
+++ b/src/lib/io/include/sol-memmap-storage.h
@@ -36,9 +36,10 @@
 #include <stdint.h>
 
 #include "sol-buffer.h"
+#include "sol-log.h"
 #include "sol-str-table.h"
 #include "sol-types.h"
-#include "sol-log.h"
+#include "sol-util.h"
 
 #ifdef __cplusplus
 extern "C" {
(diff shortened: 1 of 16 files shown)


barbieri:
@edersondisouza please cherry pick your `int/accumulator` patch and merge it to master. This patchset is already too big and that patch is good to be merged long ago :-)

src/lib/io/include/sol-memmap-storage.h
```diff
@@ -143,105 +167,166 @@ int sol_memmap_add_map(const struct sol_memmap_map *map);
...
+ * @param timeout new timeout, in milliseconds.
+ *
+ * @return true if successfuly set map timeout.
+ *
+ * @note This change will take effect after current active timer expires.
+ * Active ones will remain unchanged
+ */
+bool sol_memmap_set_timeout(struct sol_memmap_map *map, unsigned int timeout);
+
+unsigned int sol_memmap_get_timeout(struct sol_memmap_map *map);
```
barbieri: should be `const struct sol_memmap_map *` as we don't modify it.

src/lib/io/sol-fs-storage.c
```diff
@@ -41,35 +41,95 @@
...
-    int ret = 0;
+    FILE *file = NULL;
+    struct cb_data *cb_data;
+    int ret = 0, status = 0;
 
     SOL_NULL_CHECK(name, -EINVAL);
-    SOL_NULL_CHECK(buffer, -EINVAL);
+    SOL_NULL_CHECK(blob, -EINVAL);
+
+    if (cb) {
```
barbieri: I don't like this approach of different execution flows if a callback is given or not.

For these async callbacks I'd force user to provide a callback, otherwise he will be missing error handling. If he wants to ignore error handling, he will have to provide a dummy callback -- likely he will take the time and handle the error code :-)

in every way, please execute the write from the timer. It will also make it easier in future versions to move this to a worker thread... which reminds me: you should be keeping an array of pending writes and cancel the old if a new one comes. While keeping the timer, it will not be much of a problem, but if you add threads one could step on each other foot!

src/lib/io/sol-memmap-storage.c
```diff
@@ -251,15 +301,151 @@ check_version(const struct sol_memmap_map *map)
...
+
+    file = fopen(map->path, "r+e");
+    if (!file) {
+        SOL_WRN("Error opening file [%s]: %s", map->path,
+            sol_util_strerrora(errno));
+        return false;
+    }
+
+    SOL_VECTOR_FOREACH_IDX (&map_timeout->pending_writes, pending, i) {
+        sol_memmap_write_raw_do(map->path, pending->name, pending->entry,
```
barbieri: I think this could call a `pending->cb` that may modify `map_timeout->pending_writes`, causing a bug here. Since we're going to clear the vector later, I suggest you steal that vector and reset the old pointer:

```c
struct sol_vector tmp_vector = map_timeout->pending_writes;
sol_vector_init(&map_timeout->pending_writes, sizeof(struct pending_write_data));

SOL_VECTOR_FOREACH_IDX (&tmp_vector, ...) {...}
sol_vector_clear(&tmp_vector);
```

src/lib/io/sol-memmap-storage.c
```diff
@@ -489,20 +708,41 @@ sol_memmap_add_map(const struct sol_memmap_map *map)
...
+    struct map_timeout *map_timeout;
     int i;
 
     SOL_NULL_CHECK(map, -EINVAL);
 
+    SOL_VECTOR_FOREACH_IDX (&write_timeouts, map_timeout, i) {
+        if (map_timeout->map->entries == map->entries) {
+            if (map_timeout->timeout) {
+                perform_pending_writes((void *)map_timeout->map);
+                sol_timeout_del(map_timeout->timeout);
```
barbieri: you should be doing the `sol_timeout_del()` before calling `perform_pending_writes()` as that function will make the pointer to `struct sol_timeout` NULL (or keep a copy of the pointer in your stack).

src/lib/io/sol-memmap-storage.c
```diff
@@ -513,3 +753,47 @@ sol_memmap_remove_map(const struct sol_memmap_map *map)
...
+    struct sol_memmap_map *iter;
+    int i;
+
+    SOL_NULL_CHECK(map, false);
+
+    /* Rememeber, as we may have a copy of map (due to device resolving),
+     * we need to update our proper copy */
+    SOL_PTR_VECTOR_FOREACH_IDX (&memory_maps, iter, i) {
+        if (iter->entries == map->entries) {
+            map->timeout = timeout;
```
barbieri: what if this is not a `resolved_map`? Then you're writing to a read-only memory and it will crash if proper protection is set.

I think a good approach to simplify this is to create a new internal structure that extends the `sol_memmap_map` with the timeout and always use from there. This could serve the purpose of resolve the path as well, then you keep a single structure and simplify other paths.

src/modules/flow/persistence/persistence.c
```diff
@@ -107,15 +114,73 @@ static const struct sol_str_table_ptr storage_fn_table[] = {
...
+
+        goto end;
+    }
+
+    /* No packet_data_size means dynamic content (string). Let's reallocate if needed */
+    if (!mdata->packet_data_size) {
+        if (!mdata->value_ptr || strlen(mdata->value_ptr) + 1 < blob->size) {
+            void *tmp = realloc(mdata->value_ptr, blob->size);
+            SOL_NULL_CHECK_GOTO(tmp, end);
+            mdata->value_ptr = tmp;
```
barbieri: we have `sol_util_replace_str_if_changed()`, maybe it would help?

src/lib/io/sol-efivarfs-storage.c
```diff
@@ -62,28 +71,74 @@ check_realpath(const char *path)
...
+    FILE *file = NULL;
     char path[PATH_MAX];
     int r;
+    struct cb_data *cb_data = NULL;
 
     SOL_NULL_CHECK(name, -EINVAL);
-    SOL_NULL_CHECK(buffer, -EINVAL);
+    SOL_NULL_CHECK(blob, -EINVAL);
+
+    if (cb) {
```
barbieri: similar to what I wrote for file, this one should be forced.

src/lib/io/sol-fs-storage.c
edersondisouza: So I should do the same as memmap, and any readings performed after a schedule write shall return the last valid schedled write value?

barbieri: I'd say so.

edersondisouza:
int/accumulator patch merged.

edersondisouza:
Replaced by #995 and #996

PR #967, opened 2015-10-27
Create a node to build URLs. v2

cabelitos:
Changes since v1:

* Initializing encoded_key and encoded_value variables outside the loop.
* Replacing param value with a key is already present in the params array.

------
Some applications may need to construct an URL before a request may be performed.
This new node will make it easier to construct an URL.

### DIFF
--- a/src/lib/comms/include/sol-http.h
+++ b/src/lib/comms/include/sol-http.h
@@ -240,6 +240,10 @@ void sol_http_param_free(struct sol_http_param *params);
 
 int sol_http_escape_string(char **escaped, const char *value);
 
+int sol_http_build_uri(char **uri, const char *base, const struct sol_http_param *params);
+
+int sol_http_encode_params(char **encoded_params, enum sol_http_param_type type, const struct sol_http_param *params);
+
 /**
  * @}
  */
--- a/src/lib/comms/sol-http-client-impl-curl.c
+++ b/src/lib/comms/sol-http-client-impl-curl.c
@@ -543,101 +543,6 @@ free_buffer:
     return NULL;
 }
 
-static char *
-encode_key_values(CURL *curl, enum sol_http_param_type type,
-    const struct sol_http_param *params, char *initial_value)
-{
-    struct sol_http_param_value *iter;
-    uint16_t idx;
-    bool first = true;
-
-    if (type != SOL_HTTP_PARAM_QUERY_PARAM &&
-        type != SOL_HTTP_PARAM_POST_FIELD &&
-        type != SOL_HTTP_PARAM_COOKIE) {
-        free(initial_value);
-        errno = EINVAL;
-        return NULL;
-    }
-
-    SOL_VECTOR_FOREACH_IDX (&params->params, iter, idx) {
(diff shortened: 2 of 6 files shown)


barbieri:
Taking a base URI is okay as it's the common case, but how to create that base given protocol, server, path, port, fragment? We should offer a more complete version that takes all of these in separate and create an URI. And of course one C API and node that splits.

We can have two nodes or just a take a base_uri and split it internally, overriding individual values.


As for the name, JSON used "create-" prefix, this one is "build-" (and a CREATE port). The C API is usually "new", this one is "build". Review and make it uniform, check against the other source files and if they have different names, choose one and make them all the same. 

Last, why two DECLARE statements with the same composed type signature?

cabelitos:
@barbieri I will create the "complex" node that you mention it. 
I will also sync with json nodes and keep name spaces equal

About the DECLARE: My bad, It is not necessary indeed.

barbieri:
As I said, you may opt to do it in one single node. You can do something like:

Options:
 * `protocol`: the protocol name to be used, such "http" or "ftp", defaults to "http"
 * `port`: the port number to be used, defaults to 80
 * `server`: the server to be used, defaults to null
 * `path`: the path at server, defaults to "/"
 * `fragment`: the fragment inside the given path, defaults to null.
 * `query`: the encoded query. As an example 2 parameters `a=1` and `b=My String` would be encoded as `a=1&b=My+String`. Defaults to null.
 * `base_uri`: if provided, it will be parsed and values will override `protocol`, `port`, `server`, `path` and `fragment`. As an example, the `base_uri="http://myserver.com:8080/docs/x.html#anchor?a=1&b=My+String"` will split as `protocol=http`, `port=8080`, `server=myserver.com`, `path=/docs.x.html`,  `fragment=anchor` and `query=a=1&b=My+String`. Defaults to null.

Ports would be the same, a port `BASE_URI` would be split and replace other values.

Be aware that `query` is special as one may have multiple parameters of the same name. Like in HTML forms you may get `selected=a&selected=b` for select with multiple values. If user wants to reset, then use the `CLEAR` port before.

cabelitos:
@barbieri ack.

cabelitos:
@barbieri Should we add user and password as well?

barbieri:
the user:password in the URL is not a standard and is dangerous, so I'd say no. It's an extension some user agent support, but it should be in the header. We support that in our C API.

lpereira:
[**@barbieri**](https://github.com/barbieri) commented on [Oct 27, 2015, 7:10 PM GMT-2](https://github.com/solettaproject/soletta/pull/967#issuecomment-151645224 "2015-10-27T21:10:55Z - Replied by Github Comment Enhancer"):
> the user:password in the URL is not a standard and is dangerous, so I'd say no. It's an extension some user agent support, but it should be in the header. We support that in our C API.

Username and passwords are already supported by the C API. It had a bug, but I believe @otaviobp found and fixed it already.

cabelitos:
Yeah, he fixed it.

cabelitos:
Closing in favor of #1004

Issue #747, opened 2015-09-17
LCD (String) + Buttons high-level node types


barbieri:
It is common to have IoT products with LCD display with buttons (arrows) and need to ask for user input. We should offer some high-level node types to deal with that, for instance a "selector" as required by #732 (see that issue for description on a suggested node type).

Generally speaking these node types should format a multi-line string to be used by other node types such as the `grove/lcd-string`. This is done by taking the number of rows and columns in that display and doing the formatting. 3 input ports should be used to control the component (like: previous/up, next/down and ok/select).

At least the following components should exist:
 * **Boolean:** enter true or false (on/off, labels to come as options).
 * **Integer:** enter an integer value. Input ports increase/decrease the value within a range.
 * **Integer-Large:** similar to integer above, but meant for large numbers like years. This should use 2 more ports to change the position, while increase/decrease ports will affect only that digit at given position.
 * **String:** enter a string. Similar to *integer-large* above, two ports to change the position, two to cycle chars. An option should specify the characters to allow, defaults to ASCII printable.
 * **String-formatted:** similar to string, this takes a format and allows editing "fields" from it. For instance a format could be "{3d:0,255,1}.{3d:0,255,1}.{3d:0,255,1}.{3d:0,255,1}" to enter an IP address or "{2x}:{2x}:{2x}:{2x}:{2x}:{2x}" to edit MAC Address. The node would only edit the fields contained inside it, skipping the formatting such as "." and ":" in above examples. This could be the base for some helpers such as timestamp editor.
 * **Selector:** selects from a list of items, similar to HTML's `<select>`

Nodes should be similar in ports and options. From #732 we could take the proposed interface for **selector**:
   * Options:
       * `columns` (int): the number of columns in the display. Will be used to format the string, including auto-scroll.
       * `rows` (int): the number of rows in the display. If more than 1, will wrap lines.
       * `title` (string): if more than one row, this will be the contents of the first row. Otherwise will be the prefix of the only row in the string.
       * `auto_scroll` (boolean): if true and the contents doesn't fit in the columns, then each row will scroll.
       * `circular` (boolean): if true, this is a circular menu, after the last option it goes back to the first.
   * Input Ports:
       * `ITEMS` (custom packet): the options to be selected.
       * `NEXT` (any): change to the next item.
       * `PREVIOUS` (any): change to the previous item.
       * `SELECT` (any): select the current item.
[...]

PR #971, opened 2015-10-27
form: new int form node, with tests.

glima:

### DIFF
--- a/src/modules/flow/form/form.c
+++ b/src/modules/flow/form/form.c
@@ -31,6 +31,7 @@
  */
 
 #include <errno.h>
+#include <stdio.h>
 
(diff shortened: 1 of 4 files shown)


bdilly:
I've done a few comments, but it seems good overall.

I miss an way to add items, not only using range. Let say, you may want to set the time limit of a foosball match to be 3, 5 or 10 minutes. It wouldn't be possible to do it with form/int.
But I agree this implementation satisfies what was required on issue #747

barbieri:
It Would be a selector, not an int editor. We could add a int selector, but it's not top prio

glima:
Sending v2 now. It turned out I now accept initial values.

Issue #677, opened 2015-09-02
linux-micro debug support (gdbserver)


barbieri:
linux-micro will need a special trick to be debugged: we need to check debug is needed (ie: kernel command line `sol-debug=1`) and in that case we check we're PID1, if so we need to `fork()` and keep a minimal and safe PID1, doing all the work from the child. We should also check if `gdbserver` exists and attach it to the child (or spawn the child using `gdbserver`).

dorileo:
implemented on 358df2f03d929c75c076142b1e8a547eff55f1fa

PR #972, opened 2015-10-27

dorileo:

by the gdbserver and if not provided /dev/tty0 is assumed as default.

### DIFF
--- a/src/lib/common/sol-platform-impl-linux-micro.c
+++ b/src/lib/common/sol-platform-impl-linux-micro.c
@@ -54,6 +54,13 @@
 
 #include "sol-platform-linux-micro-builtins-gen.h"
 
+#define SOL_DEBUG_ARG "sol-debug=1"
+#define SOL_DEBUG_COMM_ARG "sol-debug-comm="
+struct gdbserver {
+    pid_t pid;
+    const char *comm;
+};
+
 static enum sol_platform_state platform_state = SOL_PLATFORM_STATE_INITIALIZING;
 static int reboot_cmd = RB_AUTOBOOT;
 static const char *reboot_exec;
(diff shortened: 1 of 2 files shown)


src/lib/common/sol-platform-impl-linux-micro.c
```diff
@@ -514,16 +521,108 @@ teardown_pid1(void)
...
+    char **argv = sol_argv();
+    bool res = false;
+
+    for (idx = 1; idx < argc; idx++) {
+        if (streq(argv[idx], SOL_DEBUG_ARG))
+            res = true;
+        else if (strstartswith(argv[idx], SOL_DEBUG_COMM_ARG)) {
+            gdb->comm = argv[idx] + strlen(SOL_DEBUG_COMM_ARG);
+            if (res) {
+                SOL_DBG("Got "SOL_DEBUG_ARG" flag and comm");
```
barbieri: print the comm you're going to use.

glima:
Fix the bot checks. Apart from what @barbieri said, it looks like it'll work fine for me.

src/lib/common/sol-platform-impl-linux-micro.c
```diff
@@ -514,16 +521,108 @@ teardown_pid1(void)
...
+            }
+        } else if (!(st.st_mode & S_IXUSR)) {
+            SOL_DBG("gdbserver executable at: %s - is not executable", paths[i]);
+            sol_platform_linux_fork_run_exit(EXIT_FAILURE);
+            continue;
+        }
+
+        gdbserver = paths[i];
+        break;
+    }
```
barbieri: coverity will bitch about  this `stat()` loop, at least it did for other cases like that. I guess we can try to exec them directly, if fail, then we try the next one.

src/lib/common/sol-platform-impl-linux-micro.c
```diff
@@ -514,16 +521,108 @@ teardown_pid1(void)
...
+        sol_platform_linux_fork_run_exit(EXIT_FAILURE);
+    }
+
+    err = snprintf(pids, sizeof(pids), "%d", gdb->pid);
+    if (err < 0 || err >= (int)sizeof(pids)) {
+        SOL_ERR("Could not format pid string to attach gdbserver");
+        sol_platform_linux_fork_run_exit(EXIT_FAILURE);
+    }
+
+    if (execl(gdbserver, gdbserver, "--attach", gdb->comm, pids, (char *)0) == -1) {
```
barbieri: use `NULL` as last, we have it available.

src/lib/common/sol-platform-impl-linux-micro.c
```diff
@@ -514,16 +521,108 @@ teardown_pid1(void)
...
 }
 
+static bool
+sol_platform_linux_micro_should_debug(struct gdbserver *gdb)
+{
+    int idx, argc = sol_argc();
+    char **argv = sol_argv();
+    bool res = false;
+
+    for (idx = 1; idx < argc; idx++) {
```
barbieri: are you sure this works for PID1? I guess we're called without argvs... in that case we'd need to get it from kcmdline.

src/lib/common/sol-platform-impl-linux-micro.c
```diff
@@ -514,16 +521,108 @@ teardown_pid1(void)
...
         int err = setup_pid1();
         SOL_INT_CHECK(err, < 0, err);
 
         want_load_initial_services = true;
+
+        gdb.pid = pid;
+        gdb.comm = "/dev/tty0";
+
+        if (sol_platform_linux_micro_should_debug(&gdb))
+            sol_platform_linux_fork_run(on_fork, on_fork_exit, &gdb);
```
barbieri: hum... I didn't test, but I did think it would not work. Did you confirm it's working?

Usually if you're PID1 you have some restrictions and I thought PTRACE would be forbidden, also not sure what happens if we crash it somehow.

What I initially envisioned was to keep PID1 as a very slim process that spawned the gdbserver and soletta application itself. It would `setenv("SOL_LOAD_INITIAL_SERVICES=1")` to force the application to do its stuff.

We need to be careful to have this slim process to setup the system, such as mounting `/proc`, `/dev`... (that's what `setup_pid1()` does) but not more than that, otherwise we may lose important code paths (ie: one may want  to debug a linux-micro initial service).

The other bit to do in the slim process is to handle SIGCHLD to `waitpid()`... as before, we should reduce code paths, thus no use of main loop stuff, just do a loop with `waitpid()` and it should be good.

summary:

```c
if (pid == 1 && getppid() == 0) {
   int err = setup_pid1();
   SOL_INT_CHECK(err, < 0, err);

   want_load_initial_services = true;
   gdb.comm =  "/dev/tty0";
   if (should_debug(&gdb)) {
      pid_t gdbserver_pid = fork();
      if (gdbserver_pid == 0) {
          args = alloca((_argc + 2) * sizeof(char*));
          args[0] = gdbserver_path;
          args[1] = gdb.comm;
          memcpy(args + 2, _argv, _argc * sizeof(char *));
          setenv("SOL_LOAD_INITIAL_SERVICES=1"); // so the re-execution will still run the services
          execv(gdbserver_path, args); // replace this with a loop that tries all paths. call alloca() BEFORE the loop!
          // does not return. should exit(EXIT_FAILURE)
      } else if (gdbserver_pid > 0) {
          // we enter a minimal system, do not return, just keep collecting zombie processes that goes to pid1
          while (true) {
             int status;
             pid_t child = wait(&status);
             // error handling
            SOL_DBG("child pid=%"PRIu64" exit=%d", (uint64_t)child, status);
          }
          // does not return
      } else { // error handling. I'd just SOL_WRN() and keep going without gdbserver }
   }
}
```

Another reason to do so is to avoid race conditions. If you call `gdbserver --attach PID`, then you may lose important debug paths while gdbserver starts. The only way to avoid them is to have gdbserver to start the process. Then what we do is to re-exec by taking our argv.

The trick is to set `SOL_LOAD_INITIAL_SERVICES=1` so the application will still execute linux micro services even not being pid1.

src/lib/common/sol-platform-impl-linux-micro.c
dorileo: ok

src/lib/common/sol-platform-impl-linux-micro.c
dorileo: ok

src/lib/common/sol-platform-impl-linux-micro.c
dorileo: No, I haven't tested as PID1 yet, I just read the kernel source code and it will pass the "after dash" args as argv.... anyway, it's worth testing and make sure it works somehow.

src/lib/common/sol-platform-impl-linux-micro.c
dorileo: I considered this scenario - mostly due to the spec you wrote on the issue #677 - also discussed with some guys like @glima, and we realized that for that "having a slim pid1 process..." we would need to add some pid1 and forking knowledge to the mainloop and it would be cumbersome for cases of platforms like contiki and riot since they don't have a multi process notion - we would also need to expose a non-linux-specific (I mean a generic api to wrap and fallback to specific platform's implementation) fork_run api and so on, that's the why I moved on with the current approach.

I'm testing this code right now as PID1 to make sure that it works - at least.

glima:
The fact that @dorileo assured me that the server would freeze the process when it attaches was a strong point, but indeed, we know nothing about how much time it will take to the fork to complete (and how much execution of Soletta itself we can lose).

dorileo:
ok @barbieri I got it, the most relevant bits here is the while() to wait for the gdbserver. It's ok, sounds reasonable.

src/lib/common/sol-platform-impl-linux-micro.c
barbieri: As I showed you don't need any fork knowledge in the core, it's all inside linux micro init. What you do is never return from linux micro's platform init, keeping the `wait()` (or `waitpid()`) loop forever. Everything else is handled by the new application that is executed from gdbserver.

dorileo: ok, got it.

dorileo:
v2 coming...

PR #973, opened 2015-10-27
Use "irange-spec" and "int" options

bdilly:
Integer part of task #724.

There are still some node types to be converted / tested,
but the ammount of modules converted is enough to validate
the work done.


glima:
+1

bdilly:
merged

PR #974, opened 2015-10-28
http:client: Fix PARAM_POST_DATA usage

ceolin:
SOL_HTTP_PARAM_POST_DATA is not a key/value and should not be treat as
one.


cabelitos:
+1, however you need to fix the semaphore

ceolin:
semaphore is complaining with something not related with this patch, i guess something wrong happened there.

ceolin:
merged, thanks!

PR #975, opened 2015-10-28
str-slice: implement trailing + leading whitespace helpers

dorileo:
## Changes ##
  - fixed the issues pointed by @lpereira on PR #960;
  - added the underflow test suggested by @lpereira  on PR #960;
  - added ```sol_str_slice_trim()``` function so once can remove trailing and leading white space in a single call;

## Notes ##
This patch was got from my ```linux-micro``` series so we can individually integrate this one so @cabelitos can use in his work.

## Rationale ##
This patch introduces the sol_str_slice_remove_leading_whitespace() and
sol_str_slice_remove_trailing_whitespace() helpers.

### DIFF
--- a/src/lib/datatypes/include/sol-str-slice.h
+++ b/src/lib/datatypes/include/sol-str-slice.h
@@ -32,6 +32,7 @@
 
 #pragma once
 
+#include <ctype.h>
 #include <stdbool.h>
 #include <stdlib.h>
 #include <string.h>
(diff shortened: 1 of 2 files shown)


lpereira:
LGTM. (I'd use an equality comparison instead of `if (slice.len - copy.len)`, but since `copy.len` will never be larger than `slice.len`, this seems fine.)

dorileo:
ok guys, thank you. applied.

PR #976, opened 2015-10-28
generator: it depends on flow, so enforce this dep

dorileo:
The generator code needs the flow support, it will not compile if flow
is disabled since it requires sol-flow-static.h and sol-flow-metatype.h
due the use of SOL_FLOW_NODE_PORT_ERROR_NAME, SOL_FLOW_NODE_PORT_ERROR
and so on.


ibriano:
Correct

PR #977, opened 2015-10-28
samples: c-api samples should depend on flow

dorileo:
Since they use the flow api it doesn't make any sense to build it if
we've disabled flow support.


ibriano:
+2

dorileo:
thanks... applied.

PR #978, opened 2015-10-28
Pin map v3

mbelluzzo:
Changes from **v2**:
* Remove forgotten debug
* Fix json indentation to match it's file
* Change calamari 7seg and rgb led options from int to string to match child node options
  * Solves either @edersondisouza leak and @glima asprintf suggestion
 
Changes from **v1**:
* Changed 'label_open' to 'open_by_label' on API
* 'Un-bump' the version of pin-mux struct
* Fix log_init
* Use the new API on hw io nodes (and fix the nodes that used this as sub-node)
 * Fix conf files, samples and tests


mbelluzzo:
Merging!

PR #979, opened 2015-10-28
HTTP get-json node. v2

cabelitos:
Changes since v1:

* using @dorileo  sol_str_slice_trim()
* s/url/URL/ in http-client.json
* Added one commit fixing a type in sol-flow-composed.c


bdilly:
It's good overall, just a few comments.

cabelitos:
merged with @bdilly suggestions.

PR #980, opened 2015-10-28
Fix FBP_GENERATOR dependency to correct name

otaviobp:
Building soletta with generator support was not possible because
generator dependency was incorrect


otaviobp:
merged with fix proposed by ibriano

PR #982, opened 2015-10-28
flow: options parser shouldn't allow more subfields than expected

bdilly:
It fixes such issue for subfields written in linear form, like:
 value=0|10|10|100

Issue remains for named subfields, like:
 value=min:10|max:50|invalid:100

Where extra fields can be written, or wrong subfields. In both
cases such keys are just ignored and default values are used
for subfields not set.

@glima please take a look on that

### DIFF
--- a/src/lib/flow/sol-flow-node-options.c
+++ b/src/lib/flow/sol-flow-node-options.c
@@ -98,6 +98,7 @@ get_member_memory(const struct sol_flow_node_options_member_description *member,
             if (errno != 0) goto err;                                   \
             field_cnt++;                                                \
             *end = backup;                                              \
+            start = NULL; \
             if (backup == '\0') break;                                  \
             start = end + 1;                                            \
             if (!start) break;                                          \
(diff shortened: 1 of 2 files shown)


glima:
```
error: too many arguments to function ‘mosquitto_loop_read’
     r = mosquitto_loop_read(mqtt->mosq, 1);
```

is the bot crazy or we got mosquitto version not specified?

bdilly:
@glima @anselmolsm tried to install libmosquito but bot's version was too old and we lacked proper version checks.

pr #985 fixes that. thanks

bdilly:
@glima we don't have testes for named subfields, probably, because we don't detect the error =)

bdilly:
merged

glima:
Please consider detecting, then :P

bdilly:
@glima the only way I see to do it is re-writing the way named subfields are parsed. We don't cycle between them matching knows keys, we just look for know keys in the string, one by one. We don't have a way to verify if all of them were picked, etc.

PR #983, opened 2015-10-28
build: fix dependency chain for Soletta objects

glima:
The objects were depending on *all headers*, according to the PRE_GEN
makefile variable, which is not needed there. We leave it only to the
general default_target rule.

There was another serious dependency bug too: the .dep files, listing
specific dependencies for each module, were dangling in the
Makefile.rules file. Now each module object depends on them again, and
the specific headers for the object are checked for dependency, as it
should be.

Test plan: build till the end, touch some flow module's .json file,
build again. Only the module's .o should be rebuilt and, if it's
built-in, Soletta lib's .so and the rest of the modules are them
re-linked only.

Thanks to Dorileo for the help.


ibriano:
Running `make check` straight breaks with this.

glima:
True, let's see...

glima:
v2 coming

PR #984, opened 2015-10-28
Add a http + json sample - Beta

otaviobp:
This pull req depends on #967 pull req.
I added a commit in the list from an older version of UrlBuilder, just to make sure the samples will run, but I won't merge this while UrlBuilder branch is not merged.

### DIFF
--- a/src/lib/comms/sol-http-client-impl-curl.c
+++ b/src/lib/comms/sol-http-client-impl-curl.c
@@ -599,7 +599,7 @@ set_auth_basic(CURL *curl, struct sol_arena *arena,
 
     if (!user || !password)
         return false;
-    if (curl_easy_setopt(curl, CURLOPT_HTTPAUTH, CURLAUTH_BASIC | CURLAUTH_ONLY) != CURLE_OK)
+    if (curl_easy_setopt(curl, CURLOPT_HTTPAUTH, CURLAUTH_BASIC) != CURLE_OK)
         return false;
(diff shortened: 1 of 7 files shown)


otaviobp:
I was not sure what was the best way to get AppID from user. We discussed some days ago and it looked that having a port for was not the best way to go and I had forget that it was possible to export OPTIONS.
In a final product the AppID is constant in the code, so I don't think that using persistence to keep it is a good idea either. 
I think I'll use the conffile approach.

barbieri:
Indeed the conffile would be more real for a product, but it requires users changing git-controlled files OR exporting an envvars. That's why persistence is better for the sample.


With some thinking we could support both, having an app_id(constant:string) that has its option exported. If the default value is not changed, then filter out the propagation of the constant string (do not forward ""). This way real deploys could use conffile + options. 

And you export the port where app_id is connected. This way user can connect a persistence. 

Would be also nice to document the node type API somehow, maybe in the README.

And another thing we're missing is to get a JSON for a metatype. This way if user want to know exported options and ports he could call a tool and get the same format as we have for C nodes. Useful to tools such as Soletta-dev-app, the doc generator... I talked to @cabelitos or @bottazzini to change sol-fbp-generator to take an extra parameter to output the metatypes as a JSON file such as those we install in the system.

bottazzini:
?? how are you?

2015-10-29 13:31 GMT+01:00 Gustavo Sverzut Barbieri <
notifications@github.com>:

PR #985, opened 2015-10-28
mqtt: Fix dependency check for libmosquitto

rchiossi:
mosquitto_lib_init() is not enough to check for the proper version of
libmosquitto since it has the same signature in incompatible versions.

mosquitto_opts_set is a new API added in mosquitto 1.4. All versions
after 1.4 are compatible with soletta.


bdilly:
+2

worked fine for 1.4.3 and didn't tried to build with mosquitto support for 0.15 =)

PR #987, opened 2015-10-28

dorileo:
## Changes ##
  - v2 (since #972)

## Rationale ##

by the gdbserver and if not provided /dev/tty0 is assumed as default.

### DIFF
--- a/src/lib/common/sol-platform-impl-linux-micro.c
+++ b/src/lib/common/sol-platform-impl-linux-micro.c
@@ -54,6 +54,14 @@
 
 #include "sol-platform-linux-micro-builtins-gen.h"
 
+#define SOL_DEBUG_ARG "sol-debug=1"
+#define SOL_DEBUG_COMM_ARG "sol-debug-comm="
+
+struct gdbserver {
+    pid_t pid;
+    char *comm;
+};
+
 static enum sol_platform_state platform_state = SOL_PLATFORM_STATE_INITIALIZING;
 static int reboot_cmd = RB_AUTOBOOT;
 static const char *reboot_exec;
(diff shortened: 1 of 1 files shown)


src/lib/common/sol-platform-impl-linux-micro.c
```diff
@@ -514,16 +522,137 @@ teardown_pid1(void)
...
+
+    file = sol_file_reader_get_all(fr);
+    // remove the \n in the end of the file (this is a single line file - always)
+    file.len--;
+    tokens = sol_util_str_split(file, " ", 0);
+    SOL_VECTOR_FOREACH_IDX (&tokens, token, i) {
+        char arg[PATH_MAX] = { 0 };
+        int err;
+
+        err = snprintf(arg, sizeof(arg), "%.*s", SOL_STR_SLICE_PRINT(*token));
```
barbieri: this is not required, use `sol_str_slice_eq()` (instead of a macro with a string, use a str slice) and instea of `strdup()` use `sol_str_slice_to_string()`.

src/lib/common/sol-platform-impl-linux-micro.c
```diff
@@ -514,16 +522,137 @@ teardown_pid1(void)
...
 
         want_load_initial_services = true;
+
+        gdb.pid = pid;
+        gdb.comm = NULL;
+
+        if (sol_platform_linux_micro_should_debug(&gdb)) {
+            struct sol_platform_linux_fork_run *fork;
+
+            fork = sol_platform_linux_fork_run(on_fork, on_fork_exit, &gdb);
```
barbieri: this is not required, it will do some stuff you don't need, and while it seem to help, it doesn't much and it will introduce more code paths (like malloc and adding a child to be monitored by the main loop -- that is not going to run).

do the fork yourself, no need for synchronization pipes, but I did forget to reset the sigprocmask in my version. This way we reduce code paths for this case.

src/lib/common/sol-platform-impl-linux-micro.c
```diff
@@ -514,16 +522,137 @@ teardown_pid1(void)
...
+            if (!fork) {
+                SOL_ERR("Could not launch gdbserver - %s",
+                    sol_util_strerrora(errno));
+            } else {
+                while (true) {
+                    int status;
+                    pid_t child = wait(&status);
+
+                    SOL_DBG("child pid=%" PRIu64 " status=%d",
```
barbieri: I think you should check for error or at least initialize status to zero, otherwise it may be undefined.

src/lib/common/sol-platform-impl-linux-micro.c
```diff
@@ -514,16 +522,137 @@ teardown_pid1(void)
...
+                        (uint64_t)child, status);
+                }
+                free(gdb.comm);
+                gdb.comm = NULL;
```
barbieri: this part should never be reached. You can `free(gdb.comm)` before doing the `while (true)` and put a `__builtin_unreachable()` (gcc >=4.5) after that... or an `exit(EXIT_FAILURE)`.

src/lib/common/sol-platform-impl-linux-micro.c
```diff
@@ -514,16 +522,137 @@ teardown_pid1(void)
...
+    unsigned int i;
+    char pids[PATH_MAX];
+    struct gdbserver *gdb = data;
+
+    const char *paths[] = {
+        "/usr/bin/gdbserver",
+        "/bin/gdbserver",
+    };
+
+    err = snprintf(pids, sizeof(pids), "%d", gdb->pid);
```
barbieri: no idea why you opted to keep the `--attach $PID`, it's not going to work. As I explained, there may be problems PTRACE PID1, you're missing debug paths as the gdbserver is going to take a while to start.

Actually I doubt this patch will work. Even if PTRACE of PID1 succeeds, the main process, the one that you're attaching to, is in the busy loop `while (true) { ... wait() ... }`. Then nothing to be debugged.

Really, go with the `gdbserver $COMM $ARGS`. It is the only way this is going to work.

src/lib/common/sol-platform-impl-linux-micro.c
```diff
@@ -514,16 +522,137 @@ teardown_pid1(void)
...
+    if (err < 0 || err >= (int)sizeof(pids)) {
+        SOL_ERR("Could not format pid string to attach gdbserver");
+        sol_platform_linux_fork_run_exit(EXIT_FAILURE);
+    }
+
+    if (setenv("SOL_LOAD_INITIAL_SERVICES", "1", 1) == -1) {
```
barbieri: just as a note, the way it is right now, attaching to an existing process, the `setenv()` has no use. It's going to be propagated to `gdbserver` process, that is not going to use it... as it's not spawing/exec a new process!

You must keep the `setenv()` as you will need to change to `gdbserver $COMM $ARGS` as I said below, removing the `--attach $PID`. But I wanted to ensure you understand the scope and usage of the envvar.

src/lib/common/sol-platform-impl-linux-micro.c
dorileo: Yep,  this is in the path of things I still need to test like I mentioned.

src/lib/common/sol-platform-impl-linux-micro.c
dorileo: Huuum,  ok makes sense,  just wanted to reused the most I could,  but ok this is an exception.

src/lib/common/sol-platform-impl-linux-micro.c
dorileo: Ok

src/lib/common/sol-platform-impl-linux-micro.c
dorileo: Yep,  this is also in the path of things I still need to test,  I just included the busy loop to not have this thing missing in the patch but tested without that,  so I was clearly running gdbserver attached to the main process,  ok I got it,  I plan to change that as soon as I can test it.  About the ptrace I couldn't reach that condition -  at least not yet. I intend to finish that and have the version closer to all your propositions by tomorrow.  Thanks for reviewing.

src/lib/common/sol-platform-impl-linux-micro.c
dorileo: Yes,  I noticed that as well,  like above this was included just an answer to your review but I'm sure it's useless at this point.  I see also a need to a SOL_PID1_DEBUG or sort of to avoid the pid1 setup at that point.

src/lib/common/sol-platform-impl-linux-micro.c
dorileo: Huuum,  ok,  missed that one.

barbieri: ah, and one idea would be to turn this `fork() -> exec(gdbserver)` into a loop. Then if gdbserver crashes or finishes we restart it. I guess it's more useful than sit in a `wait()` forever (or constant error if there are no more PIDs to wait).

```c
static void
gdb_wait(pid_t gdb_pid)
{
    while (true) {
        int status;
        pid_t child = wait(&status);
        if (child < 0) {
            if (errno == EINTR)
                continue;
            SOL_WRN("wait() failed: %s", sol_util_strerrnoa(errno));
            return; // no more child process, they are all dead, restart gdbserver -- it is just to be complete, it should never happen as we check if gdbserver is dead to restart it
        } else {
            SOL_DBG("child pid=%" PRIu64 " status=%d", (uint64_t)child, status);
            if (child == gdb_pid) {
                SOL_INF("gdbserver exited, restart it");
                return;
            }
        }
    }
}

static void
gdb_exec(const char *gdb_comm)
{
    const char **argv;
    int argv;
    const char **args;
    const char *paths[] = {
        "/usr/bin/gdbserver",
       "/bin/gdbserver",
    };

    if (setenv("SOL_LOAD_INITIAL_SERVICES", "1", 1) == -1) {
        SOL_ERR("Could not set SOL_LOAD_INITIAL_SERVICES");
        _exit(EXIT_FAILURE);
    }

    argc = sol_argc();
    argv = sol_argv();
    if (argc < 1 || !argv || !argv[0]) {
        SOL_ERR("Invalid argc=%d, argv=%p, argv[0]=%p", argc, argv, argv ? argv[0] : NULL);
        _exit(EXIT_FAILURE);
    }

    args = alloca((argc + 3) * sizeof(char *));
    args[1] = gdb_comm;
    memcpy(args + 2, argv, argc * sizeof(char *));
    args[argc + 2] = NULL;

    for (i = 0; i < ARRAY_SIZE(paths); i++) {
        args[0] = paths[i];
        if (execv(paths[i], args) == -1)
            SOL_DBG("failed to exec %s - %s", paths[i], sol_util_strerrora(errno)); // not an error, we'll fallback...
    }
    
   SOL_WRN("no gdb server found, execute the application without it");
   execv(argv[0], argv);
   SOL_CRIT("could not execute the application %s: %s", argv[0], sol_util_strerrora(errno));
   _exit(EXIT_FAILURE);
}

static void gdb_debug(const char *gdb_comm) SOL_ATTR_NORETURN;

static void
gdb_debug(const char *gdb_comm)
{
    while (true) {
        pid_t gdb_pid = fork();
        if (gdb_pid < 0) {
            SOL_WRN("could not fork(): %s", sol_util_strerrnoa(errno));
            sleep(1); // give the system some time to breath
        } else if (gdb_pid > 0) {
            gdb_wait(gdb_pid);
        } else {
            gdb_exec(gdb_comm);
        }
    }
}

int
sol_platform_impl_init(void)
{
...
char *gdb_comm;
if (should_gdb_debug(&gdb_comm)) // no need to prefix the static func names, not being used in this file
    gdb_debug(&gdb_comm);
...
}
```

dorileo:
ok, barbieri' v3 coming...

PR #991, opened 2015-10-29
Add oauth v1 authenticator

ceolin:
- Change the api that serve static files to use a namespace
- Add api to set static file for error's response
- OAuth v1 node


src/modules/flow/oauth/oauth.c
```diff
@@ -0,0 +1,596 @@
...
+        .api_version = SOL_HTTP_RESPONSE_API_VERSION,
+        .content = SOL_BUFFER_INIT_EMPTY,
+        .param = SOL_HTTP_REQUEST_PARAM_INIT,
+        .response_code = SOL_HTTP_STATUS_FOUND,
+    };
+
+    token = strstr(response->content.data, "oauth_token=");
+    token_secret = strstr(response->content.data, "oauth_token_secret=");
+
+    if (token_secret) {
```
lpereira: `token` might be NULL here and dereferenced in line 243.

src/modules/flow/oauth/oauth.c
```diff
@@ -0,0 +1,596 @@
...
+        token = strndup(token, token_secret - token - 1);
```
lpereira: This won't work if `oauth_token=` and `oauth_token_secret=` are in a different order.

src/modules/flow/oauth/oauth.c
```diff
@@ -0,0 +1,596 @@
...
+            break;
+        }
+    }
+
+    SOL_NULL_CHECK_GOTO(verifier, err);
+
+    req_data->request = request;
+    req_data->mdata = data;
+    if (!sol_http_param_add(&params,
+        SOL_HTTP_REQUEST_PARAM_POST_FIELD("oauth_token", token)) ||
```
lpereira: `token` is not NULL-checked here, is this OK?

src/modules/flow/oauth/oauth.c
```diff
@@ -0,0 +1,596 @@
...
+    struct v1_request_data *req_data = calloc(1, sizeof(struct v1_request_data));
+    SOL_NULL_CHECK_GOTO(req_data, err);
+
+    SOL_HTTP_PARAM_FOREACH_IDX (sol_http_request_get_params(request), value, idx) {
+        switch (value->type) {
+        case SOL_HTTP_PARAM_QUERY_PARAM:
+            if (streq(value->value.key_value.key, "oauth_verifier")) {
+                verifier = strdup(value->value.key_value.value);
+            }
+            if (streq(value->value.key_value.key, "oauth_token")) {
```
lpereira: `else` missing?

src/modules/flow/oauth/oauth.c
```diff
@@ -0,0 +1,596 @@
...
+    free(req_data);
+}
+
+static char *
+get_url_callback(const char *namespace)
+{
+    char *temp, *url;
+    int r;
+
+    r = asprintf(&temp, "http://localhost:%d/%s/oauth_callback",
```
lpereira: How is this supposed to work? I mean -- it'll work fine on your machine, but how is this going to work if, say, Soletta is running on a board, and I'm going through OAuth steps there with my cell phone?

src/modules/flow/oauth/oauth.c
```diff
@@ -0,0 +1,596 @@
...
+generate_nonce(void)
+{
+    int r;
+    ssize_t size;
+    char *nonce = NULL;
+    struct sol_random *engine;
+    struct sol_str_slice slice;
+    struct sol_buffer buffer = SOL_BUFFER_INIT_EMPTY;
+
+    engine = sol_random_new(SOL_RANDOM_DEFAULT, 0);
```
lpereira: For nonces, I'd use `SOL_RANDOM_URANDOM`. This falls back to `MT19937` (the default) if `/dev/urandom` can't be opened for any reason (which, if available, will initialize the seed using `getrandom()` or `/dev/urandom`, but the seed has only 4 bytes).

lpereira: Also, after you're done with the nonce, I'd `memset()` it with `\0`, ensuring that the `memset()` will not get elided by the compiler due to optimizations. See my `tinydtls-v2` branch, there's a `clean_memory()` function that does this.

src/modules/flow/oauth/oauth.c
ceolin: oow i just forgot to change it. I've stashed the function with that get the right address

src/modules/flow/oauth/oauth.c
ceolin: nope

src/modules/flow/oauth/oauth.c
ceolin: yep, that's true. I just have check it with twitter and the response has this order, but other service can be different. Going to change

barbieri:
I Didn't review in depth, but split patch set so that static files parts are merged sooner.

The oauth comments related to URL building is valid, we should use one of our IP address, likely the one that is originating (i.e.: got a request from 192.168.0.1/eth0 or 10.0.0.3/wlan0, use that ip as callback)

We should also use proper functions to create and split URL and parameters, not strstr() as done. Sync with @cabelitos as he is doing such work for his node type.

ceolin:
about url, i've something stashed that uses the device' s ip to create the callback, i just forgot to change

from the strstr() side, i'm using it on service's response, i'll check with @cabelitos but it seems to be something different

barbieri:
A note related to static files is that we need to detect their mimetype to provide proper "Content-type" headers. There are libraries such as libmagic to do it, but we should offer a simpler fallback like checking extensions, fallback to octet-stream.

barbieri:
The response is an URL query parameter. Just comes in a different place (body)

ceolin:
fixes on https://github.com/solettaproject/soletta/pull/1005

PR #992, opened 2015-10-29
Random improvements

lpereira:
This PR improves the random number generator slightly: generates tighter code for the MT19937 PRNG, and ensures the state of a `sol_random` is cleaned up before its memory is freed.


bdilly:
Series seems ok. I've triggered a rebuild because first one was done while our bot was in a bad shape.

ibriano:
Ping.

lpereira:
Merged using the function in `sol-util.h` to securely clear memory that has been introduced as part of the DTLS series. Thanks for the review!

PR #994, opened 2015-10-29
form: new int form node, with tests. [v2]

glima:
Changes since #971:

- initial option now accepted for the node
- more range checks where applicable


bdilly:
:thumbsup:

glima:
Thanks, merging that fix.

PR #995, opened 2015-10-29
Persistence fixes

edersondisouza:
This PR replaces part of #966 - it does the 'simple fixes', so it can be merged faster.


glima:
+1

edersondisouza:
If no more comments appear, I'll merge with proposed fixed by the end of day.

ibriano:
Fine by me

edersondisouza:
Integrated.

PR #996, opened 2015-10-29
Async v6

edersondisouza:
This commit replaces the async part of #966 

v6:
Every write happens on timer, even on filesystem and efivars, so callbacks are required;
Fixed callback changing sol-vector on the fly;
Fixed timeout deletion;
Added structure to keep track of memmap internal data, like resolved path and timeout - this simplified some code;

v5:
Less memory allocation (and leaks)

v4:

v3:

v2:
Fixed pointed issues;


edersondisouza:
If no more comments appear, I'll merge with proposed fixed by the end of day.

edersondisouza:
Integrated.

PR #997, opened 2015-10-29

dorileo:
## Changes ##
  - v2 (since #972)
  - v3 (since #987)
    - changed the overall approach, instead of attaching to a process intead launch a new one with gdbserver, got almost all the pseudo code @barbieri posted in the last review, just adjusted a few minor issues, but the design is barely the same;

## Rationale ##

by the gdbserver and if not provided /dev/ttyS0 is assumed as default.

Tested with kvm, adding a serial chardev, and using it as target remote.

### DIFF
--- a/src/lib/common/sol-platform-impl-linux-micro.c
+++ b/src/lib/common/sol-platform-impl-linux-micro.c
@@ -54,6 +54,9 @@
 
 #include "sol-platform-linux-micro-builtins-gen.h"
 
+#define SOL_DEBUG_ARG "sol-debug=1"
+#define SOL_DEBUG_COMM_ARG "sol-debug-comm="
+
 static enum sol_platform_state platform_state = SOL_PLATFORM_STATE_INITIALIZING;
 static int reboot_cmd = RB_AUTOBOOT;
 static const char *reboot_exec;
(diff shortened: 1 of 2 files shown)


src/lib/common/sol-platform-impl-linux-micro.c
```diff
@@ -514,16 +517,141 @@ teardown_pid1(void)
...
+    SOL_VECTOR_FOREACH_IDX (&tokens, token, i) {
+        if (sol_str_slice_str_eq(*token, SOL_DEBUG_ARG)) {
+            res = true;
+        } else if (strstartswith(token->data, SOL_DEBUG_COMM_ARG)) {
+            size_t id_len = sizeof(SOL_DEBUG_COMM_ARG);
+            *gdb_comm = strndup(token->data + id_len, token->len - id_len);
+        }
+    }
+
+    if (!*gdb_comm) {
```
barbieri: maybe check `if (res && !*gdb_comm)`?

src/lib/common/sol-platform-impl-linux-micro.c
```diff
@@ -514,16 +517,141 @@ teardown_pid1(void)
...
+        SOL_ERR("No comm set, trying to set default one: /dev/ttyS0");
+        *gdb_comm = strdup("/dev/ttyS0");
+        if (!*gdb_comm) {
+            SOL_ERR("Could not allocate comm string memory, not debugging.");
+            res = false;
+        }
+    }
+
+    sol_file_reader_close(fr);
```
barbieri: we're leaking tokens vector, need to clear it, no?

barbieri:
+1 overall, just handle the case where `!res`. We shouldn't `strdup("/dev/ttyS0")`in that case.

and if we got a `gdb_comm` but `res == false` we should `free(*gdb_comm); *gdb_comm = NULL;` as it's useless, would just leak memory.

src/lib/common/sol-platform-impl-linux-micro.c
dorileo: yep

src/lib/common/sol-platform-impl-linux-micro.c
dorileo: yep true...

dorileo:
ok

PR #1000, opened 2015-10-29
Some changes to http-server node types

bdilly:
* allow port configuration
 * always send packets when state changes
 * add samples using multiple paths (and ports)

@ceolin please review it when possible

I'll add some tests covering them soon making use of
these features.


bdilly:
merged with proposed changes

PR #1001, opened 2015-10-29
string: Do not link against ICU and PCRE libs if they are disabled

edersondisouza:


ibriano:
Looks ok

edersondisouza:
Integrated.

PR #1002, opened 2015-10-29
build: Be more permissive about extra-{c,ld}flags

ibriano:
If we have an obj-something built as a module, and optional
dependencies that are always built-in, the mismatch in
obj-something-m-extra-cflags and obj-something-y-extra-cflags could
break the build.
Allow these dependencies to be added to the base obj without requiring
cumbersome Makefile conditionals.


bdilly:
@ibriano *if* it fixes the build (please check semaphoreci branch status before), please merge it.

dorileo:
ok

PR #1003, opened 2015-10-29
Build [v2]

glima:
Changes since #986:
- the second part of the fix is now made right: Makefile.gen does not depend on .config anymore.


dorileo:
good to go -- +2

glima:
Cool, thanks. I'll merge it soon.

PR #1004, opened 2015-10-30
Create a node to build URLs. v3

cabelitos:
Changes since v2:
* Added a more complex node that @barbieri suggested. (See suggestion at #967)

### DIFF
--- a/src/lib/comms/include/sol-http.h
+++ b/src/lib/comms/include/sol-http.h
@@ -239,7 +239,15 @@ bool sol_http_param_add(struct sol_http_param *params,
     struct sol_http_param_value value) SOL_ATTR_WARN_UNUSED_RESULT;
 void sol_http_param_free(struct sol_http_param *params);
 
-int sol_http_escape_string(char **escaped, const char *value);
+int sol_http_escape_string(char **escaped, const struct sol_str_slice value);
+
+int sol_http_create_uri(char **uri, const char *protocol, const char *server, const char *path, const char *fragment, int port, const struct sol_http_param *params);
+
+int sol_http_create_simple_uri(char **uri, const char *base_url, const struct sol_http_param *params);
+
+int sol_http_encode_params(char **encoded_params, enum sol_http_param_type type, const struct sol_http_param *params);
+
+int sol_http_split_uri(const char *uri, struct sol_str_slice *protocol, struct sol_str_slice *server, struct sol_str_slice *path, struct sol_str_slice *fragment, struct sol_http_param *query_params, int *port);
 
 /**
  * @}
--- a/src/lib/comms/include/sol-http.h
+++ b/src/lib/comms/include/sol-http.h
@@ -34,6 +34,7 @@
 
 #include <sol-buffer.h>
 #include <sol-vector.h>
+#include <sol-str-slice.h>
 
 #ifdef __cplusplus
 extern "C" {
@@ -101,12 +102,12 @@ struct sol_http_param_value {
     enum sol_http_param_type type;
     union {
         struct {
-            const char *key;
-            const char *value;
+            struct sol_str_slice key;
+            struct sol_str_slice value;
         } key_value;
         struct {
-            const char *user;
-            const char *password;
+            struct sol_str_slice user;
+            struct sol_str_slice password;
         } auth;
         struct {
             bool value;
@@ -170,8 +171,8 @@ struct sol_http_response {
     (struct sol_http_param_value) { \
         .type = type_, \
         .value.key_value = { \
-            .key = (key_), \
-            .value = (value_) \
+            .key = sol_str_slice_from_str((key_)),      \
+            .value = sol_str_slice_from_str((value_))   \
         } \
     }
 
@@ -191,8 +192,8 @@ struct sol_http_response {
     (struct sol_http_param_value) { \
         .type = SOL_HTTP_PARAM_AUTH_BASIC, \
         .value.auth = { \
-            .user = (username_), \
-            .password = (password_) \
+            .user = sol_str_slice_from_str((username_)),        \
+            .password = sol_str_slice_from_str((password_))     \
         } \
     }
 
--- a/src/lib/comms/sol-http-client-impl-curl.c
+++ b/src/lib/comms/sol-http-client-impl-curl.c
@@ -86,14 +86,6 @@ struct sol_http_client_connection {
 static void
 destroy_connection(struct sol_http_client_connection *c)
 {
-    uint16_t i;
-    struct sol_http_param_value *param;
-
-    SOL_HTTP_PARAM_FOREACH_IDX (&c->response_params, param, i) {
-        curl_free((char *)param->value.key_value.key);
-        curl_free((char *)param->value.key_value.value);
-    }
-
     curl_multi_remove_handle(global.multi, c->curl);
     curl_slist_free_all(c->headers);
     curl_easy_cleanup(c->curl);
@@ -402,7 +394,7 @@ header_cb(char *data, size_t size, size_t nmemb, void *connp)
     struct sol_http_client_connection *connection = connp;
     struct sol_http_param_value param;
     size_t data_size, key_size, discarted, i, cookie_name_size;
-    char *sep;
+    char *sep, *key, *value;
     int r;
 
     r = sol_util_size_mul(size, nmemb, &data_size);
@@ -431,33 +423,29 @@ header_cb(char *data, size_t size, size_t nmemb, void *connp)
         param.type = SOL_HTTP_PARAM_COOKIE;
         cookie_name_size = 0;
         while (sep[cookie_name_size++] != '=') ;
-        param.value.key_value.key = curl_easy_unescape(connection->curl,
-            sep, cookie_name_size - 1, NULL);
+        key = sol_arena_strndup(connection->arena, sep, cookie_name_size - 1);
         sep += cookie_name_size + 1;
         discarted += cookie_name_size + 1;
     } else {
         param.type = SOL_HTTP_PARAM_HEADER;
-        param.value.key_value.key = curl_easy_unescape(connection->curl,
-            data, key_size, NULL);
+        key = sol_arena_strndup(connection->arena, data, key_size);
     }
-    SOL_NULL_CHECK(param.value.key_value.key, 0);
+    SOL_NULL_CHECK(key, 0);
 
-    param.value.key_value.value = curl_easy_unescape(connection->curl, sep,
-        data_size - key_size - discarted, NULL);
-    SOL_NULL_CHECK_GOTO(param.value.key_value.value, err_value);
+     value = sol_arena_strndup(connection->arena, sep,
+        data_size - key_size - discarted);
+    SOL_NULL_CHECK_GOTO(value, err_exit);
 
+    param.value.key_value.key = sol_str_slice_from_str(key);
+    param.value.key_value.value = sol_str_slice_from_str(value);
     if (!sol_http_param_add(&connection->response_params, param)) {
-        SOL_ERR("Could not add the http param - key:%s:value%s",
-            param.value.key_value.key, param.value.key_value.value);
-        goto err_add;
+        SOL_ERR("Could not add the http param - key: %s value: %s", key, value);
+        goto err_exit;
     }
 
     return data_size;
 
-err_add:
-    curl_free((char *)param.value.key_value.value);
-err_value:
-    curl_free((char *)param.value.key_value.key);
+err_exit:
     return 0;
 }
 
@@ -560,18 +548,18 @@ encode_key_values(CURL *curl, enum sol_http_param_type type,
     }
 
     SOL_VECTOR_FOREACH_IDX (&params->params, iter, idx) {
-        const char *key = iter->value.key_value.key;
-        const char *value = iter->value.key_value.value;
+        struct sol_str_slice key = iter->value.key_value.key;
+        struct sol_str_slice value = iter->value.key_value.value;
         char *tmp, *encoded_key, *encoded_value;
         int r;
 
         if (iter->type != type)
             continue;
 
-        encoded_key = curl_easy_escape(curl, key, strlen(key));
+        encoded_key = curl_easy_escape(curl, key.data, key.len);
         if (!encoded_key)
             goto cleanup;
-        encoded_value = curl_easy_escape(curl, value, strlen(value));
+        encoded_value = curl_easy_escape(curl, value.data, value.len);
         if (!encoded_value) {
             curl_free(encoded_key);
             goto cleanup;
@@ -647,8 +635,8 @@ set_headers_from_params(CURL *curl, struct sol_arena *arena,
     uint16_t idx;
 
     SOL_VECTOR_FOREACH_IDX (&params->params, iter, idx) {
-        const char *key = iter->value.key_value.key;
-        const char *value = iter->value.key_value.value;
+        struct sol_str_slice key = iter->value.key_value.key;
+        struct sol_str_slice value = iter->value.key_value.value;
         struct curl_slist *tmp_list;
         char key_colon_value[512];
         char *tmp;
@@ -658,7 +646,7 @@ set_headers_from_params(CURL *curl, struct sol_arena *arena,
             continue;
 
         r = snprintf(key_colon_value, sizeof(key_colon_value),
-            "%s: %s", key, value);
+            "%.*s: %.*s", SOL_STR_SLICE_PRINT(key), SOL_STR_SLICE_PRINT(value));
         if (r < 0 || r >= (int)sizeof(key_colon_value))
             goto fail;
 
@@ -689,8 +677,8 @@ static bool
 set_auth_basic(CURL *curl, struct sol_arena *arena,
     const struct sol_http_param_value *value)
 {
-    char *user = sol_arena_strdup(arena, value->value.auth.user);
-    char *password = sol_arena_strdup(arena, value->value.auth.password);
+    char *user = sol_arena_strdup_slice(arena, value->value.auth.user);
+    char *password = sol_arena_strdup_slice(arena, value->value.auth.password);
 
     if (!user || !password)
         return false;
@@ -785,14 +773,13 @@ set_post_data_from_params(CURL *curl, struct sol_arena *arena,
 
     SOL_VECTOR_FOREACH_IDX (&params->params, iter, idx) {
         struct sol_str_slice value = SOL_STR_SLICE_EMPTY;
-
         if (iter->type == SOL_HTTP_PARAM_POST_FIELD) {
             SOL_WRN("SOL_HTTP_PARAM_POST_FIELD and SOL_HTTP_PARAM_POST_DATA found in parameters."
                 " Only one can be used at a time");
             return false;
         } else if (iter->type == SOL_HTTP_PARAM_HEADER) {
-            const char *key = iter->value.key_value.key;
-            type_set = type_set || !strcasecmp(key, "content-type");
+            struct sol_str_slice key = iter->value.key_value.key;
+            type_set = type_set || !strncasecmp(key.data, "content-type", key.len);
         } else if (iter->type == SOL_HTTP_PARAM_POST_DATA) {
             value = iter->value.data.value;
             if (data.len != 0) {
--- a/src/lib/comms/sol-http-client-impl-curl.c
+++ b/src/lib/comms/sol-http-client-impl-curl.c
@@ -531,101 +531,6 @@ free_buffer:
     return NULL;
 }
 
-static char *
-encode_key_values(CURL *curl, enum sol_http_param_type type,
-    const struct sol_http_param *params, char *initial_value)
-{
-    struct sol_http_param_value *iter;
-    uint16_t idx;
-    bool first = true;
-
-    if (type != SOL_HTTP_PARAM_QUERY_PARAM &&
-        type != SOL_HTTP_PARAM_POST_FIELD &&
-        type != SOL_HTTP_PARAM_COOKIE) {
-        free(initial_value);
-        errno = EINVAL;
-        return NULL;
-    }
-
-    SOL_VECTOR_FOREACH_IDX (&params->params, iter, idx) {
-        struct sol_str_slice key = iter->value.key_value.key;
-        struct sol_str_slice value = iter->value.key_value.value;
-        char *tmp, *encoded_key, *encoded_value;
-        int r;
-
-        if (iter->type != type)
-            continue;
-
-        encoded_key = curl_easy_escape(curl, key.data, key.len);
-        if (!encoded_key)
-            goto cleanup;
-        encoded_value = curl_easy_escape(curl, value.data, value.len);
-        if (!encoded_value) {
-            curl_free(encoded_key);
-            goto cleanup;
-        }
-
-        if (type == SOL_HTTP_PARAM_COOKIE) {
-            r = asprintf(&tmp, "%s%s%s=%s;", initial_value, first ? "" : " ",
-                encoded_key, encoded_value);
-        } else {
-            r = asprintf(&tmp, "%s%s%s=%s", initial_value, first ? "" : "&",
-                encoded_key, encoded_value);
-        }
-
-        curl_free(encoded_key);
-        curl_free(encoded_value);
-
-        if (r < 0)
-            goto cleanup;
-
-        free(initial_value);
-        initial_value = tmp;
-        first = false;
-    }
-
-    return initial_value;
-
-cleanup:
-    free(initial_value);
-    errno = ENOMEM;
-    return NULL;
-}
-
-static char *
-build_uri(CURL *curl, const char *base, const struct sol_http_param *params)
-{
-    char *initial_value;
-    char *built_uri;
-
-    if (asprintf(&initial_value, "%s?", base) < 0) {
-        errno = ENOMEM;
-        return NULL;
-    }
-
-    built_uri = encode_key_values(curl, SOL_HTTP_PARAM_QUERY_PARAM, params,
-        initial_value);
-    if (built_uri == initial_value) {
-        free(initial_value);
-        return strdup(base);
-    }
-
-    return built_uri;
-}
-
-static char *
-build_cookies(CURL *curl, const struct sol_http_param *params)
-{
-    return encode_key_values(curl, SOL_HTTP_PARAM_COOKIE, params, strdup(""));
-}
-
-static char *
-build_post_fields(CURL *curl, const struct sol_http_param *params)
-{
-    return encode_key_values(curl, SOL_HTTP_PARAM_POST_FIELD, params,
-        strdup(""));
-}
-
 static bool
 set_headers_from_params(CURL *curl, struct sol_arena *arena,
     const struct sol_http_param *params, struct curl_slist **headers)
@@ -734,9 +639,13 @@ static bool
 set_cookies_from_params(CURL *curl, struct sol_arena *arena,
     const struct sol_http_param *params)
 {
-    char *cookies = build_cookies(curl, params);
+    char *cookies;
+    int r;
 
-    if (cookies && !*cookies) {
+    r = sol_http_encode_params(&cookies, SOL_HTTP_PARAM_COOKIE, params);
+    SOL_INT_CHECK(r, < 0, false);
+
+    if (!*cookies) {
         free(cookies);
         return true;
     }
@@ -747,7 +656,11 @@ static bool
 set_uri_from_params(CURL *curl, struct sol_arena *arena, const char *base,
     const struct sol_http_param *params)
 {
-    char *full_uri = build_uri(curl, base, params);
+    char *full_uri;
+    int r;
+
+    r = sol_http_create_simple_uri(&full_uri, base, params);
+    SOL_INT_CHECK(r, < 0, false);
 
     return set_string_option(curl, CURLOPT_URL, arena, full_uri);
 }
@@ -756,7 +669,11 @@ static bool
 set_post_fields_from_params(CURL *curl, struct sol_arena *arena,
     const struct sol_http_param *params)
 {
-    char *post = build_post_fields(curl, params);
+    char *post;
+    int r;
+
+    r = sol_http_encode_params(&post, SOL_HTTP_PARAM_POST_FIELD, params);
+    SOL_INT_CHECK(r, < 0, false);
 
     return set_string_option(curl, CURLOPT_POSTFIELDS, arena, post);
 }
--- a/src/lib/comms/sol-http-common.c
+++ b/src/lib/comms/sol-http-common.c
@@ -77,19 +77,17 @@ sol_http_param_free(struct sol_http_param *params)
 }
 
 SOL_API int
-sol_http_escape_string(char **escaped, const char *value)
+sol_http_escape_string(char **escaped, const struct sol_str_slice value)
 {
     int r;
-    size_t value_len;
+    size_t i;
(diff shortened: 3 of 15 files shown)


cabelitos:
The semaphore is failing because of code style issues, they are fixed locally.

src/lib/comms/include/sol-http.h
```diff
@@ -238,7 +239,15 @@ bool sol_http_param_add(struct sol_http_param *params,
     struct sol_http_param_value value) SOL_ATTR_WARN_UNUSED_RESULT;
 void sol_http_param_free(struct sol_http_param *params);
 
-int sol_http_escape_string(char **escaped, const char *value);
+int sol_http_escape_string(char **escaped, const struct sol_str_slice value);
+
+int sol_http_create_uri(char **uri, const char *protocol, const char *server, const char *path, const char *fragment, int port, const struct sol_http_param *params);
```
barbieri: to make it more consistent, should we take all slices?

we can offer a helper:

```c
static inline int
sol_http_create_uri_from_str(char **uri, const char *protocol, const char *server, const char *path, const char *fragment, , int port, const struct sol_http_param *params)
{
    return sol_http_create_uri(uri, sol_str_slice_from_str(protocol), sol_str_slice_from_str(server), sol_str_slice_from_str(path), sol_str_slice_from_str(fragment ? fragment : ""), port, params);
} 
```

this assumes fragment may be `NULL`.

src/lib/comms/include/sol-http.h
```diff
@@ -238,7 +239,15 @@ bool sol_http_param_add(struct sol_http_param *params,
     struct sol_http_param_value value) SOL_ATTR_WARN_UNUSED_RESULT;
 void sol_http_param_free(struct sol_http_param *params);
 
-int sol_http_escape_string(char **escaped, const char *value);
+int sol_http_escape_string(char **escaped, const struct sol_str_slice value);
```
barbieri: I wonder if we could benefit from `sol_buffer` in these APIs.

In the case of `sol_http_escape_string()` we could do a trick similar to what we did in JSON, where we take an unintialized buffer, see if we need to escape and do it using default buffer flags and if no escape was needed, we set flags to `MEMORY_NOT_OWNED` and set buffer to `value` (data/len). At the end users should always use buffer length and then call `sol_buffer_fini()` to release memory if it was escaped.

Talk to @otaviobp to see if he think it may help.

src/lib/comms/sol-http-client-impl-curl.c
```diff
@@ -785,14 +690,13 @@ set_post_data_from_params(CURL *curl, struct sol_arena *arena,
...
-
         if (iter->type == SOL_HTTP_PARAM_POST_FIELD) {
             SOL_WRN("SOL_HTTP_PARAM_POST_FIELD and SOL_HTTP_PARAM_POST_DATA found in parameters."
                 " Only one can be used at a time");
             return false;
         } else if (iter->type == SOL_HTTP_PARAM_HEADER) {
-            const char *key = iter->value.key_value.key;
-            type_set = type_set || !strcasecmp(key, "content-type");
+            struct sol_str_slice key = iter->value.key_value.key;
+            type_set = type_set || !strncasecmp(key.data, "content-type", key.len);
```
barbieri: maybe we should have a `sol_str_slice_casecmp()` (and `sol_str_slice_caseeq()` on top, to be used here)?

And you should also check for size if you're using `strncasecmp()`, otherwise a shorter key will match:

`strncasecmp("content", "content-type", strlen("content")) == 0`, something we don't want.

src/lib/comms/sol-http-common.c
```diff
@@ -77,19 +77,17 @@ sol_http_param_free(struct sol_http_param *params)
...
     SOL_NULL_CHECK(escaped, -EINVAL);
 
-    value_len = strlen(value);
 
-    while (value_len--) {
-        unsigned char c = *value;
+    for (i = 0; i < value.len; i++) {
+        unsigned char c = value.data[i];
         switch (c) {
         case '0': case '1': case '2': case '3': case '4':
```
barbieri: ugh, not related to your patch, but this switch is ugly. In a future patch we could simplify it with `isalnum()` (0-9, a-z, A-Z) followed by the extra digits we allow, or we do `if ('0' <= c && c <= '9') ... ` explicitly... or we use the same trick as @lpereira did with JSON scanner that uses `strchr("0123456789abc..._~.-", c)`that will search if the string contains the current char. I like `isalnum()` better.

src/lib/comms/sol-http-common.c
```diff
@@ -123,3 +120,453 @@ sol_http_escape_string(char **escaped, const char *value)
...
+    int r;
+
+    SOL_NULL_CHECK(encoded_params, -EINVAL);
+    SOL_NULL_CHECK(params, -EINVAL);
+
+    if (type != SOL_HTTP_PARAM_QUERY_PARAM &&
+        type != SOL_HTTP_PARAM_POST_FIELD &&
+        type != SOL_HTTP_PARAM_COOKIE &&
+        type != SOL_HTTP_PARAM_HEADER) {
+        return -EINVAL;
```
barbieri: would be nice to `SOL_WRN("type %d is not supported", type)`

src/lib/comms/sol-http-common.c
```diff
@@ -123,3 +120,453 @@ sol_http_escape_string(char **escaped, const char *value)
...
+        if (iter->type != type)
+            continue;
+
+        r = sol_http_escape_string(&encoded_key, iter->value.key_value.key);
+        SOL_INT_CHECK_GOTO(r, < 0, clean_up);
+
+        r = sol_http_escape_string(&encoded_value, iter->value.key_value.value);
+        SOL_INT_CHECK_GOTO(r, < 0, clean_up);
+
+        if (type == SOL_HTTP_PARAM_COOKIE) {
```
barbieri: ugh, this is constant in the loop. Please simplify it:

```c

const char *separator = (type == SOL_HTTP_PARAM_COOKIE) ? " " : "&";

...

SOL_HTTP_PARAM_FOREACH_IDX (...) {
...
    r = sol_buffer_append_printf(&buf, "%s%s=%s", first ? "" : separator);
...
}


```

src/lib/comms/sol-http-common.c
```diff
@@ -123,3 +120,453 @@ sol_http_escape_string(char **escaped, const char *value)
...
+    }
+
+    if (path && (strlen(path) >= 1 && path[0] != '/')) {
+        r = sol_buffer_append_printf(&buf, "%s%s",
+            path[0] != '/' ? "/" : "", path);
+        SOL_INT_CHECK_GOTO(r, < 0, exit);
+    }
+
+    if (encoded_params && *encoded_params) {
+        r = sol_buffer_append_printf(&buf, "?%s", encoded_params);
```
barbieri: if you change `sol_http_encode_params()` to take an already-initialized buffer where you just `sol_buffer_append()`then you will simplify it here, no need to `encode_params` as a separate memory that is later printf to the uri buffer. You do all in a single buffer.

Just be sure to document such behavior in the `sol-http.h` file.

src/lib/comms/sol-http-common.c
```diff
@@ -123,3 +120,453 @@ sol_http_escape_string(char **escaped, const char *value)
...
+            params);
+        SOL_INT_CHECK(r, < 0, r);
+    }
+
+    sol_buffer_init(&buf);
+
+    if (protocol)
+        r = sol_buffer_append_printf(&buf, "%s://%s", protocol, server);
+    else
+        r = sol_buffer_append_printf(&buf, "http://%s", server);
```
barbieri: come on, would be simpler to `if (!protocol) protocol = "http"` at the beginning. Or just don't accept a null/empty protocol.

and as an optimization hint, whenever possible use slices and append them, doing the `append_printf()` is potentially more costly as it will try to format and see if there is space. If there is space, then not much overhead, otherwise it will realloc and call printf again :-/

src/lib/comms/sol-http-common.c
```diff
@@ -123,3 +120,453 @@ sol_http_escape_string(char **escaped, const char *value)
...
+    SOL_INT_CHECK_GOTO(r, < 0, exit);
+
+    if (port != 80) {
+        r = sol_buffer_append_printf(&buf, ":%d", port);
+        SOL_INT_CHECK_GOTO(r, < 0, exit);
+    }
+
+    if (path && (strlen(path) >= 1 && path[0] != '/')) {
```
barbieri: this is not correct. If path is "/a/b", then `(strlen("/a/b") >= 1 && '/' != '/')` will evaluate to false... not appending the path.

As above for protocol (and assume you changed to slices):

```c
    if (!path.data)
       path = sol_str_slice_from_str("/");

...
    if (path.data[0] != '/') {
        r = sol_buffer_append_char(&buf, '/');
    }
    r = sol_buffer_append_slice(&buf, path);
```

src/lib/comms/sol-http-common.c
```diff
@@ -123,3 +120,453 @@ sol_http_escape_string(char **escaped, const char *value)
...
+_is_pchar(char c)
+{
+    if (!_is_unreserved(c) && !_is_pct_encoded(c) &&
+        !_is_sub_delimin(c) && c != ':' && c != '@')
+        return false;
+    return true;
+}
+
+static int
+_get_protocol(const char *uri, struct sol_str_slice *protocol, char **end)
```
barbieri: for these internal functions always work with slices, so `struct sol_str_slice *uri`.

src/lib/comms/sol-http-common.c
```diff
@@ -123,3 +120,453 @@ sol_http_escape_string(char **escaped, const char *value)
...
+    for (i = 0; uri[i] && uri[i] != ':'; i++, p.len++) {
+        if (!isalnum(uri[i]) && uri[i] != '+' &&
+            uri[i] != '-' && uri[i] != '.') {
+            SOL_ERR("Invalid character at the URI's scheme: %c", uri[i]);
+            return -EINVAL;
+        }
+    }
+
+    if (uri[i] != ':') {
+        SOL_ERR("Could not find the \":\" scheme");
```
barbieri: print uri in the logs (here and above).

src/lib/comms/sol-http-common.c
```diff
@@ -123,3 +120,453 @@ sol_http_escape_string(char **escaped, const char *value)
...
+        return -EINVAL;
+    }
+
+    SOL_DBG("URI Protocol: %.*s", SOL_STR_SLICE_PRINT(p));
+    *end = (char *)p.data + p.len + 1;
```
barbieri: this cast is bad. And you can force `protocol` to be given, do the `*end = p.data + p.len + 1` outside, when needed.

src/lib/comms/sol-http-common.c
```diff
@@ -123,3 +120,453 @@ sol_http_escape_string(char **escaped, const char *value)
...
+        }
+
+        if (*partial_uri != '/') {
+            SOL_ERR("Expecting a \"/\" character to parse the server. "
+                "Found: %c", *partial_uri);
+            return -EINVAL;
+        }
+    }
+
+    if (!*partial_uri) {
```
barbieri: all of this block is over-complicated because `partial_uri` is not a slice. If it was a slice:

```c
    if (partial_uri.len <=2 || memcmp(partial_uri.data, "//", 2) != 0) {
       SOL_WRN("...");
    }
```

Also, just noticed that you're using `SOL_ERR()`... we're using `SOL_WRN()`. Change this in the whole patch and review your last patches to see if some got in.

src/lib/comms/sol-http-common.c
```diff
@@ -123,3 +120,453 @@ sol_http_escape_string(char **escaped, const char *value)
...
+        SOL_ERR("Missing host information.");
+        return -EINVAL;
+    }
+
+    s.data = partial_uri;
+
+    for (i = partial_uri; *i && *i != '/' && *i != ':' && *i != '#'
+        && *i != '?'; i++) {
```
barbieri: readability hint: use `for ()` with the main condition you're walking, like `for (i = partial_uri; *i; i++)` or `for (idx = 0; idx < partial_uri.len; idx++)`. Then inside the for you check for the other stop condition.

It will be easier to spot problems and to review.

src/lib/comms/sol-http-common.c
```diff
@@ -123,3 +120,453 @@ sol_http_escape_string(char **escaped, const char *value)
...
+    if (!*i)
+        goto exit;
+
+    if (*i != '/')
+        goto exit;
+
+    i++;
+
+    //Oops, this is not a path.
+    if (*i == '?')
```
barbieri: I don't think this comment is correct and this special case shouldn't be handled. "/" is a valid path.

src/lib/comms/sol-http-common.c
```diff
@@ -123,3 +120,453 @@ sol_http_escape_string(char **escaped, const char *value)
...
+            SOL_ERR("The param: %.*s has no \"=\" separator",
+                SOL_STR_SLICE_PRINT(*token));
+            r = -EINVAL;
+            break;
+        }
+
+        param.value.key_value.key.data = token->data;
+        param.value.key_value.key.len = sep - token->data;
+
+        param.value.key_value.value.data = sep + 1;
```
barbieri: you're not unescaping value, this is problematic.

However, doing so would allocate memory. To solve it I'd not split the query, only return a slice with the query contents and point the query split function, it would allocate memory and point to free function (clears vector and values/keys).

src/lib/comms/sol-http-common.c
```diff
@@ -123,3 +120,453 @@ sol_http_escape_string(char **escaped, const char *value)
...
+_get_fragment(char *partial_uri, struct sol_str_slice *fragment)
+{
+    struct sol_str_slice f = SOL_STR_SLICE_EMPTY;
+    char *i = partial_uri;
+
+    if (!*i)
+        goto exit;
+
+    if (*i != '#') {
+        SOL_ERR("A \"#\" is required in order to identify the fragment");
```
barbieri: again, not an error. :-) just no fragment.

src/lib/comms/sol-http-common.c
```diff
@@ -123,3 +120,453 @@ sol_http_escape_string(char **escaped, const char *value)
...
+    char *next;
+    int r;
+
+    SOL_NULL_CHECK(uri, -EINVAL);
+
+    SOL_DBG("Splitting URI: %s", uri);
+
+    r = _get_protocol(uri, protocol, &next);
+    SOL_INT_CHECK(r, < 0, r);
+    r = _get_server(next, server, &next);
```
barbieri: we're not handling local paths ("file:/xx" --- no server), we're not handling userinfo ("http://user@server" or "http://user:pass@server" -- these should be extra parameters to the split function (sorry, I said something else in person, but I had to check the RFC and it's written in there, see https://tools.ietf.org/html/rfc3986#section-3).

As for the implementation, my suggested approach is to have these internal functions to always take and return a slice. 

```c
   struct sol_str_slice partial_uri;

   partial_uri = uri;
   r = _get_scheme(partial_uri, &partial_uri, scheme); // rename protocol to scheme to be close to RFC
   SOL_INT_CHECK(r, < 0, r);
   r = _get_hier_part(partial_uri, &partial_uri, user, password, server, port);
   SOL_INT_CHECK(r, < 0, r);
   r = _get_query(partial_uri, &partial_uri, query);
   if (r != -ENOENT) { // it may not exist, so don't bail... or don't return error, your choice
       SOL_INT_CHECK(r, < 0, r);
   }   
   r = _get_fragment(partial_uri, &partial_uri, fragment);
   if (r != -ENOENT) { // it may not exist, so don't bail... or don't return error, your choice
       SOL_INT_CHECK(r, < 0, r);
   }   
```

Please check with all combinations or URI, add tests to it. Things like:
 * foo://user:pass@server.com:123/a/b?p=1&c=2#/a/b
 * foo://user@server.com:123/a/b?p=1&c=2#/a/b
 * foo://server.com:123/a/b?p=1&c=2#/a/b
 * foo://server.com/a/b?p=1&c=2#/a/b
 * foo://server.com/?p=1&c=2#/a/b (path "/")
 * foo://server.com/?p=1&c=2 (path "/")
 * foo://server.com/#/a/b
 * foo://server.com?p=1&c=2 (path empty)
 * foo://server.com#/a/b (path empty)
 * foo:/a/b?p=1&c=2#/a/b
 * foo:/a/b?p=1&c=2
 * foo:/a/b#/a/b
 * foo:#a/b (path empty)
 * foo:?a/b (path empty)
 * mailto:user@server.com (path "user@server.com")
 * /a/b (fail no scheme)
 * //a/b (fail no scheme)
 * //user:pass@server.com:123/a/b?p=1&c=2#/a/b (fail no scheme)
 * foo://user:pass@server.com:123/a/b?p=1&c=2#/a/b
...

there may be a public list of uris, both positive and negative, to be tested. See http://greenbytes.de/tech/tc/uris/ or http://www.w3.org/wiki/UriTesting.

src/lib/comms/sol-http-server-impl-microhttpd.c
```diff
@@ -117,21 +119,28 @@ build_mhd_response(const struct sol_http_response *response)
 
     SOL_HTTP_PARAM_FOREACH_IDX (&response->param, value, idx) {
         int ret;
-        char buffer[512];
+        char buffer[512], key[512], value_str[512];
```
barbieri: I'm worried about this 512 constant. Does these really have such a low maximum defined?

If not, I'd suggest change both buffer/key/value to a `sol_buffer` that is initialized before the loop. Then, inside the loop you keep reseting `buffer.used = 0`. You can then append whatever you want, it will grow the buffer as required.

For key-value, you use the same buffer, just append a null byte:
```c
buf.used = 0; // reset to buffer start, but doesn't free internal memory (keeps capacity)
ret = sol_buffer_append_slice(&buf, value->value.key_value.key);
if (ret < 0) goto err;
ret = sol_buffer_append_char(&buf, 0); // force a real \0, this is accounted as used, so is not overriden by the next append slice
if (ret < 0) goto err;
ret = sol_buffer_append_slice(&buf, value->value.key_value.value); // if buffer is default flags, then we have a '\0' here, no need to force it.
if (ret < 0) goto err;
if (MHD_add_response_header(r, sol_buffer_at(&buf, 0), sol_buffer_at(&buf, value->value.key_value.key.len + 1)) == MHD_NO) {
...
}
```

src/lib/comms/sol-http-server-impl-microhttpd.c
```diff
@@ -117,21 +119,28 @@ build_mhd_response(const struct sol_http_response *response)
...
+            if (ret < 0 || ret >= (int)sizeof(value_str))
+                goto err;
+            if (MHD_add_response_header(r, key, value_str) == MHD_NO) {
+                SOL_WRN("Could not add the header: %.*s", SOL_STR_SLICE_PRINT(value->value.key_value.key));
                 goto err;
             }
             break;
         case SOL_HTTP_PARAM_COOKIE:
-            ret = snprintf(buffer, sizeof(buffer), "%s=%s", value->value.key_value.key, value->value.key_value.value);
+            ret = snprintf(buffer, sizeof(buffer), "%.*s=%.*s", SOL_STR_SLICE_PRINT(value->value.key_value.key),
```
barbieri: is `value->value.key_value.key` and value->value.key_value.value` encoded? If not, then we have problems.

if we move to sol_buffer, we can also offer a way to encode to a buffer, making it even simpler to use.

barbieri: btw, a good approach is to always have values such as "a = b", if it's not encoded it will show easily where it's broken.

src/lib/comms/sol-http-server-impl-microhttpd.c
```diff
@@ -202,11 +205,11 @@ headers_iterator(void *data, enum MHD_ValueKind kind, const char *key, const cha
     struct sol_http_request *request = data;
     char *v, *k = NULL;
 
-    v = strdup(value);
+    v = sol_arena_strdup(request->str_arena, value);
     SOL_NULL_CHECK(v, MHD_NO);
 
-    k = strdup(key);
-    SOL_NULL_CHECK_GOTO(k, err);
+    k = sol_arena_strdup(request->str_arena, key);
```
barbieri: do we need to decode/unescape those?

src/modules/flow/http-client/http-client.c
```diff
@@ -986,22 +1000,14 @@ translate_http_method(const char *method)
...
-    }
-
-    if (timeout < 0) {
-        SOL_ERR("Timeout can not be a negative number!");
-        return -EINVAL;
-    }
-
-    *out_timeout = timeout;
-    return 0;
+    if (src > INT_MAX) {
```
barbieri: this seems to be causing alerts in coverity since it's always false if sizeof(int) == sizeof(int32_t). You should guard against that case.

src/modules/flow/http-client/http-client.c
```diff
@@ -1384,4 +1395,286 @@ get_key_process(struct sol_flow_node *node, void *data,
...
+    SOL_INT_CHECK(r, < 0, r);
+
+    r = sol_util_replace_str_from_slice_if_changed(&mdata->protocol, protocol);
+    SOL_INT_CHECK(r, < 0, r);
+
+    r = sol_util_replace_str_from_slice_if_changed(&mdata->server, server);
+    SOL_INT_CHECK(r, < 0, r);
+
+    if (fragment.len) {
+        r = sol_util_replace_str_from_slice_if_changed(&mdata->fragment,
```
barbieri: you should always replace fragment... otherwise getting an empty fragment after a non-empty one will keep the old (and incorrect). Same for path.

src/modules/flow/http-client/http-client.c
barbieri: btw, this function shouldn't exist here. All the API should be int32_t (or uint32_t). Internally, if MHD only supports int, then it should clamp there -- or doesn't care, since MHD is linux-only and requires 32bits or bigger, so int == int32_t.

in any way it should be in the node type.

(BTW reminds me that our sol_timeout_add() need an API fix :-()

src/modules/flow/http-client/http-client.c
```diff
@@ -1384,4 +1395,286 @@ get_key_process(struct sol_flow_node *node, void *data,
...
+    SOL_NULL_CHECK(ckey, -ENOMEM);
+
+    cvalue = sol_arena_strdup_slice(arena, value);
+    SOL_NULL_CHECK(cvalue, -ENOMEM);
+
+    if (!sol_http_param_add(params,
+        SOL_HTTP_REQUEST_PARAM_QUERY(ckey, cvalue))) {
+        SOL_ERR("Could not add the HTTP param %s:%s", ckey, cvalue);
+        return -ENOMEM;
+    }
```
barbieri: hey, I'm seeing this pattern of arena + param_add in many places: node-type, tests, internal to sol-http. Please make a helper structure that does it in one place:

```c
struct sol_http_param {
#define SOL_HTTP_PARAM_API_VERSION (1)
    uint16_t api_version;
    uint16_t reserved;

    struct sol_vector params;
    struct sol_arena *arena; // <--- addition
};

SOL_API bool
sol_http_param_add_copy(struct sol_http_param *params, struct sol_http_param_value value)
{
    struct sol_http_param_value *ptr;

    SOL_NULL_CHECK(params, -EINVAL);

    if (params->api_version != SOL_HTTP_PARAM_API_VERSION) {
        SOL_ERR("API version mistmatch; expected %u, got %u",
            SOL_HTTP_PARAM_API_VERSION, params->api_version);
        return false;
    }

    if (!params->arena) {
       params->arena = sol_arena_new();
       SOL_NULL_CHECK(params->arena, false);
    }

   if (value.type == SOL_HTTP_PARAM_QUERY_PARAM || ... SOL_HTTP_PARAM_COOKIE .. SOL_HTTP_PARAM_HEADER ... SOL_HTTP_PARAM_POST_FIELD) {
      int r;
      r = sol_arena_slice_dup(params->arena, &value.key_value.key, value.key_value.key);
      SOL_INT_CHECK(r, < 0, false);
      r = sol_arena_slice_dup(params->arena, &value.key_value.value, value.key_value.value);
      SOL_INT_CHECK(r, < 0, false);
   } else if (value.type == SOL_HTTP_PARAM_AUTH_BASIC) {
      int r;
      r = sol_arena_slice_dup(params->arena, &value.auth.user, value.auth.user);
      SOL_INT_CHECK(r, < 0, false);
      r = sol_arena_slice_dup(params->arena, &value.auth.password, value.auth.password);
      SOL_INT_CHECK(r, < 0, false);
   } else if (value.type == SOL_HTTP_PARAM_POST_DATA) {
      int r;
      r = sol_arena_slice_dup(params->arena, &value.data.value, value.data.value);
      SOL_INT_CHECK(r, < 0, false);  
   }

    ptr = sol_vector_append(&params->params);
    if (!ptr) {
        SOL_WRN("Could not append option to parameter vector");
        return false;
    }

    memcpy(ptr, &value, sizeof(value));
    return true;
}

SOL_API void
sol_http_param_free(struct sol_http_param *params)
{
    SOL_NULL_CHECK(params);

    if (params->api_version != SOL_HTTP_PARAM_API_VERSION) {
        SOL_ERR("API version mistmatch; expected %u, got %u",
            SOL_HTTP_PARAM_API_VERSION, params->api_version);
        return;
    }
    sol_vector_clear(&params->params);
    if (params->arena) { // <--- addition ----------------
       sol_arena_del(params->arena);
       params->arena = NULL;
    }
}

File: src/modules/flow/http-server/http-server.c [hunk 25]
```diff
@@ -212,8 +212,9 @@ common_response_cb(void *data, struct sol_http_request *request)
             SOL_INT_CHECK_GOTO(r, < 0, end);
             break;
         case SOL_HTTP_PARAM_HEADER:
-            if (streq(value->value.key_value.key, HTTP_HEADER_ACCEPT)) {
-                if (strstr(value->value.key_value.value, HTTP_HEADER_CONTENT_TYPE_JSON))
+            if (sol_str_slice_str_eq(value->value.key_value.key, HTTP_HEADER_ACCEPT)) {
```
barbieri: this should be case insensitive.

src/modules/flow/http-server/http-server.c
```diff
@@ -212,8 +212,9 @@ common_response_cb(void *data, struct sol_http_request *request)
...
+                if (memmem(value->value.key_value.value.data, value->value.key_value.value.len,
```
barbieri: use str slice functions, you can keep HTTP_HEADER_CONTENT_TYPE_JSON as a static const slice.

src/shared/sol-util.c
```diff
@@ -86,6 +86,17 @@ sol_util_memdup(const void *data, size_t len)
...
 
+SOL_API long int
+sol_util_strtol(const char *nptr, char **endptr, ssize_t len, int base)
+{
+    char *tmpbuf;
+
+    if (len < 0)
+        len = (ssize_t)strlen(nptr);
+    tmpbuf = strndupa(nptr, len);
+    return strtol(tmpbuf, endptr, base);
```
barbieri: buggggggg...

endptr is pointing inside tmpbuf! You should calculate the offset and then make `endptr = nptr + offset`.

src/test/test-http.c
```diff
@@ -0,0 +1,183 @@
...
+        "!231312312#$http$//www.intel.com",
+        "http//www.intel.com",
+        "      http   :    //www.intel.com",
+        "http:www.intel.com"
+        "http://www.intel.com:asd!%%%!!23332182/",
+        "http://",
+        "www.intel.com",
+        "http://\"",
+        "http://www.intel.com:80:80",
+    };
```
barbieri: you better use the URLs I pointed out with a more extensive test case. URI is important and easy to get wrong, so we better do it right from the beginning :-)

while at that, would be nice to see gcov results to see if we're exploring all uri split in our code.

src/lib/comms/sol-http-common.c
barbieri: btw, ipv6 will allow `:` in the server name within `[]`, see http://greenbytes.de/tech/tc/uris/#cw-ipv6-i216: `[::192.168.0.1]`

src/lib/comms/sol-http-common.c
cabelitos: I will make a patch for this. I also prefer isalnum().

src/lib/comms/include/sol-http.h
cabelitos: Sounds good.

src/lib/comms/sol-http-common.c
cabelitos: ok

src/lib/comms/sol-http-common.c
cabelitos: k

src/lib/comms/sol-http-client-impl-curl.c
cabelitos: k

src/lib/comms/include/sol-http.h
cabelitos: k

src/lib/comms/sol-http-common.c
cabelitos: Indeed this is a better approach, I will document.

src/lib/comms/sol-http-common.c
cabelitos: k

src/lib/comms/sol-http-common.c
cabelitos: My bad! I will change the code. (using slices)

src/lib/comms/sol-http-common.c
cabelitos: Sounds good.

src/lib/comms/sol-http-common.c
cabelitos: I totally forgot about ipv6. I will add the support for it!
Thanks for the reminder

src/lib/comms/sol-http-common.c
cabelitos: =)

src/lib/comms/sol-http-common.c
cabelitos: you're right.

src/lib/comms/sol-http-common.c
cabelitos: Well, we made a naive parser ;).
I will improve it, so it can handle all the possible URI formats.

src/lib/comms/sol-http-server-impl-microhttpd.c
cabelitos: To be honest, I don't like the "fixed size buffer approach", I did not change it before, because I would like to make less changes as possible. However, the current approach can lead us to annoying bugs.
I will change the code and start using sol_buffer instead.

src/lib/comms/sol-http-server-impl-microhttpd.c
cabelitos: We assume that they are encoded.

src/lib/comms/sol-http-server-impl-microhttpd.c
cabelitos: Nope, they are not encoded.

src/modules/flow/http-client/http-client.c
cabelitos: Sounds good.

src/modules/flow/http-server/http-server.c
cabelitos: You're right. headers are always case inssentive :/

src/shared/sol-util.c
cabelitos: oops.

cabelitos:
See #1022

PR #1005, opened 2015-10-30
Oauth v2

ceolin:
from v1

 - Removed commit that adds default pages for common errors
 - [oauth] Fix url callback
 - [oauth] Use urandom
 - [oauth] Keep digests's reference (to delete on close)
 - [oauth] Removed authenticated port (using the error port)
 - [oauth] Fix possibles de-reference null pointer
 - [server]  Fix namespace validation
 - Replace namespace with basename due uncrustify complains
 - Several small improvements/fixes spotted on review


ceolin:
@barbieri i've checked microhttpd and seems to not have how to check in which interface the request comes. The given url does not have neither the hostname and the port (starts from /)
I'll see to add it in microhttpd.

barbieri:
I guess you'd need to use the connection fd and get the address from it:

```c
const MHD_ConnectionInfo *info = MHD_get_connection_info(conn, MHD_CONNECTION_INFO_CONNECTION_FD);
union { sockaddr_in6 in6, sockaddr_in in4 } addr;
socklen_t addrlen = sizeof(addr);
r = getsockname(info->connect_fd, &addr, &addrlen);
if (r < 0 || addrlen > sizeof(addr)) return -EINVAL;
if (addr.in4.sin_family == AF_INET) { use addr.in4 .. } else if (addr.in6.sin6_family == AF_INET6)  { use addr.in6; }
```

ceolin:
@edersondisouza `V1_parse_response` is only called from `v1_request_finished` that is the callback of a request

edersondisouza:
@ceolin so, there's no way one could fake an Oauth response and send it to device and soletta trying to parse it? If so, it's ok. I was concerned if something sent via network could trick soletta into doing something unexpected.

ceolin:
@edersondisouza actually someone can make a request to authorize_token url and give to us bogus data, but the worst thing will happen is the data that come be invalidated because the service will reject that token

ceolin:
@barbieri https://github.com/ceolin/soletta/commit/7efb474863f8b5e1912ec1b596b3bb3e9e3f1cb8
is that api ok for you ? I intend to push the oauth branch with this additional commit.

barbieri:
I think it should be `struct sol_network_addr_to_str` instead of a `char *`.

ceolin:
pushed with the suggestion.
Thanks guys for review

PR #1007, opened 2015-10-30
Add DTLS wrapper for sol_socket & update OIC client/server

lpereira:
This series adds a DTLS (Datagram Transport Layer Security) layer on top of sol_socket. The idea is that, regardless of the underlying socket implementation (be it standard POSIX sockets, RIOT, etc.), DTLS will be available for Soletta users without we having to implement it for each and every supported platform.

I chose the TinyDTLS library for the implementation simply because it's small and used by IoTivity, which we keep a close eye on due to the need to be compatible with their OIC implementation, and also for the fact that it's indeed a tiny library, as far as a DTLS implementation goes, which fits Soletta use cases very well.

To accomplish this, sol_socket has been virtualized. The standard implementation merely forwards calls to the system's socket APIs, while the DTLS version does its magic in TinyDTLS, which in turn sends and receives encrypted data using the native socket implementation.

The CoAP implementation had to be slightly changed as well, to pass a pointer to the server struct to the handler callbacks. This was required by the changes in the OIC server implementation in order to respond requests with the right channel.

OIC servers and clients were changed, and now they're aware of DTLS. Changes at the OIC API level should be (sans very small details) transparent. The OIC implementation for provisioning, transfer methods, and whatnot are still not compatible with IoTivity, but work on that will soon follow.

Many thanks to (in no particular order) @vcgomes, @rchiossi and @otaviobp for the help in debugging this contraption.

### DIFF
--- a/.gitmodules
+++ b/.gitmodules
@@ -4,3 +4,6 @@
 [submodule "src/thirdparty/tinycbor"]
 	path = src/thirdparty/tinycbor
 	url = https://github.com/01org/tinycbor/
+[submodule "src/thirdparty/tinydtls"]
+	path = src/thirdparty/tinydtls
+	url = git://git.code.sf.net/p/tinydtls/code
--- a/data/jsons/dependencies.json
+++ b/data/jsons/dependencies.json
@@ -498,6 +498,17 @@
       "dependency": "openssl",
       "type": "pkg-config",
       "pkgname": "openssl"
+    },
+    {
+      "dependency": "tinydtls_src",
+      "type": "filesystem",
+      "files": [
+        "dtls.h"
+      ],
+      "path": {
+        "in-tree": "{TOP_SRCDIR}/src/thirdparty/tinydtls",
+        "out-of-tree": "{TINYDTLS_SRC}"
+      }
     }
   ]
 }
--- a/data/scripts/sol-oic-gen.py
+++ b/data/scripts/sol-oic-gen.py
@@ -1201,7 +1201,7 @@ server_resource_init(struct server_resource *resource, struct sol_flow_node *nod
 {
     log_init();
 
-    if (sol_oic_server_init(DEFAULT_UDP_PORT) != 0) {
+    if (sol_oic_server_init() != 0) {
         SOL_WRN("Could not create %%.*s server", SOL_STR_SLICE_PRINT(resource_type));
         return -ENOTCONN;
     }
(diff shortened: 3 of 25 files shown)


lpereira:
[**@edersondisouza**](https://github.com/edersondisouza) commented on [Nov 3, 2015, 11:39 AM GMT-2](https://github.com/solettaproject/soletta/pull/1007#commitcomment-14154001 "2015-11-03T13:39:22Z - Replied by Github Comment Enhancer"):
> Cool. And the fact that `buffer.data` is incremented doesn't mean that it leaks? Or two or more `sol_socket_dtls_recvmsg` on the same data would result on invalid free when reading the last block?

Yup. Fixed here to create a new buffer on short reads.

lpereira:
[**@bdilly**](https://github.com/bdilly) commented on [Nov 3, 2015, 1:59 PM GMT-2](https://github.com/solettaproject/soletta/pull/1007#commitcomment-14158528 "2015-11-03T15:59:06Z - Replied by Github Comment Enhancer"):
> @lpereira why do we need more retries on dtls?
> Don't you think these constants deserve documentation?
> 
> ACK_TIMEOUT_MS 2345
> MAX_RETRANSMIT_UDP 4
> MAX_RETRANSMIT_DTLS 8
> 
> Are they defined on coap specification? Were they adopted from other projects? Are they random? Should they be set by API or build config in the future?

The ack timeout, I believe, was chosen empirically. It doubles every time there's a re-transmission failure. Maybe @vcgomes can shed a light here.

As for the `MAX_RETRANSMIT` tries: maybe that's not needed for CoAP on top of DTLS. I've changed that for testing purposes and forgot it was there. This has been since removed from my local branch.

otaviobp:
+1. also tested the oic samples and they look ok

bdilly:
I've got some questions about these commits, but overall it's good.

I didn't test it, but looks like @otaviobp double checked it and everything is working fine.

So my suggestion to merge it, since there are dependency issues, is to let it disabled by default and document such issues. Later, when we have a good to go upstream tinydtls we could enable it by default.

otaviobp:
I agree that is better to make it disabled by default. There are also some things that needed to be fixed before using it for real, like creds_init implementation. I would like to have a list somewhere (besides having TODO/FIXME in the code) of "must do" items.

lpereira:
I've merged with proposed changes. It's default off. Thanks for the review!

PR #1013, opened 2015-11-03
Don't use api_version stuff on static builds

ibriano:
This drastically reduces the size of the binary in some cases where api_version
was heavily used, such as flows with several node types.


ibriano:
Merged

PR #1014, opened 2015-11-03

dorileo:
## Changes ##
  - v2 (since #972)
  - v3 (since #987)
    - changed the overall approach, instead of attaching to a process intead launch a new one with gdbserver, got almost all the pseudo code @barbieri posted in the last review, just adjusted a few minor issues, but the design is barely the same;
  - v4 (since #997)
    - fixed small issues pointed by @barbieri in the last review;
    - upstream rebased;

## Rationale ##

by the gdbserver and if not provided /dev/ttyS0 is assumed as default.

Tested with kvm, adding a serial chardev, and using it as target remote.

### DIFF
--- a/src/lib/common/sol-mainloop.c
+++ b/src/lib/common/sol-mainloop.c
@@ -385,6 +385,9 @@ sol_mainloop_default_main(const struct sol_main_callbacks *callbacks, int argc, 
 {
     int r;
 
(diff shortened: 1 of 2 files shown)


dorileo:
ok, check style failing, fixed locally.

dorileo:
ok, applied.

PR #1016, opened 2015-11-04
http:server: Add api to set default pages for errors

ceolin:


ceolin:
merged with the fixes, thanks guys

PR #1017, opened 2015-11-04
sol-oic-gen: add SCAN port on clients

bdilly:
With that it's possible to request a scan for all servers
that matches the same interface of clients.

IDs will be dispatched on DEVICE_ID port, so later
they can be used to select the device which the client
will talk.

Ideally a human-friendly name should be sent on packet
as well, not only the ID string, but I didn't found
anything like that in our implementation so far.

sol_oic_client_get_server_info return some information,
but only stuff related to a model, so users wouldn't
be able to use it to differentiate devices.

Maybe manufacturer_name + model_number could be used.

TODO: ID must be converted back from binary to
hex ascii.

@otaviobp please take a look on this. I'm pushing
it this way because I'm not sure I'll be able
to work more on that today.

It requires the conversion, and accepting IDs
by port instead of only option to conclude this
task.

### DIFF
--- a/data/scripts/sol-oic-gen.py
+++ b/data/scripts/sol-oic-gen.py
@@ -518,7 +518,8 @@ def object_open_fn_client_c(state_struct_name, resource_type, name, props):
         .to_repr_vec = %(to_repr_vec_fn)s,
         .from_repr_vec = %(struct_name)s_from_repr_vec,
         .inform_flow = %(struct_name)s_inform_flow,
-        .found_port = SOL_FLOW_NODE_TYPE_%(STRUCT_NAME)s__OUT__FOUND
+        .found_port = SOL_FLOW_NODE_TYPE_%(STRUCT_NAME)s__OUT__FOUND,
+        .device_id_port = SOL_FLOW_NODE_TYPE_%(STRUCT_NAME)s__OUT__DEVICE_ID
     };
     struct %(struct_name)s *resource = data;
     int r;
@@ -695,6 +696,20 @@ def object_setters_fn_client_c(state_struct_name, name, props):
 def object_setters_fn_server_c(state_struct_name, name, props):
     return object_setters_fn_common_c(state_struct_name, name, props, False)
 
+def object_scan_fn_client_c():
+    return '''
+static int
+scan(struct sol_flow_node *node, void *data, uint16_t port,
+    uint16_t conn_id, const struct sol_flow_packet *packet)
+{
+    struct client_resource *resource = data;
+
+    send_scan_packets(resource);
+
+    return 0;
+}
+'''
+
 def generate_enums_common_c(name, props):
     output = []
     for field, descr in props.items():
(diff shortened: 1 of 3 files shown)


data/scripts/sol-oic-gen.py
```diff
@@ -1130,6 +1166,46 @@ def master_c_as_string(generated, oic_gen_c, oic_gen_h):
     return true;
 }
 
+static char *
+binary_to_hex_ascii(const char *binary)
+{
+    /* TODO: make conversition */
+
+    char *ascii = (char *)binary;
```
barbieri: maybe use `base16` or `base64` encode? see `sol-buffer.h`

barbieri:
Why did you choose to have the scan per-device, as opposed to have a scan type that takes an interface (or something like that) as parameter/inport?

My feeling is that it will be smaller code/binary-wise since this is not replicated in every generated node, also not adding 2 new ports of the node.

As for the pretty-name, @lpereira can tell you how to get. Until you get it done, use a composed packet with 2-3 strings and define which one is the ID, which one is the pretty name, etc. Use empty strings in those we don't support yet.

bdilly:
@barbieri I don't see use cases for such a generic scanner. For a flow user, what he does is something like, in client-side:

button(gpio/writer:pin=X) OUT -> ENABLED oic/client-brightlight(id=ABC)

She doesn't know it implement a "core.brightlight" interface.
Also, she won't be able to feed the node with the wrong interface.

The client knows what interfaces it implement, so it scan for all reachable servers for itself.
In the end, what you want is to find some server to talk to.

We could generate an oic/scan-brightlight (that would be used with oic/client-brightlight and oic/server-brighlight). But regarding code replication, all these scan machinery already exists, all it does so far is to filter by "id", and now we want to filter by "interface". 

The difference is that if scan ports are on client itself, you don't need an extra node.
Things would work like:

light(oic/client-brighlight) DEVICE_ID -> IN Selector(selector/string)
Selector SELECTED -> DEVICE_ID light

instead of:

scan(oic/scan-brighlight) DEVICE_ID -> IN Selector(selector/string)
Selector SELECTED -> DEVICE_ID light(oic/light-brightlight).

What use cases do you see using a generic-interfaces scanner?

bdilly:
All these options are trivial to be implemented, it's just a matter of choosing which one would be better for a user, imo.

barbieri:
I don't like the `scan-` node types, indeed it would make it more complicated (and bigger), not smaller.

By keeping a single node that does it, it could avoid adding to the binary size with 2 extra ports and functions per node.

OTOH having to know about the interface name is really bad :-(

I have to think a bit more about it. The overhead of the 2 extra ports is not that big... at least not if we disable descriptions, then for now keep like you did.

bdilly:
OK, I've discussed with leandro and otavio now, and we're inclined to follow the implemented approach for now.

Leandro highlighted that resources may implement multiple interfaces, and sometimes despite names doesn't match, they're compatible (apparently they have stuff like aliases, inheritance, etc). A lot of things that users would need to know to use a generic scanner.

I'm sending a v2 enconding to base16 and @otaviobp will continue working on this task.

bdilly:
merged

PR #1018, opened 2015-11-04
Fix bugs in JSON nodes and add new nodes to create JSON using JSON path

otaviobp:


src/modules/flow/json/json.c
```diff
@@ -1008,11 +1029,286 @@ json_serialize(struct sol_buffer *buffer, struct json_element *element)
...
+
+static struct json_element *
+json_array_get_or_create_child_element(struct json_element *element, uint16_t i)
+{
+    struct json_element *new;
+
+    while (i >= element->children.len) {
+        new = sol_vector_append(&element->children);
+        new->type = JSON_TYPE_UNKNOWN;
+    }
```
barbieri: there is `sol_vector_append_n()` to allocate in a single go, then you walk and set the type on them.

src/modules/flow/json/json.c
```diff
@@ -405,6 +405,21 @@ json_path_is_array_key(struct sol_str_slice slice)
...
         json_path_get_next_slice(&scanner, &key_slice, &end_reason);)
 
+static int32_t
+json_array_get_key_index(struct sol_str_slice key)
+{
+    long int index_val;
+    char *endptr;
+
+    errno = 0;
+    index_val = strtol(key.data + 1, &endptr, 10);
```
barbieri: @cabelitos did a safe version of this. Your version could overflow if key.data is not null-terminated. I'm not sure that was merged already.

src/modules/flow/json/json.c
```diff
@@ -805,13 +816,25 @@ json_element_clear(struct json_element *element)
...
+}
+
 static int
 json_array_create_open(struct sol_flow_node *node, void *data, const struct sol_flow_node_options *options)
 {
-    struct json_element *mdata = data;
-
-    mdata->type = JSON_TYPE_ARRAY;
-    sol_vector_init(&mdata->children, sizeof(struct json_element));
+    init_json_array_element((struct json_element *)data);
```
barbieri: no need to cast.

PR #1019, opened 2015-11-04
Improve warning about node not found in module

mbelluzzo:
Also, add a macro to make it easier to have a custom wrn msg on null checks, when apply this macro as many places I could.

Motivation was that we were slowing removing many 'SOL_NULL_CHECKS' just in order to improve our logging messages,


barbieri:
+1

PR #1008, opened 2015-10-30
Linux micro kmod v6

dorileo:
## Changes
  + v2 (since #844)
    - impemented coldplug;
    - implemented hotplug;
    - fixed issues pointed in the last version;
    - introduced patches:
         - str-slice: implement trailing + leading whitespace helpers
         - file-reader: introduce sol_file_reader_fd() API
         - linux-micro: fix sysctl directory traversal

  + v3 (since #877)
      - fixed issues pointed by @lpereira and @barbieri in the last review (#877);

  + v4 (since #916)
     - fixed issues pointed by @lpereira in the last review (#916);
     - split the uevent handling function to its own source code (sharing between modules);
     - added the ```linux-micro: automount``` patch;
     - added unit tests to the patch ```str-slice: implement trailing + leading whitespace helpers```;

  + v5 (since #935)
    - fixed issues pointed by @lpereira in the last review (#935);
    - moved uevent api to ```sol_platform_linux```;
    - added ```mount/umount/list mount points``` api's - with async options;
    - migrated the fstab linux-micro service to use the ```sol_platform_mount()``` api;
    - migrated the automount linux-micro service to use the ```sol_platform_mount()``` api;
    - changed the general approach for the ```str-slice: implement trailing + leading whitespace helpers``` and fixed the addr sanitizer issue;

  + v6(since #960)
    - fixed issues pointed in the last review;
    - dropped patch ```str-slice: implement trailing + leading whitespace helpers``` since it was merged in a different pull request;
    - dropped patch ```linux-micro: fstab``` since our mount api's are not that generic and are mostly interested in the hotpluged mount points;
    - the get_mount_points() api was simplified and it simple populates a vector of ```sol_buffer```;
    - whole bunch of simplifications....

## Rationale
Implements the modprobe process following the systemd's modules-load.d
specification, also does the coldplug module loading - which consists in
looking up devices described in the sysfs and loading their respective
modules based on the modalias descriptors.

This module will also handle the hotplug events, notified by the kernel
via uevent netlink channel.

PR #1020, opened 2015-11-05
Linux micro kmod v9

dorileo:
## Changes
  + v2 (since #844)
    - impemented coldplug;
    - implemented hotplug;
    - introduced patches:

  + v3 (since #877)

  + v4 (since #916)

  + v5 (since #935)
    - fixed issues pointed by @lpereira in the last review (#935);
    - moved uevent api to ```sol_platform_linux```;
    - added ```mount/umount/list mount points``` api's - with async options;
    - migrated the fstab linux-micro service to use the ```sol_platform_mount()``` api;
    - migrated the automount linux-micro service to use the ```sol_platform_mount()``` api;
    - changed the general approach for the ```str-slice: implement trailing + leading whitespace helpers``` and fixed the addr sanitizer issue;

  + v6 (since #960)
    - dropped patch ```str-slice: implement trailing + leading whitespace helpers``` since it was merged in a different pull request;
    - dropped patch ```linux-micro: fstab``` since our mount api's are not that generic and are mostly interested in the hotpluged mount points;
    - the get_mount_points() api was simplified and it simple populates a vector of ```sol_buffer```;
    - whole bunch of simplifications....

  + v7 (since #1008)
    - fixed check-style issue;

  + v8 (since #1010)
    - fixed issues pointed by @glima in the last review;
    - introduced the patch ```platform: get rid of dummy impl```;
    - upstream rebased;

  + v9 (since #1015)
    - fixed issues pointed by @ibriano in the last review;
    - upstream rebased;

## Rationale

via uevent netlink channel.


src/modules/linux-micro/kmod/kmod.c
```diff
@@ -0,0 +1,410 @@
...
+            int fd;
+            char buff[PATH_MAX];
+
+            fd = openat(dfd, ent.d_name, O_RDONLY | O_CLOEXEC);
+            if (fd < 0) {
+                SOL_ERR("Could not obtain uevent file descriptor");
+                goto next;
+            }
+
+            if (read(fd, buff, sizeof(buff)) > 0) {
```
lpereira: Did you test this when building with `-O0`? Because you might be getting lucky here and `buff` might be 0-initialized, thus the string read from the file ends up being 0-terminated. To avoid problems, store the return value of the call to `read()`, and pass the slice with that as the length to `kmod_apply_value()`.

src/modules/linux-micro/automount/automount.c
```diff
@@ -0,0 +1,368 @@
...
+
+struct filesystem_magic {
+    const char *magic;
+    int length;
+    int offset;
+};
+
+struct filesystem {
+    const char *id;
+    const struct filesystem_magic *magics;
```
lpereira: Nitpick: the work magic is a mass noun and has no plural form.

lpereira:
LGTM sans a minor issue that could be fixed in a followup PR.

dorileo:
ok, applied. Thanks.

Issue #732, opened 2015-09-16
OIC realistic example


barbieri:
We should provide a realistic OIC sample where devices are discovered and paired to be used together without using node options to specify the other endpoint.

We should have a node type to discover OIC devices implementing a given interface. Once these are discovered, they are sent to a "selector" node type that will pick one. The selected one is persisted and then used by further actions specific to that device interface.

A suggested approach is for the OIC device discoverer to emit a custom information packet with an array of items, each being composed of an identifier, a human readable name and possibly an icon (or a type). Alternatively it may produce packets of a single item that are accumulated by the selector -- in this case the selector would need a `RESET` or `CLEAR` port to flush items.

The selectors should be generic and work with the packets specified above (they could live in Soletta's core and not in oic module). Two selectors we could offer by default are:
 * **HTTP selector:** it will export a HTTP resource using `sol_http_server_register_handler()`. The `GET` can return a JSON with the available options and the current selection. The `POST` handler receives the device ID that should be selected. An externally provided HTML/JS/CSS set may be used to present this to the user.
    * Options:
       * `path` (string): the URL path to handle HTTP requests.
    * Input Ports:
       * `ITEMS` (custom packet): the options to be selected.
       * `SELECTED` (string): the previously selected item, if any.
    * Output Ports:
       * `SELECTED` (string): the selected identifier. Sent when user chooses via HTTP.
 * **String output + Buttons Input selector:** it will format a string output with the available options. Input ports will handle "previous", "next" and "select".
    * Options:
       * `columns` (int): the number of columns in the display. Will be used to format the string, including auto-scroll.
       * `rows` (int): the number of rows in the display. If more than 1, will wrap lines.
       * `title` (string): if more than one row, this will be the contents of the first row. Otherwise will be the prefix of the only row in the string.
       * `auto_scroll` (boolean): if true and the contents doesn't fit in the columns, then each row will scroll.
       * `circular` (boolean): if true, this is a circular menu, after the last option it goes back to the first.
    * Input Ports:
       * `ITEMS` (custom packet): the options to be selected.
       * `NEXT` (any): change to the next item.
       * `PREVIOUS` (any): change to the previous item.
       * `SELECT` (any): select the current item.
       * `ENABLED` (bool): make the interaction possible, otherwise don't produce strings or process next/previous/select inputs. This is used because the node is not created and destroyed on demand, rather it will be enabled and disabled.
       * `SELECTED` (string): the previously selected item, if any.
    * Output Ports:
[...]

PR #1021, opened 2015-11-05
New OIC demo and commits necessary to support it

otaviobp:
Improve oic-client-demo by adding a selector to choose which light
deivice will be controled by the client node. After choosing a device,
user is able to turn on/off the light.

Persistence is also used to save the favorite device and to auto-select
it when possible. Changing the device being controlled is also
supported.

So far, the json conffile is configured to use gtk output to run the
demo, but recreating a possible real scenario using a real board.

Related to #732

### DIFF
--- a/src/samples/flow/oic/light-client-scan.fbp
+++ b/src/samples/flow/oic/light-client-scan.fbp
@@ -29,18 +29,45 @@
 # (INCLUDING NEGLIGENCE OR OTHERWISE) ARISING IN ANY WAY OUT OF THE USE
 # OF THIS SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
 
-# This will timely (every 3 seconds) trigger scan for OIC lights. This node type
-# is resolved using the light-client.json, please adjust to your network
-# configuration:
+# This will timely (every 3 seconds) trigger scan for OIC lights.
+# The oic node type and IO node types are resolved using the
+# light-client-scan-gtk.json or light-client-scan-edison.json files.
 #
-#    $ export SOL_FLOW_MODULE_RESOLVER_CONFFILE=light-client-mynet.json
+# To run this sample using a gtk backend, use light-client-scan-gtk.json.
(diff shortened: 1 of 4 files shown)


otaviobp:
Probably not, I'll check this. I didn't know we had that. That's nice.

src/samples/flow/oic/light-client-scan.fbp
```diff
@@ -29,18 +29,45 @@
...
+next OUT -> PULSE_IF_TRUE _(converter/boolean-to-empty) OUT -> NEXT light_selector
+select OUT -> PULSE_IF_TRUE _(converter/boolean-to-empty) OUT -> SELECT light_selector
+light FOUND -> IN status
+light_button OUT -> PULSE_IF_TRUE _(converter/boolean-to-empty) OUT -> IN toggle(boolean/toggle) OUT -> IN_STATE light
 
-timer OUT -> IN Scanning(console)
-timer OUT -> SCAN light
-light DEVICE_ID -> IN DeviceFound(console)
+#DeviceID Persistence
+timer OUT -> IN selected_send(converter/empty-to-string) OUT -> SELECTED light_selector
```
barbieri: seems nobody defines timer anymore...did you check?

Also, no idea why a timer to send an empty string to the selector... we want to timely reset it? It seems that then the converter/empty-to-string value is changed to the persisted one, but doesn't look good. Why was it required?

bdilly: It's missing indeed. Also, the way it is, it won't work if client is launched before server.

Maybe adding a timer triggering the scan from time to time until receiving the first answer may be a good idea. Or adding a "scan" button. Or just explaining it.

Also it may be interesting to have a led with light status, not only if the machine id was found.

Displaying a secondary menu with resource features and controls to handle them is too much? So you could go through all of them changing values (for example, color, brightness, on / off) and make use of int selectors, etc? And eventually we could go back to previous menu (and newtwork would be re-scanned).

I mean, for a v2, I'm not opposing to start with a simple sample.

PR #1022, opened 2015-11-05
Create a node to build URLs. v4

cabelitos:
Changes since v3.

* Added suggestions made by @barbieri 
* Add more tests for http split and create functions
* Added commits 001f716, ec6a856, 338a1cb,  e91d052

### DIFF
--- a/src/lib/comms/include/sol-http.h
+++ b/src/lib/comms/include/sol-http.h
@@ -112,7 +112,7 @@ struct sol_http_param_value {
             bool value;
         } boolean;
         struct {
-            int value;
+            int32_t value;
         } integer;
         struct {
             const struct sol_str_slice value;
--- a/src/lib/comms/include/sol-http.h
+++ b/src/lib/comms/include/sol-http.h
@@ -34,6 +34,8 @@
 
 #include <sol-buffer.h>
 #include <sol-vector.h>
+#include <sol-str-slice.h>
+#include <sol-arena.h>
 
 #ifdef __cplusplus
 extern "C" {
@@ -95,18 +97,19 @@ struct sol_http_param {
     uint16_t reserved;
 
     struct sol_vector params;
+    struct sol_arena *arena;
 };
 
 struct sol_http_param_value {
     enum sol_http_param_type type;
     union {
         struct {
-            const char *key;
-            const char *value;
+            struct sol_str_slice key;
+            struct sol_str_slice value;
         } key_value;
         struct {
-            const char *user;
-            const char *password;
+            struct sol_str_slice user;
+            struct sol_str_slice password;
         } auth;
         struct {
             bool value;
@@ -170,8 +173,8 @@ struct sol_http_response {
     (struct sol_http_param_value) { \
         .type = type_, \
         .value.key_value = { \
-            .key = (key_), \
-            .value = (value_) \
+            .key = sol_str_slice_from_str((key_)),      \
+            .value = sol_str_slice_from_str((value_))   \
         } \
     }
 
@@ -191,8 +194,8 @@ struct sol_http_response {
     (struct sol_http_param_value) { \
         .type = SOL_HTTP_PARAM_AUTH_BASIC, \
         .value.auth = { \
-            .user = (username_), \
-            .password = (password_) \
+            .user = sol_str_slice_from_str((username_)),        \
+            .password = sol_str_slice_from_str((password_))     \
         } \
     }
 
@@ -236,6 +239,7 @@ sol_http_param_init(struct sol_http_param *params)
 
 bool sol_http_param_add(struct sol_http_param *params,
     struct sol_http_param_value value) SOL_ATTR_WARN_UNUSED_RESULT;
+bool sol_http_param_add_copy(struct sol_http_param *params, struct sol_http_param_value value);
 void sol_http_param_free(struct sol_http_param *params);
 
 int sol_http_escape_string(char **escaped, const char *value);
--- a/src/lib/comms/include/sol-http.h
+++ b/src/lib/comms/include/sol-http.h
@@ -135,7 +135,18 @@ struct sol_http_response {
     int response_code;
 };
 
-#define SOL_HTTP_RESPONSE_CHECK_API(response_, ...) \
+struct sol_http_url {
+    struct sol_str_slice scheme;
+    struct sol_str_slice user;
+    struct sol_str_slice password;
+    struct sol_str_slice host;
+    struct sol_str_slice path;
+    struct sol_str_slice query;
+    struct sol_str_slice fragment;
+    int port; //If < 0 ignore
+};
+
+#define SOL_HTTP_RESPONSE_CHECK_API(response_, ...)  \
     do { \
         if (unlikely(!response_)) { \
             SOL_WRN("Error while reaching service."); \
@@ -242,7 +253,23 @@ bool sol_http_param_add(struct sol_http_param *params,
 bool sol_http_param_add_copy(struct sol_http_param *params, struct sol_http_param_value value);
 void sol_http_param_free(struct sol_http_param *params);
 
-int sol_http_escape_string(char **escaped, const char *value);
+int sol_http_encode_slice(struct sol_buffer *buf, const struct sol_str_slice value);
+int sol_http_decode_slice(struct sol_buffer *buf, const struct sol_str_slice value);
+
+int sol_http_create_uri(char **url_out, const struct sol_http_url url, const struct sol_http_param *params);
+
+int sol_http_create_simple_uri(char **uri, const struct sol_str_slice base_url, const struct sol_http_param *params);
+
+int sol_http_encode_params(struct sol_buffer *buf, enum sol_http_param_type type, const struct sol_http_param *params);
+int sol_http_decode_params(const struct sol_str_slice params_slice, enum sol_http_param_type type, struct sol_http_param *params);
+
+int sol_http_split_uri(const struct sol_str_slice full_uri, struct sol_http_url *url);
+
+static inline int
+sol_http_create_simple_uri_from_str(char **uri, const char *base_url, const struct sol_http_param *params)
+{
+    return sol_http_create_simple_uri(uri, sol_str_slice_from_str(base_url ? base_url : ""), params);
+}
 
 /**
  * @}
--- a/src/lib/comms/sol-http-client-impl-curl.c
+++ b/src/lib/comms/sol-http-client-impl-curl.c
@@ -531,101 +531,6 @@ free_buffer:
     return NULL;
 }
 
-static char *
-encode_key_values(CURL *curl, enum sol_http_param_type type,
-    const struct sol_http_param *params, char *initial_value)
-{
-    struct sol_http_param_value *iter;
-    uint16_t idx;
-    bool first = true;
-
-    if (type != SOL_HTTP_PARAM_QUERY_PARAM &&
-        type != SOL_HTTP_PARAM_POST_FIELD &&
-        type != SOL_HTTP_PARAM_COOKIE) {
-        free(initial_value);
-        errno = EINVAL;
-        return NULL;
-    }
-
-    SOL_VECTOR_FOREACH_IDX (&params->params, iter, idx) {
-        struct sol_str_slice key = iter->value.key_value.key;
-        struct sol_str_slice value = iter->value.key_value.value;
-        char *tmp, *encoded_key, *encoded_value;
-        int r;
-
-        if (iter->type != type)
-            continue;
-
-        encoded_key = curl_easy_escape(curl, key.data, key.len);
-        if (!encoded_key)
-            goto cleanup;
-        encoded_value = curl_easy_escape(curl, value.data, value.len);
-        if (!encoded_value) {
-            curl_free(encoded_key);
-            goto cleanup;
-        }
-
-        if (type == SOL_HTTP_PARAM_COOKIE) {
-            r = asprintf(&tmp, "%s%s%s=%s;", initial_value, first ? "" : " ",
-                encoded_key, encoded_value);
-        } else {
-            r = asprintf(&tmp, "%s%s%s=%s", initial_value, first ? "" : "&",
-                encoded_key, encoded_value);
-        }
-
-        curl_free(encoded_key);
-        curl_free(encoded_value);
-
-        if (r < 0)
-            goto cleanup;
-
-        free(initial_value);
-        initial_value = tmp;
-        first = false;
-    }
-
-    return initial_value;
-
-cleanup:
-    free(initial_value);
-    errno = ENOMEM;
-    return NULL;
-}
-
-static char *
-build_uri(CURL *curl, const char *base, const struct sol_http_param *params)
-{
-    char *initial_value;
-    char *built_uri;
-
-    if (asprintf(&initial_value, "%s?", base) < 0) {
-        errno = ENOMEM;
-        return NULL;
-    }
-
-    built_uri = encode_key_values(curl, SOL_HTTP_PARAM_QUERY_PARAM, params,
-        initial_value);
-    if (built_uri == initial_value) {
-        free(initial_value);
-        return strdup(base);
-    }
-
-    return built_uri;
-}
-
-static char *
-build_cookies(CURL *curl, const struct sol_http_param *params)
-{
-    return encode_key_values(curl, SOL_HTTP_PARAM_COOKIE, params, strdup(""));
-}
-
-static char *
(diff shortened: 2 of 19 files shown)


src/lib/comms/include/sol-http.h
```diff
@@ -132,7 +135,18 @@ struct sol_http_response {
...
-#define SOL_HTTP_RESPONSE_CHECK_API(response_, ...) \
+struct sol_http_url {
+    struct sol_str_slice scheme;
+    struct sol_str_slice user;
+    struct sol_str_slice password;
+    struct sol_str_slice host;
+    struct sol_str_slice path;
+    struct sol_str_slice query;
+    struct sol_str_slice fragment;
+    int port; //If < 0 ignore
```
barbieri: int32_t... and I guess 0 is already not a valid port, so you can make it `uint32_t` and say `if == 0, ignore`

src/lib/comms/include/sol-http.h
```diff
@@ -236,9 +250,26 @@ sol_http_param_init(struct sol_http_param *params)
 
 bool sol_http_param_add(struct sol_http_param *params,
     struct sol_http_param_value value) SOL_ATTR_WARN_UNUSED_RESULT;
+bool sol_http_param_add_copy(struct sol_http_param *params, struct sol_http_param_value value);
 void sol_http_param_free(struct sol_http_param *params);
 
-int sol_http_escape_string(char **escaped, const char *value);
+int sol_http_encode_slice(struct sol_buffer *buf, const struct sol_str_slice value);
```
barbieri: document all these functions... in the header! (also valid for those you didn't do, but do those in a different commit later)

src/lib/comms/include/sol-http.h
ibriano: On Fri, 06 Nov 2015 09:03:01 -0800, Gustavo Sverzut Barbieri wrote:
> > @@ -132,7 +135,18 @@ struct sol_http_response {
> >      int response_code;
> >  };
> >  
> > -#define SOL_HTTP_RESPONSE_CHECK_API(response_, ...) \
> > +struct sol_http_url {
> > +    struct sol_str_slice scheme;
> > +    struct sol_str_slice user;
> > +    struct sol_str_slice password;
> > +    struct sol_str_slice host;
> > +    struct sol_str_slice path;
> > +    struct sol_str_slice query;
> > +    struct sol_str_slice fragment;
> > +    int port; //If < 0 ignore
> 
> int32_t... and I guess 0 is already not a valid port, so you can make it `uint32_t` and say `if == 0, ignore`

Ports go up to 65535, so uint16_t should be enough, no? But agreed that
int alone is not good, on 16 bit int we lose half the ports.

src/lib/comms/sol-http-client-impl-curl.c
```diff
@@ -431,33 +423,29 @@ header_cb(char *data, size_t size, size_t nmemb, void *connp)
         param.type = SOL_HTTP_PARAM_COOKIE;
         cookie_name_size = 0;
         while (sep[cookie_name_size++] != '=') ;
-        param.value.key_value.key = curl_easy_unescape(connection->curl,
-            sep, cookie_name_size - 1, NULL);
+        param.value.key_value.key.data = sep;
```
barbieri: is this right? AFAIU you got a raw cookie from CURL, after all you had to find "=" in it, we're keeping parameters unescaped... then we should have to unescape here, no?

src/lib/comms/sol-http-client-impl-curl.c
```diff
@@ -746,20 +639,31 @@ static bool
...
+    int err;
+
+    sol_buffer_init(&buf);
+    err = sol_http_encode_params(&buf, SOL_HTTP_PARAM_COOKIE, params);
+    SOL_INT_CHECK(err, < 0, false);
+
+    cookies = sol_buffer_steal_or_copy(&buf, NULL);
+    if (cookies)
+        r = set_string_option(curl, CURLOPT_COOKIE, arena, cookies);
+    sol_buffer_fini(&buf);
```
barbieri: we should revisit `set_string_option()`. I just looked at it and it is duplicating the string in the arena, and also freeing the given pointer. Either it should stop to free the given pointer, or there should be a way to give ownership of the pointer to the arena, avoiding the "strdup() + free()". The second case looks better, but we if choose to do the first one, this function I'm commenting shouldn't do "steal or copy"

barbieri: Just checked and strings are already copied by cURL, we do not have to keep them in our arena, see:

http://curl.haxx.se/libcurl/c/curl_easy_setopt.html

Thus we really should revisit it, stop calling "steal or copy".

src/lib/comms/sol-http-client-impl-curl.c
```diff
@@ -768,9 +672,20 @@ static bool
...
+    int err;
+    bool r = true;
+
+    sol_buffer_init(&buf);
+    err = sol_http_encode_params(&buf, SOL_HTTP_PARAM_POST_FIELD, params);
+    SOL_INT_CHECK(err, < 0, false);
+
+    post = sol_buffer_steal_or_copy(&buf, NULL);
+    if (post)
+        r = set_string_option(curl, CURLOPT_POSTFIELDS, arena, post);
```
barbieri: change this to `CURLOPT_COPYPOSTFIELDS`, see the comment above about `set_string_option()` usage.

src/lib/comms/include/sol-http.h
cabelitos: k

src/lib/comms/include/sol-http.h
cabelitos: Changing to uin32_t

src/lib/comms/sol-http-common.c
```diff
@@ -74,52 +130,645 @@ sol_http_param_free(struct sol_http_param *params)
...
+
+    SOL_NULL_CHECK(buf, -EINVAL);
+
+    escaped = false;
+    last_append = 0;
+    for (i = 0; i < value.len; i++) {
+        unsigned char c = value.data[i];
+        if (!isalnum(c) && c != '_' && c != '~' && c != '.' && c != '-') {
+            if (!escaped) {
+                sol_buffer_init(buf);
```
barbieri: you can init before the loop and avoid !escaped. Just assume it's not escaped if after the loop end you didn't append anything (ie: would append the whole slice as-is), then you change the flag and set the pointer.

See https://github.com/solettaproject/soletta/blob/master/src/lib/parsers/sol-json.c#L696

src/lib/comms/sol-http-common.c
```diff
@@ -74,52 +130,645 @@ sol_http_param_free(struct sol_http_param *params)
...
+
+    SOL_NULL_CHECK(buf, -EINVAL);
+
+    decoded = false;
+    last_append = 0;
+
+    for (i = 0; i < value.len; i++) {
+        unsigned char c = value.data[i];
+        if (c == '%' && value.len - i >= 2 && isxdigit(value.data[i + 1])
+            && value.data[i + 2]) {
```
barbieri: shouldn't it be `value.len - i > 2` and `isxdigit(value.data[i+2])`?

cabelitos: Yep, I forgot the isxdigit() there, I already added locally. And yes it should be > 2

src/lib/comms/sol-http-common.c
```diff
@@ -74,52 +130,645 @@ sol_http_param_free(struct sol_http_param *params)
...
+        scheme = url.scheme;
+    else
+        scheme = sol_str_slice_from_str("http");
+
+    r = sol_buffer_append_slice(&buf, scheme);
+    SOL_INT_CHECK_GOTO(r, < 0, exit);
+
+    if (!sol_str_slice_str_caseeq(scheme, "mailto") &&
+        !sol_str_slice_str_caseeq(scheme, "news"))
+        sep = sol_str_slice_from_str("://");
```
barbieri: this is not right, there should be no special case... and mailto AFAIK has no "://".

Even "news", it requires a server to get the "//", so just do:

```c
    r = sol_buffer_append_slice(&buf, scheme);
    ...
    r = sol_buffer_append_char(&buf, ':');

    if (url.host.len) {
       r = sol_buffer_append_slice(&buf, SOL_STR_SLICE("//"));
       ...
       if (url.user.len) {
           r = sol_buffer_append_slice(&buf, url.user); // we may have to encode this? plz check
           if (url.password.len) {
               r = sol_buffer_append_char(&buf, ':');
               r = sol_buffer_append_slice(&buf, url.password); // we may have to encode this? plz check
         }
          r = sol_buffer_append_char(&buf, '@');
    }
    r = sol_buffer_append_slice(&buf, url.host);
    if (url.port) {
       r = sol_buffer_append_char(&buf, ':');
       r = sol_buffer_append_printf(&buf, "%u", url.port);
   }
```

src/lib/comms/sol-http-common.c
```diff
@@ -74,52 +130,645 @@ sol_http_param_free(struct sol_http_param *params)
...
+    }
+
+    sol_buffer_init(&buf);
+
+    r = sol_buffer_append_slice(&buf, base_url);
+    SOL_INT_CHECK(r, < 0, r);
 
-    *escaped = sol_buffer_steal(&buf, NULL);
+    if (params && params->params.len) {
+        r = sol_buffer_append_slice(&buf, sol_str_slice_from_str("/?"));
```
barbieri: why '/' in there?, just append `?`

src/lib/comms/sol-http-common.c
```diff
@@ -74,52 +130,645 @@ sol_http_param_free(struct sol_http_param *params)
...
+    struct sol_str_slice *pass,
+    int *port)
+{
+
+    size_t offset = 2;
+    struct sol_str_slice port_slice = SOL_STR_SLICE_EMPTY;
+    char *sep, *port_sep, *fragment_sep, *path_sep, *query_sep;
+    struct sol_str_slice auth = partial_uri;
+
+    if (!auth.len) {
```
barbieri: you should check if `auth.len > 2`, as you do `strncmp()` later using that amount of bytes... or add 'sol_str_slice_startswith(slice_data, slice_prefix)` that does this in a single step and save errors in the future.

src/lib/comms/sol-http-common.c
cabelitos: Ok,

cabelitos: No good reason for it. I will just append '?'

src/lib/comms/sol-http-common.c
```diff
@@ -74,52 +130,645 @@ sol_http_param_free(struct sol_http_param *params)
...
+
+    //Not an URL
+    if (strncmp(auth.data, "//", 2))
+        return 0;
+
+    *port = -1;
+    auth.data += 2;
+    auth.len -= 2;
+
+    sep = memchr(auth.data, '@', auth.len);
```
barbieri: this looks easier, but will do useless work and is not that easier. It's better to just do a single for() until the authority terminates and while doing so, set pointers to parts:

```c
   const char *itr = auth.data;
   const char *itr_end = itr + auth_data.len;
   struct sol_str_slice user = SOL_STR_SLICE_STR(NULL, 0); /* so we get NULL instead of "" */
   struct sol_str_slice pass = SOL_STR_SLICE_STR(NULL, 0);
   struct sol_str_slice host = SOL_STR_SLICE_STR(NULL, 0);
   struct sol_str_slice port = SOL_STR_SLICE_STR(NULL, 0);
   const char *last = auth.data;
   for (; itr < itr_end; itr++) {
       const char c = *itr;
       if (c == '/') {
           auth.len = itr - auth.data;
           if (host.len == 0) {
               host.len = itr - host.data;
          } else if (port.data) {
               port.len = itr - port.data;
          }
          break;
       } else if (c == ':') {
           /* could be host:port or user:pass, store in host:port and move later if '@' appears */
           if (host.data) {
               return -EINVAL; // already got a host:port (or user:port)
           }
           host.data = last;
           host.len = itr - host.data;
           last = itr + 1;
           port.data = last;
           port.len = 0;
       } else if (c == '@') {
           if (user.len) {
               return -EINVAL; // already got an user
           } else {
               user = host;
               pass = port;
               host = SOL_STR_SLICE_STR(NULL, 0);
               port = SOL_STR_SLICE_STR(NULL, 0);
               last = itr + 1;
           }
       }
   }
```

I didn't test it, but should be something like that.

src/lib/comms/sol-http-common.c
```diff
@@ -74,52 +130,645 @@ sol_http_param_free(struct sol_http_param *params)
...
+    const struct sol_str_slice full_uri,
+    struct sol_str_slice *next,
+    struct sol_str_slice *path)
+{
+    char *sep_query, *sep_fragment;
+
+    if (!partial_uri.len)
+        return 0;
+
+    if (partial_uri.data[0] == '#' || partial_uri.data[0] == '?')
```
barbieri: in these cases we have an empty path, should we make sure path points to an empty slice?

src/lib/comms/sol-http-common.c
```diff
@@ -74,52 +130,645 @@ sol_http_param_free(struct sol_http_param *params)
...
+    path->data = partial_uri.data;
+
+    sep_query = memchr(partial_uri.data, '?', partial_uri.len);
+    sep_fragment = memchr(partial_uri.data, '#', partial_uri.len);
+
+    if (sep_query)
+        path->len = sep_query - path->data;
+    else if (!sep_query && sep_fragment)
+        path->len = sep_fragment - path->data;
+    else if (!sep_query && !sep_fragment)
```
barbieri: as above, you should walk the path and stop at `?` or `#`.  Then you decide exactly what do OUTSIDE of `_get_path()` by checking the first char of updated uri... if len > 0.

src/lib/comms/sol-http-common.c
cabelitos: Nope, I'm doing memset(url, 0, sizeof(struct..)) in sol_http_split_url

src/lib/datatypes/sol-buffer.c
```diff
@@ -344,6 +347,20 @@ sol_buffer_steal(struct sol_buffer *buf, size_t *size)
     return r;
 }
 
+SOL_API void *
+sol_buffer_steal_or_copy(struct sol_buffer *buf, size_t *size)
```
barbieri: you can add this as a `static inline` in the header, this way we keep the function without generating an extra symbol

cabelitos: I removed this function (locally) I'm not using it anymore.

src/modules/flow/http-client/http-client.json
```diff
@@ -777,6 +777,161 @@
...
+        "members": [
+          {
+            "data_type": "string",
+            "default": "http",
+            "description": "The URL protocol",
+            "name": "scheme"
+          },
+          {
+            "data_type": "int",
+            "default": 80,
```
barbieri: make it default to zero, so if we change the protocol/scheme we won't be getting 80 as their port ;-)

src/modules/flow/http-client/http-client.json
```diff
@@ -777,6 +777,161 @@
...
+          {
+            "data_type": "string",
+            "default": null,
+            "description": "The URL fragment",
+            "name": "fragment"
+          },
+          {
+            "data_type": "string",
+            "default": null,
+            "description": "The URL parameters to be encoded",
```
barbieri: say it should be already encoded in the proper format (ie: a=b&c=%30). The message says it wil be encoded, which is wrong.

src/lib/comms/sol-http-common.c
cabelitos: I will adapt the code to your suggestion.
Thanks.

src/lib/comms/include/sol-http.h
lpereira: On 11/06/2015 03:03 PM, Gustavo Sverzut Barbieri wrote:
>>
>> -#define SOL_HTTP_RESPONSE_CHECK_API(response_, ...) \
>> +struct sol_http_url {
>> +    struct sol_str_slice scheme;
>> +    struct sol_str_slice user;
>> +    struct sol_str_slice password;
>> +    struct sol_str_slice host;
>> +    struct sol_str_slice path;
>> +    struct sol_str_slice query;
>> +    struct sol_str_slice fragment;
>> +    int port; //If < 0 ignore
>
> int32_t... and I guess 0 is already not a valid port, so you can make it
> |uint32_t| and say |if == 0, ignore|
>

TCP ports are 16-bit unsigned integers.



	Leandro

cabelitos:
Closing in favor of #1028

PR #1023, opened 2015-11-05
form: new int-custom form node, with tests.

glima:
This is a "edit-digit-by-digit" version of the int form.

I'll add a commit on top of it soon with a sample exercising the ``OUT_OF_RANGE`` port.


glima:
The style was fixed locally. One of the tests really failed after a last minute change in the code, now fixing that. If no comments arise, a push -f will happen.

glima:
Moving to v2, since some fixes of Ederson comments were made (and rebasing).

PR #1024, opened 2015-11-06
Form int custom [v2]

glima:
Changes since #1023:

 - fixed "total_range" calculation

### DIFF
--- a/src/modules/flow/form/form.c
+++ b/src/modules/flow/form/form.c
@@ -392,35 +392,35 @@ format_title(struct sol_buffer *buf,
         return 0;
     }
 
+    if (!title || !title_tag)
+        return 0;
+
     /* Format title */
-    if (title && title_tag) {
-        const char *value = title;
-        r = format_chunk(buf, n_rows, n_cols, value, title + strlen(title),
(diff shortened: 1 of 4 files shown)


src/modules/flow/form/form.c
```diff
@@ -1345,5 +1363,572 @@ integer_enabled_set(struct sol_flow_node *node,
...
+        r = snprintf(format, sizeof(format), "%%0%" PRIu8 "" PRId32 "",
+            mdata->n_digits);
+    else
+        r = snprintf(format, sizeof(format), "%% 0%" PRIu8 "" PRId32 "",
+            mdata->n_digits + 1);
+    if (r < 0 || r > (int)sizeof(format))
+        return -EINVAL;
+
+#pragma GCC diagnostic ignored "-Wformat-nonliteral"
+    r = asprintf(&it_value, format, mdata->base.state.val);
```
lpereira: Why not simply use `%.*d` to specity the number of digits and move the `asprintf()` call to the previous `if` block? Then there's no need to use a nasty `#pragma`.

glima: Alright! Tested and it works, thanks a lot :)

glima:
Merged with the due fixes.

PR #1026, opened 2015-11-06
New OIC demo and commits necessary to support it - v2

otaviobp:
Changelog:
 - Adding commit a000a76af

Scanning was not working as expected in previous branch because coap was ignoring all replies after the first one. So multicast packets that expects response from several servers was failing.
We had some discussions on how to have a good way of informing sol-coap that a reply is no longer needed. The fist step, that is implemented in a000a76af is to look at the return of reply_cb.
We still need to have a function to let users cancel a request, a way of stoping waiting for replies when a timeout occurs and support for cancelling observer requests. And also there are a leak in reply elements that need to be fixed.
I sent the pr because with this commit because now demo is working, the light-client.fbp demo was fixed and this make sense even without implementing the full strategy we discussed.

Replaces #1021

### DIFF
--- a/data/scripts/sol-oic-gen.py
+++ b/data/scripts/sol-oic-gen.py
@@ -1031,6 +1031,7 @@ struct client_resource {
 
     const char *rt;
     char *device_id;
+    struct sol_ptr_vector scanned_ids;
 };
 
 struct server_resource {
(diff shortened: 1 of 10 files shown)


lpereira:
[**@bdilly**](https://github.com/bdilly) commented on [Nov 6, 2015, 6:58 PM GMT-2](https://github.com/solettaproject/soletta/pull/1026#commitcomment-14251739 "2015-11-06T20:58:50Z - Replied by Github Comment Enhancer"):
> How is this oic_res lifecycle? Do you really need to duplicate this string? No way to ref it or something like that?

For each device that's found on the network, a resource is created with `refcount = 1`, the user callback is called, and then the ref count is decremented. If the user wants to use the resource, they need to grab their own reference, or else this will destroy the resource.

PR #1027, opened 2015-11-09
Fix bugs in JSON nodes and add new nodes to create JSON using JSON path - v2

otaviobp:
Changelog from v1:
 - fixing issues reported in v1
 - Always replace a value when user set a new value to a key, even when types differ. For example, if $.a = [1,2,3], and user tries to set $.a.b = 2, the array will be overrided by the object.

### DIFF
--- a/src/modules/flow/json/json.c
+++ b/src/modules/flow/json/json.c
@@ -1008,11 +1008,46 @@ json_serialize(struct sol_buffer *buffer, struct json_element *element)
     return -EINVAL;
 }
 
+static struct json_key_element *
+json_object_get_child_element(struct json_element *element, const char *key)
+{
+    uint16_t i;
+    struct json_key_element *key_element;
+
+    SOL_VECTOR_FOREACH_IDX(&element->children, key_element, i)
+        if (streq(key, key_element->key))
+            return key_element;
+
+    return NULL;
+}
+
+static int
+json_object_add_new_element(struct json_element *base_element, const char *key, struct json_element *new_element)
+{
+    struct json_key_element *new;
+
+    new = json_object_get_child_element(base_element, key);
+    if (!new) {
+        new = sol_vector_append(&base_element->children);
+        SOL_NULL_CHECK(new, -errno);
+        new->key = strdup(key);
+        SOL_NULL_CHECK_GOTO(new->key, str_error);
+    } else
+        json_element_clear(&new->element);
+    new->element = *new_element;
+
+    return 0;
+
+str_error:
+    sol_vector_del(&base_element->children, base_element->children.len - 1);
+    return -ENOMEM;
+}
+
 static int
 json_object_in_process(struct sol_flow_node *node, void *data, uint16_t port, uint16_t conn_id, const struct sol_flow_packet *packet)
 {
-    struct json_key_element *new;
     struct json_element *mdata = data;
+    struct json_element new_element;
     const char *key;
     uint16_t len;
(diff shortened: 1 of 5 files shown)


src/modules/flow/json/json.c
```diff
@@ -405,6 +405,22 @@ json_path_is_array_key(struct sol_str_slice slice)
...
+json_array_get_key_index(struct sol_str_slice key)
+{
+    long int index_val;
+    char *endptr;
+
+    errno = 0;
+    //TODO: use safe strtol from sol-util when available
+    index_val = strtol(key.data + 1, &endptr, 10);
+    if (endptr != key.data + key.len - 1 ||
+        index_val < 0 || errno > 0 || index_val > INT32_MAX)
```
lpereira: You should also check if the value's smaller than `INT32_MIN`. A way to do that without adding another branch, is to perform `if ((long int)(int32_t)index_val != index_val)`, and remove the `if (index_val > INT32_MAX)` test.

src/modules/flow/json/json.c
```diff
@@ -405,6 +405,22 @@ json_path_is_array_key(struct sol_str_slice slice)
...
+        return errno > 0 ? -errno : -EINVAL;
```
lpereira: Since `errno` is tested twice, just move that to a condition just for it: `if (errno) return -errno;`. No need to compare if it's greater than zero.

src/modules/flow/json/json.c
```diff
@@ -1008,11 +1029,318 @@ json_serialize(struct sol_buffer *buffer, struct json_element *element)
...
+    SOL_NULL_CHECK(key_element, NULL);
+    key_element->key = sol_str_slice_to_string(key_slice);
+    SOL_NULL_CHECK_GOTO(key_element->key, str_error);
+
+    key_element->element.type = JSON_TYPE_UNKNOWN;
+
+    return key_element;
+
+str_error:
+    sol_vector_del(&element->children, element->children.len - 1);
```
lpereira: This kind of error handling where the last item is removed happens a lot in with the vector stuff. Might be a nice idea, for another patch, to have a `sol_vector_del_element()` that takes a pointer to an element, calculates the offset, and finally calls `sol_vector_del()`.

src/modules/flow/json/json.c
```diff
@@ -1008,11 +1029,318 @@ json_serialize(struct sol_buffer *buffer, struct json_element *element)
...
+        return sol_vector_get(&element->children, i);
+
+    last_len = element->children.len;
+    new = sol_vector_append_n(&element->children,
+        i - element->children.len + 1);
+    SOL_NULL_CHECK(new, NULL);
+
+    for (i = last_len; i < element->children.len; i++) {
+        new = sol_vector_get(&element->children, i);
+        SOL_NULL_CHECK(new, NULL);
```
lpereira: Won't the items added in line `1066` leak here?

src/modules/flow/json/json.c
```diff
@@ -1008,11 +1029,318 @@ json_serialize(struct sol_buffer *buffer, struct json_element *element)
...
+    }
+
+    return 0;
+
+error_parse:
+error_key:
+    free(new->key);
+
+error_append:
+    if (r == 0)
```
lpereira: `r` might be uninitialized here. (It's not initialized in line `1132`.)

otaviobp:
merged with proposed fixes

PR #1028, opened 2015-11-09
Create a node to build URLs. v5

cabelitos:
Changes since v4:
* Added some docs at sol-http.h
* Added commit fbe649b
* Added a functions in 9cefe03 (sol_str_slice_starts_with/ sol_str_slice_str_starts_with)
* Changed how URI authority and path is parsed.
* Dropped the commit that created sol_buffer_steal_or_copy(), since is not needed anymore.
* Some minor bugs there were found during the review.

### DIFF
--- a/src/lib/comms/include/sol-http.h
+++ b/src/lib/comms/include/sol-http.h
@@ -112,7 +112,7 @@ struct sol_http_param_value {
             bool value;
         } boolean;
         struct {
-            int value;
+            int32_t value;
         } integer;
         struct {
             const struct sol_str_slice value;
--- a/src/lib/comms/include/sol-http.h
+++ b/src/lib/comms/include/sol-http.h
@@ -135,7 +135,18 @@ struct sol_http_response {
     int response_code;
 };
 
-#define SOL_HTTP_RESPONSE_CHECK_API(response_, ...) \
+struct sol_http_url {
+    struct sol_str_slice scheme;
+    struct sol_str_slice user;
+    struct sol_str_slice password;
+    struct sol_str_slice host;
+    struct sol_str_slice path;
+    struct sol_str_slice query;
+    struct sol_str_slice fragment;
+    uint32_t port; //If == 0 ignore
+};
+
+#define SOL_HTTP_RESPONSE_CHECK_API(response_, ...)  \
     do { \
         if (unlikely(!response_)) { \
             SOL_WRN("Error while reaching service."); \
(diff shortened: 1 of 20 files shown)


cabelitos:
Merged with changes.

src/lib/comms/include/sol-http.h
```diff
@@ -236,9 +250,108 @@ sol_http_param_init(struct sol_http_param *params)
...
     struct sol_http_param_value value) SOL_ATTR_WARN_UNUSED_RESULT;
+bool sol_http_param_add_copy(struct sol_http_param *params, struct sol_http_param_value value);
 void sol_http_param_free(struct sol_http_param *params);
 
-int sol_http_escape_string(char **escaped, const char *value);
+/**
+ * Encodes an URL string.
+ *
+ * If the string was not encoded this function will simple copy
+ * the slices content to the buffer and set @c SOL_BUFFER_FLAGS_MEMORY_NOT_OWNE.D
```
barbieri: there is a "." in the flag name.

Also check Doxygen generated file, I think that you can use # before the name to link to that flag, like `#SOL_BUFFER_FLAGS_MEMORY_NOT_OWNED`

src/lib/comms/include/sol-http.h
```diff
@@ -236,9 +250,108 @@ sol_http_param_init(struct sol_http_param *params)
...
+ */
+int sol_http_create_uri(char **uri_out, const struct sol_http_url url, const struct sol_http_param *params);
+
+/**
+ * A simpler version of sol_http_create_uri().
+ *
+ *
+ * @param uri The created URI - it should be freed using free().
+ * @param base_uri The base uri to be used.
+ * @param params The query and cookies params.
```
barbieri: `and cookies`? There is no meaning to have cookies in an URI :-D (same for the `sol_http_create_uri()`)

src/lib/comms/include/sol-http.h
```diff
@@ -236,9 +250,108 @@ sol_http_param_init(struct sol_http_param *params)
...
+ *
+ * @return 0 on success, negative number on error.
+ */
+int sol_http_create_simple_uri(char **uri, const struct sol_str_slice base_uri, const struct sol_http_param *params);
+
+/**
+ * Encodes http parameters of a given type.
+ *
+ * @param buf Where the encoded parameters will be stored as string.
```
barbieri: make it clear the buffer will not be reset, data will be appended to it. Thus any previously set data will be kept.

src/lib/comms/include/sol-http.h
```diff
@@ -236,9 +250,108 @@ sol_http_param_init(struct sol_http_param *params)
...
+ * @param params The parameters to be encoded.
+ *
+ * @return 0 on success, negative number on error.
+ */
+int sol_http_encode_params(struct sol_buffer *buf, enum sol_http_param_type type, const struct sol_http_param *params);
+
+/**
+ * Decodes http parameters of a given type.
+ *
+ * @param buf Where the decoded parameters will be stored as string.
```
barbieri: what? buf? Copy and pasta hell

src/lib/comms/include/sol-http.h
```diff
@@ -236,9 +250,108 @@ sol_http_param_init(struct sol_http_param *params)
...
+ * Split an URI.
+ *
+ * This function will receive a complete URI and split and store its values in the sol_http_url struct.
+ *
+ * @param full_uri An URI to be splitted.
+ * @param url Where the URL parts will be stored.
+ *
+ *
+ * @return 0 on success, negative number on error.
+ * @note This function will not decoded the URI.
```
barbieri: almost good! But say what one should use to decode. You should list these functions with doxygen's `@see function()` commands.

src/lib/comms/include/sol-http.h
```diff
@@ -236,9 +250,108 @@ sol_http_param_init(struct sol_http_param *params)
...
+ */
+int sol_http_split_uri(const struct sol_str_slice full_uri, struct sol_http_url *url);
+
+/**
+ * An wrapper of over sol_http_create_simple_uri()
```
barbieri: ... taking a char pointer instead of a slice.

And in `sol_http_create_simple_uri()` you add a `@see sol_http_create_simple_uri_from_str()`, always try to add cross references to help the user.

src/lib/comms/include/sol-http.h
```diff
@@ -236,9 +250,108 @@ sol_http_param_init(struct sol_http_param *params)
 
 bool sol_http_param_add(struct sol_http_param *params,
     struct sol_http_param_value value) SOL_ATTR_WARN_UNUSED_RESULT;
+bool sol_http_param_add_copy(struct sol_http_param *params, struct sol_http_param_value value);
```
barbieri: missing documentation! And take the time to add documentation to other functions, including cross references between related.

src/lib/comms/include/sol-http.h
```diff
@@ -236,9 +250,108 @@ sol_http_param_init(struct sol_http_param *params)
...
 void sol_http_param_free(struct sol_http_param *params);
 
-int sol_http_escape_string(char **escaped, const char *value);
+/**
+ * Encodes an URL string.
+ *
+ * If the string was not encoded this function will simple copy
+ * the slices content to the buffer and set @c SOL_BUFFER_FLAGS_MEMORY_NOT_OWNE.D
+ *
+ * @param buf The buffer that will hold the encoded string.
```
barbieri: Say this buffer is being initialized in this function. If buffer contained some data, it will be lost as no `sol_buffer_fini()` is lost.

Make it clear the resulting buffer, if no encode was needed, will point to value.data and the flag will be set to SOL_BUFFER_FLAGS_MEMORY_NOT_OWNED, as you mentioned in the description of the function. Say that buffer should be finalized with `sol_buffer_fini()` and that if a copy is needed, users should call `sol_buffer_steal_or_copy()`.

src/lib/comms/include/sol-http.h
cabelitos: Which char pointer are you talking about? If it is URI, I think it makes no sense to change to slice, the URI variable will point to the URI created by sol_http_create_simple_uri() that the user should free it.

src/lib/comms/include/sol-http.h
cabelitos: OOps, sorry, forgot to document this one.

src/lib/comms/include/sol-http.h
cabelitos: Ok, I will.

src/lib/comms/include/sol-http.h
cabelitos: haha, Oops.

src/lib/comms/include/sol-http.h
cabelitos: K

src/lib/comms/include/sol-http.h
cabelitos: :/

src/lib/comms/include/sol-http.h
cabelitos: Sounds good.

src/lib/comms/include/sol-http.h
barbieri: oops, I mean to add that to the documentation! No changes to the code/behavior!

```
- * An wrapper of over sol_http_create_simple_uri()
+ * An wrapper of over sol_http_create_simple_uri() taking a char pointer base_url instead of a slice.
```

cabelitos: Oh, ok.

PR #1030, opened 2015-11-09
New OIC demo and commits necessary to support it - v3

otaviobp:
changelog:
 - using a timer to start scan
 - if making a device un-observable fails when selecting a new device, don't block the operation. just warn
 - documenting demo in README
 - Adding scan demo to makefile

Replaces #1026


otaviobp:
merged.

PR #1031, opened 2015-11-09
tests: count get responses before making a POST request

bdilly:
Otherwise a post may be received before get and returned value
will be wrong.


otaviobp:
+1

bdilly:
merged, thanks

PR #1032, opened 2015-11-10
Fix coverity issues

bdilly:
@otaviobp , @dorileo please take a look on this


bdilly:
merged dropping false positive commit

PR #1034, opened 2015-11-10
Create sol_vector_del_element function

otaviobp:
Add function to remove an element from vector. So far, this is usefull
for error handling in json.c file.

Function suggested by @lpereira in json review.

### DIFF
--- a/src/lib/datatypes/include/sol-vector.h
+++ b/src/lib/datatypes/include/sol-vector.h
@@ -95,6 +95,7 @@ sol_vector_get(const struct sol_vector *v, uint16_t i)
 }
 
 int sol_vector_del(struct sol_vector *v, uint16_t i);
+int sol_vector_del_element(struct sol_vector *v, const void *elem);
 
 void sol_vector_clear(struct sol_vector *v);
 
@@ -164,6 +165,12 @@ static inline int
 sol_ptr_vector_del(struct sol_ptr_vector *pv, uint16_t i)
 {
     return sol_vector_del(&pv->base, i);
+}
+
+static inline int
+sol_ptr_vector_del_element(struct sol_ptr_vector *pv, const void *elem)
+{
+    return sol_vector_del_element(&pv->base, elem);
 }
 
 static inline void *
--- a/src/lib/datatypes/sol-vector.c
+++ b/src/lib/datatypes/sol-vector.c
@@ -265,3 +265,19 @@ sol_ptr_vector_init_n(struct sol_ptr_vector *pv, uint16_t n)
         return -errno;
     return 0;
 }
+
+SOL_API int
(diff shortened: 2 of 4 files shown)


barbieri:
+1, I could see lots of usage for that... grep for `sol_vector_del(&v, v.len - 1)` for many cases where we delete the last entry.

We also have that for ptr-vectors... in such case it's hard to find the index from the pointer, as you would need to give the pointer-to-pointer to avoid O(n) lookup. To these cases a simple 'del_last()` could help

src/lib/datatypes/include/sol-vector.h
```diff
@@ -166,6 +167,12 @@ sol_ptr_vector_del(struct sol_ptr_vector *pv, uint16_t i)
     return sol_vector_del(&pv->base, i);
 }
 
+static inline int
+sol_ptr_vector_del_element(struct sol_ptr_vector *pv, const void *elem)
+{
+    return sol_vector_del_element(&pv->base, elem);
```
barbieri: this is not what I'd expect, as elem in a ptr-vector is the pointed-to... you'd need to walk the vector looking for elem.

src/lib/datatypes/sol-vector.c
```diff
@@ -265,3 +265,19 @@ sol_ptr_vector_init_n(struct sol_ptr_vector *pv, uint16_t n)
...
+{
+    ssize_t offset;
+
+    offset = (const char *)elem - (const char *)v->data;
+    if (offset % v->elem_size)
+        return -EINVAL;
+
+    offset = offset / v->elem_size;
+    if (offset < 0 || offset >= v->len)
+        return -EINVAL;
```
barbieri: maybe -ENOENT?

src/test/test-vector.c
```diff
@@ -360,4 +361,51 @@ vector_initializes_elements_to_zero(void)
 }
 
 
+DEFINE_TEST(test_vector_del_element);
+
+static void
+test_vector_del_element(void)
```
barbieri: please also add a test to `sol_ptr_vector_del_element()`, I think your wrapper is not correct.

src/lib/datatypes/include/sol-vector.h
otaviobp: That's true. I'll fix that and add the del_end you proposed too.

src/lib/datatypes/sol-vector.c
otaviobp: I used the same return code from sol_vector_del when the index was invalid. But for this case maybe NOENT is better

src/test/test-vector.c
otaviobp: ok

PR #1035, opened 2015-11-10
form: new string form node, with tests.

glima:
This is a good, old text entry.


bdilly:
@glima it seems nice. Just check this blinking time please.

PR #1037, opened 2015-11-11
storage headers: Do not include sol-util.h

edersondisouza:
CREATE_BUFFER macro as using sol_util_memdup, but as sol-util.h is not
installed, it was not possible to compile sol-fbp-generator generated
code.
Using malloc and memcpy instead.


otaviobp:
+1

edersondisouza:
Integrated. Tks.

PR #1041, opened 2015-11-11
Fix segfault when request is triggered without setting an url

otaviobp:


otaviobp:
merged.

PR #1042, opened 2015-11-11
Add support for Amazon IoT and IBM Bluemix

rchiossi:
This patchset adds connectivity support for Amazon IoT and IBM Bluemix through MQTT nodes. It doesn't provide specific nodes, but adds all the infrastructure needed to connect to those services using Soletta.

In order to have an standard way to handle certificates, this patchset introduces an abstraction to lead certificates. Currently it is only available for the Linux platform. As in this platform certificates are usually handled by third party libraries, we only really need the path to those certificates, so the linxu implementation just search for the file in known paths and keep its full path. Implementation for other platforms may need to keep the certificate data itself within this structure.

To reduce the amount of allocations when performing path search in sol-certificate, the sol_buffer_take_data was created. It mimics the behavior of sol_vector_take_data, but for sol_buffer.

Connection to Amazon AWS IoT cloud requires MQTT to use a TLS connection, so we add APIs to allow the certificates do be configured in sol-mqtt and internally provide that info to mosquitto.

This patch also features the previously reviewed patches (like the MQTT flow module) with minor changes to support TLS connections and updates to documentation according to previous review.


src/lib/common/sol-certificate-impl-linux.c
```diff
@@ -0,0 +1,162 @@
...
+find_cert(const char *filename, const char *const paths[])
+{
+    const char *ssl_cert_dir = getenv("SSL_CERT_DIR");
+    struct sol_buffer buffer = SOL_BUFFER_INIT_EMPTY;
+    struct stat st;
+    int idx;
+    int r;
+
+    /* Check absolute path */
+    if (stat(filename, &st) == 0)
```
barbieri: check if it's a file at least? return NULL and WRN if a directory or no permissions.

src/lib/common/sol-certificate-impl-linux.c
```diff
@@ -0,0 +1,162 @@
...
+        return strdup(filename);
+
+    /* Check SSL_CERT_DIR */
+    r = sol_buffer_append_printf(&buffer, "%s/%s", ssl_cert_dir, filename);
```
barbieri: if (ssl_cert_dir)... it may be null if no env is set.

src/lib/common/sol-certificate-impl-linux.c
```diff
@@ -0,0 +1,162 @@
...
+    SOL_INT_CHECK(r, != 0, NULL);
+
+    if (stat(buffer.data, &st) == 0)
+        return sol_buffer_take_data(&buffer);
+
+    sol_buffer_reset(&buffer);
+
+    /* Search known paths */
+    for (idx = 0; search_paths[idx] != 0; idx++) {
+        r = sol_buffer_append_printf(&buffer, "%s/%s/%s", SYSCONF, search_paths[idx], filename);
```
barbieri: minor optimization here, but in linux you could `fstatat()`, given a fd to SYSCONF (opendir() + dirfd()). This will save you parts of the `sol_buffer_append_printf()` and may help the kernel during lookup, since it will not need to resolve the whole path (and check access).

barbieri: btw, no need to change, just a heads up for future work

src/lib/common/sol-certificate-impl-linux.c
```diff
@@ -0,0 +1,162 @@
...
+
+insert_error:
+    free(cert);
+cert_alloc_error:
+    free(path);
+    return NULL;
+}
+
+SOL_API void
+sol_cert_free(struct sol_cert *cert)
```
barbieri: this should be called `unref()` instead of `free()`.

src/modules/flow/mqtt/mqtt.c
```diff
@@ -0,0 +1,452 @@
...
+{
+    struct client_data *mdata = data;
+    int r;
+    const char *in_value;
+
+    r = sol_flow_packet_get_string(packet, &in_value);
+    SOL_INT_CHECK(r, < 0, r);
+
+    free(mdata->host);
+    mdata->host = strdup(in_value);
```
barbieri: https://github.com/solettaproject/soletta/blob/master/src/shared/sol-util.h#L465 we have an efficient helper for that pattern.

src/modules/flow/mqtt/mqtt.c
```diff
@@ -0,0 +1,452 @@
...
+    int r;
+    struct sol_blob *in_value;
+
+    r = sol_flow_packet_get_blob(packet, &in_value);
+    SOL_INT_CHECK(r, < 0, r);
+
+    if (mdata->payload)
+        sol_blob_unref(mdata->payload);
+
+    sol_blob_ref(in_value);
```
barbieri: shouldn't be an issue here, but it's always better to do ref of the new value before doing the unref of the old one. In some cases you may get the same handle and it may be at the last reference, thus doing the unref before will free the handle, that you later ref.

src/modules/flow/mqtt/mqtt.json
```diff
@@ -0,0 +1,203 @@
...
+          "data_type": "string",
+          "description": "MQTT client ID. Must be valid if {clean_session} is not set",
+          "methods": {
+            "process": "id_process"
+          },
+          "name": "ID"
+        },
+        {
+          "data_type": "int",
+          "description": "Time in seconds between PING messages from the Broker to the client",
```
barbieri: we are trying to keep all timeouts as milliseconds, keep it consistent :-)

src/test/test-certificate.c
```diff
@@ -0,0 +1,100 @@
...
+
+static void
+create_dummy_certificate(void)
+{
+    FILE *fp = fopen("dummy.pem", "w");
+    int r;
+
+    ASSERT(fp != NULL);
+
+    r = fprintf(fp, dummy_cert);
```
barbieri: this should be either fwrite() or fprintf(fp, "%s", dummy_cert)

src/lib/common/sol-certificate-impl-linux.c
rchiossi: if its null you get a valid string "(null)/filename" which will make the code just skip the following check. But anyway, checking for NULL is definitely faster. I'll update.

src/lib/common/sol-certificate-impl-linux.c
rchiossi: ack

src/modules/flow/mqtt/mqtt.c
rchiossi: Hmm... this actually uncovered a problem. sol_util_replace_str_if_changed() should be applicable here only if mdata->host is always valid, but in some scenarios, like in Bluemix example, the host is unknown when the flow start.

What would be better: let host be compulsory and keep it as is, which will make the user put a dummy value in the node config, or let if be optional and default to something like 127.0.0.1?

src/modules/flow/mqtt/mqtt.c
rchiossi: Makes sense. I will just swap those lines.

src/modules/flow/mqtt/mqtt.json
rchiossi: Ok, I will change it in the mqtt lib itself and update documentation.

src/test/test-certificate.c
rchiossi: It shouldn't be an issue here since the format string is a const string and has no format notations in it, but better keep it clean. I will update it.

src/lib/common/sol-certificate-impl-linux.c
rchiossi: Indeed, but in that case I cant just return sol_buffer_steal() since the data in buffer does not contain SYSCONF, so I would have to do an extra sol_buffer_append_printf() before returning.

src/lib/common/sol-certificate-impl-linux.c
rchiossi: if (stat(filename, &st) == 0 && S_ISREG(st.st_mode) && st.st_mode & S_IRUSR)

or 

if (stat(filename, &st) == 0) {
    if (S_ISREG(st.st_mode) && st.st_mode & S_IRUSR) {
          SOL_WRN("Not found")
          return NULL;
    }

}

rchiossi:
merged

PR #1045, opened 2015-11-11
Basic twitter node

ceolin:
It allows one post (tweet) and get the timeline


src/modules/flow/twitter/twitter.c
```diff
@@ -0,0 +1,617 @@
...
+    struct sol_buffer buffer = SOL_BUFFER_INIT_EMPTY;
+    struct sol_buffer encode_buf;
+    struct twitter_data *mdata = sol_flow_node_get_private_data(cb_data->node);
+
+    r = sol_buffer_append_as_base64(&buffer, sol_str_slice_from_blob(output), SOL_BASE64_MAP);
+    SOL_INT_CHECK_GOTO(r, < 0, err);
+
+    r = sol_http_encode_slice(&encode_buf, sol_str_slice_from_str((char *)buffer.data));
+    SOL_INT_CHECK_GOTO(r, < 0, err_escape);
+    escaped_signature = sol_str_slice_to_string(sol_buffer_get_slice(&encode_buf));
```
barbieri: use sol_buffer_steal().

barbieri: actually just use encode_buf.data and move `sol_buffer_fini()` later on.

src/modules/flow/twitter/twitter.c
```diff
@@ -0,0 +1,617 @@
...
+{
+    int r = -EINVAL;
+    struct sol_blob *blob;
+    struct callback_data *cb_data;
+    struct sol_message_digest *digest;
+    struct sol_buffer buf;
+    struct twitter_data *mdata = sol_flow_node_get_private_data(node);
+    char *signature, *params, *escaped_params, *escaped_status;
+    struct sol_message_digest_config *digest_config = &(struct sol_message_digest_config) {
+        .api_version = SOL_MESSAGE_DIGEST_CONFIG_API_VERSION,
```
barbieri: use @ibriano macro to set api version.

src/modules/flow/twitter/twitter.c
```diff
@@ -0,0 +1,617 @@
...
+    cb_data = calloc(1, sizeof(*cb_data));
+    SOL_NULL_CHECK(cb_data, -ENOMEM);
+
+    cb_data->node = node;
+    cb_data->cb = post_request;
+
+    cb_data->nonce = generate_nonce();
+    SOL_NULL_CHECK_GOTO(cb_data->nonce, err);
+
+    r = asprintf(&cb_data->timestamp, "%li", (long int)time(NULL));
```
barbieri: why timestamp is kept as a string? Just format it to string when needed.

src/modules/flow/twitter/twitter.c
```diff
@@ -0,0 +1,617 @@
...
+    r = asprintf(&cb_data->key, "%s&%s", mdata->consumer_secret,
+        mdata->token_secret);
+    SOL_INT_CHECK_GOTO(r, < 0, err_key);
+
+    cb_data->status = strdup(status);
+    SOL_NULL_CHECK_GOTO(cb_data->status, err_status);
+
+    r = sol_http_encode_slice(&buf, sol_str_slice_from_str(cb_data->status));
+    SOL_INT_CHECK_GOTO(r, < 0, err_escape_status);
+    escaped_status = sol_str_slice_to_string(sol_buffer_get_slice(&buf));
```
barbieri: move `sol_buffer_fini()` or use `sol_buffer_steal()`, this way you avoid one alloc.

barbieri: be aware that if no encode was done, then steal will fail. I guess moving the fini later is the best option

src/modules/flow/twitter/twitter.c
```diff
@@ -0,0 +1,617 @@
...
+        "&oauth_version=1.0"
+        "&status=%s",
+        mdata->consumer_key, cb_data->nonce,
+        cb_data->timestamp, mdata->token, escaped_status);
+    SOL_INT_CHECK_GOTO(r, < 0, err_params);
+
+    r = sol_http_encode_slice(&buf, sol_str_slice_from_str(params));
+    SOL_INT_CHECK_GOTO(r, < 0, err_escape_params);
+    escaped_params = sol_str_slice_to_string(sol_buffer_get_slice(&buf));
+    sol_buffer_fini(&buf);
```
barbieri: same here about fini,

src/modules/flow/twitter/twitter.c
```diff
@@ -0,0 +1,617 @@
...
+    SOL_INT_CHECK(r, < 0, r);
+
+    token_len = sizeof("oauth_token=") - 1;
+    token_secret_len = sizeof("oauth_token_secret=") - 1;
+
+    tokens = sol_util_str_split(
+        sol_str_slice_from_str(val), "&", 0);
+    SOL_VECTOR_FOREACH_IDX (&tokens, slice, idx) {
+        struct sol_str_slice token;
+        if (streqn(slice->data, "oauth_token_secret=", token_secret_len) &&
```
barbieri: use the new function that @cabelitos added slice-starts with.

@cabelitos, please review this commit as you added lots of new features that could improve it :-)

src/modules/flow/twitter/twitter.c
```diff
@@ -0,0 +1,617 @@
...
+    free(mdata->token);
+    free(mdata->token_secret);
+    sol_vector_clear(&tokens);
+    return r;
+}
+
+static int
+timeline_process(struct sol_flow_node *node, void *data, uint16_t port, uint16_t conn_id,
+    const struct sol_flow_packet *packet)
+{
```
barbieri: this is almost a replica of the other function above, please refactor to share these common bits.

src/modules/flow/twitter/twitter.c
```diff
@@ -0,0 +1,617 @@
...
+    struct sol_buffer buf;
+    struct sol_flow_node_type_twitter_client_options *opts =
+        (struct sol_flow_node_type_twitter_client_options *)options;
+
+    mdata->consumer_key = strdup(opts->consumer_key);
+    SOL_NULL_CHECK(mdata->consumer_key, -ENOMEM);
+
+    r = sol_http_encode_slice(&buf, sol_str_slice_from_str(opts->consumer_secret));
+    SOL_INT_CHECK_GOTO(r, < 0, err);
+    mdata->consumer_secret = sol_str_slice_to_string(sol_buffer_get_slice(&buf));
```
barbieri: `sol_buffer_steal_or_copy()`

src/modules/flow/twitter/twitter.c
```diff
@@ -0,0 +1,617 @@
...
+        sol_str_slice_from_str(BASE_POST_URL));
+    SOL_INT_CHECK_GOTO(r, < 0, err_escape);
+    mdata->escaped_post_url = sol_str_slice_to_string(sol_buffer_get_slice(&buf));
+    sol_buffer_fini(&buf);
+    SOL_NULL_CHECK_GOTO(mdata->escaped_post_url, err_escape);
+
+    r = sol_http_encode_slice(&buf,
+        sol_str_slice_from_str(BASE_TIMELINE_URL));
+    SOL_INT_CHECK_GOTO(r, < 0, err_escape_get);
+    mdata->escaped_get_url = sol_buffer_steal(&buf, NULL);
```
barbieri: for correctness purposes, use `steal_or_copy()`, in this case escape should always happen and thus steal will work, but this will avoid mistakes in the future.

src/modules/flow/twitter/twitter.json
```diff
@@ -0,0 +1,75 @@
...
+  "name": "twitter",
+  "meta": {
+    "author": "Intel Corporation",
+    "license": "BSD-3-Clause",
+    "version": "1"
+  },
+  "types": [
+    {
+      "name": "twitter/client",
+      "category": "output/network",
```
barbieri: socialnetwork/client is a better category

src/modules/flow/twitter/twitter.json
```diff
@@ -0,0 +1,75 @@
...
+          },
+          "name": "TOKEN"
+        },
+	{
+          "data_type": "string",
+          "description": "Post the string that comes into twitter",
+          "methods": {
+            "process": "post_process"
+          },
+          "name": "TWEET"
```
barbieri: we'll never cover all APIs by adding ports. I believe a more extensible approach is to create one node per supported function, like `twitter/tweet` and `twitter/get-timeline`. This also allow us to offer some relevant input or output ports in cases where these are easier to do.

Like `twitter/tweet` could take multiple ports, one to specify a location (that we could get from gps), one to get the text and another `POST` to do the action.

src/modules/flow/twitter/twitter.json
```diff
@@ -0,0 +1,75 @@
...
+      "out_ports": [
+        {
+          "data_type": "json-object",
+          "description": "The result as json-objectt",
+          "name": "OBJECT"
+        },
+	{
+          "data_type": "json-array",
+          "description": "The result as json-array",
+          "name": "ARRAY"
```
barbieri: `twitter/tweet` would have only OBJECT, while `twitter/get-timeline` would have only ARRAY?

src/modules/flow/twitter/twitter.c
```diff
@@ -0,0 +1,617 @@
...
+    nonce = calloc(1, size + 1);
+    SOL_NULL_CHECK_GOTO(nonce, end);
+
+    r = sol_util_base16_encode(nonce, size, slice, false);
+    if (r < 0) {
+        SOL_WRN("Could not enconde the oauth_nonce");
+        free(nonce);
+        nonce = NULL;
+    } else {
+        nonce[size] = '\0';
```
edersondisouza: Is this needed since you allocated `nonce` with `calloc`? But maybe allocating with `malloc` is better, then leaving this `\0` set as is.

src/modules/flow/twitter/twitter.c
```diff
@@ -0,0 +1,617 @@
...
+        cb_data->timestamp, mdata->token);
+    SOL_INT_CHECK_GOTO(r, < 0, err_header);
+
+    connection = cb_data->cb(cb_data, authorization_header);
+    SOL_NULL_CHECK_GOTO(connection, err_connection);
+
+    r = sol_ptr_vector_append(&mdata->pending_conns, connection);
+    SOL_INT_CHECK_GOTO(r, < 0, err_append);
+
+    goto err_connection;
```
edersondisouza: Curious that normal flow goes to error =D

edersondisouza:
Will we have samples?

ceolin:
no, we won't. An example depends of an twitter's application and distribute the tokens here, we are avoiding it

src/modules/flow/twitter/twitter.json
ceolin: yep, for those requests. But i've added those ports with generic names to be used with other kinds of requests

src/modules/flow/twitter/twitter.json
ceolin: this a different approach and a i really liked

src/modules/flow/twitter/twitter.json
ceolin: ack

src/modules/flow/twitter/twitter.c
ceolin: it will not work,  ``sol_http_encode_slice`` receives a slice now, which means the encoded buffer will not contains ``'\0`` :(

src/modules/flow/twitter/twitter.c
ceolin: the problem is the lack of ``\0`` on buffer

src/modules/flow/twitter/twitter.c
ceolin: it's used in two different places, i prefer keep the string instead of have to create/release it twice

src/modules/flow/twitter/twitter.c
ceolin: there are some problems here
- sol_buffer_steal can fail if no encoded (and we still need the data)
- there is no '\0' in the encoded buffer

I'm going to add new functions on buffer to reach these problems:

``sol_bufffer_steal_or_copy``
``sol_buffer_steal_as_string``
``sol_buffer_steal_or_copy_as_string``

src/modules/flow/twitter/twitter.c
ceolin: it's just a label, i can change the name

src/modules/flow/twitter/twitter.c
edersondisouza: Aren't we overcomplicating the API? Here, all (that I saw) these `char*` that we are get are being printed into strings, so a simple `sol_buffer_get_slice` and `SOL_STR_SLICE_PRINT` would solve the problem. Maybe use more `sol_str_slice` on our APIs?
Maybe we have to give a step back and think more about our API.

ceolin: maybe ... but in this case (and several others as figured out talking with iscaro and otavio) i don't only need to print the data, i need the memory data. I can not trust in ``sol_buffer_steal`` because if nothing was encoded it will return ``NULL`` and the data is still necessary so ... the steal_or_copy is still valid.

The other two are still thinking about

ceolin: s/are/i am/

ceolin:
@barbieri for this patch, only the ``sol_buffer_steal_or_copy`` is good enough, then the data can be used as slice and avoid what you said.

I'll check if for curl and microhttpd implementation the ``sol_buffer_steal_as_string`` is valid as these code are dealing with external libraries that expect strings.

src/modules/flow/twitter/twitter.c
cabelitos: Just a note If enconding was necessary, the \0 will be present.

src/modules/flow/twitter/twitter.c
cabelitos: I will take a look! =)

src/modules/flow/twitter/twitter.c
ceolin: yep, but if not it isn't and it's pretty bad because we will always have to check/treat it.

ceolin:
fixes on https://github.com/solettaproject/soletta/pull/1052

src/modules/flow/twitter/twitter.c
barbieri: You never need to use the timestamp as string alone, all usages I saw is to feed it to `asprintf()`,  thus you don't nee to give it %s, use %li instead.

src/modules/flow/twitter/twitter.c
barbieri: I don't get this. You're only using `escaped_status` in the `asprintf()` above, so why don't you just use the `sol_buffer_get_slice(&buf)` and print the slice? An empty slice should print nothing, even if `slice.data == NULL`.

src/modules/flow/twitter/twitter.c
barbieri: look at your code:
 - if encoded, then we guarantee a null byte
 - if not encoded, it's the same buffer you gave it, that is opts->consumer_secret, that is known to be null-terminated, after all you created the slice from that string using the null terminator.

You can keep like that since it's the open() node and happens only once, also because you need to dup opts->consumer_secret since we shouldn't keep references to opts pointers. But consider this for other places

src/modules/flow/twitter/twitter.json
barbieri: you liked it because you did a very simplistic version of only 2 nodes. This doesn't scale.

If we come back to this module to add all supported API version, then what we're going to do? Increase its version and add ports like crazy? :-D

It's better to reuse the C functions but provide specific nodes that allow using Twitter's API, like I said tweet could get a location, etc.

src/modules/flow/twitter/twitter.json
barbieri: you can keep those in a twitter/client, or twitter/generic, that takes a JSON_OBJECT and JSON_ARRAY and an endpoint, outputs JSON_OBJECT or JSON_ARRAY.

ceolin: ok, so for this basic client you are proposing change the name of the out ports ?

src/modules/flow/twitter/twitter.json
ceolin: nope, i thought we were asking to create several different nodes with specific functionalities, for example:

``twitter/tweet`` -> just post a tweet (with its possible options)
``twitter/timeline`` -> just deal with timeline
``twitter/account`` -> change photo, personal information, and so on

Of course, we can have a generic version that receives a json, but i guess those small nodes will be easier to use on a flow

src/modules/flow/twitter/twitter.c
ceolin: you are true, i forgot the ``%.*s`` format and overcomplicate the problem. Sorry for that

src/modules/flow/twitter/twitter.c
ceolin: yep, i already have changed that in v2. I was using that as string previously and forgot to change. All the parameters were being set as http_param separated and not as an only string on header. For that reason i kept that strings, when changes to use only one string header it did not make sense anymore.

src/modules/flow/twitter/twitter.json
barbieri: I want both kinds of node. We can do a `twitter/generic` that doesn't know about format, all it does is take an array or object, output an array or object.

Then we can have `twitter/tweet`, `twitter/timeline` and friends, that are easier to use.

As for the port name,s I'm okay with your names.

PR #1047, opened 2015-11-11
build: more bad obj dependency issues fixed.

glima:
The fixes here were mostly spotted by make -j runs of some targets (like
samples), which would sometimes reach use of some generated dependencies
before they were fully made, leading to errors that were recoverable
from a subsequent make invocation.

First of all, the inclusion of the .dep files was totally broken, so the
files listed in them were not being checked for dependencies at all! We
now moved the inclusion to a place where "all-objs" is already declared
and we're fine.

Some files were including "sol_config.h" by hand, and by fixing some
dependency declarations this will not only be unecessary but impossible
-- only the automatic compiler line inclusion will pass it forward.

"extra-headers" usage in "parse-common-module" was taken off -- it was
passing forward wrong objects and had only one user, that now users the
general "headers-$(whatever)" syntax.

Finally, "make samples" invocation right away after kconfig is working
again, though we still have samples depending on the whole library
objects set (we'll fix that later if we feel like declaring the
individual sample dependencies by hand). On the other hand, now that
.dep files *do* work, changes in headers, modules or any dependency of
another module will trigger its build again.

Thanks a lot for more help of Dorileo here :)

----

Please help to test this!!

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -28,10 +28,6 @@ else
 include $(top_srcdir)tools/build/Makefile.kconfig
 endif
 
-ifeq (,$(filter $(dep-avoid-targets),$(MAKECMDGOALS)))
--include ${all-objs:.o=.o.dep}
-endif
-
 ifneq (,$(NOT_FOUND))
 warning:
 	$(Q)echo -e "The following (required) dependencies were not met:\n"
@@ -58,6 +54,10 @@ endif # HAVE_KCONFIG_CONFIG
 endif # NOT_FOUND
 
 $(KCONFIG_CONFIG): $(KCONFIG_GEN)
+
+ifeq (,$(filter $(dep-avoid-targets),$(MAKECMDGOALS)))
+-include ${all-objs:.o=.o.dep}
+endif
 
 .DEFAULT_GOAL = all
 .PHONY = $(PHONY) default_target
--- a/src/lib/comms/Makefile
+++ b/src/lib/comms/Makefile
@@ -87,26 +87,26 @@ obj-networking-$(MQTT) += \
 obj-networking-$(MQTT)-extra-ldflags += \
 	$(MOSQUITTO_LDFLAGS)
 
-headers-networking-$(NETWORK) += \
+headers-$(NETWORK) += \
     include/sol-network.h
 
-headers-networking-$(COAP) += \
+headers-$(COAP) += \
     include/sol-coap.h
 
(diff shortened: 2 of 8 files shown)


ibriano:
Looks good and seems to have fixed mostly everything. But straight `make samples` still fails for static builds, with a different error now.

dorileo:
LGTM -  worth checking the error pointed by @ibriano but as far as I can see seems to be easy to fix.

dorileo:
BTW,  also, I think it's worth it setting the bots to also test for static build,  hum?

dorileo:
@glima it seems we'll need to explicitly declare ```$(1).dep``` to depend on ```PRE_GEN```, the following patch (on top of yours) fixes the static build issue:

```patch
diff --git a/tools/build/Makefile.rules b/tools/build/Makefile.rules
index d28f2f5..f7498ea 100644
--- a/tools/build/Makefile.rules
+++ b/tools/build/Makefile.rules
@@ -324,7 +324,7 @@ $(1): $($(1)-src) $(1).dep
        $(Q)$(MKDIR) -p $(dir $(1))
        $(Q)$(TARGETCC) $(OBJ_CFLAGS) $($(2)-gen-cflags) $(sort $($(2)-cflags)) $($(1)-src) -c -o $(1) $(3)
 
-$(1).dep: $($(1)-src) $(all-hdr) $(find-gen-hdrs) $(call find-deps,$(2)) $(all-dest-hdr) $(KCONFIG_CONFIG) $(HEADER_GEN)
+$(1).dep: $($(1)-src) $(all-hdr) $(find-gen-hdrs) $(call find-deps,$(2)) $(all-dest-hdr) $(KCONFIG_CONFIG) $(HEADER_GEN) $(PRE_GEN)
        $(Q)$(MKDIR) -p $(dir $(1))
        $(Q)$(TARGETCC) $(OBJ_CFLAGS) $($(2)-gen-cflags) $(sort $($(2)-cflags)) -MM -MG $($(1)-src) -MT "$(1)" > $(1).dep
```

glima:
@dorileo, this diff of yours is a regression WRT ``touch build/soletta_sysroot/usr/include/soletta/sol-flow/gpio.`` and having only gpio and calamari rebuild under allmodconfig -- everything get rebuilt :( Let me try more here.

glima:
Sending v2.

PR #1048, opened 2015-11-11
sol-fbp: Add SOL_ATTR_PRINTF() to sol_fbp_log_print()

lpereira:
By declaring the prototype of this function with this attribute, the
compiler will perform a couple of checks to ensure that the varargs are
actually of the correct type for the printf-style formatting strings,
in addition to only allowing literal strings being passed to the format
string argument.

This fixes a few crashes in the generator while printing errors,
including a few call sites that passed user-controlled strings as the
formatting string.

All known places that had problems were fixed, and this attribute
will ensure these won't happen in the future.


lpereira:
Merged, thanks!

PR #1049, opened 2015-11-11
allover: Disambiguate int size on public facing API

ibriano:
In some cases, int won't necessary be the of the size the code using it
expects, so we need to make sure to specify that size where it matters.


barbieri:
+1

ibriano:
Merged

PR #1051, opened 2015-11-12
build: more bad obj dependency issues fixed [v2]

glima:
The fixes here were mostly spotted by make -j runs of some targets (like
samples), which would sometimes reach use of some generated dependencies
before they were fully made, leading to errors that were recoverable
from a subsequent make invocation.

First of all, the inclusion of the .dep files was totally broken, so the
files listed in them were not being checked for dependencies at all! We
now moved the inclusion to a place where "all-objs" is already declared
and we're fine.

Some files were including "sol_config.h" by hand, and by fixing some
dependency declarations this will not only be unecessary but impossible
-- only the automatic compiler line inclusion will pass it forward.

"extra-headers" usage in "parse-common-module" was taken off -- it was
passing forward wrong objects and had only one user, that now users the
general "headers-$(whatever)" syntax.

"make samples" invocation right away after kconfig is working again,
though we still have samples depending on the whole library objects
set (we'll fix that later if we feel like declaring the individual
sample dependencies by hand). On the other hand, now that .dep
files *do* work, changes in headers, modules or any dependency of
another module will trigger its build again.

One other thing is that explicit module dependency declaration had to
move back to our Makefiles, since at least on static builds we only have
that info to stall module builds till their dependencies are met on
parallel builds.

Thanks a lot for more help of Dorileo here :)

-------------

Changes since #1047:
- fixed issue raised on static builds

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -28,10 +28,6 @@ else
 include $(top_srcdir)tools/build/Makefile.kconfig
 endif
 
-ifeq (,$(filter $(dep-avoid-targets),$(MAKECMDGOALS)))
--include ${all-objs:.o=.o.dep}
-endif
-
 ifneq (,$(NOT_FOUND))
 warning:
 	$(Q)echo -e "The following (required) dependencies were not met:\n"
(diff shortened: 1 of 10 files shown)


glima:
Lots of build runs here and all going beautifully. It's Friday, what a better day to merge this? Doing that now.

PR #1052, opened 2015-11-12
Twitter v2

ceolin:
from v1
 - Added ``sol_buffer_steal_or_copy``
 - added a commit with a integer overflow fix on oauth
 - Use the returned buffer instead of create string (when the result of the encode is use directly)
 - save timestamp as ``long int`` instead of string
 - save the result of the encode as slice (using ``sol_buffer_steal_or_copy``) instead of as string to avoid deal with ``\0``
 - fix typo
 - added capabilities description
 - fix integer overflow
 - removed unecessary include
 - use ``SOL_SET_API_VERSION`` macro
 - Change the node's category


ceolin:
@barbieri i intend to split the node in several small nodes each one with just one functionality as you asked.
Just sending this version with the suggested fixes to not accumulate things

ceolin:
arr,  i forgot re-use the common code between both process, timeline and tweet. Doing it

bdilly:
@ceolin I would suggest (for next prs) to separate prs with fixes and prs with features.
On this PR you have 3 fixes that are independent from "adding twitter client". Even if you found them while you were writing this client.

ceolin:
@bdilly sure, i used to do this, sorry for this one

ceolin:
closed https://github.com/solettaproject/soletta/pull/1182

PR #1053, opened 2015-11-12
Sol-HTTP: Properly handle URLs with a password and without user.

cabelitos:
A bug in the parser was preventing the URL to be splitted correctly.

There was also a bug when the URL was being created, the password
was only inserted when the user was set.


otaviobp:
+1

cabelitos:
merged.

PR #1054, opened 2015-11-13
Some Fixes for HTTP. V2

cabelitos:
Changes since v2:

* Dropped commit "Sol-HTTP-Client: Username and password should be encoded.", since Curl already encodes the username and password.
* Added a FBP test for commit 0740590
* Added a TODO at commit 0740590 , since we're not if SOL_HTTP_PARAM_FRAGMENT should be considered a sol_http_param_type.
* set_cookies_from_params() no longers steals encoded buffer data.


cabelitos:
Merged with the fixes.

PR #1055, opened 2015-11-13
Board detect + assorted fixes

mbelluzzo:


src/lib/common/Kconfig
```diff
@@ -8,7 +8,7 @@ config BOARD_NAME
 
             Usually the board name is detected using either
             SOL_BOARD_NAME environment variable or instructions from
-            'board_detect.json' if DETECT_BOARD_NAME is enabled. Both
+            '99_board_detect.json' if DETECT_BOARD_NAME is enabled. Both
```
barbieri: sed too much? I don't think this is to say "99" in there

src/lib/common/sol-board-detect.c
```diff
@@ -207,8 +207,31 @@ _json_open_doc(const char *path, struct sol_json_scanner *scanner)
...
+    int r;
+    char *file_path;
+
+    if (!ent) //Just skip, no need to trigger a warning.
+        return false;
+
+    if (ent->d_type != DT_REG && ent->d_type != DT_LNK) //It won't be recursive
+        return false;
+
+    if (!ent->d_name || !ent->d_name[0])
```
barbieri: please filter to have only ".json", `strendswith(ent->d_name, ".json")`

tools/build/Makefile.vars
```diff
@@ -335,8 +335,8 @@ $(call add-template,$(FLOW_METATYPE_BUILTINS_H_IN),$(FLOW_METATYPE_BUILTINS_H))
 NODE_TYPE_SCHEMA := $(top_srcdir)data/schemas/node-type-genspec.schema
 NODE_TYPE_SCHEMA_DEST := $(build_flowdatadir)schemas/node-type-genspec.schema
 
-BOARD_DETECT := $(top_srcdir)data/jsons/board_detect.json
-BOARD_DETECT_DEST := $(build_datadir)board_detect.json
+BOARD_DETECT := $(top_srcdir)data/jsons/99_board_detect.json
+BOARD_DETECT_DEST := $(build_datadir)/boards/99_board_detect.json
```
barbieri: last one is always bad, usually people call this "50-default.json", so we know it is the default we provide and sits right in the "middle" of the range. People can add higher or lower priority versions as they see fit.

mbelluzzo: Roger that

mbelluzzo:
See V2

PR #1056, opened 2015-11-13
Build: Properly export makefile variables.

cabelitos:
The soletta package config file was being incorrectly create as
some variables were not being exported. This patch exports
the variables so the template.py can correctly create the
package config file.

Thanks Dorileo.


cabelitos:
Merged.

PR #1057, opened 2015-11-13
flow: Fix memory leak on initial packet

ceolin:
When the node's open fail the initial packet was leaking.


ibriano:
Ack

ceolin:
merged, thanls

PR #1058, opened 2015-11-13
Coverity - small fixes

ceolin:
Two things spotted by coverity

 - Complaining about usage after free
 - Not check a return a function's return


ibriano:
Ack

ceolin:
merged, thanks

PR #1059, opened 2015-11-13
app:argv: Not failing on open if index out of range

ceolin:
It's sending an error packet to be possible the flow treat the problem
correctly.


ceolin:
fixes on  https://github.com/solettaproject/soletta/pull/1065

PR #1060, opened 2015-11-13
Some gpio changes

ibriano:
- Pass the value on the event callback
 - On RIOT, fallback to polling if interruptions are not available
 - Extra: Fix a leak on UART interruptions (malloc there is bad enough already)

### DIFF
--- a/src/lib/io/sol-gpio-impl-contiki.c
+++ b/src/lib/io/sol-gpio-impl-contiki.c
@@ -52,7 +52,7 @@ struct sol_gpio {
     struct sensors_sensor *button_sensor;
     bool active_low;
     struct {
-        void (*cb)(void *data, struct sol_gpio *gpio);
+        void (*cb)(void *data, struct sol_gpio *gpio, bool value);
         void *data;
     } irq;
 };
@@ -61,8 +61,10 @@ static void
 event_handler_cb(void *user_data, process_event_t ev, process_data_t ev_data)
 {
     struct sol_gpio *gpio = user_data;
+    bool val;
 
-    gpio->irq.cb(gpio->irq.data, gpio);
+    val = sol_gpio_read(gpio);
+    gpio->irq.cb(gpio->irq.data, gpio, val);
 }
 
 struct sol_gpio *
--- a/src/lib/io/sol-gpio-impl-riot.c
+++ b/src/lib/io/sol-gpio-impl-riot.c
@@ -52,6 +52,10 @@ struct sol_gpio {
         void (*cb)(void *data, struct sol_gpio *gpio, bool value);
         const void *data;
         void *int_handler;
+        struct sol_timeout *timeout;
+        bool last_value : 1;
+        bool on_raise : 1;
+        bool on_fall : 1;
     } irq;
 };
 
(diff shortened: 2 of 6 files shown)


src/lib/io/sol-gpio-impl-contiki.c
```diff
@@ -61,8 +61,10 @@ static void
 event_handler_cb(void *user_data, process_event_t ev, process_data_t ev_data)
 {
     struct sol_gpio *gpio = user_data;
+    bool val;
 
-    gpio->irq.cb(gpio->irq.data, gpio);
+    val = sol_gpio_read(gpio);
```
barbieri: just to check, do we need to read it here? Can't we put that in the ISR and save the value in `ev` or `ev_data`?

src/lib/io/sol-gpio-impl-riot.c
```diff
@@ -59,8 +63,26 @@ static void
 gpio_process_cb(void *data)
 {
     struct sol_gpio *gpio = data;
+    bool val;
 
-    gpio->irq.cb((void *)gpio->irq.data, gpio);
+    val = sol_gpio_read(gpio);
+    gpio->irq.cb((void *)gpio->irq.data, gpio, val);
```
barbieri: same question as for as Contiki, can we read it in the ISR and queue the value somehow?

src/lib/io/sol-gpio-impl-contiki.c
ibriano: On Fri, 13 Nov 2015 14:15:49 -0800, Gustavo Sverzut Barbieri wrote:
> >  
> > -    gpio->irq.cb(gpio->irq.data, gpio);
> > +    val = sol_gpio_read(gpio);
> 
> just to check, do we need to read it here? Can't we put that in the ISR and save the value in `ev` or `ev_data`?
> 

On Contiki it wouldn't make any difference, as the moment you get to
poll the ISR, it will immediately call the callback. And we don't even
know at poll time how to read it anyway, because we don't know what kind
of sensor it is.

src/lib/io/sol-gpio-impl-riot.c
ibriano: On Fri, 13 Nov 2015 14:16:42 -0800, Gustavo Sverzut Barbieri wrote:
> >  
> > -    gpio->irq.cb((void *)gpio->irq.data, gpio);
> > +    val = sol_gpio_read(gpio);
> > +    gpio->irq.cb((void *)gpio->irq.data, gpio, val);
> 
> same question as for as Contiki, can we read it in the ISR and queue the value somehow?

Technically yes, but then we'll have RIOT behaving differently than
Linux and Contiki. On those we'll only report the last value available
at the time the user function is called.

On Riot we can implement a queue for the ISR to store the values and
then call the user function repeatedly for each value, but I'd rather
keep it consistent with the others. On that note, it should avoid
putting a new message for each ISR, and instead mark some pending_isr
flag that the main loop will clear when it finally gets to it, but that
should probably be done regardless of whether we queue the values or
not.

Opinions?

barbieri: keep current behavior and document this last part you clarified in the docs? (i didn't check the docs, maybe its there already)

ibriano:
What's the sane response to "add that explanation to the docs"?
a) Add the explanation to the docs
b) Add 6 more commits to the pull request.

Guess which I picked?

PR #1063, opened 2015-11-16
Do not enforce nul byte in sol-buffer since they are added by default.

cabelitos:
There's no need to ensure nul byte in these cases, since the sol_buffer
when initialized with the default flags will already insert the nul byte
automatically.


cabelitos:
Merged this commit with changes:

ensure_nul_byte was not removed from sol-http-common.c, since it may be required to add it.

PR #1066, opened 2015-11-16
Some GPIO changes - V2

ibriano:
- Pass the value on the event callback
 - On RIOT, fallback to polling if interruptions are not available
 - Extra: Fix a leak on UART interruptions (malloc there is bad enough already)

Differences since V1:
 - Added more documentation to sol-gpio.h
 - Fixed build on RIOT
 - Moved gpio::pin to uint32_t
 - Made GPIO interruptions consistent when active_low is set


ibriano:
Merged

PR #1067, opened 2015-11-16
Add C API for power supply, node types and samples

bdilly:
Power Supply API provides a way to get information about battery,
UPS, AC or DC power supply.

This initial version is a synchronous API heavily biased by
Linux implementation.

It doesn't provide events.

Way more information could be provided, like energy,
charge, health and other stuff that eventually few drivers
will provide and hardly will be consumed.

So I'm trying to define a set of relevant properties to expose,
and we should try to provide the same for other platforms.

@ibriano helped me with a quick research about it on RIOT
and it seems there isn't anything like that, no defined
interface. Eventually some
drivers provide some information, like voltage.

Later I intend to investigate how it's done on yet another
platform / project to make sure we get an API that at
least work for Linux and this other one.


glima:
The API overall and the commits look good.

src/lib/common/sol-power-supply-impl-linux.c
```diff
@@ -0,0 +1,412 @@
...
+
+static int
+_get_file_path(char *file_path, size_t size, const char *name,
+    enum sol_power_supply_prop prop)
+{
+    int len;
+
+    len = snprintf(file_path, size, SYSFS_POWER_SUPPLY "/%s/%s", name,
+        prop_files[prop]);
+    if (len < 0 || len > PATH_MAX)
```
lpereira: Need to account for `\0`, so check for `len >= PATH_MAX`.

src/lib/common/sol-power-supply-impl-linux.c
```diff
@@ -0,0 +1,412 @@
...
+
+    r = sol_util_read_file(file_path, " %c", &char_value);
+    SOL_INT_CHECK(r, < 0, r);
+
+    if (char_value == '0')
+        *value = false;
+    else if (char_value == '1')
+        *value = true;
+    else {
+        SOL_WRN("Unknow boolean state %c for %s", char_value, prop_files[prop]);
```
lpereira: `Unknown`.

src/lib/common/sol-power-supply-impl-linux.c
```diff
@@ -0,0 +1,412 @@
...
+    char file_path[PATH_MAX];
+    struct stat st;
+    int len;
+
+    SOL_NULL_CHECK(name, -EINVAL);
+    SOL_NULL_CHECK(exist, -EINVAL);
+
+    len = snprintf(file_path, sizeof(file_path), SYSFS_POWER_SUPPLY "/%s/",
+        name);
+    if (len < 0 || len > PATH_MAX) {
```
lpereira: Same here.

ibriano:
Looks ok to me, fixing the comments.

bdilly:
merged with fixes

src/lib/common/include/sol-power-supply.h
```diff
@@ -0,0 +1,360 @@
...
+ * Retrieves a list of names of all power supplies found. Types may vary.
+ *
+ * @param list A vector that will be initialized and filled with all
+ * power supply names found. It's required to free names pointed by
+ * it and clear the vector after usage.
+ *
+ * @return On success, it returns @c 0. On error, a negative value is returned
+ * and properties won't be set (pointers won't be valid);
+ */
+int sol_power_supply_get_list(struct sol_ptr_vector *list);
```
barbieri: it is nice to provide a "free" function to call doing loop + free + clear. Even if it's a common one in sol-vector.h, you point it here.

src/lib/common/include/sol-power-supply.h
```diff
@@ -0,0 +1,360 @@
...
+ *
+ * @param name Name of power supply. A list of all board supplies names may be
+ * fetched with sol_power_supply_get_list().
+ *
+ * @param type Type of power supply will be set on this pointer on success.
+ *
+ * @return On success, it returns @c 0. On error, a negative value is returned
+ * and properties won't be set (pointers won't be valid);
+ */
+int sol_power_supply_get_type(const char *name, enum sol_power_supply_type *type);
```
barbieri: this and the following API that are basically integers, use the return. Make the return int32_t and force type to start at "0". Then you document "<0" is -errno, and ">=0" are the actual value.

This API is a bit strange to use.

src/lib/common/include/sol-power-supply.h
```diff
@@ -0,0 +1,360 @@
...
+ * @param name Name of power supply. A list of all board supplies names may be
+ * fetched with sol_power_supply_get_list().
+ *
+ * @param exist On success, if power supply is found,
+ * exist will be set to @c True. Otherwise it'll be set to false.
+ *
+ * @return On success, it returns @c 0. On error, a negative value is returned
+ * and properties won't be set (pointers won't be valid);
+ */
+int sol_power_supply_exist(const char *name, bool *exist);
```
barbieri: return 0=false, 1=true, <0 is -errno.

src/lib/common/include/sol-power-supply.h
```diff
@@ -0,0 +1,360 @@
...
+ *
+ * @param name Name of power supply. A list of all board supplies names may be
+ * fetched with sol_power_supply_get_list().
+ *
+ * @param capacity Capacity percentage will be set on this pointer on success.
+ *
+ * @return On success, it returns @c 0. On error, a negative value is returned
+ * and properties won't be set (pointers won't be valid);
+ */
+int sol_power_supply_get_capacity(const char *name, int *capacity);
```
barbieri: if it is percentage, then you can use a "char" (or int8_t)

barbieri: and return it

src/lib/common/include/sol-power-supply.h
barbieri: after reviewing this file, to me it looks too much linux specific by using names. Go with handles, refcount them and so on.

On most small OS you're not going to have a meaningful name at all... and you'll need to do some queries that may return more than one value, so if you try to do one query at time, by name, you're doing much more effort.

on linux your implementation doesn't need to cache anything, just read the same files as methods are called.

src/modules/flow/power-supply/power-supply.json
```diff
@@ -0,0 +1,166 @@
...
+          "methods": {
+            "process": "set_type_process"
+          },
+          "name": "TYPE"
+        }
+      ],
+      "methods": {
+        "open": "get_list_open"
+      },
+      "name": "power-supply/get-list",
```
barbieri: we should keep a standard name with oic as well, like discover or get-list all around.

src/lib/common/include/sol-power-supply.h
bdilly: As I've mentioned on PR message, it's biased by Linux implementation.

We don't have something like that for RIOT.

I've didn't check for contiki and the other one. *If* for them it makes sense to have a handler, I'll change that. But for GPIOs, for example, they are referenced on RIOT by name, not numbers. So I'm not going to try to guess how other platforms work.

bdilly: I'll add the free function, good point.

src/lib/common/include/sol-power-supply.h
bdilly: It may returns < 0 that are valid for some integers, like charge family, as stated on kernel documentation.
It was the reason I decided to go with this approach. Also it guarantees an uniform API regarding error checks, and I don't need to make hacks like this one you're suggesting, to return integers for enums and booleans.

What makes it "strange" or hard to use in your opnion? I don't get it.

src/modules/flow/power-supply/power-supply.json
bdilly: I completely agree on keeping consistency between node type names, but I wasn't able to find specific nodes to get lists or scan stuff... oic, for example, doesn't have a scan / discovery node. It's one of it clients port.

I can change the name from "get-list" to "discover" if it seems more descriptive, but apparently it'll be the first of this kind anyway.

src/lib/common/include/sol-power-supply.h
bdilly: yeah, it should be in range 0-100.
I'll change that.

src/lib/common/include/sol-power-supply.h
barbieri: To get a simple integer you need to have one variable in the stack to give the function a pointer. The other way you could do a switch:

```c
switch (sol_power_supply_get_type(...)) {
   case XXX:
       ...;
       break;
   default: 
       error
}
```

barbieri: also, you are returning your enum... not what the kernel gives you, then you can guarantee it's >=0.

PR #1068, opened 2015-11-17
flow/test: Fix use after free in boolean-generator with invalid inputs

lpereira:
If a timer callback returns false, the timer is destroyed
automatically.  This caused a user after free bug in
boolean_generator_close(), which would try to call sol_timeout_del() on
a previously-destroyed timer.


lpereira:
Merged, thanks

PR #1070, opened 2015-11-17
samples/oic: don't let fixed types on conffiles

bdilly:
Only let on conffiles node types that may vary between platforms,
or same node types with different configuration.


otaviobp:
+1

bdilly:
merged

PR #1071, opened 2015-11-17
modules/timer: don't reset timer if enabled state doesn't change

bdilly:
If sucessive packets are received to enable the timer, it will
reset it at each received packet. But it shouldn't.


bdilly:
merged

PR #1072, opened 2015-11-17
WIP (do not integrate yet) -- form/string-formatted

glima:
The node is 100% functional already.

----------------

I'm taking 10 days of vacation now, but if this node is needed for anything, it's just a matter of writing tests (testing the same variations of the converter-string-{int,float}.fbp files here would be cool too, besides the normal format shenanigans) and splitting some pre-fixes of the commit into another commit.


glima:
Final version coming. There's still an (unrelated) log issue I found on Soletta, I'll double check it now.

PR #1073, opened 2015-11-18
Some changes to the javscript metatype

cabelitos:


cabelitos:
Merged the commits
* Javascript: Use nodes to test the results.
*  HTTP-client: The key option in get-cookies and get-headers should have a default value.

The commit Javascript: Add the missing Soletta types to javascript metatype can be reviwed here: #1082

PR #1075, opened 2015-11-18
Refactor RIOT's interrupt scheduler

ibriano:
Reduce the number of messages queued due to interruptions, and avoid
calling malloc from an interrupt context.


bdilly:
@ibriano I couldn't find anything weird. But I don't know if it's correct nor can I test it now.

I would suggest to merge it as it is.

ibriano:
Thanks, merged.

PR #1076, opened 2015-11-18
build: Allow the environment to define some standard build variables

ibriano:
The ClearLinux guys reported that since the v1_beta11 release, Soletta
insists on installing itself in /usr/lib, instead of respecting that
the directory should be /usr/lib64. Other variables that may be set in
the environment to affect the build were also being ignored.

This is far from an ideal solution, but it at least brings back the
behavior we had when these variables were set by the
dependency-resolver, at least for the more standard ones.


barbieri:
+1

PR #1077, opened 2015-11-18
Board detect v2

mbelluzzo:
Changes from v1:
- merged minor commits
- s/99_board_detect.json/50-default.json/
- Check file name for '.json' extension
- Add variable to turn "data" explicit


mbelluzzo:
Merged with the proposed changes.

PR #1079, opened 2015-11-18
coap: changes related to canceling replies created by sol_coap_send_packet_with_reply - v3

otaviobp:
Changelog from v2:
 - Using correct RFC to implement unobserve packets
 - fix a possible overflow
 - Fix mistake in "Fix memory leak if error" commit

Replaces #1061

### DIFF
--- a/src/lib/comms/sol-coap.c
+++ b/src/lib/comms/sol-coap.c
@@ -107,6 +107,7 @@ struct pending_reply {
         const struct sol_network_link_addr *cliaddr, void *data);
     const void *data;
     bool observing;
+    char *path;
     uint16_t id;
     uint8_t tkl;
     uint8_t token[0];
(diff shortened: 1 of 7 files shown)


src/lib/comms/sol-coap.c
```diff
@@ -293,17 +294,58 @@ sol_coap_uri_path_to_buf(const struct sol_str_slice path[],
...
+    uint16_t count;
+
+    SOL_NULL_CHECK(req, false);
+
+    r = coap_find_options(req, SOL_COAP_OPTION_URI_PATH, options, max_count);
+    SOL_INT_CHECK(r, < 0, NULL);
+    SOL_INT_CHECK(r, > max_count, NULL);
+    count = r;
+
+    path = alloca(sizeof(struct sol_str_slice) * (count + 1));
```
lpereira: Dangerous use of `alloca()` with a user-controllable value.

otaviobp: I am checking that count is between 0 and 16. Is it dangerous anyway? I could use a fixed size of 16 (maximum) if so.

otaviobp:
Some of them merged. The one left behind is now on #1106

PR #1081, opened 2015-11-19
oic-demo: Use default conffile name for light-client sample

otaviobp:
As we don't have light-server conffile anymore, we can use the default
conffile name for light-client, so now we don't need to export conffile
before running any oic flow sample.

@bdilly, can you review that, as it is related to some changes you did in this sample conffiles.


bdilly:
@otaviobp it's good
+1

PR #1085, opened 2015-11-19
dtls: Fix a memory access and a memory leak problem in dtls impl

otaviobp:


bdilly:
series is good

otaviobp:
merged

PR #1087, opened 2015-11-19
Build - Fixes related pk-config

ceolin:


otaviobp:
it looks ok to me

ceolin:

PR #1088, opened 2015-11-19
Update

edersondisouza:
New Soletta module update.
An API is introduced to update Soletta apps. The overall idea is that
different update modules can be used on different Soletta scenarios.
The API is quite straightforwad, with methods to check for updates,
fetch and install them.
A module to update Soletta app statically built running as PID1 is also
presented, to check and showcase the API. This is a simple update, just
Soletta app file is changed.
Node-types were made so one can update the app via fbp - and a sample was
also added.
Note that update module expects that device boots via EFI using an `startup.nsh`
script, showcased on samples. I'm not sure if we will integrate this somehow
with image generation - but there are more steps to make Linux micro one static
file a reality, though.
Documentation should give a better insight of this new feature.


src/lib/common/include/sol-update-modules.h
```diff
@@ -0,0 +1,88 @@
...
+extern "C" {
+#endif
+
+struct sol_update {
+#ifndef SOL_NO_API_VERSION
+#define SOL_UPDATE_API_VERSION (1)
+    uint16_t api_version; /**< API version */
+#endif
+
+    struct sol_update_handle * (*check)(const char *url,
```
barbieri: i'm not sure we should expose the URL, it should come built in the system somehow -- not sure where, but more often than not the backends will have their own storage for that, such as ClearLinux.

Systems like single-app/small-OS can define that during compile time, just allow them to specify an URL in kconfig or -DSOL_UPDATE_URL="..."

src/lib/common/include/sol-update-modules.h
```diff
@@ -0,0 +1,88 @@
...
+        void (*cb)(void *data, int status,
+        const struct sol_update_info *response),
+        const void *data);
+
+    struct sol_update_handle * (*fetch)(const struct sol_update_info *response,
+        void (*cb)(void *data, int status, const char *file_path),
```
barbieri: similar to URL, the file path should not be in this API. This will be a linux-only notion, thus backend specific.

src/lib/common/include/sol-update-modules.h
```diff
@@ -0,0 +1,88 @@
...
+        const void *data, bool resume);
+
+    bool (*cancel)(struct sol_update_handle *handle);
+
+    int (*progress)(struct sol_update_handle *handle);
```
barbieri: `get_progress`?

src/lib/common/include/sol-update-modules.h
```diff
@@ -0,0 +1,88 @@
...
+
+    struct sol_update_handle * (*install)(const char *file_path,
```
barbieri: same here, do not expose path.

To make testing simpler, Linux "single file" can take some envvars. But do not put those in the API.

src/lib/common/include/sol-update-modules.h
```diff
@@ -0,0 +1,88 @@
...
+        void (*cb)(void *data, int status),
+        void *data);
+
+    void (*init)(void);
```
barbieri: return int and check for errors.

barbieri: and provide a shutdown :-)

src/lib/common/include/sol-update.h
```diff
@@ -0,0 +1,185 @@
...
+ *
+ * @param handle handle of update task to be cancelled.
+ *
+ * @return true if current task was successfuly cancelled.
+ *
+ * @note If cancel was successful, then @c handle becomes invalid. If not,
+ * task callback shall still be called.
+ */
+/* TODO note above may yeld to memory leak if app is closed before
+ * completing the task. Shall we worry about that? */
```
barbieri: not clear what you mean, you should be able to cancel it. If the backend can't for some reason, then it should use some other trick such as "atexit()"

src/modules/flow/update/update.c
```diff
@@ -0,0 +1,495 @@
...
+    struct sol_blob *blob;
+
+    if (status < 0) {
+        sol_flow_send_error_packet(mdata->node, -status,
+            "Error while checking for updates");
+        goto end;
+    }
+
+    /* TODO maybe a custom packet instead of JSON? Maybe sol_update_info have
+     * the raw json? */
```
barbieri: you may rely on composed-packet with basic types in a certain order. Then you document the order so people can use composed type meta-type to generate extractors. Talk to @cabelitos

src/modules/flow/update/update.c
```diff
@@ -0,0 +1,495 @@
...
+    struct check_data *mdata = data;
+
+    if (!mdata->url) {
+        SOL_WRN("No URL to check for updates");
+        return -EINVAL;
+    }
+
+    if (mdata->base.handle) {
+        sol_flow_send_error_packet(node, EINVAL,
+            "Check already in progress. You may want to CANCEL before issuing a new CHECK");
```
barbieri: I don't see why we should fail. Just ignore, if there is a check, then he will get a packet at some point.

src/modules/flow/update/update.c
```diff
@@ -0,0 +1,495 @@
...
+}
+
+static void
+fetch_cb(void *data, int status, const char *file_path)
+{
+    struct base_data *mdata = data;
+
+    if (status < 0) {
+        sol_flow_send_error_packet(mdata->node, -status,
+            "Error while fetching update file");
```
barbieri: sol_util_strerrora() in the message?

src/modules/flow/update/update.c
```diff
@@ -0,0 +1,495 @@
...
+    }
+
+    sol_flow_send_string_packet(mdata->node,
+        SOL_FLOW_NODE_TYPE_UPDATE_FETCH__OUT__FILE_PATH, file_path);
+end:
+    mdata->handle = NULL;
+}
+
+static void
+free_update_info(struct sol_update_info *info)
```
barbieri: call it "clear", not "free", otherwise looks like you're freeing the `info` pointer.

src/modules/flow/update/update.json
```diff
@@ -0,0 +1,205 @@
...
+  "$schema": "http://solettaproject.github.io/soletta/schemas/node-type-genspec.schema",
+  "name": "update",
+  "meta": {
+    "author": "Intel Corporation",
+    "license": "BSD-3-Clause",
+    "version": "1"
+  },
+  "types": [
+    {
+      "category": "input/network",
```
barbieri: should be "update" or "system/update"

src/modules/update/Kconfig
```diff
@@ -0,0 +1,12 @@
+menuconfig USE_AUTO_UPDATE
+    bool "Auto Update Options"
```
barbieri: I have no idea why you added "auto" in here :-D makes no sense.

I'd keep it only "update", or call it "system update" and change everywhere.

src/modules/update/Kconfig
```diff
@@ -0,0 +1,12 @@
...
+    default y
+    help
+        Allow Soletta built apps to be updated via HTTP connection.
```
barbieri: do not mention HTTP. That's backend-specific.

Say a small OS device may use OIC or CoAP or simply UDP to get the packets... that are downloaded straight to EEPROM.

src/modules/update/common/update-common.c
```diff
@@ -0,0 +1,488 @@
...
+free_response_members(struct sol_update_info *info)
+{
+    free((void *)info->version);
+    free((void *)info->url);
+    free((void *)info->hash);
+    free((void *)info->hash_algorithm);
+}
+
+static void
+task_check_response(void *data, const struct sol_http_client_connection *conn,
```
barbieri: common shouldn't be addressing HTTP stuff.

src/modules/update/common/update-common.c
```diff
@@ -0,0 +1,488 @@
...
+    if (!(response.url && response.version
+        && response.hash && response.hash_algorithm)) {
+        SOL_WRN("Malformed check response");
+        handle->cb_check((void *)handle->user_data, -EINVAL, NULL);
+        free_response_members(&response);
+        delete_handle(handle);
+        return;
+    }
+
+    handle->cb_check((void *)handle->user_data, 0, &response);
```
barbieri: need to flag somehow, otherwise calling "cancel" from handle->cb_check will crash... you know, people do these things :-D

src/modules/update/common/update-common.c
```diff
@@ -0,0 +1,488 @@
...
+{
+    struct sol_update_handle *handle = data;
+    struct sol_buffer buffer = SOL_BUFFER_INIT_EMPTY;
+    struct sol_str_slice slice = sol_str_slice_from_blob(output);
+    int r = 0;
+
+    r = sol_buffer_append_as_base16(&buffer, slice, false);
+    SOL_INT_CHECK_GOTO(r, < 0, end);
+
+    if (buffer.used != strlen(handle->hash) ||
```
barbieri: both are null-terminated strings, just go with `streq()`

src/modules/update/common/update-common.c
```diff
@@ -0,0 +1,488 @@
...
+        strncmp(buffer.data, handle->hash, buffer.used != 0)) {
+
+        r = -EINVAL;
+        SOL_WRN("Expected hash differs of download file hash, expected [%s], found [%.*s]",
+            handle->hash, (int)buffer.used, (char *)buffer.data);
+    }
+
+end:
+    handle->cb_hash(handle, r);
```
barbieri: same as with check, flag that we're calling the cb so we don't double free. We can do something like store `handle->cb_hash` to a stack variable, set `handle->cb_hash = NULL` and check that condition inside `_cancel()`

src/modules/update/common/update-common.c
```diff
@@ -0,0 +1,488 @@
...
+     * after send last chunk */
+    if (!size && last) {
+        SOL_WRN("Nothing more to feed hash algorithm, ignoring on_feed_done request");
+        return;
+    }
+
+    blob_backend = malloc(size);
+    SOL_NULL_CHECK_GOTO(blob_backend, err);
+
+    blob = sol_blob_new(SOL_BLOB_TYPE_DEFAULT, NULL, blob_backend, size);
```
barbieri: what's wrong with `sol_file_reader` and the helper to convert it to a blob? Will be much more efficient, say the FS supports MMAP, then there is no need to allocate the whole thing in memory.

And as with HTTP, file shouldn't be in the common.

src/lib/common/include/sol-update.h
```diff
@@ -0,0 +1,185 @@
...
+struct sol_update_handle;
+
+/**
+ * Contains update info got via sol_update_check call
+ */
+struct sol_update_info {
+    const char *url; /**< URL to download update file */
+    const char *version; /**< Current version of update file */
+    const char *hash; /**< Hash of update file, so download can be checked */
+    const char *hash_algorithm; /**< Algorithm used to get file hash */
```
barbieri: you should provide `size_t size`, so we can better inform progress, estimate download size, etc...

src/modules/update/common/update-common.c
```diff
@@ -0,0 +1,488 @@
...
+    if (ferror(handle->file)) {
+        SOL_WRN("Could not read file for feed hash algorithm");
+        goto err;
+    }
+
+    last = feof(handle->file);
+
+    /* TODO Maybe this is a bug on sol_message_digest? Keeps calling on_feed_done
+     * after send last chunk */
+    if (!size && last) {
```
barbieri: not clear to me what is happening. Are you feeding after "final=true"? Check with @ibriano as he changed some of the thread part.

src/modules/update/common/update-common.c
```diff
@@ -0,0 +1,488 @@
...
+    f = fopen(SOL_UPDATE_FILE_NAME, "w+e");
+    if (!f) {
+        SOL_WRN("Could not open file for writing: %s", sol_util_strerrora(errno));
+        handle->cb_fetch((void *)handle->user_data, -errno, NULL);
+        delete_handle(handle);
+        return;
+    }
+
+    fwrite(http_response->content.data, http_response->content.used, 1, f);
+    fflush(f);
```
barbieri: I know this is an initial implementation and it's over simplified since our HTTP client does not support streaming -- which would be required for real-world updates. But keep in mind these writes would impact the main loop and are better done from a thread. Maybe you can use your async work here?

I'd say you keep a reference to the response and async write its contents to disk. Once that finishes you release the response reference.

Also, to make the process more reliable use ftruncate() to force allocation of your required size (this does have implications on how to resume as you can't rely on last file size anymore). Start with a temporary name (ie: ".part") and do fsync/fsyncdata (http://linux.die.net/man/2/fsync) after write is done. Then rename, then sync the dir. I guess you can find some updaters to base your, such as the clearlinux swupd

barbieri: ah, and just rename when hash is validated!

src/lib/common/include/sol-update-modules.h
edersondisouza: Ack. But how to address this? Yet another opaque struct? I'd like to let 'install' receive the 'path/id/anythingthatidentifies' so one could do a restore to a previous version - just install this pre-downloaded file. Or for something along this, another API or option on install, so backend takes care of 'reinstall/restore' stuff?

src/lib/common/include/sol-update-modules.h
edersondisouza: Definitely.

src/lib/common/include/sol-update-modules.h
edersondisouza: Shutdown appears to also solve the 'cancel issue' =D

src/lib/common/include/sol-update.h
edersondisouza: Backend can free memory on shutdown on last case - I think.

src/modules/update/Kconfig
edersondisouza: In the beginning I thought this as 'an auto update' because requires no final user intervention - but I dropped the idea and the 'auto' thing survived only here =D

src/modules/update/common/update-common.c
edersondisouza: But downloading via http shall be common/handful, don't you think? Maybe a 'sol_util_[http_]download' is worth?

src/modules/update/common/update-common.c
edersondisouza: ack

src/modules/update/common/update-common.c
edersondisouza: Actually, I'm not reading the whole thing in memory, it's just the opposite: I'm using `feed_done_cb` to feed small chunks to digest algorithm. I didn't want to load everything in memory - and `sol_file_reader_to_blob` does just that.

src/lib/common/include/sol-update.h
edersondisouza: ack

src/modules/update/common/update-common.c
edersondisouza: I'm kind of doing the `fsync` stuff, but after renaming/moving file.
And I'll update this to use streaming as soon as out http client support that.

src/modules/flow/update/update.json
edersondisouza: Yeah, I'd no idea about what to write here.

src/lib/common/include/sol-update-modules.h
barbieri: the easiest is for you to give the "class" to the method (ie: `const struct sol_update *`), then each can "subclass" the struct and extend it with their own context, if they need more than one.

In your case, where all should go to the same place (compile-time or runtime defined, but never changing), then you use a global and ignore the given class.

src/modules/update/common/update-common.c
barbieri: yes, just the name is misleading. My suggestion is to make it a "mix-in", so you implement the http in its own file, the "file" (save/apply) to its own file, etc. Then one platform that will use "http + file" (ie: yours), you include both and use their symbols. Others could define "coap" or "oic", but would still use "file". Or go with nvram/flash/eeprom/whatever.

src/modules/update/common/update-common.c
barbieri: Indeed, I did misread it! I didn't notice you're doing fread() of only sizeof(buf)!

src/modules/update/common/update-common.c
```diff
@@ -0,0 +1,488 @@
...
+    SOL_INT_CHECK_GOTO(r, < 0, err);
+
+    sol_blob_unref(blob);
+    return;
+
+err:
+    SOL_WRN("Could not feed data to check update file hash");
+    free(blob_backend);
+    sol_blob_unref(blob);
+    sol_message_digest_del(md);
```
barbieri: hum... what about other users/references to the `md`? I guess you should have a way to make these `NULL` after this delete.

barbieri: ah, forget it. I just noticed this is called from the md itself. So if it dies, it's not going to call it again :-D

src/modules/update/common/update-common.c
barbieri: I saw that, but need more. Write contents, sync. rename, sync.

src/lib/common/include/sol-update-modules.h
edersondisouza: Ok. So we would have something like: `sol_update_fetch` giving a `struct sol_update *` to feed `sol_update_install`, a la `struct sol_update_info *` which goes from `sol_update_check` to `sol_update_fetch` - but without the `const char url *` there, it would be hidden on an extended one of reference backend. Is that right? If so, how to do this flow on flow?
Nowadays, `update/check` outputs the JSON (soon-to-be composed packet) with information (including URL, that should not be there) to feed `update/fetch`, which in turn outputs the `file_path` to `update/install`. What could be sent in those cases from one node to another? A custom packet? Can they contain the extensible structs `sol_update_info` and `sol_update`?
About them,  I'd rather `sol_update_fetch_info` and `sol_update_install_info` - easy on C API, but not sure about on flow...

src/modules/update/common/update-common.c
```diff
@@ -0,0 +1,488 @@
...
+    handle->cb_hash(handle, -EINVAL);
+}
+
+static bool
+check_hash(struct sol_update_handle *handle,
+    void (*cb)(struct sol_update_handle *handle, int status))
+{
+    struct sol_message_digest_config cfg = {
+        SOL_SET_API_VERSION(.api_version = SOL_MESSAGE_DIGEST_CONFIG_API_VERSION, )
+        .algorithm = handle->hash_algorithm,
```
lpereira: I would strongly advise adding a few of these hashing algorithms to a white list. Some of them should never be used anymore, and this seems to be user controllable.

src/lib/common/include/sol-update.h
```diff
@@ -0,0 +1,185 @@
...
+ * may involve update/install newer versions of used libs. So, Soletta provides
+ * different update modules to update Soletta based apps on different scenarios.
+ * Soletta will try to use the first module it can use, but if having different
+ * modules to be loaded, environment variable SOL_UPDATE_MODULE can be used to
+ * define a specific one, like SOL_UPDATE_MODULE=linux-micro-efi-update.
+ * When Soletta update module cheks an URL, it expects to get a JSON like
+ * @code{.js}
+ * {
+ *   "version":"1.0.0",
+ *   "url":"http://127.0.0.1/update",
```
lpereira: Please make this HTTPS-only (refuse to work with HTTP). Also check that the certificate is valid. If Soletta is built in debugging mode, might be a good idea for a key to bypass the certificate validity check check.

edersondisouza: Does Soletta support the 'check certificate' part?

src/modules/update/common/update-common.c
edersondisouza: Whitelist on Soletta? @cabelitos approves =D

edersondisouza: On a serious side, depends on backend implementation. But the reference one might expect for some specific ones.

src/lib/common/include/sol-update-modules.h
barbieri: You didn't get it. You don't specify the URL or location anywhere in the API. These are "out of band".

In your current implementation these would come from compile-time defines (ie: Kconfig) and for ease of test you can also consume envvars, so you can `export SOL_UPDATE_URL=http://localhost:8080 SOL_UPDATE_FILE=/tmp/bla`. 

Most of the systems will handle this elsewhere, such as in ArchLinux it would be /etc/pacman.conf and mirrors list, Debian is /etc/apt/apt.conf and repositories list, etc.

That's why I said the extension of the `struct sol_update` is an option. But we don't even need to go that route. I guess I wasn't clear about that in my comment :-) (just re-read and got that impression)

src/modules/update/common/update-common.c
barbieri: I'd not bother to be too smart here. This is whatever the server and customer feels like they need to support, say they have a hardware that does md5 and not sha256, if they feel confident that is enough, then great. Creating another image that generates the same md5 is unlike to be a valid image/binary anyway.

That said, our recommended implementation would be to use some existing update solution whenever possible, such as apt-get/yum. We're providing a simple version that could be used in simpler setups such as micro-linux, also to explore the API :-)

src/lib/common/include/sol-update.h
barbieri: we should via libcurl, but nobody ever did try that. I know @rchiossi added `sol_cert` so he could use those in Amazon's MQTT, so we should offer the same support.

That said, I don't think HTTPS should be enforced. We should offer a way to check for signed files and that could cope with with plain HTTP, like ClearLinux does, see https://download.clearlinux.org/update/5160/Manifest.MoM.signed and https://download.clearlinux.org/update/5160/Manifest.MoM if your system can check for the signature of that file, great. The benefit is that this is cache-friendly (ie: squid) :-)

But we also lack the signature support in sol-crypto. My take is to do not block to be https-only (as we're even removing this from node-type and the C-API) and add a task to implement sol-crypto features required to so signature checking, then we add that to our reference implementation.

src/lib/common/include/sol-update-modules.h
edersondisouza: So, to go 'check' -> 'fetch' -> 'install' is 'backend defined', in the sense that if someone call 'install' without 'fetch', the operation would just fail. No information needs to be passed from one node/api to another? If so, I got it.

barbieri: yes, of course if they are expected to behave in a good way, like if they miss information from fetch, then they could simply do the fetch themselves.

edersondisouza:
v2 on #1122 (but it's old already, check #1133 )

PR #1089, opened 2015-11-19
lib/power-supply: add free_list()

bdilly:
Also change capacity to check if value is out of range
and use uint8_t.


bdilly:
merged with fixes

PR #1090, opened 2015-11-21
Http stream

ceolin:
Add support for transfer chunks of data

### DIFF
--- a/src/lib/comms/sol-http-client-impl-curl.c
+++ b/src/lib/comms/sol-http-client-impl-curl.c
@@ -75,8 +75,7 @@ struct sol_http_client_connection {
     struct curl_slist *headers;
     struct sol_buffer buffer;
     struct sol_http_params response_params;
-
-    void (*cb)(void *data, const struct sol_http_client_connection *connection, struct sol_http_response *response);
+    struct sol_http_request_interface *interface;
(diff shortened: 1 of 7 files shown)


src/lib/comms/sol-http-client-impl-curl.c
```diff
@@ -914,3 +967,26 @@ sol_http_client_connection_cancel(struct sol_http_client_connection *pending)
...
+    const struct sol_http_request_interface *interface,
+    const void *data)
+{
+    struct sol_http_request_interface *iface;
+    struct sol_http_client_connection *pending;
+
+    iface = calloc(1, sizeof(*iface));
+    SOL_NULL_CHECK(iface, NULL);
+
+    iface = memcpy(iface, interface, sizeof(*iface));
```
lpereira: Why not use `sol_util_memdup()`, instead of allocating memory, clearing it out, then copying over to the cleared-out block?

ceolin: yep, i should use. Just forgot it.

edersondisouza:
Other than the comments, lgtm

ceolin:
fixes on  https://github.com/solettaproject/soletta/pull/1094

PR #1094, opened 2015-11-23
Http stream v2

ceolin:
from v1:

- use sol_util_memdup
- ``write_cb`` returns a ``size_t``
- Added more http samples

### DIFF
--- a/src/lib/comms/include/sol-http-client.h
+++ b/src/lib/comms/include/sol-http-client.h
@@ -55,10 +55,21 @@ extern "C" {
 
 struct sol_http_client_connection;
 
+struct sol_http_request_interface {
+    size_t (*write_cb)(void *userdata, const struct sol_http_client_connection *connection, struct sol_buffer *buffer);
+    size_t (*read_cb)(void *userdata, const struct sol_http_client_connection *connection, void *buffer, size_t buffer_size);
+    void (*response_cb)(void *userdata, const struct sol_http_client_connection *connection, struct sol_http_response *response);
+};
+
 struct sol_http_client_connection *sol_http_client_request(enum sol_http_method method,
     const char *base_uri, const struct sol_http_params *params,
     void (*cb)(void *data, const struct sol_http_client_connection *connection,
     struct sol_http_response *response),
+    const void *data) SOL_ATTR_NONNULL(2, 4) SOL_ATTR_WARN_UNUSED_RESULT;
+
+struct sol_http_client_connection *sol_http_client_request_with_interface(enum sol_http_method method,
+    const char *base_uri, const struct sol_http_params *params,
+    const struct sol_http_request_interface *interface,
     const void *data) SOL_ATTR_NONNULL(2, 4) SOL_ATTR_WARN_UNUSED_RESULT;
 
 void sol_http_client_connection_cancel(struct sol_http_client_connection *pending);
--- a/src/lib/comms/sol-http-client-impl-curl.c
+++ b/src/lib/comms/sol-http-client-impl-curl.c
@@ -80,8 +80,7 @@ struct sol_http_client_connection {
     struct sol_buffer buffer;
     struct sol_vector watches;
     struct sol_http_params response_params;
-
-    void (*cb)(void *data, const struct sol_http_client_connection *connection, struct sol_http_response *response);
+    struct sol_http_request_interface *interface;
     const void *data;
 
     bool error;
@@ -106,6 +105,7 @@ destroy_connection(struct sol_http_client_connection *c)
 
     sol_vector_clear(&c->watches);
 
+    free(c->interface);
     free(c);
 }
 
@@ -180,11 +180,11 @@ call_connection_finish_cb(struct sol_http_client_connection *connection)
     response->param = connection->response_params;
     response->response_code = (int)response_code;
 
-    connection->cb((void *)connection->data, connection, response);
+    connection->interface->response_cb((void *)connection->data, connection, response);
     goto end;
 
 err:
-    connection->cb((void *)connection->data, connection, NULL);
+    connection->interface->response_cb((void *)connection->data, connection, NULL);
 end:
     sol_buffer_fini(&response->content);
     destroy_connection(connection);
@@ -204,7 +204,28 @@ write_cb(char *data, size_t size, size_t nmemb, void *connp)
         SOL_STR_SLICE_STR(data, data_size));
     SOL_INT_CHECK(r, < 0, 0);
 
+    if (connection->interface->write_cb) {
+        data_size = connection->interface->write_cb((void *)connection->data, connection, &connection->buffer);
+    }
+
+
     return data_size;
+}
+
+static size_t
+read_cb(char *buffer, size_t size, size_t nitems, void *connp)
+{
+    struct sol_http_client_connection *connection = connp;
+    size_t data_size, ret;
+    int r;
+
+    r = sol_util_size_mul(size, nitems, &data_size);
+    SOL_INT_CHECK(r, < 0, 0);
+
+    ret = connection->interface->read_cb((void *)connection->data,
+        connection, buffer, data_size);
+
+    return ret;
 }
 
 static void
@@ -484,8 +505,7 @@ err_exit:
 
 static struct sol_http_client_connection *
 perform_multi(CURL *curl, struct curl_slist *headers,
-    void (*cb)(void *data, const struct sol_http_client_connection *connection,
-    struct sol_http_response *response),
+    struct sol_http_request_interface *interface,
     const void *data)
 {
     struct sol_http_client_connection *connection;
@@ -499,7 +519,7 @@ perform_multi(CURL *curl, struct curl_slist *headers,
 
     connection->headers = headers;
     connection->curl = curl;
-    connection->cb = cb;
+    connection->interface = interface;
     connection->data = data;
     connection->error = false;
     sol_vector_init(&connection->watches, sizeof(struct connection_watch));
@@ -509,6 +529,11 @@ perform_multi(CURL *curl, struct curl_slist *headers,
 
     curl_easy_setopt(curl, CURLOPT_WRITEFUNCTION, write_cb);
     curl_easy_setopt(curl, CURLOPT_WRITEDATA, connection);
+
+    if (interface->read_cb) {
+        curl_easy_setopt(curl, CURLOPT_READFUNCTION, read_cb);
+        curl_easy_setopt(curl, CURLOPT_READDATA, connection);
+    }
 
     curl_easy_setopt(curl, CURLOPT_HEADERFUNCTION, header_cb);
     curl_easy_setopt(curl, CURLOPT_HEADERDATA, connection);
@@ -581,6 +606,14 @@ set_headers_from_params(CURL *curl, const struct sol_http_params *params,
 
         if (iter->type != SOL_HTTP_PARAM_HEADER)
             continue;
+
+        if (sol_str_slice_str_caseeq(key, "Content-Length")) {
+            long int len;
+            len = sol_util_strtol(iter->value.key_value.value.data, NULL,
+                iter->value.key_value.value.len, 0);
+
+            curl_easy_setopt(curl, CURLOPT_POSTFIELDSIZE_LARGE, len);
+        }
 
         buf.used = 0;
         r = sol_buffer_append_slice(&buf, key);
@@ -778,11 +811,10 @@ check_param_api_version(const struct sol_http_params *params)
     return true;
 }
 
-SOL_API struct sol_http_client_connection *
-sol_http_client_request(enum sol_http_method method,
+static struct sol_http_client_connection *
+client_request_internal(enum sol_http_method method,
     const char *base_uri, const struct sol_http_params *params,
-    void (*cb)(void *data, const struct sol_http_client_connection *connection,
-    struct sol_http_response *response),
+    struct sol_http_request_interface *interface,
     const void *data)
 {
     static const struct sol_http_params empty_params = {
@@ -868,17 +900,17 @@ sol_http_client_request(enum sol_http_method method,
         goto invalid_option;
     }
 
-    if (!set_headers_from_params(curl, params, &headers)) {
-        SOL_WRN("Could not set custom headers from params");
-        goto invalid_option;
-    }
-
     if (method == SOL_HTTP_METHOD_POST) {
         if (!set_post_fields_from_params(curl, params) ||
             !set_post_data_from_params(curl, params)) {
             SOL_WRN("Could not set POST fields or data from params");
             goto invalid_option;
         }
+    }
+
+    if (!set_headers_from_params(curl, params, &headers)) {
+        SOL_WRN("Could not set custom headers from params");
+        goto invalid_option;
     }
 
     SOL_VECTOR_FOREACH_IDX (&params->params, value, idx) {
@@ -920,13 +952,38 @@ sol_http_client_request(enum sol_http_method method,
         }
     }
 
-    pending = perform_multi(curl, headers, cb, data);
+    pending = perform_multi(curl, headers, interface, data);
     if (pending)
         return pending;
 
 invalid_option:
     curl_slist_free_all(headers);
     curl_easy_cleanup(curl);
+    return NULL;
+}
+
+
+SOL_API struct sol_http_client_connection *
+sol_http_client_request(enum sol_http_method method,
+    const char *base_uri, const struct sol_http_params *params,
+    void (*cb)(void *data, const struct sol_http_client_connection *connection,
+    struct sol_http_response *response),
+    const void *data)
+{
+    struct sol_http_client_connection *pending;
+    struct sol_http_request_interface *interface;
+
+    interface = calloc(1, sizeof(*interface));
+    SOL_NULL_CHECK(interface, NULL);
+
+    interface->response_cb = cb;
+    pending = client_request_internal(method, base_uri, params, interface, data);
+    SOL_NULL_CHECK_GOTO(pending, err);
+
+    return pending;
+
+err:
+    free(interface);
     return NULL;
 }
 
@@ -937,4 +994,26 @@ sol_http_client_connection_cancel(struct sol_http_client_connection *pending)
 
     SOL_INT_CHECK(sol_ptr_vector_remove(&global.connections, pending), < 0);
     destroy_connection(pending);
+}
+
+SOL_API struct sol_http_client_connection *
+sol_http_client_request_with_interface(enum sol_http_method method,
+    const char *base_uri, const struct sol_http_params *params,
+    const struct sol_http_request_interface *interface,
+    const void *data)
+{
+    struct sol_http_request_interface *iface;
+    struct sol_http_client_connection *pending;
+
+    iface = sol_util_memdup(interface, sizeof(*iface));
+    SOL_NULL_CHECK(iface, NULL);
+
+    pending = client_request_internal(method, base_uri, params, iface, data);
+    SOL_NULL_CHECK_GOTO(pending, err);
+
+    return pending;
+
+err:
+    free(iface);
+    return NULL;
 }
--- a/src/samples/http/.gitignore
+++ b/src/samples/http/.gitignore
@@ -1,0 +1,1 @@
+download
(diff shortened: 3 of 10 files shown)


src/lib/comms/sol-http-client-impl-curl.c
```diff
@@ -80,8 +80,7 @@ struct sol_http_client_connection {
     struct sol_buffer buffer;
     struct sol_vector watches;
     struct sol_http_params response_params;
-
-    void (*cb)(void *data, const struct sol_http_client_connection *connection, struct sol_http_response *response);
+    struct sol_http_request_interface *interface;
```
barbieri: nah, remove the "*", keep it as a struct instead a pointer to a struct, you're always allocating it and thus you can avoid doing that.

src/lib/comms/sol-http-client-impl-curl.c
```diff
@@ -106,6 +105,7 @@ destroy_connection(struct sol_http_client_connection *c)
 
     sol_vector_clear(&c->watches);
 
+    free(c->interface);
```
barbieri: then you can remove this free.

src/lib/comms/sol-http-client-impl-curl.c
```diff
@@ -204,9 +204,30 @@ write_cb(char *data, size_t size, size_t nmemb, void *connp)
...
+{
+    struct sol_http_client_connection *connection = connp;
+    size_t data_size, ret;
+    int r;
+
+    r = sol_util_size_mul(size, nitems, &data_size);
+    SOL_INT_CHECK(r, < 0, 0);
+
+    ret = connection->interface->read_cb((void *)connection->data,
+        connection, buffer, data_size);
```
barbieri: for consistency, give `const struct sol_buffer *` instead of `buffer` and `data_size`, of course set flags to memory not owned and the pointer should be `const`.

src/lib/comms/sol-http-client-impl-curl.c
```diff
@@ -499,7 +519,7 @@ perform_multi(CURL *curl, struct curl_slist *headers,
 
     connection->headers = headers;
     connection->curl = curl;
-    connection->cb = cb;
+    connection->interface = interface;
```
barbieri: here you change to `connection->interface = *interface`

src/lib/comms/sol-http-client-impl-curl.c
```diff
@@ -930,6 +962,31 @@ sol_http_client_request(enum sol_http_method method,
...
+    struct sol_http_response *response),
+    const void *data)
+{
+    struct sol_http_client_connection *pending;
+    struct sol_http_request_interface *interface;
+
+    interface = calloc(1, sizeof(*interface));
+    SOL_NULL_CHECK(interface, NULL);
+
+    interface->response_cb = cb;
```
barbieri: here you'd have `struct sol_http_request_interface interface` (no pointer) in the stack, then handle a pointer to it to the internal function.

src/lib/comms/sol-http-client-impl-curl.c
```diff
@@ -938,3 +995,25 @@ sol_http_client_connection_cancel(struct sol_http_client_connection *pending)
...
+SOL_API struct sol_http_client_connection *
+sol_http_client_request_with_interface(enum sol_http_method method,
+    const char *base_uri, const struct sol_http_params *params,
+    const struct sol_http_request_interface *interface,
+    const void *data)
+{
+    struct sol_http_request_interface *iface;
+    struct sol_http_client_connection *pending;
+
+    iface = sol_util_memdup(interface, sizeof(*iface));
```
barbieri: and avoid this memdup.

barbieri: btw, the struct should be versioned and the version checked here if compiled with `api_version` support.

src/samples/http/.gitignore
```diff
@@ -0,0 +1,4 @@
+download
+static-files
+server
+client
```
barbieri: missing trailing newline

src/samples/http/client.c
```diff
@@ -0,0 +1,202 @@
...
+                fprintf(stdout, "[HEADER] %.*s : %.*s\n",
+                    SOL_STR_SLICE_PRINT(value->value.key_value.key),
+                    SOL_STR_SLICE_PRINT(value->value.key_value.value));
+            default:
+                break;
+            }
+        }
+    }
+
+    fprintf(stdout, "%.*s\n", (int)response->content.used, (char *)response->content.data);
```
barbieri: what's wrong with `fwrite()` + `fputchar()`? This way if you receive a binary content it will barf as will stop on the first null byte.

src/samples/http/client.c
```diff
@@ -0,0 +1,202 @@
...
+    sol_quit_with_code(EXIT_SUCCESS);
+
+    return;
+
+err:
+    sol_quit_with_code(EXIT_FAILURE);
+}
+
+static int
+create_post_field_params(struct sol_http_params *params, const char *value)
```
barbieri: really? I bet you should be using @cabelitos functions to parse these things ;-)

src/samples/http/client.c
```diff
@@ -0,0 +1,202 @@
...
+    char **argv = sol_argv();
+    char *postfields = NULL, *url = NULL;
+    int i, argc = sol_argc();
+    bool verbose = false;
+    struct sol_http_params params = SOL_HTTP_REQUEST_PARAMS_INIT;
+    enum sol_http_method method = SOL_HTTP_METHOD_GET;
+
+    if (argc < 2) {
+        fprintf(stderr,
+            "Usage:\n\t%s [-v verbose] [-d <\"post fields\">] <url>\n", argv[0]);
```
barbieri: `-v verbose` looks like it takes an argument.

src/samples/http/download.c
```diff
@@ -0,0 +1,157 @@
...
+write_func(void *userdata, const struct sol_http_client_connection *connection,
+    struct sol_buffer *buffer)
+{
+    int fd = (int)(long)userdata;
+    void *data;
+    size_t len;
+    ssize_t ret;
+
+    data = sol_buffer_steal(buffer, &len);
+    ret = write(fd, data, len);
```
barbieri: you shouldn't need to steal if you're only using it to print. The caller of this function should be calling buffer_fini, that would do that.

barbieri: also `write(2)` is too lowlevel, you'd need to handle `EINTR` and such. Better to use `fwrite()` instead.

src/samples/http/download.c
```diff
@@ -0,0 +1,157 @@
...
+static void
+response_func(void *userdata, const struct sol_http_client_connection *connection,
+    struct sol_http_response *response)
+{
+    int fd = (int)(long)userdata;
+
+    close(fd);
+    if (response->response_code != SOL_HTTP_STATUS_OK) {
+        fprintf(stderr, "Finished with error, response code: %d\n", response->response_code);
+        goto err;
```
barbieri: `sol_quit_with_code(EXIT_FAILURE)` and return.

src/samples/http/download.c
```diff
@@ -0,0 +1,157 @@
...
+
+    fprintf(stdout, "Download concluded successfully\n");
+    sol_quit_with_code(EXIT_SUCCESS);
+    return;
+
+err:
+    sol_quit_with_code(EXIT_FAILURE);
+}
+
+static struct sol_http_request_interface iface = {
```
barbieri: add version and use the macro to set it (so it works when api-version is disabled).

barbieri: also add `const` keyword to the struct.

src/samples/http/server.c
```diff
@@ -0,0 +1,179 @@
...
+            if ((errno == EAGAIN) || (errno == EINTR))
+                return true;
+            fprintf(stderr,
+                "ERROR: Failed to read file descriptor: %d\n", fd);
+            goto err;
+        }
+
+        buffer[r] = '\0';
+        free(value);
+        value = strdup(buffer);
```
barbieri: ugh... keep this as buffer, reset `buffer.used = 0` and read to it... we could even use the `sol-util-file.h` if it was exposed (which we should, it's pending for a while).

src/samples/http/server.c
```diff
@@ -0,0 +1,179 @@
...
+    return false;
+}
+
+static int
+request_cb(void *data, struct sol_http_request *request)
+{
+    int r;
+    struct sol_http_response response = {
+        SOL_SET_API_VERSION(.api_version = SOL_HTTP_RESPONSE_API_VERSION, )
+        .content = SOL_BUFFER_INIT_EMPTY,
```
barbieri: since we're using pointers we're not going to change and they will live during this function (as `value` may be modified later, I'm assuming that), then we can use different macro to initialize the buffer to a memory-not-owned value and give the data + size, be the constant string or the `value` (which should be a buffer anyway).

src/lib/comms/include/sol-http-client.h
```diff
@@ -55,12 +55,23 @@ extern "C" {
 
 struct sol_http_client_connection;
 
+struct sol_http_request_interface {
```
barbieri: missing api_version (declare with macros)

src/lib/comms/include/sol-http-client.h
```diff
@@ -55,12 +55,23 @@ extern "C" {
...
+    size_t (*write_cb)(void *userdata, const struct sol_http_client_connection *connection, struct sol_buffer *buffer);
+    size_t (*read_cb)(void *userdata, const struct sol_http_client_connection *connection, void *buffer, size_t buffer_size);
```
barbieri: `const struct sol_buffer *buffer`. Also document these methods.

src/samples/http/download.c
```diff
@@ -0,0 +1,157 @@
...
+{
+    int fd = (int)(long)userdata;
+    void *data;
+    size_t len;
+    ssize_t ret;
+
+    data = sol_buffer_steal(buffer, &len);
+    ret = write(fd, data, len);
+    if (ret != (ssize_t)len) {
+        fprintf(stderr, "ERROR: Failed to write, just have written: %d\n", (int)ret);
```
barbieri: format with `%zd` and remove the cast

src/lib/comms/sol-http-client-impl-curl.c
```diff
@@ -180,11 +180,11 @@ call_connection_finish_cb(struct sol_http_client_connection *connection)
     response->param = connection->response_params;
     response->response_code = (int)response_code;
 
-    connection->cb((void *)connection->data, connection, response);
+    connection->interface->response_cb((void *)connection->data, connection, response);
```
barbieri: it looks like you don't check if given `interface->response_cb` (or the `cb` in the version that takes a single cb instead of an interface) are NULL, then it would crash here. Either check it here or at the `sol_http_client_request()`  and `sol_http_client_request_with_interface()`

src/samples/http/Kconfig
```diff
@@ -0,0 +1,23 @@
...
+	default y
+
+config STATIC_FILES_SAMPLE
+	bool "Serve static files sample"
+	depends on HTTP_SAMPLES && HTTP_SERVER
+	default y
+
+config SERVER_SAMPLE
+	bool "Server sample"
+	depends on HTTP_SAMPLES && HTTP_SERVER
```
barbieri: these samples deals with files, so I think you should use that on `depends` here. Otherwise if w add http server for riot, contiki and the likes we'll have a test that doesn't work.

src/samples/http/client.c
```diff
@@ -0,0 +1,202 @@
...
+ * THEORY OF LIABILITY, WHETHER IN CONTRACT, STRICT LIABILITY, OR TORT
+ * (INCLUDING NEGLIGENCE OR OTHERWISE) ARISING IN ANY WAY OUT OF THE USE
+ * OF THIS SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
+ */
+
+#include <fcntl.h>
+#include <stdio.h>
+#include <string.h>
+#include <sys/stat.h>
+#include <sys/types.h>
```
barbieri: review to use `stdio.h` (`fopen()`, `fwrite()`, `fclose()`) and remove these headers.

src/samples/http/client.c
```diff
@@ -0,0 +1,202 @@
...
+
+#include "sol-mainloop.h"
+#include "sol-http.h"
+#include "sol-http-client.h"
+
+static void
+response_cb(void *userdata, const struct sol_http_client_connection *connection,
+    struct sol_http_response *response)
+{
+    bool verbose = (bool)(long)userdata;
```
barbieri: for these things it's okay to use a global, will make code cleaner.

src/samples/http/client.c
```diff
@@ -0,0 +1,202 @@
...
+
+    if (response->response_code != SOL_HTTP_STATUS_OK) {
+        fprintf(stderr, "Finished with error, response code: %d\n", response->response_code);
```
barbieri: for all errors start the line with `ERROR:`, this will allow some tools (ie: emacs) to highlight lines that failed.

src/samples/http/client.c
```diff
@@ -0,0 +1,202 @@
...
+
+        SOL_HTTP_PARAMS_FOREACH_IDX (&response->param, value, idx) {
+            switch (value->type) {
+            case SOL_HTTP_PARAM_COOKIE:
+                fprintf(stdout, "[COOKIE] %.*s : %.*s\n",
+                    SOL_STR_SLICE_PRINT(value->value.key_value.key),
+                    SOL_STR_SLICE_PRINT(value->value.key_value.value));
+                break;
+            case SOL_HTTP_PARAM_HEADER:
+                fprintf(stdout, "[HEADER] %.*s : %.*s\n",
```
barbieri: no big deal, but if we could offer the same output as wget or curl, it would rock :-)

src/samples/http/client.c
```diff
@@ -0,0 +1,202 @@
...
+        if (!ptr) {
+            fprintf(stderr, "Invalid key/value in %s\n", value);
+            goto err;
+        }
+
+        *ptr = '\0';
+        val = ++ptr;
+
+        if (!sol_http_param_add_copy(params,
+            SOL_HTTP_REQUEST_PARAM_POST_FIELD(key, val))) {
```
barbieri: btw, here you'll fail to send headers with encoding, as you'd get an encoded parameter and handle that to a function that will encode it again. Thus `a%20b` will turn into `a%2520b`

src/samples/http/client.c
```diff
@@ -0,0 +1,202 @@
...
+    enum sol_http_method method = SOL_HTTP_METHOD_GET;
+
+    if (argc < 2) {
+        fprintf(stderr,
+            "Usage:\n\t%s [-v verbose] [-d <\"post fields\">] <url>\n", argv[0]);
+        sol_quit_with_code(EXIT_FAILURE);
+        return;
+    }
+
+    for (i = 1; i < argc; i++) {
```
barbieri: why not using `getopt()` for these samples? it's pretty standard.

src/samples/http/download.c
barbieri: also, for errors use `strerrora()` or our `sol-util.h`, that should be exposed (also pending task).

src/samples/http/download.c
```diff
@@ -0,0 +1,157 @@
...
+        fprintf(stderr, "ERROR: Failed to create the request\n");
+        sol_quit_with_code(EXIT_FAILURE);
+        return;
+    }
+}
+
+
+static void
+shutdown(void)
+{
```
barbieri: did you test the case where you terminate the application before the download is finished? Is `response_func()` always being called? Or do you need to cancel the pending here?

src/samples/http/server.c
```diff
@@ -0,0 +1,179 @@
...
+    int port = 80, i, argc = sol_argc();
+
+    if (argc < 2) {
+        fprintf(stderr,
+            "Usage:\n\t%s [-p <port>]\n", argv[0]);
+        sol_quit_with_code(EXIT_FAILURE);
+        return;
+    }
+
+    for (i = 1; i < argc; i++) {
```
barbieri: why not `getopt()`? Also, why not take the string from command line argument instead? You can keep the stdin if you wish, but IMO makes sense to give it via the cmdline.

src/lib/comms/include/sol-http-client.h
```diff
@@ -55,12 +55,23 @@ extern "C" {
...
+    void (*response_cb)(void *userdata, const struct sol_http_client_connection *connection, struct sol_http_response *response);
+};
+
 struct sol_http_client_connection *sol_http_client_request(enum sol_http_method method,
     const char *base_uri, const struct sol_http_params *params,
     void (*cb)(void *data, const struct sol_http_client_connection *connection,
     struct sol_http_response *response),
     const void *data) SOL_ATTR_NONNULL(2, 4) SOL_ATTR_WARN_UNUSED_RESULT;
 
+struct sol_http_client_connection *sol_http_client_request_with_interface(enum sol_http_method method,
```
barbieri: document both functions (actually all in this header), make sure you document the expected behavior, like if `interface` can be null, if `interface->response_cb` can be null, if `interface->write_cb` can be null... etc.

Cross link using `@see XXX` to things like `sol_http_client_connection_cancel()` and the "simpler"  version `sol_http_client_request()` (as well as the other way around).

barbieri:
I also miss a some features in the sample:

 * **streaming**: This is useful to simulate what people may do like broadcast video from a camera. Don't care about payload, you can keep reading from stdin and send to client or server. No `content-length` is to be defined, new data is pushed as it shows -- so for stdin tests, you can give it few seconds before sending more commands, like this: `(echo "first batch"; sleep 10; echo "second batch") | ./stream-send`
 * **resume**: This is useful to simulate resume, basically you `stat()` to get current size and provide that in the `Range` header, check the returned header to see if you got what you asked (I don't recall the semantics, maybe it may send you the whole file, in that case you'd `ftruncate()` and start from the beginning).

src/lib/comms/sol-http-client-impl-curl.c
ceolin: i do not know, this function expects the user fills the ``buffer`` with the ``data_size``, i guess give a ``const sol_buffer`` with the  capacity ``data_size`` will not be clear about what is expected. As a const it's expected to not modify, what you think ?

src/lib/comms/sol-http-client-impl-curl.c
ceolin: ack

src/samples/http/client.c
ceolin: really, this the most useful function would be ``sol_util_str_split`` to replace the ``strtok`` but this is in the shared, @cabelitos what you see here ?

I think i'll move the ``sol_util_str_split`` to sol-str-slice.h

src/samples/http/client.c
ceolin: yep, by default thi is what is done by curl, one must explictly call --data-urlencoded if want to pass data encoded.

src/samples/http/client.c
ceolin: no reason, i can change that

src/lib/comms/sol-http-client-impl-curl.c
barbieri: As you explained to me in person, this should not be const, sure. But we can still go with memory-not-owned/fixed-capacity.

as I said in person, I think the method names are confusing. We shouldn't mimic cURL, just try to match the action user expects to do (like in "write_cb" you expect user to receive some data, in "read_cb" you expect user to send some data)

src/samples/http/client.c
barbieri: this is confusing and likely to generate the double-encode I said. If you don't take url-encoded from parameter, then you're like to get errors, like people providing "a=b&c=1" where they meant "a=b%26c%3d=1". I've seen that in the past, like if you're trying to use an URL as the value, such as "subscribe=http://xxx?a=b"

src/samples/http/client.c
cabelitos: @barbieri @ceolin 

Hello, I created a function to split a string into HTTP params (without trying to encode/decode the params), however this function is not exposed, since only the http-client/create-url node is using it (check http-client.c, function split_query())

I think the best option is to is expose this function.

ceolin:
fixes on https://github.com/solettaproject/soletta/pull/1107

PR #1096, opened 2015-11-23
doc: document missing I/O API

bdilly:
Also fix some parts.

It covers all missing documentation as spotted by doxy-coverage.

Idea here is that we'll cover everything and set our bots
to guarantee no further API get in without documentation.

@mbelluzzo , @edersondisouza , could you guys please review
this? I believe you had more contact with such API.


mbelluzzo:
+2

bdilly:
ok, thanks.
merged with changes suggested by Ederson

PR #1100, opened 2015-11-25
Added new functions to buffer

ceolin:
Functions to remove part of the buffer's data and to steal (or copy if not owner) buffer's data.


ceolin:
fixes on https://github.com/solettaproject/soletta/pull/1103

PR #1103, opened 2015-11-25
Buffer v2

ceolin:
* from v1:
 - Uses ``size_t`` instead of using ``long``
 - Check for overflow
 - Remove all data up to ``buffer->used``

### DIFF
--- a/src/lib/datatypes/sol-buffer.c
+++ b/src/lib/datatypes/sol-buffer.c
@@ -873,3 +873,30 @@ sol_buffer_append_from_base16(struct sol_buffer *buf, const struct sol_str_slice
     return 0;
 }
 
+SOL_API int
+sol_buffer_remove_data(struct sol_buffer *buf, size_t size, size_t offset)
+{
+    int r;
+    size_t total;
+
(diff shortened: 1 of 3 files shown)


otaviobp:
lgtm

ceolin:
merged, thanks

src/lib/datatypes/sol-buffer.c
```diff
@@ -50,8 +50,7 @@ sol_buffer_resize(struct sol_buffer *buf, size_t new_size)
     char *new_data;
 
     SOL_NULL_CHECK(buf, -EINVAL);
-    SOL_EXP_CHECK(buf->flags & SOL_BUFFER_FLAGS_FIXED_CAPACITY, -EPERM);
-    SOL_EXP_CHECK(buf->flags & SOL_BUFFER_FLAGS_NO_FREE, -EPERM);
+    SOL_EXP_CHECK(buf->flags & SOL_BUFFER_FLAGS_MEMORY_NOT_OWNED, -EPERM);
```
barbieri: I've used 2 checks so the error message would be more meaningful :-D

PR #1107, opened 2015-11-25
http stream v3

ceolin:
- from v2
 - Use fwrite/fopen/fclose on sample
 - Rename write_cb/read_cb to recv_cb/send_cb
 - Use buffer in both callbacks
 - Add api_versio (with checks)
 - Documentation
 - Interface is not a pointer anymore (avoid memory allocation)


ceolin:
fixes on https://github.com/solettaproject/soletta/pull/1109

PR #1109, opened 2015-11-25
Http stream v4

ceolin:
- from v3
 - Change interface's callbacks to return ssize_t (negative values abort connection)
 - Use getopt on sample

- from v2
 - Use fwrite/fopen/fclose on sample
 - Rename write_cb/read_cb to recv_cb/send_cb
 - Use buffer in both callbacks
 - Add api_versio (with checks)
 - Documentation
 - Interface is not a pointer anymore (avoid memory allocation)


ceolin:
pushed with the fix, thanks a lot

PR #1111, opened 2015-11-25
dtls: Use correct server when sending it through dtls

otaviobp:
Instead of using server parameter, function send_notification_to_server was
using global oic_server.server so we were not sending notifications to
the dtls_server


lpereira:
+2

otaviobp:
merged

PR #1117, opened 2015-11-26
Javascript metatype code generation. v2

cabelitos:
Changes.

* Added commit aa18efd
* Added commit 5be92d4
* Fix possible leak pointed by @bdilly in 7235d6a
* The javascript common code is defined as a macro in js_code_start.h and it's appended at once during code generation


cabelitos:
The append rollbacks are not necessary. Because when an error happens the calling code will start to clear the vectors.

cabelitos:
Merged.

PR #1118, opened 2015-11-26
scripts/gen-html: escape double quotes on default values

bdilly:
Otherwise it may break the generated page.


bdilly:
merged after an ack of @cabelitos on irc

PR #1119, opened 2015-11-26
Cleanup and features needed by the OIC security module

lpereira:
This PR contains small patches adding features necessary to implement the OIC security module:

* Functions specific to DTLS sockets, such as changing the cipher suite, enabling anonymous ECDH handshakes, etc.
* Removing unused members from sol_coap_resource
* Ensuring sol_random with URANDOM engine won't accept a seed being passed
* Add a flag to sol_buffer to clear the memory before freeing the buffer

@otaviobp, could you please review these?


lpereira:
Merged with fixes, thanks!

PR #1120, opened 2015-11-26
tests: avoid wrong order of packets on timeout delays

bdilly:
Apparently if a delay happens and sequence generator
send packets together with buffer (validator) timeout,
results diverge.

Changing order of node types declaration should
avoid this problem.


bdilly:
merged after discussion with @cabelitos and @edersondisouza on irc channel

PR #1121, opened 2015-11-26
Docs, docs, docs

mbelluzzo:
* Improve pin mux modules text a little bit (take the opportunity to add @briefs)
* Add 'get_board_name' docs that was missing from platform (and add more @briefs)
* s/''/@c/ in aio and add missing brief in group because was the only one without description in the 'Modules' list


otaviobp:
+1

PR #1122, opened 2015-11-26
Update v2

edersondisouza:
v2:
I've tried to address all issues of v1. Main highlights:

- No information is passed along one API call to another - it's backend task to keep state. So, no URL or file_paths anymore;
- Using http 'chunked' transfer;
- Example module uses Kconfig defined data to get URL to get relevant data, but also accepts some envvars;
- Common helpers divided into 'file' and 'http' ones: include functions that download file, check its hash, move files, etc.

New Soletta module update.
An API is introduced to update Soletta apps. The overall idea is that
different update modules can be used on different Soletta scenarios.
The API is quite straightforwad, with methods to check for updates,
fetch and install them.
A module to update Soletta app statically built running as PID1 is also
presented, to check and showcase the API. This is a simple update, just
Soletta app file is changed.
Node-types were made so one can update the app via fbp - and a sample was
also added.
Note that update module expects that device boots via EFI using an startup.nsh
script, showcased on samples. I'm not sure if we will integrate this somehow
with image generation - but there are more steps to make Linux micro one static
file a reality, though.
Documentation should give a better insight of this new feature.


edersondisouza:
v3 on #1133

PR #1123, opened 2015-11-26
Dbus improvements v2

vcgomes:
Changes from last version:
 - Fixed comments from review

This series add a way to watch interfaces, and monitor when a name enters or exits the bus.

This was part of a bigger series adding support for the Parrot Flower Power, if this looks disorganized it is because the development was much more organic.

### DIFF
--- a/src/lib/common/sol-bus.c
+++ b/src/lib/common/sol-bus.c
@@ -47,9 +47,11 @@
 struct property_table {
     const struct sol_bus_properties *properties;
     const void *data;
-    void (*changed)(void *data, uint64_t mask);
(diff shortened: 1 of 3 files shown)


edersondisouza:
Documentation of API, i.e. `sol-bus.h`, is something that will be covered after?

vcgomes:
I will write the documentation for the next version, to avoid the case that I forget about it and it goes for months without any documentation.

Thanks for the review.

vcgomes:
Just submitted https://github.com/solettaproject/soletta/pull/1140 , I will update there if there are any points unresolved.

PR #1124, opened 2015-11-27
buffer: Only perform 'clear memory' if buffer has requested so

edersondisouza:
A buffer shall use `sol_util_secure_clear_memory` only if
SOL_BUFFER_FLAGS_CLEAR_MEMORY flag is set.
However, current comparison had a mistake, as this flag contains
SOL_BUFFER_FLAGS_MEMORY_NOT_OWNED bits, so it was taking not owned as
requiring secure free.

Shall fix segfaults on make samples.

### DIFF
--- a/src/lib/comms/sol-http-common.c
+++ b/src/lib/comms/sol-http-common.c
@@ -158,6 +158,11 @@ sol_http_encode_slice(struct sol_buffer *buf, const struct sol_str_slice value)
     SOL_NULL_CHECK(buf, -EINVAL);
 
     sol_buffer_init(buf);
+
+    /* Empty slice. Just return an empty buffer */
+    if (!value.len)
+        return 0;
+
     last_append = 0;
     for (i = 0; i < value.len; i++) {
         unsigned char c = value.data[i];
@@ -199,6 +204,11 @@ sol_http_decode_slice(struct sol_buffer *buf,
     SOL_NULL_CHECK(buf, -EINVAL);
 
     sol_buffer_init(buf);
+
+    /* Empty slice. Just return an empty buffer */
+    if (!value.len)
+        return 0;
+
     last_append = 0;
 
     for (i = 0; i < value.len; i++) {
--- a/src/lib/datatypes/sol-buffer.c
+++ b/src/lib/datatypes/sol-buffer.c
@@ -924,7 +924,7 @@ sol_buffer_fini(struct sol_buffer *buf)
 {
     if (!buf)
         return;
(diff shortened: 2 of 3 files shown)


barbieri:
I don't get why `SOL_BUFFER_FLAGS_MEMORY_NOT_OWNED` should be in `SOL_BUFFER_FLAGS_MEMORY_NOT_OWNED`. We should remove that. @lpereira  can you comment? Maybe it's related to `realloc()`?

For instance, I may create a sol-buffer that I'm going to operate on and I want the memory to be cleared.

This should be taken in consideration when doing `realloc()`, otherwise we may lose the previous memory handle and leave it dangling in the malloc internal pool. Then the resize logic should handle securely in some way, maybe there is a way to get this with some malloc variant, otherwise just fallback to `malloc()` + `memcpy(new,old)` + `memset(old)` + `free(old)`.

Also, what if we change this flag to `SECURE` and in addition to `memset()` we use `mlock()` (handle errors with only `SOL_WRN()` since not all processes have the required caps). See http://linux.die.net/man/2/mlock

edersondisouza:
Just added one commit to fix http test.

edersondisouza:
@barbieri Yes, @lpereira comments on its commit that added `SOL_BUFFER_FLAGS_CLEAR_MEMORY` that it's related to realloc - and he choose the simple way.

barbieri:
well, I think it's not a good choice or balance, the other option is like 5 lines of code more?

edersondisouza:
Added one more commit, to fix CoAP samples. So we get closer to no break build - but there are things on tinydtls hindering it. @otaviobp will look into that.
@barbieri This series aims to fix build. Shall I look into that for this series?

barbieri:
no, that can be done later.

lpereira:
[**@barbieri**](https://github.com/barbieri) commented on [Nov 27, 2015, 9:29 AM GMT-2](https://github.com/solettaproject/soletta/pull/1124#issuecomment-160119319 "2015-11-27T11:29:15Z - Replied by Github Comment Enhancer"):
> @lpereira  can you comment? Maybe it's related to `realloc()`?

Yes, that's because of `realloc()`.  Since all my use cases required fixed-length buffers, I decided to go to the simpler way of requiring the memory not be owned by the `sol_buffer`; it's a restriction that could be lifted any time, should an alternative `realloc()` implementation be used (that allocates the new buffer, copies the old buffer to the new buffer, zero out the old buffer and frees the old buffer).

The `mlock()` idea is nice indeed.

edersondisouza:
Integrated. Tks!

edersondisouza:
Integrated. Tks!

PR #1127, opened 2015-11-27
oic/cbor: encode cbor to/from packets on the wire

### LINKED ISSUE #1080: oic/cbor encoding to/from the wire, remove sol_vector
We changed the approach regarding memory allocation when we changed the internals to use cbor to use more memory allocation that we did before.

we should change it to iterator pattern that reads/writes directly to the "wire" (packet's payload).

hopefully before v1 so we can break api and do not keep the old behavior.

@lpereira can share some ideas.

otaviobp:
Don't allocate extra sol_vector sctrucs to hold data from cbor fields
contained in oic packets. Read and write them directly from/to the
packet payload.

To read data from oic packets the user will now get a struct
sol_oic_map_reader, intead of a sol_vector and use the macro
SOL_OIC_MAP_LOOP to loop through all elements contained in the cbor map.

To write data in an oic packet the user won't need to add all elements
in a sol_vector. The user will need to create a callback that receives
a struct sol_oic_map_writer and add all elements to it using
sol_oic_map_append function. The callback strategy was used because the
used doesn't have access to the packet and the cbor element needs to be
created in the packets before adding elements to it and closed after
adding elements.

Fixes #1080

### DIFF
--- a/data/scripts/sol-oic-gen.py
+++ b/data/scripts/sol-oic-gen.py
@@ -182,19 +182,18 @@ def generate_object_to_repr_vec_fn_common_c(state_struct_name, name, props, clie
             'key': prop_name,
             'fargs': ', '.join(fargs)
         }
-        fields.append('''f = sol_vector_append(repr);
-SOL_NULL_CHECK(f, false);
-*f = %(ftype)s("%(key)s", %(fargs)s);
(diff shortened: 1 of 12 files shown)


otaviobp:
merged.

Issue #1125, opened 2015-11-27
Public headers need to be buildable using g++


gabrielschulhof:
When using g++ to compile a file which includes soletta public headers, it dies on code like

```
In file included from /home/nix/iot/soletta-js/depbuild/soletta/build/soletta_sysroot/usr/include/soletta/sol-flow-packet.h:40:0,
                 from /home/nix/iot/soletta-js/depbuild/soletta/build/soletta_sysroot/usr/include/soletta/sol-flow.h:40,
                 from /home/nix/iot/soletta-js/depbuild/soletta/build/soletta_sysroot/usr/include/soletta/sol-flow-static.h:35,
                 from ../generated/main.h:8,
                 from ../generated/main.cc:7:
/home/nix/iot/soletta-js/depbuild/soletta/build/soletta_sysroot/usr/include/soletta/sol-str-slice.h: In function ‘sol_str_slice sol_str_slice_from_blob(const sol_blob*)’:
/home/nix/iot/soletta-js/depbuild/soletta/build/soletta_sysroot/usr/include/soletta/sol-str-slice.h:68:88: warning: invalid conversion from ‘void*’ to ‘const char*’ [-fpermissive]
 #define SOL_STR_SLICE_STR(_s, _len) (struct sol_str_slice){.len = (_len), .data = (_s) }
                                                                                        ^
/home/nix/iot/soletta-js/depbuild/soletta/build/soletta_sysroot/usr/include/soletta/sol-str-slice.h:146:12: note: in expansion of macro ‘SOL_STR_SLICE_STR’
     return SOL_STR_SLICE_STR(blob->mem, blob->size);
            ^
```

The above is only a warning and not an error because I passed ```-fpermissive``` to g++, but I'd rather pointers were cast than have to add this flag.

More importantly, it dies on
```
In file included from /home/nix/iot/soletta-js/depbuild/soletta/build/soletta_sysroot/usr/include/soletta/sol-i2c.h:40:0,
                 from ../generated/main.h:57,
                 from ../generated/main.cc:7:
/home/nix/iot/soletta-js/depbuild/soletta/build/soletta_sysroot/usr/include/soletta/sol-buffer.h: At global scope:
/home/nix/iot/soletta-js/depbuild/soletta/build/soletta_sysroot/usr/include/soletta/sol-buffer.h:279:125: error: expected primary-expression before ‘static’
[...]

PR #1130, opened 2015-11-27
common: add macro for static checking array size

cmarcelo:
In C there's a feature to improve the check of the array size being
passed to a function, by specifying 'static' before the array
size. However this feature is not supported on C++.

This patch wraps the usage into a macro to let the public headers to be
used in C++ programs.

### DIFF
--- a/src/lib/common/include/sol-macros.h
+++ b/src/lib/common/include/sol-macros.h
@@ -73,6 +73,12 @@
 #define SOL_ATTR_PURE
 #endif
 
+#ifdef __cplusplus
+#define SOL_STATIC_ARRAY_SIZE(n)
+#else
+#define SOL_STATIC_ARRAY_SIZE(n) static n
+#endif
+
 /**
  * @}
  */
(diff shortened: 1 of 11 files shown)


cmarcelo:
This should fix https://github.com/solettaproject/soletta/issues/1125.

src/lib/common/include/sol-macros.h
```diff
@@ -73,6 +73,12 @@
 #define SOL_ATTR_PURE
 #endif
 
+#ifdef __cplusplus
+#define SOL_STATIC_ARRAY_SIZE(n)
```
barbieri: shouldn't it be `n`? Or does g++ dislikes `char id[33]`?

barbieri:
other than that, +1 :-)

lpereira:
This only needs to go to the headers, not the `.c` files. No strong opposition, though.

cmarcelo:
Applied f3aa5fb82145e6dfc08a9ab58c5aa0c93a1fbe55 adding the `n` to the C++ case.

I've considered not changing the source files, but opted to do it so that is one less rule to think about, always use the macro.

PR #1131, opened 2015-11-27
Plataform+Flow: Add support to get/set hostname.

cabelitos:

@barbieri Could you please take a look in this PR?

### DIFF
--- a/src/lib/common/include/sol-platform.h
+++ b/src/lib/common/include/sol-platform.h
@@ -190,6 +190,53 @@ int sol_platform_get_mount_points(struct sol_ptr_vector *vector);
 int sol_platform_unmount(const char *mpoint, void (*cb)(void *data, const char *mpoint, int error), const void *data);
 
 /**
+ * @brief Gets the hostname.
+ *
+ * @param name The current hostname.
+ *
+ * @return 0 on success, negative errno otherwise.
+ *
+ * @see sol_plataform_set_hostname()
+ */
+int sol_plataform_get_hostname(char **name);
+
+/**
+ * @brief Changes the hostname to @c name.
+ *
+ * @param name The new hostname.
+ *
+ * @return 0 on success, negative errno otherwise.
+ *
+ * @see sol_plataform_get_hostname()
+ */
+int sol_plataform_set_hostname(const char *name);
+
+/**
+ * @brief Adds a hostname monitor.
+ *
+ * If the hostname changes @c cb will be called.
+ *
+ * @param cb
+ * @param data
+ * @return 0 on success, negative errno otherwise.
+ *
+ * @see sol_plataform_del_hostname_monitor()
+ */
+int sol_plataform_add_hostname_monitor(void (*cb)(void *data), const void *data);
+
+/**
+ * @brief Remove a hostname monitor.
+ *
+ * @param cb The registered callback
+ * @param data
+ * @return 0 on success, negative errno otherwise.
+ *
+ *
+ * @see sol_plataform_add_hostname_monitor()
+ */
+int sol_plataform_del_hostname_monitor(void (*cb)(void *data), const void *data);
+
+/**
  * @}
  */
 
--- a/src/lib/common/sol-platform-impl-contiki.c
+++ b/src/lib/common/sol-platform-impl-contiki.c
@@ -135,3 +135,31 @@ sol_platform_impl_umount(const char *mpoint, void (*cb)(void *data, const char *
     SOL_WRN("Not implemented");
     return -ENOTSUP;
 }
+
+int
+sol_plataform_unregister_hostname_monitor(void)
+{
+    SOL_WRN("Not implemented");
+    return -ENOTSUP;
+}
+
+int
+sol_plataform_register_hostname_monitor(void)
+{
+    SOL_WRN("Not implemented");
+    return -ENOTSUP;
+}
+
+int
+sol_plataform_impl_get_hostname(char **name)
+{
+    SOL_WRN("Not implemented");
+    return -ENOTSUP;
+}
+
+int
+sol_plataform_impl_set_hostname(const char *name)
+{
+    SOL_WRN("Not implemented");
+    return -ENOTSUP;
+}
--- a/src/lib/common/sol-platform-impl-linux-micro.c
+++ b/src/lib/common/sol-platform-impl-linux-micro.c
@@ -33,6 +33,7 @@
 #include <ctype.h>
 #include <dlfcn.h>
 #include <errno.h>
+#include <fcntl.h>
 #include <limits.h>
 #include <signal.h>
 #include <stdio.h>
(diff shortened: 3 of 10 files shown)


src/lib/common/include/sol-platform.h
```diff
@@ -190,6 +190,53 @@ int sol_platform_get_mount_points(struct sol_ptr_vector *vector);
...
+ *
+ * If the hostname changes @c cb will be called.
+ *
+ * @param cb
+ * @param data
+ * @return 0 on success, negative errno otherwise.
+ *
+ * @see sol_plataform_del_hostname_monitor()
+ */
+int sol_plataform_add_hostname_monitor(void (*cb)(void *data), const void *data);
```
barbieri: please specify the const char *hostname in the cb

src/lib/common/include/sol-platform.h
```diff
@@ -190,6 +190,53 @@ int sol_platform_get_mount_points(struct sol_ptr_vector *vector);
...
 /**
+ * @brief Gets the hostname.
+ *
+ * @param name The current hostname.
+ *
+ * @return 0 on success, negative errno otherwise.
+ *
+ * @see sol_plataform_set_hostname()
+ */
+int sol_plataform_get_hostname(char **name);
```
barbieri: I'd say we better return the `char *` or hide the memory management complexity and return `const char *`, then internally we can allocate or keep a fixed static buffer, it's easier to use.

src/lib/common/sol-platform-impl-contiki.c
```diff
@@ -135,3 +135,31 @@ sol_platform_impl_umount(const char *mpoint, void (*cb)(void *data, const char *
     SOL_WRN("Not implemented");
     return -ENOTSUP;
 }
+
+int
+sol_plataform_unregister_hostname_monitor(void)
+{
+    SOL_WRN("Not implemented");
+    return -ENOTSUP;
```
barbieri: @ibriano do we have a hostname concept in contiki?

src/lib/common/sol-platform-impl-linux-micro.c
```diff
@@ -1001,3 +1012,70 @@ sol_platform_linux_micro_inform_service_state(const char *service, enum sol_plat
...
     sol_platform_inform_service_monitors(service, state);
 }
+
+int
+sol_plataform_impl_set_hostname(const char *name)
+{
+    size_t len = strlen(name);
+
+    if (len > HOST_NAME_MAX) {
+        SOL_WRN("Hostname can not be bigger than %d", HOST_NAME_MAX);
```
barbieri: put the "%s" as well

src/lib/common/sol-platform-impl-linux-micro.c
```diff
@@ -1001,3 +1012,70 @@ sol_platform_linux_micro_inform_service_state(const char *service, enum sol_plat
...
+    return true;
+}
+
+int
+sol_plataform_register_hostname_monitor(void)
+{
+    if (hostname_monitor.watcher)
+        return 0;
+
+    hostname_monitor.fd = open("/proc/sys/kernel/hostname", O_RDONLY);
```
barbieri: `O_CLOEXEC`

src/lib/common/sol-platform-impl-riot.c
```diff
@@ -208,3 +208,31 @@ sol_platform_impl_umount(const char *mpoint, void (*cb)(void *data, const char *
     SOL_WRN("Not implemented");
     return -ENOTSUP;
 }
+
+int
+sol_plataform_unregister_hostname_monitor(void)
+{
+    SOL_WRN("Not implemented");
```
barbieri: @ibriano do we have hostname in riot?

src/lib/common/sol-platform-impl-systemd.c
```diff
@@ -489,3 +489,84 @@ sol_platform_impl_shutdown(void)
...
+sol_plataform_impl_get_hostname(char **name)
+{
+    sd_bus *bus;
+    sd_bus_error error = SD_BUS_ERROR_NULL;
+    int r;
+
+    bus = sol_bus_get(NULL);
+    SOL_NULL_CHECK(bus, -ENOTCONN);
+
+    r = sd_bus_get_property_string(bus, "org.freedesktop.hostname1",
```
barbieri: as I told you, it's simpler to just use `gethostname()` here, it's faster and you can't do it synchronously.

Usually set is the bitch, due permissions and so on. the daemon will do `sethosname()`, thus `gethostname()` will work

src/modules/flow/platform/platform.c
```diff
@@ -208,4 +212,75 @@ platform_machine_id_open(struct sol_flow_node *node,
...
 
+static int
+hostname_send(struct sol_flow_node *node)
+{
+    int r;
+    char *name;
+
+    r = sol_plataform_get_hostname(&name);
+    SOL_INT_CHECK(r, < 0, r);
+    return sol_flow_send_string_packet(node, 0, name);
```
barbieri: see, here you're leaking! There is a "take" variant that will own `name` and free it, this one will `strdup()` and leak :-)

src/lib/common/sol-platform-impl-contiki.c
lpereira: If it's unavailable, is it possible to generate one based on, say, the MAC address, or the machine ID, so it's always available, making it easier to obtain it (no error checking from the client-side, etc).

barbieri: we're not sure the OS have such concept. If they do not, then what we can do is a "fallback" implementation that only stores/reads from a global variable, so other soletta apps/subsys would work

src/lib/common/sol-platform-impl-linux-micro.c
cabelitos: Ok

src/lib/common/sol-platform-impl-linux-micro.c
cabelitos: Ok

src/lib/common/include/sol-platform.h
cabelitos: Ok

src/lib/common/sol-platform-impl-systemd.c
cabelitos: I agree if you, I will change that!

src/lib/common/include/sol-platform.h
cabelitos: I think I will hide the complexity and create a static buffer.

src/modules/flow/platform/platform.c
cabelitos: My bad. :(

cabelitos:
Closing in favor of #1139

PR #1132, opened 2015-11-27
build: fix make install regarding EXTRA_BINS

bdilly:
Some scripts weren't installed for install rule,
like oic-gen or node-type-validate.

### DIFF
--- a/tools/build/Makefile.targets
+++ b/tools/build/Makefile.targets
@@ -103,7 +103,7 @@ samples: $(SOL_LIB_OUTPUT) $(samples-out) $(PRE_GEN)
 
 PHONY += samples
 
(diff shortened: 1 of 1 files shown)


bdilly:
@glima we can't check that, but the point is PRE_GEN is already used by many targets, including default_target.

So if you do something like

"make && make install"

you'll have everything on PRE_GEN installed anyway.

Problem was that making

"make install" without "make" first would let some required scripts out of installation.

That's what this patch is intended to solve.

glima:
We checked personally that the aforementioned files do not get installed with that patch, so I'm OK with the fix, +1.

bdilly:
merged, thanks

PR #1135, opened 2015-12-01
oic: Call callback when timeout is reached

otaviobp:
In sol_oic_client_get_server_info, callback needs to be called when
timeout is reached, so users can perform any cleaning operation. This is
the same behavior from other equivalent methods, as
sol_oic_client_find_resource()


otaviobp:
merged.

PR #1137, opened 2015-12-01
update: 'need_update' field on `struct sol_update_info`.

edersondisouza:
So, update modules have a handy way to inform user that the update is
necessary.
Also introduced a versioning schema to `struct sol_update_info` because
it was proved it can change.
Update file helper method `check_file_hash` changed to `get_file_hash`,
turning caller task do the check, as Linux Micro EFI update module
checks if current file hash is different from update one to decide if
update is necessary (instead of comparing versions).
An output port NEED_UPDATE was also added to 'update/check' node, as
well as sample was updated to use it.

### DIFF
--- a/src/lib/common/include/sol-update.h
+++ b/src/lib/common/include/sol-update.h
@@ -81,8 +81,13 @@ struct sol_update_handle;
  * @see sol_update_check
  */
 struct sol_update_info {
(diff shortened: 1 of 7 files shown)


edersondisouza:
Integrated with fixes, tks!

PR #1138, opened 2015-12-01
Form string formatted

glima:
This is the final form node, that needed tests when I left for vacations. While fixing the the format-string.h file inclusion (+ log domain) issue, I found another, more serious, and fixed it too.


glima:
@cabelitos, you that had taken o look before, looking alright now?

glima:
Thanks a lot. Fixed and merged.

PR #1139, opened 2015-12-01
Plataform+Flow: Add support to get/set hostname. v2

cabelitos:
Changes since v1:
* The hostname monitor callback now contains the new hostname.
* The user no longer have to free the return of sol_plataform_get_hostname() 
* Linux micro and systemd uses the same get_hostname implementation.
* Some other minor fixes.

### DIFF
--- a/src/lib/common/include/sol-platform.h
+++ b/src/lib/common/include/sol-platform.h
@@ -190,6 +190,53 @@ int sol_platform_get_mount_points(struct sol_ptr_vector *vector);
 int sol_platform_unmount(const char *mpoint, void (*cb)(void *data, const char *mpoint, int error), const void *data);
 
 /**
+ * @brief Gets the hostname.
+ *
+ * @param name The current hostname.
+ *
+ * @return The hostname or NULL on error.
+ *
+ * @see sol_platform_set_hostname()
+ */
+const char *sol_platform_get_hostname(void);
+
+/**
+ * @brief Changes the hostname to @c name.
+ *
+ * @param name The new hostname.
+ *
+ * @return 0 on success, negative errno otherwise.
+ *
+ * @see sol_platform_get_hostname()
+ */
+int sol_platform_set_hostname(const char *name);
+
+/**
+ * @brief Adds a hostname monitor.
+ *
+ * If the hostname changes @c cb will be called.
+ *
+ * @param cb
+ * @param data
+ * @return 0 on success, negative errno otherwise.
+ *
+ * @see sol_platform_del_hostname_monitor()
+ */
+int sol_platform_add_hostname_monitor(void (*cb)(const char *hostname, void *data), const void *data);
+
+/**
+ * @brief Remove a hostname monitor.
+ *
+ * @param cb The registered callback
+ * @param data
+ * @return 0 on success, negative errno otherwise.
+ *
+ *
+ * @see sol_platform_add_hostname_monitor()
+ */
+int sol_platform_del_hostname_monitor(void (*cb)(const char *hostname, void *data), const void *data);
+
+/**
  * @}
  */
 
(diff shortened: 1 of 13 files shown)


src/lib/common/include/sol-platform.h
```diff
@@ -190,6 +190,53 @@ int sol_platform_get_mount_points(struct sol_ptr_vector *vector);
...
+ *
+ * If the hostname changes @c cb will be called.
+ *
+ * @param cb
+ * @param data
+ * @return 0 on success, negative errno otherwise.
+ *
+ * @see sol_platform_del_hostname_monitor()
+ */
+int sol_platform_add_hostname_monitor(void (*cb)(const char *hostname, void *data), const void *data);
```
barbieri: `void *data` should always be the first parameter, like in our other callbacks.

src/lib/common/sol-platform-impl-linux-micro.c
```diff
@@ -1001,3 +1013,55 @@ sol_platform_linux_micro_inform_service_state(const char *service, enum sol_plat
...
+sol_platform_impl_set_hostname(const char *name)
+{
+    return sol_platform_impl_linux_set_hostname(name);
+}
+
+const char *
+sol_platform_impl_get_hostname(void)
+{
+    return sol_platform_impl_linux_get_hostname();
+}
```
barbieri: IMO you should define `sol_platform_impl_set_hostname()` and `sol_platform_impl_get_hostname()` in the linux one, no need to create yet-another impl and call it from this one.

src/lib/common/sol-platform-impl-systemd.c
```diff
@@ -489,3 +490,68 @@ sol_platform_impl_shutdown(void)
...
+{
+    return sol_platform_impl_linux_get_hostname();
+}
+
+static bool
+_set_hostname(void *data, const void *value)
+{
+    static bool first_call = true;
+
+    if (first_call) {
```
barbieri: what? why do we need this?

src/lib/common/sol-platform-linux-common.c
```diff
@@ -115,6 +117,30 @@ find_handle(const struct sol_platform_linux_fork_run *handle)
     return UINT16_MAX;
 }
 
+int
+sol_platform_impl_linux_set_hostname(const char *name)
```
barbieri: this one should be in linux-micro, not in linux-common.

src/lib/common/sol-platform-linux-common.c
```diff
@@ -115,6 +117,30 @@ find_handle(const struct sol_platform_linux_fork_run *handle)
...
+        return -EINVAL;
+    }
+
+    if (sethostname(name, len) < 0)
+        return -errno;
+    return 0;
+}
+
+const char *
+sol_platform_impl_linux_get_hostname(void)
```
barbieri: this one should be here, but no need to have `_linux` in the name.

src/lib/common/sol-platform-linux-common.h
```diff
@@ -0,0 +1,14 @@
+#pragma once
+
+#ifdef __cplusplus
+extern "C" {
+#endif
+
+int sol_platform_impl_linux_set_hostname(const char *name);
+
+const char *sol_platform_impl_linux_get_hostname(void);
```
barbieri: no need to add this file, the `set_hostname()` is not in common, the `sol_platform_impl_get_hostname()` (without `_linux`!) is already in the `sol-platform-impl.h`

src/lib/common/include/sol-platform.h
cabelitos: My bad, ok.

cabelitos:
Closing in favor of #1141

PR #1140, opened 2015-12-01
D-Bus improvements v3

vcgomes:
Changes from last version:
 - Fixed comments from review

### DIFF
--- a/src/lib/common/sol-bus.c
+++ b/src/lib/common/sol-bus.c
@@ -47,9 +47,11 @@
 struct property_table {
     const struct sol_bus_properties *properties;
     const void *data;
-    void (*changed)(void *data, uint64_t mask);
     sd_bus_slot *match_slot;
+    void (*changed)(void *data, const char *path, uint64_t mask);
     sd_bus_slot *getall_slot;
+    char *iface;
+    char *path;
 };
 
 struct ctx {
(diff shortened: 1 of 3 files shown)


vcgomes:
Updated version of https://github.com/solettaproject/soletta/pull/1123

src/lib/common/sol-bus.c
```diff
@@ -397,50 +531,59 @@ _getall_properties(sd_bus_message *reply, void *userdata,
...
-    r = sol_ptr_vector_append(&_ctx.property_tables, t);
-    SOL_INT_CHECK_GOTO(r, < 0, fail_append);
+    if (!client->properties_changed) {
+        r = snprintf(matchstr, sizeof(matchstr),
+            "type='signal',"
+            "sender='%s',"
+            "interface='org.freedesktop.DBus.Properties',"
+            "member='PropertiesChanged'",
+            client->service);
+        SOL_INT_CHECK_GOTO(r, < 0, fail);
```
lpereira: `snprintf()` will return `< 0` on overflow only in older implementations; it's prudent to check for `>= sizeof(matchstr)` as well.

vcgomes: As I explained in earlier versions of this series, I have a guarantee that `client->service` is smaller than 255 bytes, so the overflow case is not possible.

ceolin:
the serie seems pretty good, +1

vcgomes:
Merged. Thanks.

PR #1141, opened 2015-12-01
Plataform+Flow: Add support to get/set hostname. v3

cabelitos:
Changes since v2:
* changed the order of the arguments in sol_platform_add/del_hostname_monitor.
* removed src/lib/common/sol-platform-linux-common.h
* moved sol_platform_impl_linux_set_hostname(const char *name) to linux micro
* renamed sol_platform_impl_linux_get_hostname(void) to sol_platform_linux_get_hostname(void)


otaviobp:
Just reconsidering my comment. Adding a test to set_hostname doesn't sound like a good idea. It would be funny to have the hostname changed in everyone's machines.

cabelitos:
Merged.

PR #1162, opened 2015-12-07
json: Export JSON Path API

otaviobp:
Export API to handle JSON Path strings and to get an element from a JSON
Array or an JSON Object using a JSON Path string.

This code was private in json flow node, but it may be usefull to export
it.

@edersondisouza, as you are a possible user, can you review that?

### DIFF
--- a/src/lib/parsers/include/sol-json.h
+++ b/src/lib/parsers/include/sol-json.h
@@ -480,6 +480,181 @@ int sol_json_token_get_unescaped_string(const struct sol_json_token *token, stru
 char *sol_json_token_get_unescaped_string_copy(const struct sol_json_token *value);
 
 /**
+ * @brief Get the value of the JSON Object child element pointed by @a key_slice.
+ *
+ * @param scanner An initialized scanner, which is pointing to a JSON Object.
+ * @param key_slice The key of the desired element.
+ * @param value A pointer to the structure that will be filled with the value
+ *        of the element pointed by @a key_slice.
+ *
+ * @return If any parameter is invalid, or if @a scanner is not pointing to a
+ *         JSON Object, return -EINVAL. If @a key_slice is not present in the
+ *         JSON Object, -ENOENT. If @a key was found, and @a value was update
+ *         successfully with the value pointed by @a key, return 0.
+ *
+ * @see sol_json_array_get_at_index()
+ */
+int sol_json_object_get_key(struct sol_json_scanner *scanner, const struct sol_str_slice key_slice, struct sol_json_token *value);
+
+/**
+ * @brief Get the element in position @a i in JSON Array pointed by @a scanner
+ *
+ * @param scanner An initialized scanner, which is pointing to a JSON Array.
+ * @param value A pointer to the structure that will be filled with the value
+ *        of the element at position @a i.
+ * @param i The position of the desired element.
+ *
+ * @return If any parameter is invalid, or if @a scanner is not pointing to a
+ *         JSON Array, return -EINVAL. If @a i is larger than the array's
+ *         length, -ENOENT. If @a value was updated successfully with the value
+ *         in position @a i, return 0.
+ *
+ * @see sol_json_object_get_key()
+ */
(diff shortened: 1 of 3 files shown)


src/lib/parsers/include/sol-json.h
```diff
@@ -480,6 +480,181 @@ int sol_json_token_get_unescaped_string(const struct sol_json_token *token, stru
...
+ * @return If any parameter is invalid, or if @a scanner is not pointing to a
+ *         JSON Object or JSON Array, or if @a path is not a valid JSON Path,
+ *         return -EINVAL. If @a path is pointing to an invalid position in a
+ *         JSON Object or in a JSON Array, -ENOENT. If @a value was
+ *         successfully updated with the value of the element pointed by @a
+ *         path, return 0.
+ *
+ * @see sol_json_path_scanner
+ */
+int sol_json_get_path(struct sol_json_scanner *scanner, struct sol_str_slice path, struct sol_json_token *value);
```
edersondisouza: What's the difference from `sol_json_object_get_key`?

src/lib/parsers/include/sol-json.h
```diff
@@ -480,6 +480,181 @@ int sol_json_token_get_unescaped_string(const struct sol_json_token *token, stru
...
+ * sol_json_path_scanner_init(&path_scanner, path);
+ * SOL_JSON_PATH_FOREACH(path_scanner, key_slice, reason) {
+ *     //Do something
+ * }
+ *
+ * if (end_reason != SOL_JSON_LOOP_REASON_OK) {
+ *     //Error Handling
+ * }
+ *
+ * @endcode
```
edersondisouza: Maybe show an example of JSON to be scanned?

src/lib/parsers/include/sol-json.h
```diff
@@ -480,6 +480,181 @@ int sol_json_token_get_unescaped_string(const struct sol_json_token *token, stru
...
+ *        of the element pointed by @a key_slice.
+ *
+ * @return If any parameter is invalid, or if @a scanner is not pointing to a
+ *         JSON Object, return -EINVAL. If @a key_slice is not present in the
+ *         JSON Object, -ENOENT. If @a key was found, and @a value was update
+ *         successfully with the value pointed by @a key, return 0.
+ *
+ * @see sol_json_array_get_at_index()
+ */
+int sol_json_object_get_key(struct sol_json_scanner *scanner, const struct sol_str_slice key_slice, struct sol_json_token *value);
```
edersondisouza: The doc says this gets the value, but the name says it gets the key, isn't this confusing?

src/lib/parsers/include/sol-json.h
otaviobp: Good.

src/lib/parsers/include/sol-json.h
otaviobp: sol_json_object_get_key gets the child pointed by a key. sol_json_get_path uses a JSON Path instead of a key to look for the child.
Maybe the name is not clear enough. I'll rename it to sol_json_object_get_value_by_key and sol_json_get_value_by_path. Is it better?

src/lib/parsers/include/sol-json.h
otaviobp: I followed the same name used in nodes. If it is confusing it is better to change them too. The nodes are called json/object-get-key and json/object-get-path and they do the same thing as these functions

src/lib/parsers/include/sol-json.h
edersondisouza: Yes, it is.

src/lib/parsers/include/sol-json.h
edersondisouza: Got it, but the name still confusing. Maybe have other opinions?

mbelluzzo:
This was merged, right?

otaviobp:
yep. sorry.
Merged with proposed changes.

PR #1163, opened 2015-12-07
http server samples

ceolin:
Added two samples:
 - Serve static files
 - Basic handler usage


mbelluzzo:
Could you document the use and expected behavior plz? Dunno how the final version of the sample docs will be, but will help a lot.

ceolin:
fixes on https://github.com/solettaproject/soletta/pull/1166

PR #1164, opened 2015-12-07
Plataform+Flow: Add timezone/system clock API/nodes. v2

cabelitos:
Since v1:

* Rebased with master.
* Added a comment explaining a magic value.
* fixed typo (timezeon)

PR #1165, opened 2015-12-07
Platform+Flow: Add locale API and nodes.

cabelitos:
Depends on #1164 

* Do not review  29ced30 here, check #1164

### DIFF
--- a/src/lib/common/include/sol-platform.h
+++ b/src/lib/common/include/sol-platform.h
@@ -145,6 +145,26 @@ enum sol_platform_service_state {
     SOL_PLATFORM_SERVICE_STATE_UNKNOWN = -1
 };
 
+/**
+ * @brief A locale category
+ */
+enum sol_platform_locale_category {
+    SOL_PLATFORM_LOCALE_LANGUAGE,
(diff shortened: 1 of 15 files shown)


cabelitos:
Strange, the sempahore is failing because it can't open /etc/locale.conf.
Does someone knows what distribution is the sempahore using it?

barbieri:
an old one... not-systemd compatible.

It may be testing micro-linux, in that case I'd say do not fail, just WRN() and keep going without any file.

Of course you won't have write permission in any way to do the "set" tests of any of those platform issues, that's all right. The proper way to test would be to spawn a container or VM and test things in there (for micro-linux or systemd).

cabelitos:
@barbieri The problem is that is not failing on a test, it is failing during platform startup (at sol_platform_init() ), because during the startup I read the conf file (/etc/locale.conf) and I make calls to setlocale() explicitly. If you can, please take a look at sol_platform_init()

Should I remove that code?

barbieri:
no, just fail gently... WRN() and keep going, do not return -ENOENT

cabelitos:
Sounds good.

cabelitos:
Closing in favor of: #1172

PR #1167, opened 2015-12-09
DO NOT MERGE YET - Javascript code generation: Dump js bytecode and use it if possible

cabelitos:
This is an experimental version and we will need to **bump the duktape version in order to merge this branch**.

As requested, now the JS meta type will dump the javascript bytecode in the generated file and use it if the duktape versions match (one that was used to generate the C file and the one that is being compiled with the generated C file).

Below one can see the binary memory footprint with and without the bytecode (without bytecode means that the plain javascript code was used):

```
*** Compiled with: O0 ***

--- App javascript-gen.c ---
With bytecode - File size: 496K memory usage:480579
Without bytecode - File size:496K memory usage:478354

--- App mycounter.c ---
With bytecode - File size: 473K memory usage:458702
Without bytecode - File size:477K memory usage:458854


*** Compiled with: O1 ***

--- App javascript-gen.c ---
With bytecode - File size: 334K memory usage:480707
Without bytecode - File size:338K memory usage:478450

--- App mycounter.c ---
With bytecode - File size: 315K memory usage:458603
Without bytecode - File size:319K memory usage:458696

--- App telegram.c ---
With bytecode - File size: 319K memory usage:479951
Without bytecode - File size:323K memory usage:479009

*** Compiled with: O2 ***

--- App javascript-gen.c ---
With bytecode - File size: 373K memory usage:480707
Without bytecode - File size:373K memory usage:478450

--- App mycounter.c ---
With bytecode - File size: 350K memory usage:458603
Without bytecode - File size:354K memory usage:458438

--- App telegram.c ---
With bytecode - File size: 354K memory usage:479951
Without bytecode - File size:358K memory usage:479009
```

FBPs used:
* javascript.fbp (under src/test-fbp)
* mycounter.fbp (under src/samples/flow/js)
* telegram.fbp (under src/samples/flow/js)


barbieri:
good to have those numbers, sad they look the same :-D

did you test memory consumption during runtime? test with massif, see the peak and also the overall consumption if it helps.

cabelitos:
@barbieri Yeah, it's really sad :(

The memory usage that is shown in the report above was retrieved using massif. However I'm displaying the memory peak only.

barbieri:
:-(

ok, so you can choose to merge it or not. Doesn't make a difference at the moment... but we can try to push Duktape guys to improve that situation. It would help if they could use the bytecode from static-const-uint8_t buffer instead of re-creating it in memory, the static memory does not use RAM in systems with execute-in-place (Small OS).

cabelitos:
@barbieri we could try to use duk_push_external_buffer() (which makes duktape do not alloc space for the buffer, it assumes that is already allocated by someone else) and check the memory usages again.

barbieri:
but that's about the code (AST/IR?) or the memory allocated at runtime to execute the code?

cabelitos:
@barbieri It's the memory heap peak.

cabelitos:
I'm closing this for now, since the bytecode did not bring any major benefits.

PR #1169, opened 2015-12-09
Http client sample

ceolin:
A simple http client sample

### DIFF
--- a/src/lib/comms/sol-http-common.c
+++ b/src/lib/comms/sol-http-common.c
@@ -810,8 +810,9 @@ sol_http_split_uri(const struct sol_str_slice full_uri,
     return 0;
 }
 
-SOL_API int
-sol_http_split_query(const char *query, struct sol_http_params *params)
+static int
+sol_http_split_str_key_value(const char *query, const enum sol_http_param_type type,
+    struct sol_http_params *params)
 {
     struct sol_vector tokens;
     struct sol_str_slice *token;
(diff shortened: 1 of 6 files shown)


src/lib/comms/sol-http-common.c
```diff
@@ -843,15 +844,27 @@ sol_http_split_query(const char *query, struct sol_http_params *params)
...
 exit:
     sol_vector_clear(&tokens);
     return 0;
 }
+
+SOL_API int
+sol_http_split_query(const char *query, struct sol_http_params *params)
+{
+    return sol_http_split_str_key_value(query, SOL_HTTP_PARAM_QUERY_PARAM, params);
+}
```
barbieri: if `sol_http_split_str_key_value()` is public, make these helpers `static inline` in the headers, so we avoid maintaining symbols.

ceolin: do you think is better make `sol_http_split_str_key_value()` public ?

I did not do it to avoid mistakes, this function has hardcoded the delimiters `&` and `=` and it will not work for all key/value types, for instance, headers use `:`.

I think to make it public it will be necessary more flexible, to my mind is better gives those specific functions like `sol_http_split_query()`

barbieri: I guessed it was public already.

As for the function, it can calculate the delimiters based on the type (`SOL_HTTP_PARAM_POST_FIELD`), no? For Headers it could use `:`.

ceolin: yes, it's possible but it's not only the delimiter, for headers, post fields and query params it's allowed empty values, but i guess it's not valid for cookies. Also, it'll make redundant the functions to split query and post fields, and still will be necessary functions for things that is not key/value.

Do you think it's worth export `sol_http_split_str_key_value()` and remove `sol_http_split_query()` and `sol_http_split_post_field()` ?

barbieri: maybe not, keep as you propose

ceolin:
merged, thanks !

PR #1171, opened 2015-12-10
[RFC] mavlink: introduce the mavlink protocol implementation

dorileo:
## Known issues
  + This is not yet a full complaint implementation, there'are lots of API's yet to come;
  + The serial connection mechanism is not yet implemented;

## Rationale
This patch introduce the basic infrasctructure and API's to control MAV
(Micro Air Vehicle) communication. More API's are to come.

### DIFF
--- a/src/lib/comms/sol-mavlink.c
+++ b/src/lib/comms/sol-mavlink.c
@@ -1,0 +1,786 @@
+/*
+ * This file is part of the Soletta Project
+ *
+ * Copyright (C) 2015 Intel Corporation. All rights reserved.
+ *
+ * Redistribution and use in source and binary forms, with or without
+ * modification, are permitted provided that the following conditions
+ * are met:
+ *
+ *   * Redistributions of source code must retain the above copyright
+ *     notice, this list of conditions and the following disclaimer.
+ *   * Redistributions in binary form must reproduce the above copyright
+ *     notice, this list of conditions and the following disclaimer in
+ *     the documentation and/or other materials provided with the
+ *     distribution.
+ *   * Neither the name of Intel Corporation nor the names of its
+ *     contributors may be used to endorse or promote products derived
+ *     from this software without specific prior written permission.
+ *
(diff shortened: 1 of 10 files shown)


src/lib/comms/sol-mavlink.c
```diff
@@ -0,0 +1,786 @@
...
+    mavlink_message_t msg;
+    mavlink_status_t status;
+    uint8_t buf[MAVLINK_MAX_PACKET_LEN];
+    int i, res;
+    sol_mavlink_status sol_status = mavlink->status;
+
+    bzero(buf, MAVLINK_MAX_PACKET_LEN);
+    res = recv(mavlink->fd, buf, MAVLINK_MAX_PACKET_LEN, 0);
+
+    for (i = 0; i < res; ++i) {
```
lpereira: What happens if `recv()` returns `-1`?

src/lib/comms/sol-mavlink.c
```diff
@@ -0,0 +1,786 @@
...
+sol_mavlink_fd_handler(void *data, int fd, uint32_t cond)
+{
+    struct sol_mavlink *mavlink = data;
+    mavlink_message_t msg;
+    mavlink_status_t status;
+    uint8_t buf[MAVLINK_MAX_PACKET_LEN];
+    int i, res;
+    sol_mavlink_status sol_status = mavlink->status;
+
+    bzero(buf, MAVLINK_MAX_PACKET_LEN);
```
lpereira: Maybe `uint8_t buf[...] = {0}` instead?

src/lib/comms/sol-mavlink.c
dorileo: i will be greater than res and will not run the loop, nothing will take effect and so on, but well, worth testing and returning...

src/lib/comms/sol-mavlink.c
dorileo: yep sure, the use of bzero and friends have been discussed by the other guys, I'm getting rid of it wherever appropriate.

dorileo:
new version coming...

PR #1172, opened 2015-12-10
Platform+Flow: Add locale API and nodes V2.

cabelitos:
Changes since V1:

* Do not apply locales at startup.
* Fail gracefully during the startup if /etc/locale.conf does not exist.
* The node platform/locale now have ports to apply the locale values.

### DIFF
--- a/src/lib/common/include/sol-platform.h
+++ b/src/lib/common/include/sol-platform.h
@@ -236,6 +236,95 @@ int sol_platform_add_hostname_monitor(void (*cb)(void *data, const char *hostnam
 int sol_platform_del_hostname_monitor(void (*cb)(void *data, const char *hostname), const void *data);
 
 /**
+ * @brief Set the system wide time.
+ *
+ * @param timestamp The new system_clock
+ * @return @c 0 on succes, negative errno otherwise
+ *
+ * @note The @c timestamp is relative to 1970-01-01 00:00:00 +0000 (UTC)
+ * @see sol_platform_get_system_clock()
+ */
+int sol_platform_set_system_clock(int64_t timestamp);
+
+/**
+ * @brief Get the current system time.
+ *
+ * @return the system time on success, negative errno on error
+ *
+ * @note The returned value is the number of seconds since 1970-01-01 00:00:00 +0000 (UTC)
+ *
+ * @see sol_platform_set_system_clock()
+ */
+int64_t sol_platform_get_system_clock(void);
+
+/**
+ * @brief Add a callback to monitor system clock changes.
+ *
+ * @param cb A callback to be called when the system clock changes
+ * @param data The data to @c cb
+ * @return @c 0 on success, negative errno otherwise
+ *
+ * @note The registered callback will not be called every second (a.k.a: this is not a timer!), it will only
+ * be called if the system clock is adjusted. If one needs a timer use sol_timeout_add()
+ * @see sol_platform_del_system_clock_monitor()
+ */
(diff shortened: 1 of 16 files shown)


src/lib/common/include/sol-platform.h
```diff
@@ -236,6 +256,155 @@ int sol_platform_add_hostname_monitor(void (*cb)(void *data, const char *hostnam
...
 /**
+ * @brief Set the system wide time.
+ *
+ * @param timestamp The new system_clock
+ * @return @c 0 on succes, negative errno otherwise
+ *
+ * @note The @c timestamp is relative to 1970-01-01 00:00:00 +0000 (UTC)
+ * @see sol_platform_get_system_clock()
+ */
+int sol_platform_set_system_clock(int64_t timestamp);
```
lpereira: Is the resolution in seconds sufficient?

src/lib/common/sol-platform-linux-common.c
```diff
@@ -785,3 +800,208 @@ sol_platform_linux_uevent_unsubscribe(const char *action, const char *subsystem,
...
+        SOL_STR_TABLE_ITEM("LC_IDENTIFICATION", SOL_PLATFORM_LOCALE_IDENTIFICATION),
+        SOL_STR_TABLE_ITEM("LC_MEASUREMENT", SOL_PLATFORM_LOCALE_MEASUREMENT),
+        SOL_STR_TABLE_ITEM("LC_MESSAGES", SOL_PLATFORM_LOCALE_MESSAGES),
+        SOL_STR_TABLE_ITEM("LC_MONETARY", SOL_PLATFORM_LOCALE_MONETARY),
+        SOL_STR_TABLE_ITEM("LC_NAME", SOL_PLATFORM_LOCALE_NAME),
+        SOL_STR_TABLE_ITEM("LC_NUMERIC", SOL_PLATFORM_LOCALE_NUMERIC),
+        SOL_STR_TABLE_ITEM("LC_PAPER", SOL_PLATFORM_LOCALE_PAPER),
+        SOL_STR_TABLE_ITEM("LC_TELEPHONE", SOL_PLATFORM_LOCALE_TELEPHONE),
+        SOL_STR_TABLE_ITEM("LC_TIME", SOL_PLATFORM_LOCALE_TIME),
+    };
```
lpereira: Shouldn't this table have a sentinel value?

cabelitos: if the value is not find in the table it will return SOL_PLATFORM_LOCALE_UNKNOWN,
check the line below
"sol_str_table_lookup_fallback(map, loc, SOL_PLATFORM_LOCALE_UNKNOWN);"

src/lib/common/include/sol-platform.h
cabelitos: I don't know, I was inspired by the time() function.

src/lib/common/sol-platform-linux-common.c
lpereira: It won't. It will crash, since it checks if the key is NULL to stop iterating. See the implementation.

cabelitos: Sad

src/lib/common/include/sol-platform.h
barbieri: it is fine, this is just used to save calling `time()` when time changes, if more resolution is wanted the user can call something else, like `clock_gettime()`

cabelitos: I aggree.

cabelitos:
Closing in favor of #1190

PR #1173, opened 2015-12-10
coap: Fix incorrect way of accessing a string position

otaviobp:
In commit 386ee723 function packet_extract_path was changed to receive a
pointer to a string instead of returing the result string. In this new
function, the string was incorrectly accessed when adding '\0' and
invalid memory access was occuring.


otaviobp:
merged. thanks

PR #1174, opened 2015-12-10
scripts:oicgen: Remove wrong free

ceolin:
device_id is not a pointer and should not be freed.


ceolin:
@otaviobp could you confirm it please ?

PR #1175, opened 2015-12-11
fix:random: urandom engine was not working

ceolin:
It gives for urandom engine the exactly value given by user. For the
other engines, it'll not ignore the value given (except in case of 0).


ceolin:
@lpereira  could you take a look please

lpereira:
LGTM

ceolin:
merged, thanks

PR #1176, opened 2015-12-11
comms:http:oauth: Fix signature

ceolin:
The signing key for request token should be followed by an ampersand
char '&'.

As the consumer key secret is only used when request the tokens, it's
safe append '&'.


ceolin:
merged, thanks!

PR #1182, opened 2015-12-14
Twitter v3

ceolin:
from v2
 - Add sample
 - Change a cas into a printf
 - follow the http-common changes


glima:
+1, looking good

ceolin:
Thanks for review it. If any other comments appear i'll push it with the suggestions

ceolin:
merged with the suggestions, thanks

PR #1183, opened 2015-12-15
flow:iio: Add Temperatue sensor type support into iio node

guchaojie:
Change-Id: I057906f941839429edf7b2e3d4b89660ef3f42bb

### DIFF
--- a/data/scripts/sol-flow-node-type-gen.py.in
+++ b/data/scripts/sol-flow-node-type-gen.py.in
@@ -434,7 +434,7 @@ def generate_header_tail(outfile, data):
 
 #ifdef SOL_LOG_ENABLED
 #define SOL_LOG_INTERNAL_DECLARE(_var, _name) \
-    struct sol_log_domain _var = { \
+    SOL_ATTR_UNUSED struct sol_log_domain _var = { \
         .name = "sol-" _name, \
         .color = SOL_LOG_COLOR_MAGENTA, \
         .level = SOL_LOG_LEVEL_WARNING \
--- a/src/lib/common/sol-log-internal.h
+++ b/src/lib/common/sol-log-internal.h
@@ -36,7 +36,7 @@
 
 #ifdef SOL_LOG_ENABLED
 #define SOL_LOG_INTERNAL_DECLARE(_var, _name)    \
-    struct sol_log_domain _var = {               \
+    SOL_ATTR_UNUSED struct sol_log_domain _var = { \
         .name = "sol-" _name,                    \
         .color = SOL_LOG_COLOR_MAGENTA,          \
         .level = SOL_LOG_LEVEL_WARNING           \


src/modules/flow/iio/nodes.c
```diff
@@ -184,4 +184,136 @@ gyroscope_tick(struct sol_flow_node *node, void *data, uint16_t port, uint16_t c
     return -EIO;
 }
 
+struct temperature_data {
+    struct sol_iio_config config;
+    double scale;
+    double offset;
+    struct sol_flow_node *node;
```
barbieri: This is not required anymore, I see you did the same as the `gyroscope_data`, but I'd rather fix that one (as a different patch).

You can skip `struct sol_flow_node *node` from `struct temperature_data` and instead of passing a pointer to this last struct, you pass the `node` and then call `sol_flow_node_get_private_data(node)`, this will return your `struct temperature_data`. An example below.

src/modules/flow/iio/nodes.c
```diff
@@ -184,4 +184,136 @@ gyroscope_tick(struct sol_flow_node *node, void *data, uint16_t port, uint16_t c
...
+    bool buffer_enabled : 1;
+    bool use_device_default_scale : 1;
+    bool use_device_default_offset : 1;
+};
+
+static void
+temp_reader_cb(void *data, struct sol_iio_device *device)
+{
+    static const char *errmsg = "Could not read channel buffer values";
+    struct temperature_data *mdata = data;
```
barbieri: as above, you could do:

```c
struct sol_flow_node *node = data;
struct temperature_data *mdata = sol_flow_node_get_private_data(node);
```

src/modules/flow/iio/nodes.c
```diff
@@ -184,4 +184,136 @@ gyroscope_tick(struct sol_flow_node *node, void *data, uint16_t port, uint16_t c
...
+    if (mdata->buffer_enabled) {
+        mdata->config.sol_iio_reader_cb = temp_reader_cb;
+        mdata->config.data = mdata;
+    }
+    mdata->use_device_default_scale = opts->use_device_default_scale;
+    mdata->use_device_default_offset = opts->use_device_default_offset;
+    mdata->scale = opts->scale;
+    mdata->offset = opts->offset;
+
+    if (!sol_iio_address_device(opts->iio_device, temp_create_device_cb, mdata)) {
```
barbieri: and here, instead of providing `mdata`,  you would pass `node`.

barbieri:
patch looks good, I didn't check style as I'm reviewing using github webui (if you didn't, try running "./tools/check-style")

@edersondisouza can you take a look, you wrote the gyroscope code. Also see the comments I've made regarding `mdata`. If possible to fix your code before @guchaojie fix his patch it would help (nice part about being in different timezones ;-))

guchaojie:
@barbieri  I  upgrade this PR according to your suggestion and test the sensor work well  , please review it again.

barbieri:
+1, looks good!

@guchaojie as a comment, we do not replace PR since Github will miss the previous comments! So do a new PR next time (we call them "v2", "v3"...)

guchaojie:
@edersondisouza I would submit another PR to fix these issuel. thx.

PR #1184, opened 2015-12-15
log: log domain variables may not be referenced.

glima:
At least when the log domain is set to one different than the global
one, if the user issues no log calls in the Soletta app, that variable
will have no use in the code, thus leading to unused variable warnings.
We now tell, then, to the compiler, that the log domain variable may be
unused in the SOL_LOG_INTERNAL_DECLARE() macro.

### DIFF
--- a/src/lib/common/sol-log-internal.h
+++ b/src/lib/common/sol-log-internal.h
@@ -36,7 +36,7 @@
 
 #ifdef SOL_LOG_ENABLED
 #define SOL_LOG_INTERNAL_DECLARE(_var, _name)    \
-    struct sol_log_domain _var = {               \
+    SOL_ATTR_UNUSED struct sol_log_domain _var = { \
         .name = "sol-" _name,                    \
         .color = SOL_LOG_COLOR_MAGENTA,          \
         .level = SOL_LOG_LEVEL_WARNING           \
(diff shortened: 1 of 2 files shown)


glima:
@otaviobp ^

glima:
Thanks, merging.

src/lib/common/sol-log-internal.h
```diff
@@ -36,7 +36,7 @@
 
 #ifdef SOL_LOG_ENABLED
 #define SOL_LOG_INTERNAL_DECLARE(_var, _name)    \
-    struct sol_log_domain _var = {               \
+    SOL_ATTR_UNUSED struct sol_log_domain _var = { \
```
barbieri: why do we need this attribute in this file? I get it for the generated one, but this one we should keep warnings so people can remove unused stuff.

Or we try to solve cases where a combination of `#ifdef` can declare logs but not actually use them?

ceolin: yep, we are declaring log on generated code that can be not used. This was happening with many samples.

glima: @barbieri , I remember there were cases where a redefinition would be pointed out. I'll revisit it and check for the inclusion orders.

glima: Yeah, by only removing the change on the 2nd file we'll have to depend on the order of inclusion in all node files (node's generated header vs. that sol-log-internal.h), which is bad.

PR #1186, opened 2015-12-15
lcd: Use less memory on the commands queue handling.

glima:
Since the porting from the old, synchronous I2C API, the correct data
structure to hold the commands of the I2C queue would be the pointer
vector. By using that, we avoid copies.

---------------------------------

Since we have to address small OSes, I attacked this item that was on my TODO-list to
lower the memory usage on this node.


bdilly:
+2

glima:
Cool, merging.

PR #1187, opened 2015-12-16
flow: Add helper to load packet types

ceolin:
Some nodes can have their own custom packets, it's necessary something
that allows one use them in different nodes even when they are not build static.

### DIFF
--- a/src/lib/flow/include/sol-flow.h
+++ b/src/lib/flow/include/sol-flow.h
@@ -507,8 +507,26 @@ struct sol_flow_port_type_in {
 #define sol_flow_get_node_type(_mod, _type, _var) sol_flow_internal_get_node_type(_mod, #_type, _var)
 
 int sol_flow_internal_get_node_type(const char *module, const char *symbol, const struct sol_flow_node_type **type);
+
+/**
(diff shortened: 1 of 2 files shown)


ceolin:
@barbieri this is the simplest way that solve the problem. Do you think worth add on generator the boilerplate work of declare the packet, use this function to obtain the symbol. I mean, who needs a custom packet from a node, can declare on its json something like, `"depends": [ {"module": "mod_name", "packet":"packet_name"}]` and then the generator generates the packet and loads the symbol

barbieri:
+1, I like it. check with @ibriano and @cabelitos as well. I recall I talked about this to Iscaro, so maybe he can comment on that.

I also like the idea of the `"depends": [ {"module": "modname"}]`.

cabelitos:
@barbieri @ceolin The patch looks good to me.
It solves our problem. I spoke with ceolin about the "depends" keyword on our json descriptions. It's a good idea, but since we have only one use case for it, I suggest we leave the implementation for later.

ceolin:
thanks, i'll commit it and send the patch for sml. We can hold this depends idea for now.

PR #1192, opened 2015-12-16
inspector: Invalid memory access when inpecting int and float options

otaviobp:
Inspector was accessing invalid memory area and printing int and float
values, because they are now a single number instead of an irange or
drange.
Also adding debug for the new option types irange-spec and drange-spec.

Error following:
=================================================================
==23290==ERROR: AddressSanitizer: heap-buffer-overflow on address 0x6030000310d4 at pc 0x00000040948a bp 0x7ffee0aa9b40 sp 0x7ffee0aa9b30
READ of size 4 at 0x6030000310d4 thread T0

0x6030000310d4 is located 0 bytes to the right of 20-byte region [0x6030000310c0,0x6030000310d4)
allocated by thread T0 here:


glima:
Looks OK.

otaviobp:
merged.

PR #1194, opened 2015-12-16
OIC: fixing some issues and other small changes to make it possible to run soletta against iotivity samples.

otaviobp:


src/lib/comms/include/sol-oic-client.h
```diff
@@ -286,13 +287,14 @@ bool sol_oic_client_resource_request(struct sol_oic_client *client, struct sol_o
...
  * @param observe If server will be obeserved or unobserved.
+ * @param non_confirmable Use non-confirmable CoAP packets to make this request.
  *
  * @return True if packet was successfully sent. False otherwise.
  */
 bool sol_oic_client_resource_set_observable(struct sol_oic_client *client, struct sol_oic_resource *res,
     void (*callback)(struct sol_oic_client *cli, const struct sol_network_link_addr *addr,
     const struct sol_str_slice *href, const struct sol_oic_map_reader *repr_map, void *data),
-    void *data, bool observe);
+    void *data, bool observe, bool non_confirmable);
```
lpereira: If at all possible, I'd like to avoid having more than one boolean parameter. From the call site, it's hard to know what `true, false` or `false, true` actually are. For this specific case, I'd create an enum with values `SOL_OIC_OBSERVE = 1<<0` and `SOL_OIC_NONCON = 1<<1` or something along these lines, so that they can be OR-ed in the call site, and thus obvious to the reader

src/lib/comms/sol-coap.c
```diff
@@ -1614,3 +1622,20 @@ sol_coap_unobserve_server(struct sol_coap_server *server, const struct sol_netwo
 
     return -ENOENT;
 }
+
+SOL_API void
+sol_coap_packet_debug(struct sol_coap_packet *pkt)
```
lpereira: I'd make these functions dummy if `NDEBUG` is defined.

barbieri: you can also use the SOL_LOG debug stuff, it can be disabled and we could remove it at this location and also in the header file (empty `static inline`)

src/lib/comms/sol-oic-client.c
```diff
@@ -910,7 +910,7 @@ _resource_request(struct sol_oic_client *client, struct sol_oic_resource *res,
 
 SOL_API bool
 sol_oic_client_resource_request(struct sol_oic_client *client, struct sol_oic_resource *res,
```
lpereira: Instead of adding another parameter to this function, I'd make another public function only for non-confirmable requests. This function already has way too many parameters.

src/lib/comms/sol-oic-common.c
```diff
@@ -92,3 +92,31 @@ sol_oic_map_append(struct sol_oic_map_writer *oic_map_writer, struct sol_oic_rep
...
+        return;
+    }
+
+    err = cbor_parser_init(payload, payload_len, 0, &parser, &root);
+    if (err != CborNoError) {
+        SOL_DBG("print_packet: Failed to get cbor payload");
+        return;
+    }
+
+    cbor_value_to_pretty(stderr, &root);
```
lpereira: Keep in mind that `stderr` is possibly not available on systems other than POSIX.

src/lib/comms/sol-coap.c
cabelitos: I aggree with @barbieri I think we should log using SOL_LOG stuff and not directly at stderr

src/lib/comms/include/sol-oic-client.h
otaviobp: nice idea

src/lib/comms/sol-coap.c
otaviobp: I'll check if SOL_LOG_ENABLED is defined and use sol_log_get_level to check if debug log is activated in this method.

src/lib/comms/sol-oic-client.c
otaviobp: ok

src/lib/comms/sol-oic-common.c
otaviobp: Any idea in how to fix this? cbor_value_to_pretty expects a FILE. I could create a tmp file and than using sol_log to print it, but I didn't like this approach.

lpereira: Fix TinyCBOR, or only enable this if stdio is available.

PR #1195, opened 2015-12-16
flow/iio/gyroscope: Do not set direction vector min and max

edersondisouza:
It was setting to lesser and greater values of x, y and z, but this
doesn't make much sense. Just leave them as -DBL_MAX and DBL_MAX.


glima:
If there are no alternatives to getting the real component's min/max specs, this is better than we got earlier.

bdilly:
@edersondisouza I agree this approach is better than the previous one. But shouldn't you add this range as an option of the node type?
A developer would know real device range, right? By default you could leave it as -MAX / MAX (option default), seems fair to me.

edersondisouza:
@bdilly Yes, I thought about that - I was just eager to fix previous approach, but is quite simple to add the range, I'll do it.

edersondisouza:
v2 on #1205

PR #1196, opened 2015-12-16
log: Fix SOL_INT_CHECK

ceolin:
int16_t x = 0;
SOL_INT_CHECK(x, < 0);

error: invalid use of void expression
     SOL_INT_CHECK(x, < 0);


cabelitos:
Looks good. In fact we should merge this ASAP, since the current master will break on 32-bit systems.

ceolin:
merged, thanks

PR #1197, opened 2015-12-16
Common docs

mbelluzzo:
Again, help reviewing it is welcoming since this job is very error prone! Thank you all!


bdilly:
@mbelluzzo it's good overall, I believe that you can merge it directly after dealing with all these comments (no v2 seems to be necessary).

PR #1202, opened 2015-12-17
OIC: fixing some issues and other small changes to make it possible to run soletta against iotivity samples.

otaviobp:
Changelog from v1:
 - Check if stdout is available before logging to it.
 - Make the debug functions empty when loggging is disabled.
 - Use different functions name instead of a boolean parameter for non-observable packets

Replaces #1194


otaviobp:
merged

PR #1203, opened 2015-12-17
http:server: Support HEAD method

ceolin:


ceolin:
merged, with the suggestion

PR #1210, opened 2015-12-21
build: Fix install message

ceolin:
The install dir consists of $DESTDIR/$PREFIX


ceolin:
merged, thanks

PR #1212, opened 2015-12-21
Mainloop improvements

barbieri:
Set of patches I wrote while flying back from Finland. They were mostly optimizations I was hoping to do for a long time, unfortunately I did not have time to write any specific benchmarks, but gains are expected particularly with small OS where the timers vector could be bouncing on "pop+insert_sorted".

### DIFF
--- a/src/lib/common/include/sol-glib-integration.h
+++ b/src/lib/common/include/sol-glib-integration.h
@@ -442,7 +442,7 @@ sol_glib_integration(void)
     mdata->timeout = -1;
     mdata->max_prio = 0;
 
-    msource = sol_mainloop_source_new(&_sol_glib_integration_source_type, gsource);
+    msource = sol_mainloop_source_add(&_sol_glib_integration_source_type, gsource);
     SOL_NULL_CHECK_GOTO(msource, failed_mainloop_source);
 
     g_main_context_ref(ctx);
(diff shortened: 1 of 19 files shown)


ibriano:
Quick look seems fine.

bdilly:
Seems good overall, but we had a fbp test depending on the following behaviour:

"If a node with timeout is declared first than other with the same timeout, it dispatch first"

Since it's not the case anymore, we need to change it, also avoiding this dependency on rules that we're not guarantying (it was an implementation detail).

So please do something like:

-ind(test/int-generator:sequence="1 0 1 0",interval=50)
+ind(test/int-generator:sequence="1 0 1 0",interval=49)

in src/test-fbp/switcher-input-select.fbp and we should be good.

Also there are some style issues, please take a look in semaphoreci build log for details.

barbieri:
I need to investigate further about the timeout. It shouldn't happen at all: timeouts are created using milliseconds, but internally they are kept as nanoseconds since the `sol_timeout_add()` call, so very likely the first node will get a smaller nanosecond than the second! :-)

But surely the proposed patch would solve it.

As for style, I expected some failures as I didn't have uncrustify to run check-style during my flight :-) Will fix and incorporate the suggestions for `insert_at()` and doxygen fixes, then send v2.

glima:
the @a + line break thing was fixed in doxygen, just tested here.

barbieri:
@bdilly it was actually a "bug" on my side, the following comparison should be `>= 0` to make it as before:

```c
    if (i + 1 < pv->base.len) {
        other = sol_ptr_vector_get_nocheck(pv, i + 1);
        if (compare_cb(ptr, other) > 0) // need to change to >=
            goto fix_it;
    }
```

bdilly:
@barbieri oh, ok. got it.

barbieri:
replaced with #1226

PR #1217, opened 2015-12-22
build: fix circular dependency

dorileo:
Some circular dependency may happen between .o and .o.dep files due
inter-dependent modules. This patch avoids that by filtering out the
actual object file from find-deps results.

A real current example is string-format.o <- string-format.o.dep.


dorileo:
@glima yep, you're right, I'm sending a new version with a proper fix. Thanks.

PR #1219, opened 2015-12-22
headers: Make headers buildable in C++ compiler

ceolin:


cmarcelo:
There are multiple changes in the same commit, e.g. removing uses of keyword C++ keywords `delete` and `namespace` from the headers, adding necessary casts because C++ doesn't automatically converts void* to other pointer types and so on.

I suggest splitting those changes, or at least explaining them in the commit message.

ceolin:
fjxes in https://github.com/solettaproject/soletta/pull/1238

PR #1220, opened 2015-12-22
oic: Make api similar with the rest of project

ceolin:
OIC server was using init/release, but over the project it's more usual
to use init/shutdown.


bdilly:
+1

@glima do you know any way to make code style checks to consider only changed lines / blocks instead of changed files?

glima:
@bdilly I don't get why would we want that

cabelitos:
+1

ceolin:
@glima in this case check-style is complaining with part of the code that was not changed, failing the semaphoreci

ceolin:
merged, thanks

PR #1223, opened 2015-12-23
style: force exactly one space before newline definition continuations.

glima:


glima:
I vote for at least committing the schema change, so we fix by chunks when needed.

ceolin:
I'm ok with the schema :)

glima:
@bdilly fine with committing just the schema file change as well?

mbelluzzo:
I vote for merging everything. Git log is your friend

bdilly:
@glima it seems a good approach. Just push the new schema and we change style when we need to modify these blocks.

glima:
Works for me, pushing.

PR #1224, opened 2015-12-23
mainloop: don't use log print function before sol_init().

glima:


glima:
Merging, thanks.

PR #1226, opened 2015-12-23
Mainloop improvements v2

barbieri:
Changes since v1 (#1212):
 * fixed doxygen, used `@brief` as recommended;
 * changed `insert_at()`, removed `after` parameter. It already handled the case where `index == (length-1)` to do an append.
 * fixed bug spotted by FBP test `switcher-input-select.fbp`, if updated item is `==` the next element, then we "fix it", adding it afterwards to preserve expected order.
 * `insert_at()` won't return the index anymore
 * functions that return index will explicitly use `int32_t`, so we're sure `uint16_t` will fit, as well as errors. Previously it could fail if `sizeof(int) == sizeof(uint16_t)`.
 * improve `find_sorted()` to not look twice at the matched index.

### DIFF
--- a/src/lib/common/include/sol-glib-integration.h
+++ b/src/lib/common/include/sol-glib-integration.h
@@ -442,7 +442,7 @@ sol_glib_integration(void)
     mdata->timeout = -1;
     mdata->max_prio = 0;
 
-    msource = sol_mainloop_source_new(&_sol_glib_integration_source_type, gsource);
+    msource = sol_mainloop_source_add(&_sol_glib_integration_source_type, gsource);
     SOL_NULL_CHECK_GOTO(msource, failed_mainloop_source);
 
     g_main_context_ref(ctx);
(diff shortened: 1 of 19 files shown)


gabrielschulhof:
I can't test this PR because
```
../build/soletta_sysroot/usr/include/soletta/sol-vector.h:355:47: error: invalid conversion from ‘void*’ to ‘void**’ [-fpermissive]
     data = sol_vector_get_nocheck(&pv->base, i);
```

If this is already fixed on master, please rebase this PR so I can continue testing! It would actually be good to rebase onto master anyway, because there are conflicts.

barbieri:
Hi Gabriel, I was out of access to my development machine, but will update this PR soon.

Until then, you can add a temporary commit that does a cast.

barbieri:
rebased and c++ fixes in #1267

barbieri:
Hi Gabriel,

This was fixed in v3 (#1267) and it was already merged to master.

Thanks for testing :-)

PR #1227, opened 2015-12-23
oic: Fix /oic/d and /oic/p url behavior

otaviobp:
Supported to well known url /oic/d was implemented in soletta, but with
the expected result to /oic/p url. So /oic/d was renamed to /oic/p and
/oic/d was implemented.

Also sending platform_id in expected way in /oic/p url.


otaviobp:
merged.

PR #1228, opened 2015-12-23
Locale: Fix missing locale symbols in some platforms.

cabelitos:
Some platforms may support more categories, in order to do not
break the build, these extented categories should be handled
in their spefic backend.


barbieri:
+1

cabelitos:
@glima Ok I will add that and merge it.

cabelitos:
merged.

PR #1230, opened 2015-12-23
Flow docs

mbelluzzo:
and with this, we hit 100% of c api documented!


edersondisouza:
Looks good overall, however there are lots of `@c` referring to function parameters, but I'm afraid that they should be referred via `@a`... =/

PR #1231, opened 2015-12-23
Fixes on keyboard node types

bdilly:
Next I'll add another mode that will make it possible
to know when a key was pressed and released.


glima:
+1

bdilly:
merged, thanks

PR #1232, opened 2015-12-24
flow:iio: Add Magnet sensor Category support into iio node

guchaojie:
Change-Id: I917796bac0b403653ef26013351a435f6ee9f26a

### DIFF
--- a/src/lib/io/include/sol-aio.h
+++ b/src/lib/io/include/sol-aio.h
@@ -40,7 +40,8 @@ extern "C" {
 
 /**
  * @file
- * @brief These routines are used for analog I/O access under Soletta.
+ * @brief These routines are used for analog I/O access (reading
+ * from analog-to-digital converters) under Soletta.
  */
 
 /**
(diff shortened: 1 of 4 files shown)


guchaojie:
@edersondisouza  This PR is about magnet category . However, We have another about 5 sensor category which is similar with previous category. May I submit one PR to support these 5 types for your review instead of submitting one by one?

edersondisouza:
@guchaojie About more categories to be added, you can send them in one PR, sure, but on different commits - one for each sensor, please. This way we keep them better organised, I think.
And now that you are working on this batch of patches, you could 'factor out' some common code, to avoid duplication. For instance, macro `ADD_CHANNEL` could take the whole name of channel, and be defined for the whole 'nodes.c' file. The 'open' function for each sensor also shares lots of similarities, maybe add some helpers to set node data `mdata` based on options data `options`. Same goes for 'tick' functions...
That's it - now I'll wait eagerly these new patches =D

guchaojie:
@edersondisouza  We think about factoring out common code for some time at the beginning to do work. 
For open function, I think we need define new function to use "switch ..case" method to define different custom data type and function name such as magnet_read_cb ...

guchaojie:
@edersondisouza  We think about factoring out common code for some time at the beginning to do work. But at that time, we found that it did not make sense to modify existing code architecture according to flow node design. In iio.json file, every type  has one to one correspondence with a serious of sensor operation and custom data type. Although the implementation is similar , every category has its own custom type such as pressure_data, temperature_data and so on.  So you can hardly differentiate with different code path in runtime. Only we can do is to define at least 2 "switch ..case" to define different custom data type and pass in different code section and the problem is that only we can know which custom type definition is to get the function name such as pressure_open from json file definition, which cannot get type name according to function name. 
All in all, I think we can keep current source code architecture to extend sensor category fastly, at the same time , we will try to factor out more common code step by step.

PR #1233, opened 2015-12-24
WIP: Node.js bindings

gabrielschulhof:
This PR adds the infrastructure for creating Node.js bindings, and binds the very simple function ```sol_platform_get_machine_id()```. Some things to note/problems to work out:

0. <s>I also needed to bind dlfcn, because I need to try loading libsoletta.so from "./build/soletta_sysroot/usr/lib" before I can load the bindings. I wanted to do that to avoid using rpath, which would make it difficult to implement installation later on.</s>

0. I don't exactly know what I should run in response to make install. Should I do ```npm install -g```?

0. <s>A<a id="build-integration">l</a>though I've tried to integrate the build process with Kconfig, building the bindings currently only works if you start via npm install. That's because if you start with make menuconfig, turn on the bindings, and then run make, it runs npm install after it's done building soletta, but because node-gyp also uses ```build/``` as the output directory, it first deletes it and then creates it. This causes ```build/soletta_sysroot``` to disappear, so I can't subsequently use the header files and the .so that I need. I can do hacky things like ```cp -a build/soletta_sysroot somewhere/else```, run npm install, and copy it back to ```build/soletta_sysroot``` from inside the node-gyp build process, but I'd rather discuss this with you. Is it possible that we change the output from ```build/soletta_sysroot``` to something else, like, maybe, ```out/soletta_sysroot```?

    If you start with npm install, it works because it runs make alldefconfig && make as part of npm install, so by the time it runs those, node-gyp will already have performed the step of removing ```build/``` and creating a new directory ```build/```.</s>

### DIFF
--- a/Kconfig
+++ b/Kconfig
@@ -216,3 +216,12 @@ menu "Test suite"
 depends on FEATURE_RUNNABLE_PROGRAMS
 source "src/test/Kconfig"
 endmenu
+
+menu "Bindings"
+config NODE_BINDINGS
+	bool "Node.js bindings"
+	depends on HAVE_NODE
+	default n
+	help
+		Enable Node.js bindings
+endmenu
--- a/Makefile
+++ b/Makefile
@@ -20,6 +20,23 @@ include $(top_srcdir)tools/build/Makefile.vars
 
 include $(top_srcdir)tools/build/Makefile.common
 
+ifneq ($(NODE_BINDINGS),)
+
+bins-out += node_bindings
+
+node_bindings: $(SOL_LIB_OUTPUT)
+
+	# Install dependencies without building the package
+	npm install --ignore-scripts
+
+	# Build the package without clobbering build/
+	SOLETTA_FROM_MAKE=true $$(dirname `which node`)/../lib/node_modules/npm/bin/node-gyp-bin/node-gyp configure
+	SOLETTA_FROM_MAKE=true $$(dirname `which node`)/../lib/node_modules/npm/bin/node-gyp-bin/node-gyp build
+
+PHONY += node_bindings
+
+endif
+
 # kconfig interface rules
 ifeq (help, $(filter help,$(MAKECMDGOALS)))
 help: soletta_help
--- a/bindings/nodejs/build-csdk.sh
+++ b/bindings/nodejs/build-csdk.sh
@@ -1,0 +1,64 @@
+#!/bin/bash
+
+# This file is part of the Soletta Project
+#
+# Copyright (C) 2015 Intel Corporation. All rights reserved.
+#
+# Redistribution and use in source and binary forms, with or without
+# modification, are permitted provided that the following conditions
+# are met:
+#
+#   * Redistributions of source code must retain the above copyright
+#     notice, this list of conditions and the following disclaimer.
+#   * Redistributions in binary form must reproduce the above copyright
+#     notice, this list of conditions and the following disclaimer in
+#     the documentation and/or other materials provided with the
+#     distribution.
+#   * Neither the name of Intel Corporation nor the names of its
+#     contributors may be used to endorse or promote products derived
+#     from this software without specific prior written permission.
+#
+# THIS SOFTWARE IS PROVIDED BY THE COPYRIGHT HOLDERS AND CONTRIBUTORS
+# "AS IS" AND ANY EXPRESS OR IMPLIED WARRANTIES, INCLUDING, BUT NOT
+# LIMITED TO, THE IMPLIED WARRANTIES OF MERCHANTABILITY AND FITNESS FOR
+# A PARTICULAR PURPOSE ARE DISCLAIMED. IN NO EVENT SHALL THE COPYRIGHT
+# OWNER OR CONTRIBUTORS BE LIABLE FOR ANY DIRECT, INDIRECT, INCIDENTAL,
+# SPECIAL, EXEMPLARY, OR CONSEQUENTIAL DAMAGES (INCLUDING, BUT NOT
+# LIMITED TO, PROCUREMENT OF SUBSTITUTE GOODS OR SERVICES; LOSS OF USE,
+# DATA, OR PROFITS; OR BUSINESS INTERRUPTION) HOWEVER CAUSED AND ON ANY
+# THEORY OF LIABILITY, WHETHER IN CONTRACT, STRICT LIABILITY, OR TORT
+# (INCLUDING NEGLIGENCE OR OTHERWISE) ARISING IN ANY WAY OUT OF THE USE
+# OF THIS SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
+
+if test "x${V}x" != "xx"; then
+	set -x
+fi
+
+DO_DEBUG=false
+
+while test $# -gt 0; do
+	if test "x$1x" = "x--debugx" -o "x$1x" = "x-dx"; then
+		DO_DEBUG=true
+	elif test "x$1x" = "x--helpx" -o "x$1x" = "x-hx"; then
+		echo "$( basename "$0" ) [options...]"
+		echo ""
+		echo "Possible options:"
+		echo "--debug or -d : Build in debug mode"
+		exit 0
+	fi
+	shift
+done
+
+unset PYTHON || exit 1
+unset PYTHON_PATH || exit 1
+make alldefconfig || exit 1
+if test "x${DO_DEBUG}x" == "xtruex"; then
+	( cat .config | awk '{
+		if ( $0 ~ /^CONFIG_CFLAGS=/ ) {
+			print( "CONFIG_CFLAGS=\"-g -Wall -O0\"" );
+		} else {
+			print;
+		}
+	}' > .config.new && mv .config.new .config ) || exit 1
+fi
+make || exit 1
--- a/bindings/nodejs/establish-flags.sh
+++ b/bindings/nodejs/establish-flags.sh
@@ -1,0 +1,61 @@
+#!/bin/bash
+
+# This file is part of the Soletta Project
+#
+# Copyright (C) 2015 Intel Corporation. All rights reserved.
+#
+# Redistribution and use in source and binary forms, with or without
+# modification, are permitted provided that the following conditions
+# are met:
+#
+#   * Redistributions of source code must retain the above copyright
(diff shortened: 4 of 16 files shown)


poussa:
@bdilly We would like to merge this. PTAL.

bdilly:
@poussa oh, sure. That's great. I'll make sure this pull request is reviewed tomorrow.

@gabrielschulhof for next pull requests, I would like to suggest to drop the "WIP" from the pr title when it's good to be merged. Some guys were avoiding to invest much time reviewing it because they considered you were still working on it.

Makefile
```diff
@@ -20,6 +20,23 @@ include $(top_srcdir)tools/build/Makefile.vars
 
 include $(top_srcdir)tools/build/Makefile.common
 
+ifneq ($(NODE_BINDINGS),)
+
+bins-out += node_bindings
```
barbieri: not sure `bins-out` is the best variable to append to.

@dorileo could you check?

bindings/nodejs/build-csdk.sh
```diff
@@ -0,0 +1,64 @@
...
+# A PARTICULAR PURPOSE ARE DISCLAIMED. IN NO EVENT SHALL THE COPYRIGHT
+# OWNER OR CONTRIBUTORS BE LIABLE FOR ANY DIRECT, INDIRECT, INCIDENTAL,
+# SPECIAL, EXEMPLARY, OR CONSEQUENTIAL DAMAGES (INCLUDING, BUT NOT
+# LIMITED TO, PROCUREMENT OF SUBSTITUTE GOODS OR SERVICES; LOSS OF USE,
+# DATA, OR PROFITS; OR BUSINESS INTERRUPTION) HOWEVER CAUSED AND ON ANY
+# THEORY OF LIABILITY, WHETHER IN CONTRACT, STRICT LIABILITY, OR TORT
+# (INCLUDING NEGLIGENCE OR OTHERWISE) ARISING IN ANY WAY OUT OF THE USE
+# OF THIS SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
+
+if test "x${V}x" != "xx"; then
```
barbieri: not that relevant but since you use `#!/bin/bash` you can use the `-n`:

```sh
if [ -n "$V" ]; then
    set -x
fi
```

bindings/nodejs/build-csdk.sh
```diff
@@ -0,0 +1,64 @@
...
+	set -x
+fi
+
+DO_DEBUG=false
+
+while test $# -gt 0; do
+	if test "x$1x" = "x--debugx" -o "x$1x" = "x-dx"; then
```
barbieri: same here, using bash you can simplify with:

```sh
if [ "$1" = "--debugx" ] || [ "$1" = "-dx" ]; then
```

bindings/nodejs/build-csdk.sh
```diff
@@ -0,0 +1,64 @@
...
+unset PYTHON_PATH || exit 1
+make alldefconfig || exit 1
+if test "x${DO_DEBUG}x" == "xtruex"; then
+	( cat .config | awk '{
+		if ( $0 ~ /^CONFIG_CFLAGS=/ ) {
+			print( "CONFIG_CFLAGS=\"-g -Wall -O0\"" );
+		} else {
+			print;
+		}
+	}' > .config.new && mv .config.new .config ) || exit 1
```
barbieri: this is nasty :-(

Can you confirm you need this here? IMO it would be better handled by people picking the correct CFLAGS in Soletta's kconfig/kbuild

bindings/nodejs/establish-flags.sh
```diff
@@ -0,0 +1,61 @@
...
+# SPECIAL, EXEMPLARY, OR CONSEQUENTIAL DAMAGES (INCLUDING, BUT NOT
+# LIMITED TO, PROCUREMENT OF SUBSTITUTE GOODS OR SERVICES; LOSS OF USE,
+# DATA, OR PROFITS; OR BUSINESS INTERRUPTION) HOWEVER CAUSED AND ON ANY
+# THEORY OF LIABILITY, WHETHER IN CONTRACT, STRICT LIABILITY, OR TORT
+# (INCLUDING NEGLIGENCE OR OTHERWISE) ARISING IN ANY WAY OUT OF THE USE
+# OF THIS SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
+
+# By default, build soletta using defaults
+SOLETTA_CFLAGS="-I$(pwd)/build/soletta_sysroot/usr/include/soletta"
+SOLETTA_LIBS="-L$(pwd)/build/soletta_sysroot/usr/lib -lsoletta -Wl,-rpath $(pwd)/build/soletta_sysroot/usr/lib"
```
barbieri: you can get these from Soletta's tools/build/Makefile.vars `LINKED_OBJS_LDFLAGS`.

bindings/nodejs/establish-flags.sh
```diff
@@ -0,0 +1,61 @@
...
+if test "x${SOLETTA_FROM_MAKE}x" == "xtruex"; then
+	PREFIX="$(cat .config | grep '^PREFIX=' | sed -r 's/^[^"]*"([^"]*)"/\1/')"
+	PREFIX="${PREFIX##/}"
+
+	SOLETTA_CFLAGS="-I$(pwd)/build/soletta_sysroot/${PREFIX}/include/soletta"
+	SOLETTA_LIBS="-L $(pwd)/build/soletta_sysroot/${PREFIX}/lib -lsoletta"
+	BUILD_SOLETTA="false"
+
+# Otherwise, try to find soletta via pkg-config
+elif pkg-config --exists soletta > /dev/null 2>&1; then
```
barbieri: you can simplify your logic if you get Soletta's `build/soletta_sysroot/$PREFIX`, then you can handle only `pkg-config` case by doing something like:

```sh

if [ "$SOLETTA_FROM_MAKE" = "true" ]; then
   export PKG_CONFIG_PATH="$SOLETTA_BUILD_DIR/lib/pkgconfig:$PKG_CONFIG_PATH"
fi

# do what you did:
if pkg-config --exists soletta /dev/null 2>&1; then
...
```

Kconfig
```diff
@@ -216,3 +216,12 @@ menu "Test suite"
 depends on FEATURE_RUNNABLE_PROGRAMS
 source "src/test/Kconfig"
 endmenu
+
+menu "Bindings"
+config NODE_BINDINGS
```
barbieri: you better call it `USE_NODEJS`, it's more in the lines of what we did. Also calling it `NODEJS` instead of just `NODE` in Makefile and Kconfig may help others to understand what it means (Soletta's Flow already use the term 'node').

Kconfig
```diff
@@ -216,3 +216,12 @@ menu "Test suite"
...
+	bool "Node.js bindings"
+	depends on HAVE_NODE
+	default n
```
barbieri: if you're testing it and it works, we better `default y`.

And we should also include this in our semaphore so it builds your bindings as well.

bindings/nodejs/build-csdk.sh
gabrielschulhof: Remember that this script only runs when you do ```npm install```. I'd like to build soletta without optimization and with debug symbols in response to
```
npm install --debug
```
I was looking for a way of manipulating the configuration in a more structured manner, but in the end I found none.

I've also noticed that the soletta CI process does something similar in the CI
```
sed -i .config -e "s/\(.*\)OIC\(.*\)=y/\1OIC\2=n/" -e "s/\(.*\)DTLS\(.*\)=y/\1DTLS\2=n/"
```
so I went ahead and did that. I suppose I could change it to ```sed -i .config ...```.

bindings/nodejs/establish-flags.sh
gabrielschulhof: I cannot use ```$SOLETTA_BUILD_DIR/lib/pkgconfig/soletta.pc``` because that file contains information that is valid only after soletta is installed. Since ```$SOLETTA_BUILD_DIR/lib/pkgconfig/soletta.pc``` refers to the prefix where soletta *will be* installed, those places are still empty during build time, so, when I try to use the includedir or the libdir given by ```$SOLETTA_BUILD_DIR/lib/pkgconfig/soletta.pc``` during the npm build process, it will fail.

bindings/nodejs/build-csdk.sh
gabrielschulhof: Good point! Maybe I shouldn't assume bash. After all, we're aiming at IoT, so we might end up in a busybox build environment.

gabrielschulhof:
Hey! I'm addressing these issues. Please hold off with merging! I'll put the "WIP" back and take it off when I'd like another review.

bdilly:
@gabrielschulhof that's fine. I believe @cabelitos and @brunobottazzini will take a look on that soon, so you can make a new version based on many different feedback.

@gabrielschulhof also, I would like to suggest you to read our contributions flow.
 https://github.com/solettaproject/soletta/wiki/Contributing#contributions-flow

bindings/nodejs/build-csdk.sh
barbieri: I doubt, go the easy path and enforce bash. Generic shells are PITa

bindings/nodejs/establish-flags.sh
barbieri: bummer... I found `pkg-config --define-variable=prefix=$PWD`, BUT we don't use `${prefix}` in our `soletta.pc`. I'll change that.

barbieri: I did a PR #1362 to solve this, the message explains how to use it, but it's basically call pkg-config with `PKG_CONFIG_PATH` set and use `--define-variable=prefix=$PWD/build/soletta_sysroot/$PREFIX` when `$SOLETTA_FROM_MAKE = true`.

Please clean your tree after updating, or remove soletta.pc so it's re-generated.

brunobottazzini:
It looks good to me.

cabelitos:
The rest looks good.

gabrielschulhof:
@cabelitos Roger that! bash it is.

gabrielschulhof:
@cabelitos Though wait ... I'm trying to use lowest-common-denominator notation here - so, like, unless you're using ```csh``` or ```zsh``` it should work. semaphore uses Ubuntu, right? So, I'll try to run the build process on semaphore and see if it work.

PR #1235, opened 2015-12-28
oic: Make oic packets payload compatible with iotivity 1.0.1 - v2

otaviobp:
There was a rebase problem in v1 fixed in v2.

Replaces #1222


otaviobp:
merged

PR #1238, opened 2015-12-28
headers: Make headers buildable in C++ compiler v2

ceolin:
from v1:
 - Split the fixes in different commits
 - Use some `#define` magics to not duplicate code


cmarcelo:
+1

Please check semaphoreci, but it seems it's just uncrustify not understanding the code... :/

Issue #1234, opened 2015-12-28
sol-platform.h uses constant not defined in public headers: NSEC_PER_MSEC


gabrielschulhof:
sol-platform.h only includes sol-vector.h and no other file, and NSEC_PER_MSEC is not defined in either of those files.

I suppose sol-platform.h should include sol-util.h, which is where NSEC_PER_MSEC is defined, but sol-util.h does not appear to be a public header.

ceolin:
Fixed on 7f4cdd6bb9a495944440d97725c4b2aac0cd525c

PR #1241, opened 2015-12-28
fix: Public header uses private thing

ceolin:
Since the value used in the public header is arbitrary and the constant
is not necessary, it's just changing the code to use the value directly.

### DIFF
--- a/src/lib/common/include/sol-platform.h
+++ b/src/lib/common/include/sol-platform.h
@@ -74,7 +74,7 @@ extern "C" {
  * When this time is reached, stops the * chunk reading and allow mainloop to run again.
  * This keeps interactivity.
  */
-#define CHUNK_MAX_TIME_NS (20 * (NSEC_PER_MSEC))
+#define CHUNK_MAX_TIME_NS (20 * 1000000)
 
 /**
  * @brief Retrieves the name of the board on which Soletta is running.


ceolin:
problem spotted in https://github.com/solettaproject/soletta/issues/1234

src/lib/common/include/sol-platform.h
```diff
@@ -74,7 +74,7 @@ extern "C" {
  * When this time is reached, stops the * chunk reading and allow mainloop to run again.
  * This keeps interactivity.
  */
-#define CHUNK_MAX_TIME_NS (20 * (NSEC_PER_MSEC))
+#define CHUNK_MAX_TIME_NS (20 * (1000000000ULL))
```
lpereira: Some platforms (such as Riot) are still defining `time_t` to be 32-bit, so the `ULL` can't be used without a check.

ceolin: i'll see it, but just to record, it is already happening, this value is defined in `sol-util.h` and there is any check there.

ibriano:
Fine by me

ceolin:
merged, thanks

barbieri:
I wonder why these symbols are defined in sol-platform.h at all, and worse: why are them missing "SOL_" prefix?

PR #1243, opened 2015-12-28
Mark noreturn function with SOL_ATTR_NORETURN

thiagomacieira:
This avoids the mistaken interpretation that the function returned to
the caller from inside the child process.

### DIFF
--- a/src/modules/linux-micro/rc-d/rc-d.c
+++ b/src/modules/linux-micro/rc-d/rc-d.c
@@ -62,7 +62,7 @@ static struct sol_ptr_vector monitors = SOL_PTR_VECTOR_INIT;
 static struct sol_timeout *monitor_timer;
 static struct sol_vector pendings = SOL_VECTOR_INIT(struct pending);
 
-static void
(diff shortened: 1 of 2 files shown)


src/modules/linux-micro/rc-d/rc-d.c
```diff
@@ -177,7 +177,7 @@ rc_d_stop(const struct sol_platform_linux_micro_module *mod, const char *service
             sigemptyset(&emptyset);
             sigprocmask(SIG_SETMASK, &emptyset, NULL);
             find_exec(service, "stop");
-            return -errno;
+            /* unreachable */
```
lpereira: Maybe `__builtin_unreachable()`, so ubsan can catch it?

thiagomacieira: I can do that. Do you want me to create a macro? If so, where?

glima: sol-macros.h, I guess

glima:
Looking good, will merge.

PR #1244, opened 2015-12-28
Use C11 <stdatomic.h> and its functions instead of GCC instrinics

thiagomacieira:
This allows the code to build with other compilers that may have
different implementations, like Clang.

### DIFF
--- a/doc/doxygen/Doxyfile
+++ b/doc/doxygen/Doxyfile
@@ -272,7 +272,7 @@ EXPAND_ONLY_PREDEF     = YES
 SEARCH_INCLUDES        = YES
 INCLUDE_PATH           =
 INCLUDE_FILE_PATTERNS  =
-PREDEFINED             = __attribute__((x))=
+PREDEFINED             = __attribute__((x))=  DOXYGEN_RUN
 EXPAND_AS_DEFINED      =
 SKIP_FUNCTION_MACROS   = YES
 #---------------------------------------------------------------------------
--- a/src/lib/common/Makefile
+++ b/src/lib/common/Makefile
@@ -100,6 +100,7 @@ requires-private-$(PLATFORM_SYSTEMD) += \
 
 
 headers-y := \
+    include/sol-atomic.h \
     include/sol-log.h \
     include/sol-macros.h \
     include/sol-mainloop.h \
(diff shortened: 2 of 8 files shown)


lpereira:
We can't use this (or other C11-only features, such as `_Generic`) because the CI server uses a pre-historic C compiler.

thiagomacieira:
GCC stdatomic.h was added in GCC 4.9. The __atomic intrinsics were added in 4.7, so it's not that long ago. Besides, we can just copy GCC's stdatomic.h and ship alongside Soletta. Do you want me to write a configure-time check plus sol_atomic.h? If so, please suggest where sol_atomic.h should be placed.

_Generic was also added in 4.9.

ulissesf:
+1 for Thiago's idea of sol_atomic.h and use C11 stdatomic.h and friends.

barbieri:
I don't think we need another file, use sol-macros.h, it's not that big and already does some check on the GCC.

If we want  to be clear and separate in another file, then remember to use dash "-" instead of underscore for file names.

Last but not least, if it may cause problems with non GCC compilers, then let's prefix them with "SOL_" namespace. I'm not sure this is required.

thiagomacieira:
I think the point was that we could well use an atomic API. We'll start with the GCC and C11 implementations. If someone needs something different, they can contribute it (e.g., Windows).

ulissesf:
Agree, Thiago. And I should have checked C11 to begin with and used stdatomic.h, my bad.

thiagomacieira:
I don't understand the failure. sol-atomic.h is there.

ulissesf:
@thiagomacieira, forgot to add sol-atomic.h to Makefile under headers-y, maybe?

thiagomacieira:
I didn't know I had to.

thiagomacieira:
Please help with the Doxygen failure. Why doesn't it see the #defines?

barbieri:
because doxygen is stupid and if it's not declared (ie: inside `#if` or `#ifdef`), Doxygen doesn't take it.

See my "solution" at https://github.com/solettaproject/soletta/blob/master/src/lib/common/include/sol-mainloop.h#L263

```c
#ifndef SOL_MAINLOOP_FD_ENABLED
/* keep doxygen happy */
#define SOL_MAINLOOP_FD_ENABLED
#undef SOL_MAINLOOP_FD_ENABLED
#endif
```

since FD may exist or not (ie: RIoT)

bdilly:
Commits merged on PR #1943 

Thanks

PR #1246, opened 2015-12-29
systemd: Fix warnings from sol-bus update

vcgomes:
Now that sol_bus_client() keeps track of the bus and service name, we
can re-use that information instead of hardcoding it, or retrieving it
every time.


vcgomes:
Merged. Thanks.

PR #1250, opened 2015-12-30
AIO - move API to be asynchronous

glima:
We have come across OSes in which this API is not synchronous, so we got to change that in Soletta as well.


mbelluzzo:
LGTM

glima:
Thanks. I just tested with your Edison image and we're fine. Merging.

PR #1252, opened 2016-01-04
Fixes

glima:


glima:
Thanks a lot, fixing that and merging.

PR #1253, opened 2016-01-05
mavlink: Fix uninitialized variable

ceolin:
mask can be used without be initialized.


dorileo:
+1

PR #1254, opened 2016-01-05
oic: Parameter not check on public function

ceolin:
It's de-referencing some parameters without any check in a public function.


ceolin:
merged

PR #1256, opened 2016-01-07
support another 5 category sensors into soletta iio node

guchaojie:
This PR supports Magnet , Pressure, Color , Huminity and Accelerate Category into iio node. And all the sensor category have been tested to work well in minnowmax using different sensors supported by their iio drivers.

### DIFF
--- a/src/lib/common/sol-mainloop-common.c
+++ b/src/lib/common/sol-mainloop-common.c
@@ -184,14 +184,16 @@ sol_mainloop_common_timeout_process(void)
     sol_mainloop_impl_unlock();
 
     now = sol_util_timespec_get_current();
-    for (i = 0; i < TIMEOUT_PROCESS.base.len; i++) {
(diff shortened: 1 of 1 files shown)


guchaojie:
@edersondisouza This PR is focus on extending sensor category support list  in soletta for QA test. Next step we will optimize the source code such as factoring out common code as your suggestion

guchaojie:
@bdilly Anyone can have a review this PR? Any comment is welcome.

barbieri:
sorry, both @bdilly and @edersondisouza are on vacation :-(

the patches looks good, keep going with other patches and they will review as soon as they are back.

edersondisouza:
I'm OK with factoring out common code on a future patch, but wouldn't do it before simplify the others patches?

guchaojie:
@edersondisouza yes, our plan is to extending sensor category to support 30 sensors before M3. In case to avoid introducing bugs in this process , so we factoring out common in next patch. That would make easy to maintain the code step by step.

PR #1258, opened 2016-01-08
Fixes http

ceolin:
- [client] Handle post data correctly (using multipar/form-data)
- [client] Allows one send file contents just giving a file path
- [client] Add support multipar/form-data on sample
- [server] remove unused code
- [server] handle post data properly


cabelitos:
Be aware that the semaphore is failing, it looks like that the machine has an old Curl version.

barbieri:
so we need to bump required curl version

ceolin:
i'll check the which version it was introduced and bump the curl's version

PR #1259, opened 2016-01-11
LWM2M: Add LWM2M server. 

cabelitos:
Changes since V1:

* Split commits.

This patch introduces a LWM2M server with the following features:

* Registration interface.
* Discovery interface.
* Management interface.
* TLV data types.

What is not supported yet.
* Secure connections. (Data is not encrypted right now)
* LWM2M JSON.
* Queue mode operation.
* SMS support.


cabelitos:
Check #1286

PR #1264, opened 2016-01-12
flow: Add robotics module

### LINKED ISSUE #1261: H-Bridge node
H-Bridge circuits are common in robotics and can be used to allow voltage to be applied across a load in either direction.

Based on 4 switches (S1 to S4) connected as an "H" with the motor in the inner link, we have the following table:

H-Bridge circuit from WikiPedia

S1 | S2 | S3 | S4 | Result
----|----|----|----|----
1 | 0 | 0 | 1 | Motor moves right
0 | 1 | 1 | 0 | Motor moves left
0 | 0 | 0 | 0 | Motor coasts
0 | 1 | 0 | 1 | Motor brakes
1 | 0 | 1 | 0 | Motor brakes
1 | 1 | 0 | 0 | Short circuit
0 | 0 | 1 | 1 | Short circuit
1 | 1 | 1 | 1 | Short circuit

I'd say this is should be either inside "switcher" or be another module similar to that one, as it should support multiple kinds of types, like irange, boolean.

References:
 - https://en.wikipedia.org/wiki/H_bridge

Requested by: @lpereira and @rchiossi

lpereira:
This commit adds the robotics module, implementing a simple H-Bridge
node type.

Closes #1261.


lpereira:
Thanks for the review! Closing for newer version.

PR #1265, opened 2016-01-12
Quadrature encoder

lpereira:
Please disregard the `Add robotics module`, which has its own pull request. Since this one is made on top of that, the commit is part of the patch set.

This adds a quadrature encoder to the robotics module, compatible with the one here: https://web.archive.org/web/20150728092206/http://letsmakerobots.com/node/24031


lpereira:
Thanks for the review! Closing for newer version.

PR #1266, opened 2016-01-12
Make headers compatible with C++

ceolin:
- add necesary casts in glib-integration
- make sol-log (int-check) compatible with c++
- added target to test public headers with g++
- added missing stdbool include

### DIFF
--- a/data/templates/headers-public.tpl
+++ b/data/templates/headers-public.tpl
@@ -1,0 +1,10 @@
+{{
+import os
+
+def gen_headers(headers):
+    headers = os.getenv(headers)
+    headers = headers.strip()
+    headers = headers.split(" ")
+    for h in headers:
+    	st.println("#include \"%s\"" % (h.split("/")[-1]))
+}}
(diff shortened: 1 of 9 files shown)


ceolin:
@thiagomacieira could you please take a look in https://github.com/solettaproject/soletta/commit/e47cf001ab7d704db2575b74d1cbeea0961e6451

I guess there are better solutions for this problem. I could not find something equivalent with `__builtin_choose_expr()` in c++.

ceolin:
we can avoid the `__builtins_types_compatible()` but we still can't have the same implementation that C. `_SOL_TYPE_FORMAT` will be processed as the function `_sol_int_format()` and we will not be able to concat it in the macro.

We will still need get the result and concat with the other parameters using the stack.

Am I wrong ?

vcgomes:
Yeah, I see the problem. I tried to come up with something, without success. @cmarcelo @thiagomacieira ideas?

thiagomacieira:
I haven't had the time to look into this yet. Give me a day or two.

ceolin:
don't worry, thank you for helping with it.

thiagomacieira:
Added a few comments.

ceolin:
committed with suggestions

PR #1267, opened 2016-01-13
Mainloop improvements v3

barbieri:
Changes since v2 (#1226):
 - rebased to master
 - C++ fixes

@ceolin will add a `make check-cxx` that will get all of our public headers, include them and compile with C++, this will avoid problems like that in the future. Grantes we still need to use some macros in that test, so the code expands, I spotted our `_FOREACH` macros would fail with C++.


barbieri:
failing style is not even due my changes, so nothing to worry about.

I will commit this

PR #1268, opened 2016-01-13
coap: Add unknown resource flag.

cabelitos:
This flag is useful when one wants to treat by oneself the requests that
could not be matched with the registered resources.

### DIFF
--- a/src/lib/comms/include/sol-coap.h
+++ b/src/lib/comms/include/sol-coap.h
@@ -242,7 +242,12 @@ typedef enum {
 enum sol_coap_flags {
     SOL_COAP_FLAGS_NONE       = 0,
     /** If the resource should be exported in the CoRE well-known registry. */
-    SOL_COAP_FLAGS_WELL_KNOWN = (1 << 1)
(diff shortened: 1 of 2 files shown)


cabelitos:
@vcgomes Could you please take a look?

barbieri:
yeah, I'd do this differently as well, with an explicit call `sol_coap_server_set_unknown_resource_handler(struct sol_coap_server *server, int (*handler)(struct sol_coap_server *server, struct sol_coap_packet *req, const struct sol_network_link_addr *cliaddr, void *data), const void *data)`, or something similar to that.

It's not a resource per se, it will handle getting the opcode with `sol_coap_header_get_code(req)` is that is needed, it will evaluate the path if needed, etc.

We can also keep a single public API, to unset it just call with `NULL`.

cabelitos:
I liked the @barbieri suggestion, the implementation is also simplier. I think I will stick with it.
What you guys think?
@vcgomes

vcgomes:
Go ahead. @barbieri solution sounds cleaner indeed.

cabelitos:
Check #1272

PR #1272, opened 2016-01-14
coap: Add unknown resource handler. v2

cabelitos:


cabelitos:
@vcgomes @barbieri Could you guys please take a look?

src/lib/comms/include/sol-coap.h
```diff
@@ -855,6 +855,21 @@ int sol_coap_cancel_send_packet(struct sol_coap_server *server, struct sol_coap_
...
+ * Everytime the @a server receives a request for a #sol_coap_resource that was not
+ * registered with sol_coap_server_register_resource() the @a handler will be called.
+ *
+ * @param server The server to register the unknown resource handler.
+ * @param handler The unknown handler callback or @c NULL to unregister.
+ * @param data The data to @a handler
+ * @return 0 on success
+ *         -EINVAL if server is @c NULL.
+ */
+int sol_coap_server_set_unknown_resource_handler(struct sol_coap_server *server, int (*handler)(struct sol_coap_server *server, struct sol_coap_packet *req, const struct sol_network_link_addr *cliaddr, void *data), void *data);
```
barbieri: I'd introduce the new API "right", with `const` here. So we need to touch less lines when we fix the header in the future (it also generate changes to more parts, like the .c)

src/lib/comms/sol-coap.c
```diff
@@ -87,6 +87,10 @@ struct sol_coap_server {
     struct sol_ptr_vector pending; /* waiting pending replies */
     struct sol_ptr_vector outgoing; /* in case we need to retransmit */
     struct sol_socket *socket;
+    int (*unknown_handler)(struct sol_coap_server *server,
+        struct sol_coap_packet *req,
+        const struct sol_network_link_addr *cliaddr, void *data);
+    void *unknown_handler_data;
```
barbieri: keep as const, cast at callback time.

src/lib/comms/sol-coap.c
```diff
@@ -1295,6 +1299,10 @@ respond_packet(struct sol_coap_server *server, struct sol_coap_packet *req,
         return cb(server, resource, req, cliaddr, (void *)c->data);
     }
 
+    if (server->unknown_handler)
+        return server->unknown_handler(server, req, cliaddr,
+            server->unknown_handler_data);
```
barbieri: hum... just noticed it here: our convention is `data` as the first parameter, not the last. (You can change it now or later and fix all of `sol-coap.h`)

src/lib/comms/sol-coap.c
cabelitos: ok.

src/lib/comms/include/sol-coap.h
cabelitos: ok.

src/lib/comms/sol-coap.c
cabelitos: I think I will change this one for now and later I come back and change the whole coap.

cabelitos:
Check #1276

PR #1274, opened 2016-01-15
support another 7 category sensors into soletta iio node

guchaojie:
This PR supports Magnet , Pressure, Color , Huminity, Accelerate, light and ADC Category into iio node. And all the sensor category have been tested to work well in minnowmax using different sensors supported by their iio drivers.

### DIFF
--- a/src/lib/common/include/sol-log.h
+++ b/src/lib/common/include/sol-log.h
@@ -93,7 +93,7 @@ extern "C" {
 #define SOL_NULL_CHECK(ptr, ...) \
     do { \
         if (SOL_UNLIKELY(!(ptr))) { \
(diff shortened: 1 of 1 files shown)


guchaojie:
@edersondisouza I have updated patches and add more  2 category into PR. Please have a review.

mbelluzzo:
Would be really nice to have a sample with the fbp you used to test your code, dont need to be in this PR, and a picture of the assembly as we have for the grove kit samples. The picture is important because a future IDE could (will ;-) ) present it when someone try to run a sample.

guchaojie:
@mbelluzzo Thank you for your suggestion. We will work with QA to add these test sample and picture in future.

PR #1275, opened 2016-01-15
Platform: Delete timezone fd watcher before re-adding it

gabrielschulhof:
This prevents a handle leak on the main loop.


barbieri:
It's not a bug since it returns `false`, that is the same as doing `sol_fd_del()`. But it doesn't hurt to take it.

PR #1277, opened 2016-01-15
macros: Add likely/unlikely in sol-macros.h

ceolin:
Those macros were defined in sol-util.h that is not exported but there
was public headers using them. Also, it was some re-define of the same
functionality, this commit attempts to sanitize it.


ibriano:
Fine by me

ceolin:
in, thanks

PR #1278, opened 2016-01-15
packet: Add missing packets to type_as_string()

ceolin:


ceolin:
error is already there

PR #1279, opened 2016-01-18
http:client: Add composed metatype

ceolin:

### DIFF
--- a/src/samples/flow/http-composed/http-composed-client.fbp
+++ b/src/samples/flow/http-composed/http-composed-client.fbp
@@ -1,0 +1,47 @@
+# This file is part of the Soletta Project
+#
+# Copyright (C) 2015 Intel Corporation. All rights reserved.
+#
+# Redistribution and use in source and binary forms, with or without
+# modification, are permitted provided that the following conditions
+# are met:
+#
+#   * Redistributions of source code must retain the above copyright
+#     notice, this list of conditions and the following disclaimer.
+#   * Redistributions in binary form must reproduce the above copyright
+#     notice, this list of conditions and the following disclaimer in
+#     the documentation and/or other materials provided with the
+#     distribution.
+#   * Neither the name of Intel Corporation nor the names of its
+#     contributors may be used to endorse or promote products derived
+#     from this software without specific prior written permission.
+#
+# THIS SOFTWARE IS PROVIDED BY THE COPYRIGHT HOLDERS AND CONTRIBUTORS
+# "AS IS" AND ANY EXPRESS OR IMPLIED WARRANTIES, INCLUDING, BUT NOT
+# LIMITED TO, THE IMPLIED WARRANTIES OF MERCHANTABILITY AND FITNESS FOR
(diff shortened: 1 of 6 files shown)


src/samples/flow/http-composed/http-composed-client.fbp
```diff
@@ -0,0 +1,47 @@
...
+# OWNER OR CONTRIBUTORS BE LIABLE FOR ANY DIRECT, INDIRECT, INCIDENTAL,
+# SPECIAL, EXEMPLARY, OR CONSEQUENTIAL DAMAGES (INCLUDING, BUT NOT
+# LIMITED TO, PROCUREMENT OF SUBSTITUTE GOODS OR SERVICES; LOSS OF USE,
+# DATA, OR PROFITS; OR BUSINESS INTERRUPTION) HOWEVER CAUSED AND ON ANY
+# THEORY OF LIABILITY, WHETHER IN CONTRACT, STRICT LIABILITY, OR TORT
+# (INCLUDING NEGLIGENCE OR OTHERWISE) ARISING IN ANY WAY OUT OF THE USE
+# OF THIS SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
+
+DECLARE=int_pair:http-client-composed:KEY(string)|VALUE(int)
+DECLARE=int_splitter:composed-split:STRING_VALUE(string)|INT_VALUE(int)
```
barbieri: "int_pair" is confusing, even more for the http client. What about:
 * http-client-composed:  client-string-int OR client-key-value
 * composed-split: split-string-int OR split-key-value

src/samples/flow/http-composed/http-composed-client.fbp
```diff
@@ -0,0 +1,47 @@
...
+
+http_client(int_pair)
+
+
+_(constant/string:value="http://localhost:8080/") OUT -> URL http_client
+_(constant/string:value="string_value") OUT -> KEY http_client
+_(constant/int:value=10) OUT -> VALUE http_client
+
+get(keyboard/boolean:binary_code=97,toggle=true) OUT -> GET http_client
+post(keyboard/boolean:binary_code=98,toggle=true) OUT -> POST http_client
```
barbieri: this is also not obvious, what about writing README or simply a brief set of commentaries in this file about what is this sample showing, how to use it.

Right now if you just run it, it will fail since there are no servers listening on localhost at 8080... also people may just be clueless that to execute something you need to reach key 97 and 98 of the keyboard... I can't tell those either :-D

It would help if we could ship with a server (possibly in FBP or Python) to interact with, maybe take the URL from command line, maybe segment the example in 2: one for GET, one for POST.

@mbelluzzo consider these comments while reviewing the other samples.

ceolin: @barbieri that's true, sorry.

I've been testing it using a modified version of the server sample here and forgot of changing this sample, i'll change it.

ceolin:
https://github.com/solettaproject/soletta/pull/1305

PR #1280, opened 2016-01-18
log: Fix invalid expression

ceolin:
It would fail if in SOL_INT_CHECK or SOL_EXP_CHECK the expression
contains '%' char.

### DIFF
--- a/src/lib/common/include/sol-log.h
+++ b/src/lib/common/include/sol-log.h
@@ -93,7 +93,7 @@ extern "C" {
 #define SOL_NULL_CHECK(ptr, ...) \
     do { \
         if (SOL_UNLIKELY(!(ptr))) { \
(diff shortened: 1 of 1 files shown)


src/lib/common/include/sol-log.h
```diff
@@ -111,7 +111,7 @@ extern "C" {
 #define SOL_NULL_CHECK_GOTO(ptr, label) \
     do { \
         if (SOL_UNLIKELY(!(ptr))) { \
-            SOL_WRN("" # ptr "== NULL"); \
+            SOL_WRN("%s == NULL", # ptr); \
```
barbieri: good, but you forgot the non-goto version `SOL_NULL_CHECK()` is still subject to that.

PR #1282, opened 2016-01-18
string-format: Initializing the log domain

ceolin:


barbieri:
+0

you can merge as is, but this macro will declare a static boolean where it is used. It would make more sense to have a single function that calls it, and use that function where he macro is used.

no big deal

ceolin:
i know that, usually this macro is used in the node's entry point (open()), but in this case both functions are used for different nodes

Issue #1269, opened 2016-01-13
review all functions taking `void *data` to use the `const` specifier.


barbieri:
In many places such as `sol-coap.h` we have functions that take `void *data`, without the `const` specifier, this will require people to add casts if they are giving a `const` object, which is wrong as they may not modify it in their callbacks.

In other cases such as in `sol-vector.h` we are also missing the specifier. In this case we're not talking about callbacks, but nonetheless people may be storing stuff they are not going to modify later, so there is no reason to require them to cast in order to store the pointers.

ceolin:
@barbieri please take a look if it is that you had in mind 
https://github.com/solettaproject/soletta/pull/1283

barbieri:
looks good

PR #1283, opened 2016-01-19
Normalizing user data pointer

ceolin:
https://github.com/solettaproject/soletta/issues/1269


cabelitos:
If I remenber right, the data should be the first parameter of the callbacks as well
@barbieri ^

ceolin:
I think so, but i prefer do not mix things here, also the issue itself it's not telling anything about the order.

cabelitos:
Well, I'm saying that, because he warned to me about this issue.
Anyway +1

ceolin:
i know, i just think they are different/separated fixes

barbieri:
+1

PR #1284, opened 2016-01-19
support another 6 category sensors into soletta iio node

guchaojie:
This PR supports Magnet , Pressure, Color , Huminity, Accelerate and ADC Category into iio node. And all the sensor category have been tested to work well in minnowmax using different sensors supported by their iio drivers.

### DIFF
--- a/src/modules/flow/iio/iio.json
+++ b/src/modules/flow/iio/iio.json
@@ -97,6 +97,95 @@
     },
     {
       "category": "input/hw",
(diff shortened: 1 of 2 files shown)


guchaojie:
@edersondisouza  This PR remove light patch this time but will submit it later when library fixed issue as we discussed.

edersondisouza:
Integrated, thanks!

edersondisouza:
@guchaojie Are you planning to change the API? Maybe, a better approach would be to check if there is the '_input' version of the channel, and fallback to '_raw' in case the first is not found. If using the processed one ('_input'), do not apply scale or offset. What do you think? This way, no changes on API actually, just internally on lib.

guchaojie:
@edersondisouza Actually I think this way is good to keep sensor category source code to be unchanged. One thing what I think different from you is to check if there is "_raw" version of channel firstly , then to check "_input" version in case the "_raw" not found. Because "_raw" version is more often to seen in most driver, this checking sequence would improve library efficiency.

PR #1285, opened 2016-01-19
Several fixes found using other toolchains

ibriano:
It seems that depending on the toolchain used and how it was
configured, there are several errors and warnings we may encounter.
'Tis an attempt to get those fixed.


ibriano:
Merged

PR #1287, opened 2016-01-20
sol-network: Add getaddrinfo() like function.

cabelitos:


cabelitos:
Check #1290

PR #1289, opened 2016-01-20
oic: Add oic sample to test iotivity compatibility - v2

otaviobp:
Add an oic sample similar to sample available at iotivity repository, so
we can test soletta and iotivitity compatibility.

Changelog from v1:
 - Just rebasing it on top of a recent master

Closes #1249


glima:
Other than those, the C logic looks OK and the OIC, from someone not very familiar with the details, looks at least sane.

otaviobp:
merged with proposed changes

PR #1290, opened 2016-01-20
sol-network: Add getaddrinfo() like function. v2

cabelitos:
Changes since v1:
* The function now returns a pointer that is used as a handle to cancel the operation.
* Added a commit fixing the code indent, so the semaphore stops to complain.
* Do not parse the hostname trying to find port and protocol.


otaviobp:
lgtm

cabelitos:
check #1291

PR #1296, opened 2016-01-21
Util export functions used on samples

ceolin:
- Export sol-util.h
 - Fix samples to not use internal headers
 - Add doc


barbieri:
-1 export it all :-)

from the remaining, now `sol-util-internal.h` just couple of macros are missing prefix, so just add to those. No need to always use `sol_util_`, such as in `sol_streq()` is enough, no need to bloat to `sol_util_streq()`.

glima:
You don't seem to install the header too

glima:
Now I see it, disregard

ceolin:
https://github.com/solettaproject/soletta/pull/1298

PR #1298, opened 2016-01-21
Util v2

ceolin:
from v1:
 - Export more functions from sol-util.h
 - Export sol-util-file.h
 - Fix sol-util-file dependency


mbelluzzo:
You will have to add 'src/shared/include' in doc/doxygen/Doxyfile 'INPUT' list so docs are generated for this new header files. This will trigger some <minor> doxygen errors.

ceolin:
fixes on https://github.com/solettaproject/soletta/pull/1304

PR #1299, opened 2016-01-21
Fix leftover of "Normalizing user data pointer" commit.

glima:


glima:
Thanks, merging.

PR #1300, opened 2016-01-22
log: Add missing header

ceolin:
check-cpp is failing without this.


cabelitos:
+1, however the semaphore complains about code style :/

PR #1301, opened 2016-01-22
LWM2M: Add LWM2M server and client. V3

cabelitos:
Changes since v2:
* Fixed problems pointed by @glima


bdilly:
I'm missing samples here.
Samples are good even to validate our API.
"Is it easy to use? Is some task cumbersome? How can I change that?"

Also it can be used by newcomers as a template for their projects.

So I would suggest to add simple samples here. You can find some on src/samples/coap and src/samples/mqtt

Sure, adding a test was nice, but it doesn't serve these purposes.

cabelitos:
@bdilly Can I add the samples in another pull req?

cabelitos:
#1431

PR #1305, opened 2016-01-22
Http composed v2

ceolin:
from v1:
 - fix possible leak of `url` on open
 - removed sample (so far, there is no way to test it using ours node servers, the sample will come with the composed server)
 - use `sol_util_replace_str_if_changed`
 - support options


cabelitos:
looks good.

ceolin:
merged, thanks for review it.

PR #1307, opened 2016-01-22
Sol-platform: Register a new inotify if /etc/locale.conf was deleted.

cabelitos:
When one changes the locale using systemd's localectl, the systemd
will not simple open the /etc/locale.conf file and update its contents.
It will create a new file file and then replace the /etc/locale.conf.
(basically mv /etc/ANeWLocaleFile.conf /etc/locale.conf)
Since the old locale.conf was deleted, a new inode watcher must
be created.


cabelitos:
check #1313

PR #1311, opened 2016-01-25
flow: Add STTS751 temperature sensor node type

edersondisouza:


edersondisouza:
@glima All others temperature sensors live in their own namespace, I've just followed that.

edersondisouza:
Integrated with fixes. Tks!

PR #1313, opened 2016-01-25
Sol-platform: Register a new inotify if /etc/locale.conf was deleted. v2

cabelitos:
When one changes the locale using systemd's localectl, the systemd
will not simple open the /etc/locale.conf file and update its contents.
It will create a new file file and then replace the /etc/locale.conf.
(basically mv /etc/ANeWLocaleFile.conf /etc/locale.conf)
Since the old locale.conf was deleted, a new inode watcher must
be created.

### DIFF
--- a/src/lib/common/include/sol-platform.h
+++ b/src/lib/common/include/sol-platform.h
@@ -484,6 +484,9 @@ const char *sol_platform_get_locale(enum sol_platform_locale_category category);
  * @param data The data to @c cb
  *
  * @return 0 on success, negative errno otherwise.
+ *
+ * @note If an error happens while the locale is being monitured the @c cb
+ * will be called and @c category will be set to #SOL_PLATFORM_SERVICE_STATE_UNKNOWN and @c locale to @c NULL.
  */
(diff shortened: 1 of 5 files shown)


src/lib/common/include/sol-platform.h
```diff
@@ -484,6 +484,9 @@ const char *sol_platform_get_locale(enum sol_platform_locale_category category);
  * @param data The data to @c cb
  *
  * @return 0 on success, negative errno otherwise.
+ *
+ * @note If an error happens while the locale is being monitured the @c cb
```
barbieri: typo "monitured"

barbieri:
looks okay, we should keep an eye and eventually add `sol_inotify` to wrap these things and handle this case if it shows in another place (like timezone, etc).

cabelitos:
@barbieri Thanks for the review. I will fix the typo before merging it.
I agree with you, if we start having to use iinotify in more parts of the code I think we should create an inotify wrapper.

cabelitos:
merged

PR #1316, opened 2016-01-26
Remove -e from echo

thiagomacieira:
POSIX echo has no options. Otherwise, we get output like:
    
          -e The following (required) dependencies were not met:
    
          -e python module: jsonschema
    
          -e If you've just installed it, run: make reconf
          -e For more information/options, run: make help


barbieri:
+1 merge it

PR #1317, opened 2016-01-26
Add missing #pragma GCC diagnostic push

thiagomacieira:
You can't pop without pushing first. Clang complains:
    
    warning: pragma diagnostic pop could not pop, no matching push [-Wunknown-pragmas]


barbieri:
+1 merge it

PR #1321, opened 2016-01-26
iio: More detail on some error messages

edersondisouza:
Also one more debug message. Should help finding IIO bugs.

### DIFF
--- a/src/lib/io/sol-iio.c
+++ b/src/lib/io/sol-iio.c
@@ -157,6 +157,8 @@ craft_filename_path(char *path, size_t size, const char *base, ...)
     len = vsnprintf(path, size, base, args);
     va_end(args);
 
+    SOL_DBG("Crafted path len: %d Buffer size: %zu", len, size);
+
     return len >= 0 && (size_t)len < size;
 }
 
@@ -352,14 +354,15 @@ static void
 set_buffer_size(struct sol_iio_device *device, int buffer_size)
 {
     char path[PATH_MAX];
+    int r;
 
     if (craft_filename_path(path, sizeof(path), BUFFER_LENGHT_DEVICE_PATH, device->device_id)) {
         SOL_WRN("Could not set IIO device buffer size");
         return;
     }
 
-    if (sol_util_write_file(path, "%d", buffer_size) < 0) {
-        SOL_WRN("Could not set IIO device buffer size");
+    if ((r = sol_util_write_file(path, "%d", buffer_size)) < 0) {
+        SOL_WRN("Could not set IIO device buffer size: %s", sol_util_strerrora(r));
     }
 
     return;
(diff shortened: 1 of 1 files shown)


src/lib/io/sol-iio.c
```diff
@@ -157,6 +157,8 @@ craft_filename_path(char *path, size_t size, const char *base, ...)
     len = vsnprintf(path, size, base, args);
     va_end(args);
 
+    SOL_DBG("Crafted path len: %d Buffer size: %zu", len, size);
```
barbieri: the important information should be the path, which is not printed. You can shrink it to: "available=%zu, used=%d, path=\"%s\""

You can omit "crafted" since it's going to show the function name.

src/lib/io/sol-iio.c
```diff
@@ -352,14 +354,15 @@ static void
...
 
     if (craft_filename_path(path, sizeof(path), BUFFER_LENGHT_DEVICE_PATH, device->device_id)) {
         SOL_WRN("Could not set IIO device buffer size");
         return;
     }
 
-    if (sol_util_write_file(path, "%d", buffer_size) < 0) {
-        SOL_WRN("Could not set IIO device buffer size");
+    if ((r = sol_util_write_file(path, "%d", buffer_size)) < 0) {
+        SOL_WRN("Could not set IIO device buffer size: %s", sol_util_strerrora(r));
```
barbieri: and path!

src/lib/io/sol-iio.c
edersondisouza: On 26-01-2016 16:45, Gustavo Sverzut Barbieri wrote:
> In src/lib/io/sol-iio.c
> <https://github.com/solettaproject/soletta/pull/1321#discussion_r50879574>:
>
>> @@ -157,6 +157,8 @@ craft_filename_path(char *path, size_t size, const char *base, ...)
>>      len = vsnprintf(path, size, base, args);
>>      va_end(args);
>>
>> +    SOL_DBG("Crafted path len: %d Buffer size: %zu", len, size);
>
> the important information should be the path, which is not printed. You
> can shrink it to: "available=%zu, used=%d, path=\"%s\""

Well, if `vsnprintf` failed, path is undefined, no? It's better not try 
to print it... I was willing to avoid ifs and branching just for logging...

>
> You can omit "crafted" since it's going to show the function name.
>
Fair.

barbieri: fair enough for the failure, BUT  on success you're not printing it :-) That's not fair.

You can start path[] = "", so if it's failed either snprintf will null-terminate or it will be already null-terminated.

edersondisouza: @barbieri  Do you mean `path[0] = '\0'`? If `vsnprintf` fails, I can be assured that this first char wasn't changed?

barbieri: I was looking at path[size] = {}; (init all to null)

If you want to be sure you can `if (r < 0) path[0] = '\0';`

edersondisouza:
Integrated with proposed fixes.

PR #1326, opened 2016-01-26
sol-coap: Properly handle observe option.

cabelitos:
According to the spec the observe option may have its length
equals to 0.

@vcgomes


otaviobp:
+1. Works fine with oic.

cabelitos:
merged

PR #1332, opened 2016-01-27
Fix use of non-POSIX errno name ENOTUNIQ

### LINKED ISSUE #1329: Use of undeclared identifier 'ENOTUNIQ'
```
./src//lib/flow/sol-flow-builder.c:484:21: error: use of undeclared identifier 'ENOTUNIQ'
            return -ENOTUNIQ;
                    ^
```

In files that aren't Linux-specific, please choose errno names *only* if they are defined in POSIX.

http://pubs.opengroup.org/onlinepubs/9699919799/basedefs/errno.h.html

thiagomacieira:
It doesn't exist in POSIX, so this can't be used outside of Linux
systems. This fixes #1329.


ceolin:
+2 merged

PR #1333, opened 2016-01-27
Fix truncation of 64-bit value during abs() call

thiagomacieira:
The abs() function takes int, while llabs() takes long long. There's a
truncation when assigned to out.{red,green,blue}, but at least that
doesn't create a warning with Clang.

warning: absolute value function 'abs' given an argument of type 'int64_t' (aka 'long long') but has parameter of type 'int' which may cause truncation of value [-Wabsolute-value]
note: use function 'llabs' instead


thiagomacieira:
ping

barbieri:
+1

PR #1335, opened 2016-01-27
Fix build: led-7seg.c requires sol-gpio.h

thiagomacieira:
This is definitely wrong, but since I have no clue how to do it properly, I am sending the temporary fix. I'd appreciate instruction on how to set the include path. I can't find where that gets set.

### DIFF
--- a/src/modules/flow/led-7seg/led-7seg.c
+++ b/src/modules/flow/led-7seg/led-7seg.c
@@ -33,7 +33,7 @@
 #include "sol-flow/led-7seg.h"
 #include "sol-flow-internal.h"
 
-#include <sol-gpio.h>
+#include "../../../lib/io/include/sol-gpio.h"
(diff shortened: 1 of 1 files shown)


barbieri:
this is weird, since it should being installed to `build/soletta_sysroot/usr/include/soletta/` that should be provided to all builds.

I see two possibilities here:
 - GPIO is being disabled and nevertherless led-7seg is still being built. It may be the case since `src/modules/flow/led-7seg/Kconfig` seems to miss `depends on USE_GPIO`. **This is more likely**
 - sol-gpio.h not being in build/soletta_sysroot/usr/include/soletta/, while it should

glima:
Yep, the Kconfig patch should be the right fix.

bdilly:
ok, I've sent a fix on #1415 
please let us know if it fixes this issue so we can close this pr.

PR #1336, opened 2016-01-27
Fix build: detect memrchr and strdupa

thiagomacieira:
These two functions are GNU libc extensions.

### DIFF
--- a/data/jsons/dependencies.json
+++ b/data/jsons/dependencies.json
@@ -268,6 +268,15 @@
       "fragment": "strndupa(0, 0);"
     },
     {
+      "dependency": "decl_strdupa",
+      "type": "ccode",
+      "comment": "Detect GNU extension strdupa()",
+      "headers": [
+        "<string.h>"
+      ],
+      "fragment": "strdupa((void *)0);"
+    },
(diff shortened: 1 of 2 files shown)


thiagomacieira:
Cannot reproduce the test failure.

barbieri:
the failure is stupid uncrustify (check-style). We cannot get it to consider only the fragments that changed and it's checking code that was already there... so ignore and merge anyway.

barbieri:
actually one small part of the style check is related to your change, better fix it: https://semaphoreci.com/solettaproject/soletta/branches/pull-request-1336/builds/1

```
-    for ( ; ptr != haystack; --ptr) {
+
+    for (; ptr != haystack; --ptr) {
```

That is, should be no space between `(` and `;`. And the coding style requires one empty line between variable declaration and the code.

thiagomacieira:
I would have thought that having the ; isolated would make it more noticeable, but sure.

thiagomacieira:
False positive on CI.

thiagomacieira:
ping

barbieri:
+1

PR #1337, opened 2016-01-27
Fix warning when XSI-compliant strerror_r is provided

thiagomacieira:
On 64-bit systems, casting from int to pointers may cause warnings.

error: cast to 'char *' from smaller integer type 'int' [-Werror,-Wint-to-pointer-cast]

### DIFF
--- a/src/shared/sol-util.c
+++ b/src/shared/sol-util.c
@@ -222,7 +222,7 @@ sol_util_strerror(int errnum, char *buf, size_t buflen)
 
     buf[0] = '\0';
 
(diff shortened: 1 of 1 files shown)


cabelitos:
Hello @thiagomacieira, I was wondering should we use the macro INT_TO_PTR() ?
(see sol-util-internal.h)

thiagomacieira:
It's equivalent. I don't see the need for the macro, though.

thiagomacieira:
Ping. Please advise.

barbieri:
no need for macro, +1

PR #1338, opened 2016-01-27
Fix Clang warnings about printf format mismatches

thiagomacieira:
All of the API version numbers are declared as plain integer literals,
so using %hu is wrong. They are int, not unsigned short.

The booleans are not to be printed as %hhu.


thiagomacieira:
Please explain the failure.

barbieri:
why boolean not to be print with `%hhu`?

If we can't find a way to print boolean, then at least provide `SOL_BOOL_STR(x) ((x) ? "true" : "false")`, but I'd like to keep them as numbers to avoid cluttering file size.

thiagomacieira:
Because bool is not uint8_t, therefore %hhu is incorrect. I can introduce a cast, instead.

thiagomacieira:
I'm postponing this until I convert more of the Linux→Unix checks. There may be more of these.

PR #1339, opened 2016-01-27
Fix build: include sol-board-detect.h with the right #ifdef

thiagomacieira:
The code that uses sol_board_detect is behind #ifdef DETECT_BOARD_NAME.


barbieri:
+1

PR #1340, opened 2016-01-27
Fix build: static_assert is C11

thiagomacieira:
So make sure __STDC_VERSION__ has the correct value.


barbieri:
maybe `#else assert` and fallback to old `assert()`?

thiagomacieira:
Is it worth it? Someone will catch this in another platform...

barbieri:
fair enough

thiagomacieira:
+1?

barbieri:
sure, +1

PR #1342, opened 2016-01-27
oic: Macro SOL_OIC_MAP_LOOP was never finishing with success

otaviobp:
Commit b2d2c2b9 added parameter checking, but added this bug in
SOL_OIC_MAP_LOOP macro.

In case the cbor value is not valid, we reached the end of the list, so
reason shouldn't be updated.


otaviobp:
merged.

PR #1343, opened 2016-01-27
builder: Use posix error code

ceolin:
Replacing ENOTUNIQ with EEXIST. Since the previous statement is
returning EEXIST if there already is the node the same can be applied
for duplicated name.


ceolin:
duplicated, https://github.com/solettaproject/soletta/pull/1332

Issue #1327, opened 2016-01-26
check_cflags iterates over each letter, not each option of the .json file


thiagomacieira:
Take this test:

```json
    {
      "dependency": "check_cflags",
      "type": "cflags",
      "cflags": [
        "-Wno-cpp",
       "-Wno-#warnings"
      ],
      "comment": "append to CFLAGS_NO_SOURCE_WARNINGS",
      "append_to": "CFLAGS_NO_SOURCE_WARNINGS"
    },
```

config.log shows:

```
Command: gcc -Werror -Wno-cpp -Wno-#warnings /var/folders/_q/byf907ss6bb130qvj9fqk2nr0000gq/T/tmp4w3xl9af.c -o /var/folders/_q/byf907ss6bb130qvj9fqk2nr0000gq/T/tmp4w3xl9af.c-bin 
exit code: 1
Command output(stderr):
error: unknown warning option '-Wno-cpp' [-Werror,-Wunknown-warning-option]

Compiling source code:

int main(int argc, char **argv){
  return 0;
}
Command: gcc -Werror  - /var/folders/_q/byf907ss6bb130qvj9fqk2nr0000gq/T/tmpw9fqd2ie.c -o /var/folders/_q/byf907ss6bb130qvj9fqk2nr0000gq/T/tmpw9fqd2ie.c-bin 
exit code: 1
Command output(stderr):
clang: error: -E or -x required when input is from standard input

Compiling source code:

int main(int argc, char **argv){
  return 0;
}
Command: gcc -Werror  W /var/folders/_q/byf907ss6bb130qvj9fqk2nr0000gq/T/tmpboayr9fu.c -o /var/folders/_q/byf907ss6bb130qvj9fqk2nr0000gq/T/tmpboayr9fu.c-bin 
exit code: 1
Command output(stderr):
clang: error: no such file or directory: 'W'

Compiling source code:

int main(int argc, char **argv){
  return 0;
}
Command: gcc -Werror  n /var/folders/_q/byf907ss6bb130qvj9fqk2nr0000gq/T/tmpcqptonzx.c -o /var/folders/_q/byf907ss6bb130qvj9fqk2nr0000gq/T/tmpcqptonzx.c-bin 
exit code: 1
Command output(stderr):
clang: error: no such file or directory: 'n'

Compiling source code:

int main(int argc, char **argv){
  return 0;
}
[...]

PR #1344, opened 2016-01-27
fix dependency-resolver flags check.

barbieri:
This simplifies the code by using Python more efficiently and fix bugs
when the first-try fails.

Always convert cflags and ldflags to lists, then we can easily merge
them with "+" operator, including -Werror prepending.

Add a wrapper to compile_test(), so our code is shorter.

Consider cflags and ldflags independently, callers will take care to
only call with one or another, so the append to variable will be
meaningful, but it should be harmless if both CFLAGS and LDFLAGS are
provided.

Last but not least fix the bug to handle unsupported flags. Before it
was iterating "flags", a string. This was producing char-by-char,
which makes no sense.

More than that, if we try flag by flag with an EMPTY body, -Wextra
will be avoided since it would trigger "argv" and "argc" warnings of
being unused, those are converted to error with -Werror we
prepend. Then we generate warning-free code by casting argc/argv to
void.

Should address #1327

### DIFF
--- a/data/scripts/dependency-resolver.py
+++ b/data/scripts/dependency-resolver.py
@@ -352,44 +352,47 @@ def handle_filesystem_check(args, conf, context):
 
 def handle_flags_check(args, conf, context, cflags, ldflags):
     append_to = conf.get("append_to")
-    source = cstub.format(headers="", fragment="")
-    test_cflags = ""
-    test_ldflags = ""
+    source = cstub.format(headers="", fragment="(void)argc; (void)argv;")
 
-    if cflags:
-        test_cflags = " ".join(cflags)
-        flags = test_cflags
-    elif ldflags:
-        test_ldflags = " ".join(ldflags)
-        flags = test_ldflags
-    else:
+    if not cflags and not ldflags:
         context.info("Neither cflags nor ldflags provided to flags_check.")
         exit(1)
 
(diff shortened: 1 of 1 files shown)


thiagomacieira:
Oops, wrong window.

Anyway, this fixes the problem for me:

```
CFLAGS_NO_SOURCE_WARNINGS += -Wno-#warnings
COMMON_CFLAGS += -pipe -Wall -W -Wextra -Wundef -Wformat=2 -Wsign-compare -Wformat-security -Wmissing-include-dirs -Wformat-nonliteral -Wold-style-definition -Wpointer-arith -Winit-self -Wdeclaration-after-statement -Wfloat-equal -Wmissing-prototypes -Wstrict-prototypes -Wredundant-decls -Wmissing-declarations -Wmissing-noreturn -Wno-unused-parameter -Wshadow -Wendif-labels -Wwrite-strings -Wno-long-long -Wno-missing-field-initializers -Wnested-externs -Wchar-subscripts -Wtype-limits -Wuninitialized -fno-common -fdiagnostics-show-option -fvisibility=hidden -ffunction-sections -fdata-sections -fno-asynchronous-unwind-tables
```

I don't understand Python, so I can't give you +1.

barbieri:
merged.

PR #1345, opened 2016-01-27
Remove extra slash in find

thiagomacieira:
This extra slash caused make rules with "src//lib", which obviously
don't match "src/lib" (make does string comparisons).

 make: *** No rule to make target `src//modules/flow/oic/oic.json', needed by `build/soletta_sysroot/usr/include/soletta/sol-flow/oic.h'.  Stop.


barbieri:
wait CI, but looks good

dorileo:
I'm running the same tests CI runs on my computer, but yeah, LGTM

dorileo:
yep, the tests passed here, go for it.

PR #1346, opened 2016-01-27
Check for the validity of the -soname flag before using

thiagomacieira:


barbieri:
+1

PR #1347, opened 2016-01-27
Remove fallback accept4 because it doesn't work

thiagomacieira:
This was obviously never tested because this function does not properly
set the flags it purported to set. F_SETFD only recognises one flag,
FD_CLOEXEC, which will never match the values of SOCK_CLOEXEC and
SOCK_NONBLOCK that are used throughout Soletta. Using F_SETFL would have
helped with SOCK_NONBLOCK / O_NONBLOCK, but we'd need to handle
SOCK_CLOEXEC separately anyway.

Anyway, there were only two uses of accept4 in Soletta and one of them
is in a Linux-specific file. So instead just fix the other file
(portable code) with the direct fallback.


barbieri:
+1

Issue #1331, opened 2016-01-26
Platform pin multiplexer is failing to load when sol_init() is called from a module


nagineni:
Platform pin multiplexer is failing to load on target device, when sol_* APIs are called from a module(e.g nodejs module) that was loaded with dlopen by an executable(e.g node) using RTLD_LOCAL flag.

Here is the log:
`INF:sol-pin-mux ./src/lib/common/sol-pin-mux.c:82 _load_mux() Could not load platform pin multiplexer '//usr/lib/soletta/modules//pin-mux//intel-edison-rev-c.so': //usr/lib/soletta/modules//pin-mux//intel-edison-rev-c.so: undefined symbol: sol_log_global_domain`

Note that the issue is not reproducible when we open an executable that linked with libsoletta directly. I already checked this by compiling a simple example that uses the same sol C API, and running it on Edison.

bdilly:
@mbelluzzo could you please take a look on this issue? thanks

mbelluzzo:
@nagineni, just to help me to find out what is goind on: what happens if you disable the pinmux during configuration?

barbieri:
This fails for sure, see the C snippet below that can be tested as:

```sh
SOL_LOG_LEVEL=4 SOL_BOARD_NAME=intel-edison-rev-c  /tmp/t libsoletta.so
```

Running without `SOL_BOARD_NAME` in my desktop it works, while using edison board name it will fail with the exact message.

```c
#include <dlfcn.h>
#include <stdio.h>
#include <stdlib.h>

int main(int argc, char *argv[])
{
    void *h;
    int (*f_init)(void);
    void (*f_shutdown)(void);
    int retcode = EXIT_FAILURE;
    int r;

    h = dlopen(argv[1], RTLD_LOCAL | RTLD_NOW);
    if (!h) {
        fprintf(stderr, "error: %s: %s\n", argv[1], dlerror());
        return 1;
    }

    f_init = dlsym(h, "sol_init");
    if (!f_init) {
        fprintf(stderr, "error: no sol_init() in %s: %s\n", argv[1], dlerror());
        goto end;
    }

    f_shutdown = dlsym(h, "sol_shutdown");
    if (!f_shutdown) {
        fprintf(stderr, "error: no sol_shutdown() in %s: %s\n", argv[1], dlerror());
        goto end;
    }

    r = f_init();
    if (r < 0) {
        fprintf(stderr, "sol_init() = %d\n", r);
        goto end;
    }

[...]

PR #1348, opened 2016-01-27
Modules fix

barbieri:
Fix module's linkage, it was missing '-lsoletta'. @dorileo please check.

A second commit is to fix an issue with log level, not really required as part of this PR, but it was the reason we found the bug. see #1331


mbelluzzo:
+1

nagineni:
This fixes the issue #1331. 

:+1:

barbieri:
merged.

PR #1352, opened 2016-01-27
Rename variables called "timezone"

thiagomacieira:
Since we build with -Wshadow, we can't use the name of the global
variable "timezone".


barbieri:
+1

PR #1355, opened 2016-01-27
Add pipe2 replacement implementation

thiagomacieira:
This depends on O_CLOEXEC being defined, which is part of POSIX.1-2008.
So there's a good chance that the target platforms will have it.

When we run across a platform that is POSIX.1-2001 or earlier, we'll
need to add a suitable #define somewhere.

### DIFF
--- a/src/lib/common/sol-missing.h
+++ b/src/lib/common/sol-missing.h
@@ -70,6 +70,54 @@ memmem(const void *haystack, size_t haystacklen, const void *needle, size_t need
 }
 #endif
 
+#if !defined(HAVE_PIPE2) && defined(HAVE_PIPE)
+#include <fcntl.h>
+#include <unistd.h>
+#include <sol-util-file.h>
+#ifndef O_CLOEXEC
+#error "I need O_CLOEXEC to work!"
+#endif
+
+static inline int
+pipe2(int pipefd[2], int flags)
+{
+    int ret = pipe(pipefd);
+
+    if (ret < 0)
+        return ret;
+
+    if (flags & O_NONBLOCK) {
+        ret = sol_util_fd_set_flag(pipefd[0], O_NONBLOCK);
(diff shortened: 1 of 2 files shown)


src/lib/common/sol-missing.h
```diff
@@ -70,6 +70,50 @@ memmem(const void *haystack, size_t haystacklen, const void *needle, size_t need
...
+    if (flags & O_NONBLOCK) {
+        ret = sol_util_fd_set_flag(pipefd[0], O_NONBLOCK);
+        if (ret >= 0)
+            ret = sol_util_fd_set_flag(pipefd[1], O_NONBLOCK);
+        if (ret < 0)
+            goto err;
+    }
+
+    if (flags & O_CLOEXEC) {
+        ret = fcntl(pipefd[0], F_GETFL);
```
barbieri: why `fcntl()` here and `sol_util_fd_set_flag()` above? also, can't we just set them both at once?

thiagomacieira: Oops, it was supposed to be F_SETFD, not F_SETFL.

thiagomacieira:
Failure is false positive.

ceolin:
:+1:

thiagomacieira:
Can someone check RIoT and Contiiki?

barbieri:
hum... riot, contiki and the likes are likely to fail since they miss `pipe2()` AND `pipe()`. You need to guard against it, not sure what's the best way, just `&& (!defined(CONTIKI) && !defined(RIOT)...` or to check for `pipe()` and only define if that is available or guard for `POSIX`.

ceolin:
i've taken a look in the riot and the pipe api is completely different

ceolin:
the patch looks good to me, +1

cabelitos:
+1

PR #1356, opened 2016-01-27
Fix Clang warning -Wmissing-noreturn

thiagomacieira:
This function doesn't return, so add the attribute.


barbieri:
+1

PR #1357, opened 2016-01-28
Remove use of non-standard errno name ENOKEY from portable code

thiagomacieira:


barbieri:
maybe -ENOENT?

thiagomacieira:
See file comments.

barbieri:
got it, then +1. commit

PR #1360, opened 2016-01-28
POSIX struct siginfo does not have si_stime or si_utime

thiagomacieira:


barbieri:
+1

PR #1361, opened 2016-01-28
test-fbp: make expected output less strict

cmarcelo:
Change the regex to look for the content that we explicitly print (the
division by zero), and don't be strict about the other content: the
exact filename and function name that leads to the warning.

These tests currently fail when compiled with Clang due to the different
output it uses for function name when SOL_WRN and friends are used.


cmarcelo:
@bdilly could you take a look?

bdilly:
that's fine.
thanks, @cmarcelo 
+1

cmarcelo:
Applied in 8391983dd355e7f18db2e1646fc7ad51c7c8b3b5.

PR #1362, opened 2016-01-28
soletta.pc: use paths relative to variables.

barbieri:
This allows us to use
build/soletta_sysroot/$PREFIX/lib/pkgconfig/soletta.pc while it's not
installed by calling:

```sh
   export BUILT_PREFIX=$PWD/build/soletta_sysroot/$PREFIX
   export PKG_CONFIG_PATH=$BUILT_PREFIX/lib/pkgconfig:$PKG_CONFIG_PATH
   pkg-config --cflags --libs --define-variable=prefix=$BUILT_PREFIX soletta
```

This may be required to simulate out-of-tree builds without installing
it.

### DIFF
--- a/pc/soletta.pc.in
+++ b/pc/soletta.pc.in
@@ -1,11 +1,11 @@
 prefix={{@PREFIX@}}
-exec_prefix={{@PREFIX@}}
-libdir={{@LIBDIR@}}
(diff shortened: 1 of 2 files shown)


barbieri:
forgot about this one... merged.

PR #1363, opened 2016-01-28
Fix build: get_current_dir_name() is a GNU extension

thiagomacieira:
So add a replacemenet function.

### DIFF
--- a/src/lib/common/sol-missing.h
+++ b/src/lib/common/sol-missing.h
@@ -140,6 +140,40 @@ err:
 }
 #endif
 
+#if !defined(HAVE_GET_CURRENT_DIR_NAME) && defined(HAVE_UNIX)
+#include <errno.h>
+#include <limits.h>         /* for PATH_MAX */
+#include <stdlib.h>
+#include <unistd.h>
+
+static inline char *
+get_current_dir_name(void)
+{
+    size_t len = 0;
+    char *buffer = NULL;
+
+    while (1) {
+        len += PATH_MAX;
+        char *nb = (char *)realloc(buffer, len);
+
+        if (nb == NULL)
+            goto err_free;
+        nb = getcwd(buffer, len);
+        if (nb == NULL) {
+            if (errno == ERANGE)
+                continue;
+            goto err_free;
+        }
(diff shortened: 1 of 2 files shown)


src/lib/common/sol-missing.h
```diff
@@ -70,6 +70,39 @@ memmem(const void *haystack, size_t haystacklen, const void *needle, size_t need
...
+#include <unistd.h>
+
+static inline char *get_current_dir_name(void)
+{
+    size_t len = 0;
+    char *buffer = NULL;
+
+    while (1) {
+        len += PATH_MAX;
+        char *nb = (char *)realloc(buffer, len);
```
barbieri: consider using our `sol-buffer.h`, much easier and handles realloc checks for you. Ask it for the size you want.

src/lib/common/sol-missing.h
```diff
@@ -70,6 +70,39 @@ memmem(const void *haystack, size_t haystacklen, const void *needle, size_t need
 }
 #endif
 
+#if !defined(HAVE_GET_CURRENT_DIR_NAME) && defined(HAVE_UNIX)
```
thiagomacieira: Note this depends on HAVE_UNIX being defined, which is in another patch not yet ready for submission. But this won't break the build.

src/lib/common/sol-missing.h
thiagomacieira: NAK. This function needs to return a pointer freeable with free().

barbieri: you can keep the `struct sol_buffer` on the stack and use `sol_buffer_steal()` :-)

realloc is a PITA

thiagomacieira: Not after you've already done it correctly :-)

The problem with doing with sol-buffer.h is that I'd have to learn it first and I don't have time for that.

barbieri: Fair, go for it. +1

bdilly:
OK, I'm merging this patch so the next guy working on this task may provide HAVE_UNIX definition.

Thank you

PR #1364, opened 2016-01-28
Add support for pollts as a replacement for ppoll

thiagomacieira:


barbieri:
+1

PR #1365, opened 2016-01-28
sol-vector: Use the correct error code if sol-vector is full.

cabelitos:


barbieri:
-ED+

PR #1366, opened 2016-01-28
Symbols - Fix unresolved symbols

ceolin:


barbieri:
+0, but check the linker flags is okay with @thiagomacieira as he ports to MacOS and our semaphore still doesn't cover it.

Also, semaphore says `flow/http-client` also misses it

ceolin:
it was complaining with http-server that i've added. I've asked iscaro to test and also worked in his system. I'm checking what is wrong

ceolin:
i know that this is a linker flag, but adding it to `COMMON_LDFLAGS` simply did not work.
probably there is another problem. I'll check before add it and see why it's failing on semaphore

thiagomacieira:
If you pass an -l flag while -c is active, some compilers will complain about arguments not used because no linking was done. Please make sure that the -l flag is passed only to the linking step.

ceolin:
i'll, thanks !

ceolin:
https://github.com/solettaproject/soletta/pull/1372

PR #1369, opened 2016-01-29
Node.js bindings: Add infrastructure and start binding platform functions (v2)

gabrielschulhof:
This PR replaces #1233


gabrielschulhof:
Ummm, what is ```__asan_unregister_globals``` and why is soletta using it :)

gabrielschulhof:
https://www.google.fi/search?q=__asan_unregister_globals brings up some interesting pages ...

gabrielschulhof:
https://github.com/google/sanitizers/issues/380 is interesting. Perhaps this is issue arises from the fact that soletta.node is built with g++ whereas soletta.so is built with gcc.

ceolin:
It's libasan thing. You can enable/disable this address sanitizer in the menuconfig

basic -> compiler options -> compiler sanitize

It's useful to detect uninitialized reads and memory leaks

gabrielschulhof:
@ceolin so, does this mean it needs to be ```apt-get install```-ed on the semaphore?

gabrielschulhof:
@ceolin is libasan getting loaded dynamically? If so, it may be running into the same issue as https://github.com/solettaproject/soletta/issues/1331

ceolin:
i think it's dynamic. I am seeing what can be happening

gabrielschulhof:
Maybe we should have a test where we load soletta with ```RTLD_LOCAL | RTLD_NOW``` - which is what happens when soletta is loaded from another library, to make sure that chaining works. Of course, this still wouldn't catch issues like #1331 ... hmmm ...

gabrielschulhof:
What if we make use of libasan only when building in debug mode?

ibriano:
On Fri, 29 Jan 2016 08:29:28 -0800, gabrielschulhof wrote:
> What if we make use of libasan only when building in debug mode?
> 

That's already the case, IIRC.

ceolin:
i've made a pull request passing -z,defs to linker. This should check if there are undefined symbols.

gabrielschulhof:
OK, I'll wait for that to land, then I'll rebase onto it and hopefully this issue will no longer be present.

cabelitos:
@gabrielschulhof Could you please rebase with the master branch and try again?
Thank you

cabelitos:
@gabrielschulhof You need to add to depedency check if the user have npm installed. I tried to run your branch and the build broke because I did not have npm.

gabrielschulhof:
@cabelitos you have node but you don't have npm? I have a check for node, and node always comes with npm.

cabelitos:
+1

PR #1370, opened 2016-01-29
generator: Fix tool dependency

ceolin:
Since metatype was introduced the generator uses code from
sol-flow-parser that only build if descriptions are enabled.


ceolin:
I intend to review if it's really necessary the metatype code is in  sol-flow-parser and then review the description's dependency. But while it's not done i think it's better we not allow this broken state.

barbieri:
+1, but add a TODO in that Kconfig

ceolin:
merged with the suggestion

PR #1372, opened 2016-01-29
Symbols - Fix unresolved symbols [v2]

ceolin:
from v1
 - Just append the flag if it's available in the system
 -  -l option to linker


thiagomacieira:
LGTM

ceolin:
https://github.com/solettaproject/soletta/pull/1374

PR #1373, opened 2016-01-29
samples:linux-micro: Do not redefine _GNU_SOURCE

ceolin:


cabelitos:
What you think about gcc providing this define?
I mean gcc -D_GNU_SOURCE ...

ceolin:
in our build it's already done. But as it''s a sample and the idea is one copy it and to it's on project, i think it's better you let it defined.

cabelitos:
Ok.

PR #1374, opened 2016-01-29
Symbols v3

ceolin:
from v2:
 - don't set -Wl,-no-undefined if linking with libasan

According with libasan faq:

`Q: When I link my shared library with -fsanitize=address, it fails due to some undefined ASan symbols (e.g. asan_init_v4)?

A: Most probably you link with -Wl,-z,defs or -Wl,--no-undefined. These flags don't work with ASan.`


thiagomacieira:
Oops, should be ldflags/LDFLAGS, not CFLAGS

ceolin:
fixes on https://github.com/solettaproject/soletta/pull/1381

PR #1379, opened 2016-02-01
Fix pipe2 in case one fcntl fails

thiagomacieira:
I missed one of the if checks.


thiagomacieira:
@lpereira

lpereira:
LGTM

PR #1381, opened 2016-02-01
Symbols v4

ceolin:
from v3:
 - Option is set on LDFLAGS instead of being set on CFLAGS
 - New commit adding common_ldflags when building modules


thiagomacieira:
LGTM.

ceolin:
thanks for review

PR #1382, opened 2016-02-02
flow/iio: Use correct data on callback direct call

edersondisouza:
When iio node stopped storing node on private data, reading callbacks
started using node directly - so we need to change direct callback
calls, used when buffer is not enabled.


yongli3:
+ 1 
Fixed the SIGSEGV issue when using the "buffer_size=-1" option

edersondisouza:
Integrated, tks!

PR #1383, opened 2016-02-02
Build: Set -ffloat-store on non 64 bits architetures.

cabelitos:
Prevent excess precision on non 64 machines, this may lead to
wrong results.

### DIFF
--- a/data/jsons/dependencies.json
+++ b/data/jsons/dependencies.json
@@ -39,6 +39,14 @@
       }
     },
     {
+      "dependency": "check_cflag_float_store",
+      "type": "cflags",
+      "cflags": [
+        "-ffloat-store"
+      ],
+      "append_to": "CFLAG_FLOAT_STORE"
+    },
+    {
       "dependency": "check_cflags",
       "type": "cflags",
       "cflags": [
@@ -386,6 +394,11 @@
         "<stdlib.h>"
       ],
       "fragment": "strtod_l(0, 0, 0);"
+    },
+    {
+      "dependency": "getconf",
+      "type": "exec",
+      "exec": "getconf"
     },
     {
       "dependency": "valgrind",
--- a/tools/build/Makefile.vars
+++ b/tools/build/Makefile.vars
@@ -176,6 +176,12 @@ GDB_AUTOLOAD_PY := $(top_srcdir)data/gdb/libsoletta.so-gdb.py
 COMMON_CFLAGS += $(CFLAGS)
 COMMON_LDFLAGS += $(LDFLAGS)
 
+ifeq (y, $(HAVE_GETCONF))
+ifneq (64, $(shell getconf LONG_BIT))
+COMMON_CFLAGS += $(CFLAG_FLOAT_STORE)
+endif
+endif
+
 ifeq (y,$(CC_SANITIZE))
 ifeq (y,$(CC_SANITIZE_UNDEFINED))
 ifeq (y,$(SHARED_LIBRARY))


ibriano:
On Tue, 02 Feb 2016 09:41:04 -0800, Guilherme Íscaro wrote:
> Prevent excess precision on non 64 machines, this may lead to
> wrong results.
> 

A bit late on my part, but what wrong results? Why is this the right
fix?

edersondisouza:
On 02-02-2016 19:32, Iván Briano wrote:
> On Tue, 02 Feb 2016 09:41:04 -0800, Guilherme Íscaro wrote:
>  > Prevent excess precision on non 64 machines, this may lead to
>  > wrong results.
>  >
>
> A bit late on my part, but what wrong results? Why is this the right
> fix?

Some persistence tests that mapped drange values may fail on x86 
(32bit), due to GCC generating fpu instructions that use 80bit floating 
points (extended double precision). With -ffloat-store, doubles are kept 
using 64bits and then results are equal to that of x86_64, where 
floating point calc use sse instructions.
We had a similar fix to soletta galileo builds, afair. Maybe time to 
revisit it also? @brunobottazzini?

brunobottazzini:
@edersondisouza, Yes we had this but it is not upstreamed, AFAIR.
https://github.com/brunobottazzini/meta-soletta/commit/6a2a1cae049f3f0751bb6f913a83883fe802a242

Yes maybe it is time to review it.

bdilly:
@brunobottazzini why didn't you pushed it to master back on november?
I believe it won't be necessary anymore since now it's done on Soletta build, right?

ibriano:
On Wed, 03 Feb 2016 02:16:23 -0800, Ederson de Souza wrote:
> On 02-02-2016 19:32, Iván Briano wrote:
> > On Tue, 02 Feb 2016 09:41:04 -0800, Guilherme Íscaro wrote:
> >  > Prevent excess precision on non 64 machines, this may lead to
> >  > wrong results.
> >  >
> >
> > A bit late on my part, but what wrong results? Why is this the right
> > fix?
> 
> Some persistence tests that mapped drange values may fail on x86 
> (32bit), due to GCC generating fpu instructions that use 80bit floating 
> points (extended double precision). With -ffloat-store, doubles are kept 
> using 64bits and then results are equal to that of x86_64, where 
> floating point calc use sse instructions.
> We had a similar fix to soletta galileo builds, afair. Maybe time to 
> revisit it also? @brunobottazzini?

Yes, I remember vaguely to have seen that before. My point is mostly
that with commits like this, we need to be a lot more explicit in the
commit message. And I know that I'm guilty as well of leaving out
details, but we need to start enforcing this on all our commits to avoid
running into problems later and not knowing, exactly, what a patch was
trying to solve.

In this case, for example, your explanation tells us what's the problem,
but it still doesn't answer my other question. Is this the right fix?
How will this affect applications that use these things? I know that
floating point is a huge pain in the bottoms, but this still feels a lot
(to me, at least) as a way to make our test units pass, rather than
making sure that we have the best possible code for the task at hand.

edersondisouza:
On 03-02-2016 21:48, Iván Briano wrote:
> On Wed, 03 Feb 2016 02:16:23 -0800, Ederson de Souza wrote:
>  > On 02-02-2016 19:32, Iván Briano wrote:
>  > > On Tue, 02 Feb 2016 09:41:04 -0800, Guilherme Íscaro wrote:
>  > > > Prevent excess precision on non 64 machines, this may lead to
>  > > > wrong results.
>  > > >
>  > >
>  > > A bit late on my part, but what wrong results? Why is this the right
>  > > fix?
>  >
>  > Some persistence tests that mapped drange values may fail on x86
>  > (32bit), due to GCC generating fpu instructions that use 80bit floating
>  > points (extended double precision). With -ffloat-store, doubles are kept
>  > using 64bits and then results are equal to that of x86_64, where
>  > floating point calc use sse instructions.
>  > We had a similar fix to soletta galileo builds, afair. Maybe time to
>  > revisit it also? @brunobottazzini?
>
> Yes, I remember vaguely to have seen that before. My point is mostly
> that with commits like this, we need to be a lot more explicit in the
> commit message. And I know that I'm guilty as well of leaving out
> details, but we need to start enforcing this on all our commits to avoid
> running into problems later and not knowing, exactly, what a patch was
> trying to solve.
>
> In this case, for example, your explanation tells us what's the problem,
> but it still doesn't answer my other question. Is this the right fix?
> How will this affect applications that use these things? I know that
> floating point is a huge pain in the bottoms, but this still feels a lot
> (to me, at least) as a way to make our test units pass, rather than
> making sure that we have the best possible code for the task at hand.
>

I do believe that keep behaviour consistent is good. So I'd say that 
this is the right fix - after all, Soletta does work with double, and 
expects things to be consistent with that. The fact that x87 
instructions use internally extended double precision looks like a 
problem. Indeed, more 'modern' instruction sets (like SSE) keeps double 
as 64 bits double precision format.
Of course, using '-ffloat-store' may reduce performance, and we always
can find a way to relax the tests. Not sure if it's worth, as I don't 
know by how much performance is affected.
And IIRC, the test that triggers this bug is definitely wrong when 
running without '-ffloat-store' - to the point that one facing it would 
use a 'sprintf' to get the integer part of a number =D
(IIRC code tries to get the integer part of a number that looks like a 5 
[and is printed as such], should be a 5 [hence the calculations that led 
to it], but get a 4.)

brunobottazzini:
@bdilly, yes this my patch is not necessary anymore.

It was a workaround to compile it for Galileo. 

Since meta-soletta didn't support this board yet, We haven't up streamed it.

PR #1384, opened 2016-02-02
oic: Scan callback shouldn't set the connected resource

otaviobp:
resource->resource variable is used to store which resource was being
observed by this node, if any. This variable was incorrectly set when
scanning.

Before trying to observer a resource that was previously scanned, we
were trying to unobserve it and in some cases this was preventing it to
be observed.

@ederson Can you test that?


otaviobp:
merged.

PR #1387, opened 2016-02-03
Fixed issue#1371 Humidity sensor result into segmentation fault.

guchaojie:
The root cause is that sol_iio_open function may access sysfile
which is not exportd by some drivers. This will result into
segmentation fault by directly calling sol_util_write_file
to write absent file without checking if the file exist or not
in the sol_iio_open process.

Change-Id: I9833bfc25dd737010b15da9cf33a0596c392d210

### DIFF
--- a/src/lib/common/include/sol-mainloop.h
+++ b/src/lib/common/include/sol-mainloop.h
@@ -62,6 +62,7 @@ extern "C" {
  * For a better reference, check the following groups:
  * @li @ref Certificate
  * @li @ref Comms
+ * @li @ref Crypto
  * @li @ref Datatypes
  * @li @ref Flow
  * @li @ref IO


guchaojie:
@edersondisouza please have a review.

src/shared/sol-util-file.c
```diff
@@ -86,6 +86,10 @@ sol_util_write_file(const char *path, const char *fmt, ...)
     va_list ap;
     int ret;
 
+    ret = access(path, 0);
+    if (ret < 0)
+        return ret;
+
```
edersondisouza: I'm afraid that this is not right. `sol_util_write_file` may create a file in some situations, so it isn't right not to allow that. Where is this segfault happening? Could you please send valgrind log of this segfault?
You can get it with:

     valgrind sol-fbp-runner test.fbp &> valgrind.log

Thanks!

guchaojie: I'm afraid that sol_util_write_file just fopen the file , not create file in this situation. So we have to check this file path exist or not.
The segfault happens in calling set_buffer_size function in sol_iio_open process. It will access BUFFER_LENGHT_DEVICE_PATH(/iio:device%d/buffer/length) by sol_util_write_file.
And we can see the light sensor as example(in fact humidity sensor has the same issue):
root@intel-corei7-64:/sys/bus/iio/devices/iio:device2# ls
dev	in_illuminance0_input	      in_intensity_both_raw	  in_intensity_ir_raw  power	  uevent
events	in_intensity_both_calibscale  in_intensity_ir_calibscale  name		       subsystem
root@intel-corei7-64:/sys/bus/iio/devices/iio:device2# cat name 
tsl2563
There is no buffer directory for sol_util_write_file to fopen, we trace here segfault happen. This is just for this case, sometimes some driver also will not expose trigger directory, so I think we have to check the file existence before sol_util_write_file fopen in case of segment fault.

To attach valgrind sol-fbp-runner test.fbp &> valgrind.log for your reference:
root@intel-corei7-64:/# valgrind sol-fbp-runner light.fbp &> valgrind.log
Segmentation fault
root@intel-corei7-64:/# cat valgrind.log 
==2032== Memcheck, a memory error detector
==2032== Copyright (C) 2002-2015, and GNU GPL'd, by Julian Seward et al.
==2032== Using Valgrind-3.11.0 and LibVEX; rerun with -h for copyright info
==2032== Command: sol-fbp-runner light.fbp
==2032== 
WRN: ./src/lib/io/sol-iio.c:699 sol_iio_open() Could not disable buffer for device1
WRN:sol-flow-iio ./src/modules/flow/iio/nodes.c:1297 light_create_device_cb() mdata->device == NULL
==2032== Invalid read of size 8
==2032==    at 0x4ECFED5: sol_iio_read_channel_value (in /usr/lib/libsoletta.so)
==2032==    by 0x845C774: light_reader_cb (in /usr/lib/soletta/modules/flow/iio.so)
==2032==    by 0x845CE51: light_tick (in /usr/lib/soletta/modules/flow/iio.so)
==2032==    by 0x4EC0A6B: dispatch_process (in /usr/lib/libsoletta.so)
==2032==    by 0x4EC159D: flow_send_do (in /usr/lib/libsoletta.so)
==2032==    by 0x4EC176A: flow_send_idle (in /usr/lib/libsoletta.so)
==2032==    by 0x4E806A7: sol_mainloop_common_timeout_process (in /usr/lib/libsoletta.so)
==2032==    by 0x4E82DFC: sol_mainloop_impl_iter (in /usr/lib/libsoletta.so)
==2032==    by 0x4E81116: sol_mainloop_impl_run (in /usr/lib/libsoletta.so)
==2032==    by 0x4E834AB: sol_run (in /usr/lib/libsoletta.so)
==2032==    by 0x4030C0: main (in /usr/bin/sol-fbp-runner)
==2032==  Address 0x0 is not stack'd, malloc'd or (recently) free'd
==2032== 
==2032== 
==2032== Process terminating with default action of signal 11 (SIGSEGV)
==2032==  Access not within mapped region at address 0x0
==2032==    at 0x4ECFED5: sol_iio_read_channel_value (in /usr/lib/libsoletta.so)
==2032==    by 0x845C774: light_reader_cb (in /usr/lib/soletta/modules/flow/iio.so)
==2032==    by 0x845CE51: light_tick (in /usr/lib/soletta/modules/flow/iio.so)
==2032==    by 0x4EC0A6B: dispatch_process (in /usr/lib/libsoletta.so)
==2032==    by 0x4EC159D: flow_send_do (in /usr/lib/libsoletta.so)
==2032==    by 0x4EC176A: flow_send_idle (in /usr/lib/libsoletta.so)
==2032==    by 0x4E806A7: sol_mainloop_common_timeout_process (in /usr/lib/libsoletta.so)
==2032==    by 0x4E82DFC: sol_mainloop_impl_iter (in /usr/lib/libsoletta.so)
==2032==    by 0x4E81116: sol_mainloop_impl_run (in /usr/lib/libsoletta.so)
==2032==    by 0x4E834AB: sol_run (in /usr/lib/libsoletta.so)
==2032==    by 0x4030C0: main (in /usr/bin/sol-fbp-runner)
==2032==  If you believe this happened as a result of a stack
==2032==  overflow in your program's main thread (unlikely but
==2032==  possible), you can try to increase the size of the
==2032==  main thread stack using the --main-stacksize= flag.
==2032==  The main thread stack size used in this run was 8388608.
==2032== 
==2032== HEAP SUMMARY:
==2032==     in use at exit: 76,662 bytes in 56 blocks
==2032==   total heap usage: 1,015 allocs, 959 frees, 348,363 bytes allocated
==2032== 
==2032== LEAK SUMMARY:
==2032==    definitely lost: 0 bytes in 0 blocks
==2032==    indirectly lost: 0 bytes in 0 blocks
==2032==      possibly lost: 0 bytes in 0 blocks
==2032==    still reachable: 76,662 bytes in 56 blocks
==2032==         suppressed: 0 bytes in 0 blocks
==2032== Rerun with --leak-check=full to see details of leaked memory
==2032== 
==2032== For counts of detected and suppressed errors, rerun with: -v
==2032== ERROR SUMMARY: 1 errors from 1 contexts (suppressed: 0 from 0)

edersondisouza: Hi,

Unfortunately, this log doesn't contain line numbers =/ To get them, run valgrind on a debug build or build Soletta with `make CFLAGS='-g'`.
Nevertheless, by putting the pieces together, I guess that the real issue is that iio device wasn't created at all - I remember a warning that you reported elsewhere about it. So I believe that the real fix is to see where iio device creation failed (some call to `sol_iio_open`), and abort the flow there.
I also understood that device creation is failing because it's attempting to do a buffered reading on a device that does not support buffers. We could fallback to 'not buffered' readings as well, instead of aborting the flow.
Either way, `sol_util_write_file` is not wrong, there's no need to change it, regarding this issue. If it can not create the file, it will just return an error. Look at this line on valgrind log:

     ==2032== at 0x4ECFED5: sol_iio_read_channel_value (in /usr/lib/libsoletta.so)

The error is on `sol_iio_read_channel_value`, not on `sol_util_write_file`. You can check the line if you build Soletta with debug information before running on valgrind.

guchaojie: Yes, segfault in sol_iio_read_channel_value because device is NULL which failed to creation in sol_iio_open. FYI, The segment fault appear when we set buffer_size=-1

guchaojie: For method 1st,  calling set_buffer_enable do not goto error when disabling buffer failed, just give warning to user and keep going instead of modifying sol_util_write_file function. What's your point?

PR #1390, opened 2016-02-03
Fix some compiler warnings.

glima:
Again, given that char does not define whether it's signed or unsigned
in C, some compilers may issue warnings on these functions (leftover
from "various: Avoid using 'char' as an index" commit). At
power-supply.c, we cast to the right type for
sol_power_supply_get_capacity() usage.


ceolin:
:+1:

glima:
Merged with fixes.

PR #1391, opened 2016-02-03
Oic - Allows more than one server's instace

ceolin:


thiagomacieira:
I'd like to ask you to wait. Your first commit will conflict with my splitting of the Linux socket code into Unix-generic and Linux-specific. Your change is easier to re-do after the conflict, mine is a little more difficult.

barbieri:
okay @thiagomacieira but do you have an early version so we can check how you're doing it?

I think it could be done by moving the common functions to a -posix.c file and use those symbols when setting the "impl".

thiagomacieira:
See #1395

ceolin:
@thiagomacieira i intend to merge this commit today later because there is an old issue opened that is fixed with it

edersondisouza:
Do I need something special to test this patch? On my machine, if I start two light-server and light-scan app, it doesn't matter to which server I try to 'pair', client will only communicate with the one that started after. So, at least for me, this series doesn't work =/

otaviobp:
In order to have 2 servers running in same machine, you need to set SOL_MACHINE_ID variable in one of them, so we will have 2 servers with different device ids.
But, anyway, you are right. I am now facing the same problem, only the last server is receiving packets. @ceolin, can you check that?

edersondisouza:
@otaviobp Yes, one of them has a different SOL_MACHINE_ID.

thiagomacieira:
I've been busy and haven't had the time to act on people's feedback on #1395 yet. I definitely won't have time for it today before EOB in your timezone, but could try to do it if you give me 24 hours.

ceolin:
@thiagomacieira sure, i can wait. I was thinking you would not have time to review you pr.

thiagomacieira:
I don't. I will try to do something today, during meetings.

ceolin:
@thiagomacieira ok,  thanks a lot !

ceolin:
closed, new version https://github.com/solettaproject/soletta/pull/1467

PR #1393, opened 2016-02-03
Re-fix the case of XSI-compliant strerror_r

thiagomacieira:
Instead of relying on whether buf[0] was modified or not, do a proper
detection.


barbieri:
+1

PR #1394, opened 2016-02-03
Fix the build when SOCK_CLOEXEC and SOCK_NONBLOCK aren't defined

thiagomacieira:

### DIFF
--- a/src/lib/comms/sol-socket-impl-linux.c
+++ b/src/lib/comms/sol-socket-impl-linux.c
@@ -152,7 +152,11 @@ static struct sol_socket *
 sol_socket_linux_new(int domain, enum sol_socket_type type, int protocol)
 {
     struct sol_socket_linux *s;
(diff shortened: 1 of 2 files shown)


barbieri:
+1

but I wonder if we should introduce AF_UNIX to `sol_socket` so we can avoid doing manual socket in `unix-socket`. What do you think? @ceolin?

thiagomacieira:
Makes sense, but the use is very limited now. Adding it to sol-socket would cause the non-Unix systems to have to show undefined for those cases...

Merge?

ceolin:
i go with @thiagomacieira. I prefer don't introduce it on sol-socket if the only usage so far is those nodes. I think it's better do this manual than have to introduce support for it in N platforms, the chance of we break something when making changes is bigger.

ceolin:
:+1:

barbieri:
+1 for me

bdilly:
Commit was merged through pr #1945 
Thanks

PR #1395, opened 2016-02-03
Make sockets work on non-Linux Unix systems

thiagomacieira:
Most of the code was generic anyway, so move it to a "common" file. For the bits and pieces that were specific to Linux, reimplement in a generic version.

This patch series includes #1394.


thiagomacieira:
@ceolin

src/lib/common/Kconfig
```diff
@@ -32,6 +32,11 @@ config DETECT_BOARD_NAME
 config PLATFORM_LINUX
 	bool
 
+config PLATFORM_UNIX
+        bool "unix"
+	depends on HAVE_UNIX
```
barbieri: different indentation with line above

src/lib/comms/sol-network-impl-unix-generic.c
```diff
@@ -0,0 +1,140 @@
...
+
+    free(network);
+    network = NULL;
+}
+
+SOL_API bool
+sol_network_subscribe_events(void (*cb)(void *data, const struct sol_network_link *link,
+    enum sol_network_event event),
+    const void *data)
+{
```
barbieri: can we find a way to do this in other platforms? Otherwise if network links are enabled/disabled the servers such as CoAP will not work properly, like joining multicast groups and so on.

src/lib/comms/sol-socket-impl.h
```diff
@@ -52,8 +52,8 @@ struct sol_socket_impl {
     int (*bind)(struct sol_socket *s, const struct sol_network_link_addr *addr);
 };
 
-#ifdef SOL_PLATFORM_LINUX
-const struct sol_socket_impl *sol_socket_linux_get_impl(void);
+#ifdef SOL_PLATFORM_UNIX
+const struct sol_socket_impl *sol_socket_unix_get_impl(void);
```
barbieri: this is a good time to review this, @ibriano why did you use this instead of a `sol_socket_get_impl()` that is implemented per platform?

barbieri:
looks good to me, just minor comments. I think we better merge this ASAP and work on details afterwards.

src/lib/common/Kconfig
thiagomacieira: Damn. It didn't show because of 8-space tabs. My OS X's Emacs isn't properly configured for space mode.

src/lib/comms/sol-network-impl-unix-generic.c
thiagomacieira: No, we cannot. Not generically anyway and most likely not at all for most OSes.

FreeBSD 8.0 had EVFILT_NETDEV, but it's gone in 9.0 (https://www.freebsd.org/cgi/man.cgi?query=kqueue&sektion=2&manpath=FreeBSD+9.0-RELEASE ). Other OSes may not have the functionality at all.

thiagomacieira: Looks like we use PF_ROUTE sockets and we'll get messages of type RTM_NEWADDR / RTM_DELADDR and RTM_IFANNOUNCE / RTM_IFINFO for state changes. It should be possible to write a network monitor that will work on the BSDs and on OS X (which is a BSD).

src/lib/comms/sol-socket-impl.h
ceolin: yep, it is pretty weird because the build handle this. We only build one implementation per time AFAIK.

ibriano: On Wed, 03 Feb 2016 15:50:11 -0800, Gustavo Sverzut Barbieri wrote:
> > @@ -52,8 +52,8 @@ struct sol_socket_impl {
> >      int (*bind)(struct sol_socket *s, const struct sol_network_link_addr *addr);
> >  };
> >  
> > -#ifdef SOL_PLATFORM_LINUX
> > -const struct sol_socket_impl *sol_socket_linux_get_impl(void);
> > +#ifdef SOL_PLATFORM_UNIX
> > +const struct sol_socket_impl *sol_socket_unix_get_impl(void);
> 
> this is a good time to review this, @ibriano why did you use this instead of a `sol_socket_get_impl()` that is implemented per platform?

'twas not me, but I can fix it.

barbieri: Didn't you fix the int/unsigned to stdint.h types? 

anyway, thanks for fixing

poussa:
Can we get this moving again. I'd like to see os x support getting in...

@thiagomacieira how to you handle the sol_platform_* APIs. I get link errors when trying this patch.

Don't see any activity in ```src/lib/common/Makefile```which I guess needs to be patched and ```sol-platform-linux-common.c```as well.

```
      LD   build/soletta_sysroot/usr/lib/libsoletta.so
Undefined symbols for architecture x86_64:
  "_sol_platform_impl_add_service_monitor", referenced from:
      _sol_platform_add_service_monitor in sol-platform.o
  "_sol_platform_impl_apply_locale", referenced from:
      _sol_platform_apply_locale in sol-platform.o
```

thiagomacieira:
@poussa: I had fixed all such errors. You need to either provide an implementation for those functions (dummy or not) or you need to make sure that their not called in the first place.

The best solution for OS X (and iOS and Android and QNX and Windows, etc.) is to not compile sol-platform.c at all, since one won't be using Soletta to control the platform. It just happens that there's a lot of code leakage inside Soletta that requires functionality from sol-platform.c today, so I provided dummies instead to make the porting easier.

barbieri:
For Sakari and Nodejs, the route Thiago was doing is okay.

For OSX/Native and iOS where you want to be a guest part of an application, it doesn't make sense to use that route of us providing a mainloop and the likes. For these we should export native (ie: ObjC) modules that use Soletta internally, but it's not exposed. In iOS/OSX this would be a Framework where you compile with XCode and pick only the C files you need from Soletta. The "sol-mainloop-impl-ios.m", for instance, would use the current `RunLoop` to forward requests. Likewise, the "sol-socket-impl-ios.m" would call recommedend socket APIs (maybe POSIX socket, I'm not sure). These cases should also define `SOL_API` to empty, so we do not export any symbols from Soletta at all.

And as thiago said, export only what makes sense or what is missing in the platform. Like OIC is missing, MQTT, HTTP are not... I/O (GPIO, I2C...) are not possible, etc.

This is a different case of what Thiago was doing, that is to run soletta applications in MacOS. This one we want Soletta as host main loop, similar to how Qt/Gtk does their porting.

poussa:
@thiagomacieira I don't see the sol_platform dummy functions anywhere. Where are they? I started to do that but wanted to ask since it seems the only viable solution.

@barbieri My request for this PR has nothing to do with Node.js. I just want to run / debug soletta on OS X since that is my default env. Now I need to start VM (linux) every time I want to use soletta. That is, my request is to run soletta apps on OS X (MacOS).

thiagomacieira:
@poussa: the specific functions you're listing didn't exist when I wrote my patches. So you're not going to find the dummy functions anywhere.

@barbieri: as far as I am concerned, this patch is done and was ready to be merged 2 months ago. Please take care of it. (don't ask me to resolve the merge conflict, I don't have time; it didn't conflict two months ago)

barbieri:
I understood you'd merge it.

Okay, I can ask someone in my team to merge and solve conflicts, but will need you to test the build on MacOS.

@cabelitos can you rebase this branch and send a PR so @thiagomacieira can test?

thiagomacieira:
From @poussa's message, we know it doesn't compile anymore. Someone will need to add the missing functions.

PR #1397, opened 2016-02-04
Fix compile error on small OSes.

glima:
The pipe2() implementation trick should not even take place if
FEATURE_FILESYSTEM is not available, as it makes use of the
sol_util_fd_set_flag() function, defined at sol-util-file.c, compiled
under the rule

obj-libshared-$(FEATURE_FILESYSTEM) += \
    sol-util-file.o

### DIFF
--- a/src/lib/common/sol-missing.h
+++ b/src/lib/common/sol-missing.h
@@ -91,7 +91,7 @@ memrchr(const void *haystack, int c, size_t n)
 }
 #endif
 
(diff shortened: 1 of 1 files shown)


glima:
The style check error will be addressed by a following commit handling just that.

edersondisouza:
lgtm

glima:
Okay, merging.

Issue #1389, opened 2016-02-03
GET request against iotivity's ocserverbasicops fails


gabrielschulhof:
iotivity seems not to be able to parse the payload.

https://gist.github.com/gabrielschulhof/b4f8a423983e83d9d7ae

gabrielschulhof:
I built iotivity with default flags.

otaviobp:
Payload was created incorrectly when using a callback to fill the payload, but no data was added. You can use NULL instead of using no_payload callback.
Fix to this issue was pushed in pr #1399 and get.c is suppose to work against iotivity servers.
Please reopen ticket if problem was not solved for you.

PR #1399, opened 2016-02-04
oic: Support no payload packets and packets with empty map payload

otaviobp:
It was impossible to have a packet with no payload when callback to fill
payload was called. No payload packets were only possible when callback
was NULL.

Add a function to set the type of the payload, so users can create no
payload packets and packets with an empty map as the payload. It is
similar but it is not handled the same way by iotivity and Soletta.

Fixes issue #1389


otaviobp:
merged with proposed changes.

PR #1405, opened 2016-02-05
oic: Coap option size is incorrect

otaviobp:
Bug introduced by commit fd65ac4c when type of variable was changed. It
looked like it was an unharmful change, but changing the variable
definition caused a change in coap option size, afting the protocol
usage.


otaviobp:
merged

PR #1408, opened 2016-02-09
Assorted fixes

ibriano:


src/lib/io/sol-uart-impl-riot.c
```diff
@@ -164,7 +164,7 @@ sol_uart_close(struct sol_uart *uart)
 }
 
 SOL_API bool
-sol_uart_write(struct sol_uart *uart, const unsigned char *tx, unsigned int length, void (*tx_cb)(void *data, struct sol_uart *uart, unsigned char *tx, int status), const void *data)
+sol_uart_write(struct sol_uart *uart, const uint8_t *tx, unsigned int length, void (*tx_cb)(void *data, struct sol_uart *uart, uint8_t *tx, int status), const void *data)
```
barbieri: should we change `unsigned int length` to more specific type like `uint32_t` or `size_t`?

ibriano: On Tue, 09 Feb 2016 15:13:19 -0800, Gustavo Sverzut Barbieri wrote:
> > @@ -164,7 +164,7 @@ sol_uart_close(struct sol_uart *uart)
> >  }
> >  
> >  SOL_API bool
> > -sol_uart_write(struct sol_uart *uart, const unsigned char *tx, unsigned int length, void (*tx_cb)(void *data, struct sol_uart *uart, unsigned char *tx, int status), const void *data)
> > +sol_uart_write(struct sol_uart *uart, const uint8_t *tx, unsigned int length, void (*tx_cb)(void *data, struct sol_uart *uart, uint8_t *tx, int status), const void *data)
> 
> should we change `unsigned int length` to more specific type like `uint32_t` or `size_t`?

Yes, but since we keep finding these all over, we need to stop and do a
proper review of our entire API regarding these things, and I didn't
feel like it at the time.

PR #1410, opened 2016-02-11
oic: Make /oic/d and /oic/p resources discoverable

otaviobp:
In iotivity both resources are discoverable. Make them discoverable in
soletta to be compatible.

### DIFF
--- a/src/lib/comms/sol-oic-server.c
+++ b/src/lib/comms/sol-oic-server.c
@@ -385,7 +385,7 @@ error:
     if (err != CborNoError) {
         char addr[SOL_INET_ADDR_STRLEN];
         sol_network_addr_to_str(cliaddr, addr, sizeof(addr));
-        SOL_WRN("Error building response for /oc/core, server %p client %s: %s",
+        SOL_WRN("Error building response for /oic/res, server %p client %s: %s",
             oic_server.server, addr, cbor_error_string(err));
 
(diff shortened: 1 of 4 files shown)


src/lib/comms/sol-oic-server.c
```diff
@@ -754,6 +681,21 @@ sol_oic_server_add_resource(const struct sol_oic_resource_type *rt,
     return NULL;
 }
 
+SOL_API struct sol_oic_server_resource *
+sol_oic_server_add_resource(const struct sol_oic_resource_type *rt,
+    const void *handler_data, enum sol_oic_resource_flag flags)
```
poussa: I would name this function to: sol_oic_server_resource_auto and the sol_oic_server_resource_full to sol_oic_server_resource.

otaviobp: good point.

gabrielschulhof: Alternatively, why not add the path to the ```struct sol_oic_resource_type``` and, if empty, just generate one. No need for two APIs then. Wouldn't even break ABI if the path is at the end of the structure.

poussa: Yes, that is better (IMO)

cabelitos: I Aggree

otaviobp: really better. I'll do it.

PR #1412, opened 2016-02-12
Coverity fixes

ceolin:
- Fix possible null pointer dereference
- Check the `sol_ptr_vector_append` result


ceolin:
thank you for review guys

PR #1414, opened 2016-02-12
Reduce memory usage by reorganizing struct members.

cabelitos:


ibriano:
Fine by me, but I don't think we have any numbers to show here, right?

cabelitos:
@ibriano what you mean?

cabelitos:
The memory savings?

ibriano:
On Fri, 12 Feb 2016 08:26:10 -0800, Guilherme Íscaro wrote:
> The memory savings?

Yes

src/lib/common/include/sol-update.h
```diff
@@ -85,9 +85,9 @@ struct sol_update_info {
 #define SOL_UPDATE_INFO_API_VERSION (1)
     uint16_t api_version;
 #endif
+    bool need_update; /**< @brief If version of update is newer than current, so the update is necessary. */
```
lpereira: There's now a hole of alignof(version)-alignof(need_update) bytes in this struct. This is 7 bytes on x86-64, and 3 bytes on x86. You might want to [read this article on struct packing](http://www.catb.org/esr/structure-packing/), since the reorganization in all the structs in this patch is not doing what's in the tin (quite the contrary, actually).

lpereira:
Just for the record: as discussed in the IRC, some of these changes, although not using the common trick of sorting struct elements by size (largest first), would yield holes because of the 16-bit api_version element.

Since the API version is very likely to be removed on release builds and on small devices, moving, for instance, a few smaller fields to the start of the struct is likely to be a problem, generating holes in the structure and making them needlessly larger.

Also, this wouldn't be so future-proof, considering that structure members might be added or removed due to new features being implemented or code cleanups being performed.

bdilly:
@cabelitos I'm dropping this PR given all these issues raised up.
Indeed re-organizing structs to reduce memory consumption is a valid concern but if we're going this way we need to assure changes are valid for multiple archs, future proof, etc.
Also measuring is good to avoid this problem of eventually actually increasing the consumption in some cases.

PR #1417, opened 2016-02-13
Fix check of snprintf() return value

lpereira:
Patch made with Coccinelle using the following spatch:

    @@
    type T;
    T ret;
    expression size;
    @@

    <...
    ret = snprintf(...);
    ...
    (
    -ret < 0 || ret > size
    +ret < 0 || ret >= size
    )
    ...>


lpereira:
Merged, thankyou!

PR #1418, opened 2016-02-15
Build: Ease integrated build of node.js bindings

gabrielschulhof:
When building node.js bindings as part of a package it is important to have
provisions for allowing the integrator to specify the path to node-gyp and
additional flags with which it should run.


gabrielschulhof:
For some reason the semaphore stopped working:

> There seems to be an authorization problem with GitHub.
> 
> This could be either due to the authorization expiring or the project owner removing Semaphore from GitHub connections.
> 
> Please ask @anselmolsm to reauthorize Semaphore on their GitHub account.

PR #1419, opened 2016-02-15
oic: Make /oic/d and /oic/p resources discoverable - v2

otaviobp:
Changelog from v2:
 - Use field from sol_oic_resource_type struct to inform resource's path to sol_oic_server_add_resource
 - Add overflow check in function that generates resource's name.
Replaces #1410


otaviobp:
merged with proposed changes.

PR #1423, opened 2016-02-15
A few more coverity fixes [v2]

bdilly:


ceolin:
serie looks awesome !

Issue #1371, opened 2016-01-29
[P3] Humidity sensor reading result into segmentation fault.


guchaojie:
./humidity.fbp
WRN: ./src/lib/io/sol-iio.c:697 sol_iio_open() Could not disable buffer for device1
WRN:sol-flow-iio ./src/modules/flow/iio/nodes.c:1003 humidity_create_device_cb() mdata->device == NULL
Segmentation fault

dmesg -c
[1128737.017182] sol-fbp-runner[8534]: segfault at 656d69b0 ip b7680739 sp bf9120d0 error ffff0004 in libsoletta.so[b7600000+fd000]

edersondisouza:
On 29-01-2016 01:26, guchaojie wrote:
> ./humidity.fbp
> WRN: ./src/lib/io/sol-iio.c:697 sol_iio_open() Could not disable buffer
> for device1

This first warning should have more warnings before, please send them 
too - it would be interesting also run using strace, like:

   strace -ff sol-fbp-runner humidity.fbp

So we can get the real issue on this.


> WRN:sol-flow-iio ./src/modules/flow/iio/nodes.c:1003
> humidity_create_device_cb() mdata->device == NULL
> Segmentation fault
>

This is really weird. The device creation failed, as shown in the 
previous warning. Anyway, crashes are not ok. Could you please run this 
code with valgrind, so we get where this segfault happens? Don't forget 
to build Soletta for debug. I think there's a Kconfig flag, but you can 
simply use
"make CFLAGS='-g'"

Cheers,

guchaojie:
@edersondisouza PR#1387 has been submitted to fix this bug.

edersondisouza:
As #1424 solved this issue - acked by @guchaojie - I'll close it.

PR #1424, opened 2016-02-16
Iio segv

edersondisouza:


edersondisouza:
It may fix issue #1371 
@guchaojie can you check if this PR fix the issues, please?

glima:
Both make sense to me (and I hope they fix the issue :))

guchaojie:
@edersondisouza This method can work, verIfied it. Thx!

src/lib/io/sol-iio.c
```diff
@@ -1052,12 +1051,14 @@ sol_iio_read_channel_value(struct sol_iio_channel *channel, double *value)
...
     int64_t raw_value;
     char path[PATH_MAX];
-    struct sol_iio_device *device = channel->device;
+    struct sol_iio_device *device;
     bool r;
 
     SOL_NULL_CHECK(channel, false);
     SOL_NULL_CHECK(value, false);
 
+    device = channel->device;
```
guchaojie: For one suggestion, maybe we can just add one line to check SOL_NULL_CHECK(device, false) which would be better and safer, because for this issue, segment fault result from struct device being NULL. Of course, in this case that set_buffer_enable do not goto error, this segfault would never happen.

edersondisouza: @guchaojie Are you sure that `device` was `NULL`? This seems unlikely, as to add a channel one needs a device. I think that the bug here was `channel` being `NULL` (as no device was created), and trying to dereference it. If there's a channel with device NULL, we have bigger problems, and `SOL_NULL_CHECK` would just hide it.

edersondisouza:
Integrated, tks!

PR #1425, opened 2016-02-16
sol-conffile: If find an error on JSON, log it as SOL_ERR

edersondisouza:
So the message is more visible to user, and not buried along lots of
other debug messages.

### DIFF
--- a/src/shared/sol-conffile.c
+++ b/src/shared/sol-conffile.c
@@ -164,7 +164,7 @@ sol_conffile_set_entry_options(struct sol_conffile_entry *entry, struct sol_json
         }
     }
     if (reason != SOL_JSON_LOOP_REASON_OK) {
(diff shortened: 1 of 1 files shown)


mbelluzzo:
+2

barbieri:
not ERR, WRN. We use WRN in our stuff, not error... no big deal, that's the convention... error and critical are not being used in our side.

WRN is the default visibility, so will have the same effect

edersondisouza:
I didn't get the part of 'our stuff' - isn't the conffile json given to us by users?

barbieri:
our code

edersondisouza:
Integrated with SOL_WRN. Tks!

mbelluzzo:
What is the point to just wrn and keep going in those cases?

PR #1428, opened 2016-02-16
Node gyp

dorileo:
### Rationale
Do some house keeping on nodejs and node-gyp build bits...


gabrielschulhof:
This must land before #1418

:+1:

dorileo:
applied, thanks.

PR #1429, opened 2016-02-16
Two simple patches

ibriano:
- Implement addr_eq for RIOT
 - Use the namespace provided to sol-flow-node-type-gen when generating
   code with sol-fbp-generator


ibriano:
Merged

PR #1430, opened 2016-02-17
Soletta IIO node and samples

guchaojie:
Add IIO node fbp sample for Accelerate, ADC, Color, Humidity,
Light, Magnet, Pressure and Temperature sensor category.

Change-Id: I0f2795bd979f817dbf976938501a0b22b1dd6300


guchaojie:
@mbelluzzo @edersondisouza All the samples have been tested to prove work well as long as user can  setup the hardware environment and own according sensor type.

barbieri:
this is getting out of hand to help the user understand. Could we make use of sol-conf.json and the schema files for all our supported boards (at least Galileo, Edison and Minnow, whenever these are supported in them)?

I know existing gyroscope test was like that, so please send a patch to change it as well.

barbieri:
https://github.com/solettaproject/soletta/tree/master/src/samples/flow/grove-kit is a nice reference:
 - README file with general instructions;
 - `sol-flow.json` configuration file as the "fallback" case, if any;
 - `sol-flow-${BOARD_NAME}.json` for each board setup, `${BOARD_NAME}` is based on the platform detect code or environment variable, right now https://github.com/solettaproject/soletta/blob/master/data/jsons/50-default.json is being installed
 - `schema-${BOARD_NAME}.jpg` with the picture on how to setup the wires and connections to get it working.

So say you get `accelerate.fbp`, replace the instance:

```
accelerate(iio/accelerate:iio_device="lis3dh create,i2c,platform/80860F41:05,0x19,lis3dh 0",buffer_size=20)
```

With a type defined in the `sol-flow*.json`. It is not clear from this patch if this is on minnowboard or galileo, so I will suppose it is from newest kernel (>3.17) minnowboard (`BOARD_NAME=intel-minnow-max-linux_gt_3_17`), you create `sol-flow-intel-minnow-max-linux_gt_3_17.json` with:
```json
{
 "$schema": "http://solettaproject.github.io/soletta/schemas/config.schema",
 "nodetypes": [
 {
   "name": "MyAccelerate",
   "options": {
      "iio_device": "lis3dh create,i2c,platform/80860F41:05,0x19,lis3dh 0",
      "buffer_size": "20"
   },
   "type": "iio/accelerate"
  }
  ]
}
```

and in `accelerate.fbp` you use that type:

```
accelerate(MyAccelerate)
```

As you see the `sol-flow-${BOARD_NAME}.json` contains an array of `nodetypes`, then you'd add one object for each sensor you test in that folder.

This way if the address is different in minnow and galileo (ie: likely the "platform/80860F41..." is going to be different), they will work regardless of the board.

If the board is not supported/detected, then Soletta will try `sol-flow.json` (catch-all). If there is a way to provide a fallback or standard address, then you do it in this file. So if the user runs this on his desktop, then it will use that.

guchaojie:
OK, Today minnow board is in use for another test, we will resubmit new PR to add sample tomorrow.

PR #1432, opened 2016-02-17
oic: Fix uninitialized memory access in no-payload requests

otaviobp:
Payload should be ignored for GET and DELETE requests, so we should use
NULL as the input parameter when calling the callback. Instead of
removing the input parameter, this approach was prefered so we can use
the same function was callback to multiple methods.


glima:
Closing as this seems to have been merged.

PR #1434, opened 2016-02-18
oic-gen: Do not include platform specific network headers

edersondisouza:
They are not necessary and may break build on different platforms.


edersondisouza:
@otaviobp Could you take a look, pls?

otaviobp:
+1. There is no reason to have those includes here.

ibriano:
Take the chance to remove the similar headers oic-server/client.

edersondisouza:
@ibriano  Do you mean from the samples?

ibriano:
No, from src/lib/comms/oic-*

edersondisouza:
v2 on #1435

PR #1435, opened 2016-02-18
oic: Do not include platform specific network headers - v2

edersondisouza:
v2: Even less includes - on oic samples and lib.

They are not necessary and may break build on different platforms.


ibriano:
Now we are talking, +1

otaviobp:
+1

edersondisouza:
Merged with removal of `#include <netinet/in.h>` on `src/lib/comms/sol-oic-server.c` that was missing.

PR #1437, opened 2016-02-18
Fix some compiler warnings pointed by clang.

cabelitos:
The semaphore will say the code style is incorrect, however it breaks on parts that I did not touch.


cabelitos:
#1446

PR #1440, opened 2016-02-18
LWM2M: Add LWM2M server and client. V5

cabelitos:
Fixed problems pointed by @edersondisouza 

Which includes:
* Missing calloc() return checks
* Added a README for the samples (I added as a comment in the source code file)
* Fixed overflow checks for signed int.
* Some other minor stuff.

### DIFF
--- a/src/lib/comms/Kconfig
+++ b/src/lib/comms/Kconfig
@@ -115,3 +115,12 @@ config MAVLINK
             MAVLink or Micro Air Vehicle Link is a protocol for communicating with
             small unmanned vehicle. It is designed as a header-only message marshaling
             library.
+
+config LWM2M
+       bool "LWM2M"
+       default y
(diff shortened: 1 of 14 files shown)


src/lib/comms/Kconfig
```diff
@@ -115,3 +115,12 @@ config MAVLINK
             MAVLink or Micro Air Vehicle Link is a protocol for communicating with
             small unmanned vehicle. It is designed as a header-only message marshaling
             library.
+
+config LWM2M
+       bool "LWM2M"
+       default y
+       depends on COAP && HTTP
+       help
+           LWM2M is a protocol for communication beetween IoT devices defined by OMA
```
brunobottazzini: s/beetween/between

cabelitos: ok

PR #1441, opened 2016-02-18
grove: If a cmd fails, we still have to mark it as done.

glima:
This way, free_commands() will be still be able to free it.


glima:
Thanks, merging

PR #1444, opened 2016-02-19
Build: Only use -fPIE and -pie if soletta is shared library.

cabelitos:


bdilly:
It doesn't seem to make much sense.
Have you checked how are the binaries generated with your patch?
You can see hardening-check log if Position Independent Executable is enabled

cabelitos:
@bdilly According to GCC these flags are used for shared libraries.
Take a look at gcc man pages.
"-fpic: Generate position-independent code (PIC) suitable for use in a shared library....."

cabelitos:
You're Right I will create a new Pullreq.

PR #1445, opened 2016-02-19
Build: Enable fPIC for all builds

cabelitos:
Since our binaries are compiled with fPIE now, when compiling
Soletta as a static library the linker will complain that Soletta
is not linked with fPIC.

Hardening output as static lib:

hardening-check build/soletta_sysroot/usr/bin/sol-fbp-runner
build/soletta_sysroot/usr/bin/sol-fbp-runner:
 Position Independent Executable: yes
 Stack protected: yes
 Fortify Source functions: yes (some protected functions found)
 Read-only relocations: yes
 Immediate binding: yes


cabelitos:
merged

PR #1446, opened 2016-02-19
Fix some compiler warnings pointed by clang. v2

cabelitos:
Changes since v1.

* On  36dbea2 - Now uses sol_str_table_fallback() 
* On 88e8af1 - Removed r = -ENOMEM from sol-bus.c, since there's a patch fixing the problem.


glima:
:+1:

cabelitos:
merged

PR #1449, opened 2016-02-19
build: Set max log level to error on release mode

bdilly:
If a user is building Soletta on release mode,
set max log level to error, so no debug or informative
messages are logged / displayed.

Otherwise sensitive information regarding their products
eventually would be displayed if the enviroment
variable that sets log levels were changed
by product users / attackers.

### DIFF
--- a/src/lib/common/Kconfig
+++ b/src/lib/common/Kconfig
@@ -327,6 +327,7 @@ config LOG
 
 choice
 	prompt "Maximum allowed log level"
+	default MAXIMUM_LOG_LEVEL_ERROR if BUILD_TYPE_RELEASE
 	default MAXIMUM_LOG_LEVEL_UNLIMITED
 	depends on LOG
 	help


ibriano:
Nothing against it, but it makes me wonder if that wouldn't hide real errors, given how we use SOL_WRN for everything.

ceolin:
lgtm

bdilly:
@ibriano possibly it would hide some errors, indeed. But I prefer to be conservative here. Users may fine tune it anyway, it's just the default maximum error for release mode. She can set release mode and change the log level for debug, if she wants to.

ibriano:
Go ahead then

On Feb 19, 2016, at 20:01, Bruno Dilly <notifications@github.com<mailto:notifications@github.com>> wrote:


@ibriano<https://github.com/ibriano> possibly it would hide some errors, indeed. But I prefer to be conservative here. Users may fine tune it anyway, it's just the default maximum error for release mode. She can set release mode and change the log level for debug, if she wants to.

�

bdilly:
Ok, merged.
Actually I was tweaking our CI a bit to make sure it will continue to log everything (required for error tests).

barbieri:
I really dislike it, removing debug is okay-ish (so up to info level) as it saves space and eventually performance. But even that I'd keep, see below about sol-fbp-runner

I'm not sure what it would disclose from a product... This is just about soletta's code. If user wants to disable to his own usage of sol-log, he should disable it in his app (CFLAGS+=-DSOL...)

Most distros will end using this default setting, turning using soletta really bad experience. Say using sol-fbp-runner. 

What we should document is that if space saving is required people should disable log altogether. This can be the case after enough testing is done

bdilly:
OK, please take a look on PR https://github.com/solettaproject/soletta/pull/1473 to see if it solves this issue in a satisfactory manner.

PR #1454, opened 2016-02-22
Sol-I2C: Fix resource leak pointed by coverity.

cabelitos:


cabelitos:
Merged

PR #1455, opened 2016-02-22
comms: add bluetooth socket and network management for Zephyr - V2

brunobottazzini:
Differences from V1:
     - Replace AF_BLUETOOTH to AF_BT_IOTIVITY to be more comprehensive as accorded 
     - Use 0-time timer in set_on_write function.
     - Use of  sol_util_strtol instead of strtol.
     - Removal of SOL_ERR replacing it for SOL_WRN.
     - Changing some SOL_WRN to SOL_DBG.
     - Add PTR missing checks.
     - Checking some bt_gatt returns.
     - Some other minor changes.

@cabelitos and @glima, could you take a look at this ?

### DIFF
--- a/src/lib/comms/include/sol-network.h
+++ b/src/lib/comms/include/sol-network.h
@@ -73,6 +73,9 @@ extern "C" {
  */
 #define SOL_INET_ADDR_STRLEN 48
 
+#ifndef AF_BT_IOTIVITY
+    #define AF_BT_IOTIVITY 31
+#endif
 
 /**
  * @struct sol_network_hostname_handle
@@ -133,6 +136,9 @@ struct sol_network_link_addr {
     union {
         uint8_t in[4];
         uint8_t in6[16];
+#ifdef SOL_PLATFORM_ZEPHYR
+        uint8_t in_ble[6]; /* 0:5 are address, 6 is the type of the address */
+#endif
     } addr; /**< @brief The address itself */
     uint16_t port; /**< @brief The port associed with the IP address */
 };
--- a/src/lib/comms/Makefile
+++ b/src/lib/comms/Makefile
@@ -38,7 +38,8 @@ obj-networking-$(PLATFORM_LINUX) += \
     sol-socket-impl-linux.o
 
 obj-networking-$(PLATFORM_ZEPHYR) += \
-    sol-network-impl-zephyr.o
+    sol-network-impl-zephyr.o \
+    sol-socket-impl-zephyr-ble.o
 
 obj-networking-$(COAP) += \
     coap.o \
--- a/src/lib/comms/Makefile
+++ b/src/lib/comms/Makefile
@@ -37,6 +37,9 @@ obj-networking-$(PLATFORM_LINUX) += \
     sol-network-impl-linux.o \
     sol-socket-impl-linux.o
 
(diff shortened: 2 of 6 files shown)


src/lib/comms/include/sol-network.h
```diff
@@ -73,6 +73,9 @@ extern "C" {
  */
 #define SOL_INET_ADDR_STRLEN 48
 
+#ifndef AF_BT_IOTIVITY
+    #define AF_BT_IOTIVITY 31
```
barbieri: This is strange, could you elaborate?

How are we expected to make this work in Linux?

brunobottazzini: You're right this is Zephyr case only. It won't work for Linux, at least for now.

I should add:
`#ifdef SOL_PLATFORM_ZEPHYR`

What do you think ?

barbieri: I still don't think it is right. What is the purpose of mixing Iotivity knowledge in sol-socket level?

brunobottazzini: This socket is made to be able to comunicate with Iotivity's instances and Soletta's instances. It contains the very same characteristics and services of Iotivity.

@glima, @vcgomes and I were talking about this, and since this patch is not for a general puporse, it would be nice to expose it as the name iotivity. 

I was thinking now I should also have pointed this out in the commit message.

ibriano: On Mon, 22 Feb 2016 12:46:16 -0800, Bruno Bottazzini wrote:
> > @@ -73,6 +73,9 @@ extern "C" {
> >   */
> >  #define SOL_INET_ADDR_STRLEN 48
> >  
> > +#ifndef AF_BT_IOTIVITY
> > +    #define AF_BT_IOTIVITY 31
> 
> This socket is made to be able to comunicate with Iotivity's instances and Soletta's instances. It contains the very same characteristics and services of Iotivity.
> 
> @glima, @vcgomes and I were talking about this, and since this patch is not for a general puporse, it would be nice to expose it as the name iotivity. 
> 
> I was thinking now I should also have pointed this out in the commit message.

So, say we have another protocol using CoAP through BLE, we'd need to
have another AF_BT_SOMEPROTOCOL? And how would sol-coap know which one
to use?

brunobottazzini: Good point.

What you guys suggest to be done ?

ceolin:
btw, i agree with @barbieri, this code contains a lot of oic specific things, it should not be in the network layer. I'm taking a look to find out a better place to put this oic part.

src/lib/comms/include/sol-network.h
barbieri: we're discussing at IRC, likely will require bigger changes, but it will pay off in the future:
 - create sol-ble.h and move BLE wrappers in there;
 - split network handling from sol-coap and sol-oic. We can keep the current functions as a helpers since socket will be the common transport, but we also work based on given buffers:
   * this allows us to use sol-socket
   * add BLE
   * load/save to files, enabling more testing like fuzzy-testers (ie: AFL)
 - sol-oic and sol-coap would handle different transports such as BLE, OIC, SMS, XMPP...

PR #1458, opened 2016-02-22
comms:build: Preserve soletta flags

ceolin:
There are a lot of warnings that are disabled when building tinydtls
that should not be used in soletta.


ceolin:
@otaviobp please take a look

barbieri:
+1

PR #1461, opened 2016-02-23
Change some apis to use struct sol_buffer insteand of char *buf, size_t buflen

cabelitos:
@barbieri Is this what you wanted?

### DIFF
--- a/src/lib/common/sol-power-supply-impl-linux.c
+++ b/src/lib/common/sol-power-supply-impl-linux.c
@@ -100,14 +100,14 @@ _get_file_path(char *file_path, size_t size, const char *name,
 
 static int
 _get_string_prop(const char *name, enum sol_power_supply_prop prop,
-    char **value)
+    struct sol_buffer *buf)
 {
     char file_path[PATH_MAX];
     char *str_value;
     int r;
 
     SOL_NULL_CHECK(name, -EINVAL);
-    SOL_NULL_CHECK(value, -EINVAL);
+    SOL_NULL_CHECK(buf, -EINVAL);
 
     r = _get_file_path(file_path, sizeof(file_path), name, prop);
     SOL_INT_CHECK(r, < 0, r);
@@ -119,7 +119,9 @@ _get_string_prop(const char *name, enum sol_power_supply_prop prop,
     if (r < 0)
         return r;
 
-    *value = str_value;
+    r = sol_buffer_append_slice(buf, sol_str_slice_from_str(str_value));
+    free(str_value);
+    SOL_INT_CHECK(r, < 0, r);
 
     return 0;
 }
@@ -392,23 +394,27 @@ sol_power_supply_get_capacity_level(const char *name,
 }
 
 SOL_API int
-sol_power_supply_get_model_name(const char *name, char **model_name)
+sol_power_supply_get_model_name(const char *name,
+    struct sol_buffer *model_name_buf)
 {
-    return _get_string_prop(name, SOL_POWER_SUPPLY_PROP_MODEL_NAME, model_name);
+    return _get_string_prop(name, SOL_POWER_SUPPLY_PROP_MODEL_NAME,
+        model_name_buf);
 }
 
 SOL_API int
-sol_power_supply_get_manufacturer(const char *name, char **manufacturer)
+sol_power_supply_get_manufacturer(const char *name,
+    struct sol_buffer *manufacturer_buf)
 {
     return _get_string_prop(name,
-        SOL_POWER_SUPPLY_PROP_MANUFACTURER, manufacturer);
+        SOL_POWER_SUPPLY_PROP_MANUFACTURER, manufacturer_buf);
 }
 
 SOL_API int
-sol_power_supply_get_serial_number(const char *name, char **serial_number)
+sol_power_supply_get_serial_number(const char *name,
+    struct sol_buffer *serial_number_buf)
 {
     return _get_string_prop(name,
-        SOL_POWER_SUPPLY_PROP_SERIAL_NUMBER, serial_number);
+        SOL_POWER_SUPPLY_PROP_SERIAL_NUMBER, serial_number_buf);
 }
 
 SOL_API int
--- a/src/lib/comms/sol-network-impl-linux.c
+++ b/src/lib/comms/sol-network-impl-linux.c
@@ -91,12 +91,22 @@ static struct sol_network *network = NULL;
 
 SOL_API const char *
 sol_network_addr_to_str(const struct sol_network_link_addr *addr,
-    char *buf, uint32_t len)
+    struct sol_buffer *buf)
 {
+    const char *r;
+    size_t len;
+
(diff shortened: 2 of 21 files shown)


src/lib/common/sol-power-supply-impl-linux.c
```diff
@@ -119,7 +119,9 @@ _get_string_prop(const char *name, enum sol_power_supply_prop prop,
     if (r < 0)
         return r;
 
-    *value = str_value;
+    r = sol_buffer_append_slice(buf, sol_str_slice_from_str(str_value));
```
barbieri: isn't `sol_util_read_file()` returning the length? So you could use it here instead of using that helper that will do the strlen() for you.

Also, eventually provide a  read-file-to-buffer, if we don't have it (I guess we do). Then you read directly to the buffer and avoid the malloc.

src/lib/comms/sol-network-impl-linux.c
```diff
@@ -91,12 +91,22 @@ static struct sol_network *network = NULL;
...
+    const char *r;
+    size_t len;
+
     SOL_NULL_CHECK(addr, NULL);
     SOL_NULL_CHECK(buf, NULL);
+    len = buf->capacity - buf->used;
+    SOL_INT_CHECK(len, == 0, NULL);
+
+    r = inet_ntop(sol_network_sol_to_af(addr->family), &addr->addr,
+        sol_buffer_at_end(buf), len);
```
barbieri: hum, not good. This way you work with pre-allocated/static buffers, but you don't for empty buffers that would allocate memory. The ideal solution is to check if the buffer can be resized and if so, grow it by some threshold (ie: 32bytes), try it and handle `ENOSPC` in a loop. You can start the resize with a reasonable size, AFAIU we can know the size based on the family, right?

src/lib/comms/sol-network-impl-riot.c
```diff
@@ -50,16 +51,27 @@ static struct sol_vector links = SOL_VECTOR_INIT(struct sol_network_link);
...
-    char *buf, uint32_t len)
+    struct sol_buffer *buf)
 {
 #if MODULE_GNRC_IPV6_NETIF
+    const char *r;
+    size_t len;
+
     SOL_NULL_CHECK(addr, NULL);
     SOL_NULL_CHECK(buf, NULL);
+    len = buf->capacity - buf->used;
```
barbieri: here as well.

src/lib/datatypes/include/sol-buffer.h
```diff
@@ -182,6 +182,25 @@ enum sol_decode_case {
...
+ * // SOL_BUFFER_DECLARE_STATIC(buf, 1024);
+ * uint8_t buf_bytes[1024] = { 0 };
+ * struct sol_buffer buf = SOL_BUFFER_INIT_FLAGS(buf_bytes, 1024, SOL_BUFFER_FLAGS_FIXED_CAPACITY);
+ * @endcode
+ *
+ * @param name_ The name of the struct sol_buffer variable
+ * @param size_ The capacity of the buffer
+ */
+#define SOL_BUFFER_DECLARE_STATIC(name_, size_) \
+    uint8_t name_ ## bytes[(size_)] = { 0 }; \
```
barbieri: `storage` would be a better name than `bytes`.

src/lib/parsers/include/sol-json.h
```diff
@@ -697,20 +696,20 @@ char *sol_json_escape_string(const char *str, char *buf, size_t len) SOL_ATTR_NO
  * @param s String to escape
  */
 #define SOL_JSON_ESCAPE_STRING(s) ({ \
-        size_t str_len ## __COUNT__  = sol_json_calculate_escaped_string_len(s); \
-        sol_json_escape_string(s, (char *)alloca(str_len ## __COUNT__), str_len ## __COUNT__); \
+        SOL_BUFFER_DECLARE_STATIC(aux_buf ## __COUNT__, strlen(s) + 1); \
```
barbieri: why are you using `strlen(s) + 1`? The buffer should be of the size of the escaped length, not the input length. Like escaping `a\n` (2 bytes)  will be `a\\n` (3 bytes).

src/lib/parsers/sol-json.c
```diff
@@ -607,62 +607,83 @@ sol_json_calculate_escaped_string_len(const char *str)
...
-    char *decimal_point;
+    char *decimal_point, *end;
+    size_t len;
 
 #ifdef HAVE_LOCALE
     struct lconv *lc = localeconv();
 #endif
 
     SOL_NULL_CHECK(buf, -EINVAL);
+    len = buf->capacity - buf->used;
```
barbieri: as before, this won't work for empty strings. And you should be using `sol_buffer_append_printf()` below :-)

Review what could be simplified with `sol-buffer.h`, eventually even the memmove can go.

src/modules/flow/http-client/http-client.c
```diff
@@ -1814,9 +1820,10 @@ create_url_create_process(struct sol_flow_node *node, void *data,
 
     r = sol_http_create_uri(&uri, url, &mdata->params);
     SOL_INT_CHECK(r, < 0, r);
-    r = sol_flow_send_string_packet(node,
-        SOL_FLOW_NODE_TYPE_HTTP_CLIENT_CREATE_URL__OUT__OUT, uri);
-    free(uri);
+    r = sol_flow_send_string_slice_packet(node,
+        SOL_FLOW_NODE_TYPE_HTTP_CLIENT_CREATE_URL__OUT__OUT,
+        sol_buffer_get_slice(&uri));
+    sol_buffer_fini(&uri);
```
barbieri: you can use string-take packet sending variant and steal the buffer, so you don't duplicate memory.

src/lib/comms/sol-network-impl-linux.c
cabelitos: We have a define for that:
#define SOL_INET_ADDR_STRLEN 48
I will use your suggestion.

src/lib/comms/sol-network-impl-riot.c
cabelitos: ok

src/modules/flow/http-client/http-client.c
cabelitos: Sounds fair, there are other parts of this PR that I will sol_flow_send_string_slice_packet(), I will change them.

src/lib/datatypes/include/sol-buffer.h
cabelitos: Ok.

src/lib/parsers/sol-json.c
cabelitos: My bad. I will fix that.

src/lib/parsers/include/sol-json.h
cabelitos: You're right. Sorry

cabelitos:
#1474

PR #1465, opened 2016-02-23
Sol-platform: Remove the '\"' and '\'' characters from the locale if …

cabelitos:
…they are present.

Changes since V1:

*Remove the '\'' character as well.

### DIFF
--- a/src/lib/common/sol-platform-linux-common.c
+++ b/src/lib/common/sol-platform-linux-common.c
@@ -972,6 +972,9 @@ sol_platform_impl_load_locales(char **locale_cache)
     line = NULL;
     r = -EINVAL;
     while (fscanf(f, "%m[^\n]\n", &line) != EOF) {
+        size_t offset, line_len;
+        char quote;
+
         sep = strchr(line, '=');
         if (!sep) {
             SOL_WRN("The locale entry: %s does not have the separator '='",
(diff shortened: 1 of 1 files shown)


src/lib/common/sol-platform-linux-common.c
```diff
@@ -979,10 +982,30 @@ sol_platform_impl_load_locales(char **locale_cache)
...
+            quote = *(sep + 1);
+            value.data = sep + 2;
+            offset++;
+        }
+
         key.len = sep - key.data;
-        value.len = strlen(line) - key.len - 1;
+        //It starts with a \" or \' so it must end with a \" or \'
+        if (quote != 0) {
+            if (*(line + line_len - 1) != quote) {
```
barbieri: +0

this will do but is not correct to cover all cases. Quotes should be handled considering extra chars after it, like trailing spaces. The correct approach would be to, once a quote is present, to scan until the end quote is present.  Even more correct to handle escaping. Note that escaping is not expected in the locale values, but anyway.

I'd say that similar to the json parse, we'd use a `sol-util.h` function that takes a buffer and a slice and if required unescapes the slice to the buffer, otherwise just copies it. Then use it here. Not mandatory for this case, but likely to save this from happening in the future with other PR.

cabelitos:
#1512

PR #1467, opened 2016-02-23
OIC - v2

ceolin:
from v1:
 - Changed server api to receive `sol_network_link_addr`. It allows one bind a specific address and then has instances in different links.
 - set/get options in dtls socket

### DIFF
--- a/src/lib/comms/include/sol-coap.h
+++ b/src/lib/comms/include/sol-coap.h
@@ -465,35 +465,33 @@ void sol_coap_header_set_id(struct sol_coap_packet *pkt, uint16_t id);
 /**
  * @brief Creates a new CoAP server instance.
  *
- * Creates a new, unsecured, CoAP server instance listening on @a port. Clients
- * may use 0 for @a port to let the system pick an available one.
- * If the server cannot be created, NULL will be returned and errno will be
- * set to indicate the reason.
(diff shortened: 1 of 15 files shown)


src/lib/comms/include/sol-coap.h
```diff
@@ -465,35 +465,33 @@ void sol_coap_header_set_id(struct sol_coap_packet *pkt, uint16_t id);
...
  *
- * @param port The port the server will listen on.
+ * @param addr The address where the server will listen on.
  *
  * @return A new server instance, or NULL in case of failure.
  *
  * @see sol_coap_secure_server_new()
  */
-struct sol_coap_server *sol_coap_server_new(uint16_t port);
+struct sol_coap_server *sol_coap_server_new(struct sol_network_link_addr addr);
```
barbieri: thinking about the other PR from @brunobottazzini where I mention we'll need different implementation (not just sol-socket based, but maybe BLE, SMS...), should we think about a different parameter, like a structure describing how to access the network? If not here in `sol_coap_server()`, at least in the `sol_coap_server_full()`

You can merge this PR without that, but it is nice to plan the change to happen before v1 is tagged.

otaviobp:
series looks good to me

ceolin:
merged with the suggestions

PR #1468, opened 2016-02-23
small changes related to crypto

bdilly:


barbieri:
+1, thanks for updating to the new openssl methods, those were deprecated. And it's better to use sha256 as an example, avoiding people to use md5 as a sample :-)

Issue #1469, opened 2016-02-24
use of python3


bapclenet:
```
The following (required) dependencies were not met:

python module: jsonschema

If you've just installed it, run: make reconf
For more information/options, run: make help
```

You should add that you use python 3 here beacause I tried to install jsonschema with pip (which was linked to python 2.7) and Soletta couldn't find it after the install because it has not been added to python3

ibriano:
Ah, those are listed as dependencies somewhere in Soletta's documentation.

On Feb 24, 2016, at 07:13, bapclenet <notifications@github.com<mailto:notifications@github.com>> wrote:


The following (required) dependencies were not met:

python module: jsonschema

If you've just installed it, run: make reconf
For more information/options, run: make help


You should add that you use python 3 here beacause I tried to install jsonschema with pip (which was linked to python 2.7) and Soletta couldn't find it after the install because it has not been added to python3

�

dorileo:
Since dependency-resolver.py is a python3 script this will consider all the python tests/checks to be python3, the message can be changed to reflect it. I'm cooking a simple patch for that.

dorileo:
pull request #1470 fixes the issue.

bapclenet:
Ok thanks

PR #1470, opened 2016-02-24
dep-resolver: show the proper python's version

dorileo:
This patch changes the dependency-resolver's python module handler to
show the python's version considered in the checks.

This was raised by issue #1469.


bdilly:
It seems to work for script purposes.

But using this call to get python version in general way may not work.
For instance, I may call python (I will end up using python2 in my machine), but it would return "python", not "python2".

Another way to get the proper version would be something like:

sys.version_info[0]

that should always work.

Anyway, issue is solved.
+1

dorileo:
adjusted as you suggested, thanks.
Applied.

PR #1471, opened 2016-02-24
Nodejs simple bindings

gabrielschulhof:
Adds a test suite for node.js bindings as well as some more simple platform bindings.


bindings/nodejs/generated/main.cc.prologue
```diff
@@ -8,5 +9,6 @@ using namespace v8;
 #include "../src/common.h"
 
 NAN_MODULE_INIT(Init) {
+  sol_init();
```
barbieri: before doing the init, you should specify your libuv mainloop implementation, ok? (when you integrate mainloops).

https://github.com/solettaproject/soletta/blob/master/src/lib/common/include/sol-mainloop.h#L1013

You will need to define a `struct sol_mainloop_implementation libuv_soletta_mainloop_impl` that will define wrappers to translate our primitives to libuv. If the signatures of our callbacks do not match, then you need to translate back when calling them. In a way it is similar to our Glib main loop implementation (similar in concept, not equal in implementation):  https://github.com/solettaproject/soletta/blob/master/src/lib/common/sol-mainloop-impl-glib.c

gabrielschulhof: @barbieri, I couldn't integrate the main loop by providing an implementation, because libuv does not support polling for hangup. Instead I call ```sol_run()``` from the uv main loop, but not before setting up file descriptor and timeout watches for libuv's ```uv_backend_fd()``` and ```uv_backend_timeout()```.

Thus, the outermost execution of the uv main loop hangs, because ```sol_run()``` never returns, but a re-entrant run-once-and-don't-hang call to the uv main loop happens whenever the uv backend file descriptor wakes up the watcher attached to the soletta main loop.

Additionally, I call ```sol_quit()``` whenever all soletta asynchronous callbacks have been released. This returns control to the uv main loop and stops the soletta main loop. This latter step is necessary, because node.js operates the main loop such that, when there are no more open handles, the process is expected to quit. In contrast, soletta apps are expected to call ```sol_quit()``` in order for them to terminate.

In conclusion, it is safe for me to simply call ```sol_init()``` from the bindings' entry point.

barbieri: Not sure this is the best path to go. But let's digest in parts:

 - what you mean "polling for hangup"? I'm not sure this is UNIX signals, children PID (ie: waitpid) or something else. If you want to monitor for signals you could use `signalfd()`, or just ignore it.
 - what happens in node.js/libuv if you have no file descriptors to monitor but you still have idlers and timers? If it still works, than that's fine. Soletta not quitting if there is nothing to do (ie: no fds, no idlers, no timers) is me being lazy to implement. I could  "justify" that we still want to listen for signals, but that's not very useful -- would be just an excuse.

your current implementation seems to work, looks like you have tested it, but I'm afraid it would cause us problems with us changing the expectations of nodejs users.

PR #1473, opened 2016-02-24
build: let log level unlimited and add more explanations

bdilly:
Since we intend that most distros and users just use our
default configuration, reducing log levels, that only
would be useful for final products doesn't seem the
best approach.

Instead, we're improving documentation about it.

### DIFF
--- a/src/lib/common/Kconfig
+++ b/src/lib/common/Kconfig
@@ -327,7 +327,6 @@ config LOG
 
 choice
 	prompt "Maximum allowed log level"
-	default MAXIMUM_LOG_LEVEL_ERROR if BUILD_TYPE_RELEASE
 	default MAXIMUM_LOG_LEVEL_UNLIMITED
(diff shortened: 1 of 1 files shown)


src/lib/common/Kconfig
```diff
@@ -337,11 +336,26 @@ choice
           disabled in compile time and won't be present in the
           final library.
 
+          It only affects log levels in the library functions.
+          If an application is using Soletta log system, it
+          needs to be changed using application CFLAGS.
+
```
barbieri: maybe say the `CFLAGS += -DSOL_LOG_LEVEL_MAXIMUM=2` to disable all log levels greater than 2 (warning)?

I'd also add that same remark to sol-log.h inside the `SOL_LOG_LEVEL_MAXIMUM` documentation.

bdilly:
merged with suggested changes

PR #1474, opened 2016-02-24
Change some apis to use struct sol_buffer insteand of char *buf, size_t buflen - V2

cabelitos:
Changes since v1:
* Added commits 382255e and 12e4b51
* Fixed problems pointed by @barbieri

### DIFF
--- a/src/lib/comms/sol-network-impl-riot.c
+++ b/src/lib/comms/sol-network-impl-riot.c
@@ -48,16 +48,32 @@ static struct sol_vector links = SOL_VECTOR_INIT(struct sol_network_link);
 
 SOL_API const char *
 sol_network_addr_to_str(const struct sol_network_link_addr *addr,
-    char *buf, uint32_t len)
+    struct sol_buffer *buf)
 {
 #if MODULE_GNRC_IPV6_NETIF
+    const char *r;
+
     SOL_NULL_CHECK(addr, NULL);
     SOL_NULL_CHECK(buf, NULL);
 
     if (addr->family != SOL_NETWORK_FAMILY_INET6)
         return NULL;
 
-    return ipv6_addr_to_str(buf, (ipv6_addr_t *)&addr->addr, len);
+    if (buf->capacity - buf->used < IPV6_ADDR_MAX_STR_LEN) {
+        int err;
+
+        err = sol_buffer_expand(buf, IPV6_ADDR_MAX_STR_LEN);
+        SOL_INT_CHECK(err, < 0, NULL);
+    }
+
+    r = ipv6_addr_to_str(sol_buffet_at_end(buf), (ipv6_addr_t *)&addr->addr,
+        IPV6_ADDR_MAX_STR_LEN);
+
+    if (r)
+        buf->used += strlen(r);
+
+    return r;
+
 #else
     return NULL;
 #endif
--- a/src/lib/datatypes/include/sol-buffer.h
+++ b/src/lib/datatypes/include/sol-buffer.h
@@ -270,6 +270,18 @@ sol_buffer_at_end(const struct sol_buffer *buf)
 int sol_buffer_resize(struct sol_buffer *buf, size_t new_size);
 
 /**
+ * @brief Increment the buffer capacity.
+ *
+ * @param buf The buffer
+ * @param bytes The number of bytes that the buffer capacity should me incremented.
+ *
+ * @return @c 0 on success, error code (always negative) otherwise
+ *
+ * @note Internally this function uses sol_buffer_ensure()
+ */
+int sol_buffer_expand(struct sol_buffer *buf, size_t bytes);
+
+/**
  * @brief Ensures that @c buf has at least @c min_size.
  *
  * It may allocate more than requested.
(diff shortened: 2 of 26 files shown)


src/lib/comms/sol-network-impl-riot.c
```diff
@@ -48,16 +48,32 @@ static struct sol_vector links = SOL_VECTOR_INIT(struct sol_network_link);
...
     SOL_NULL_CHECK(buf, NULL);
 
     if (addr->family != SOL_NETWORK_FAMILY_INET6)
         return NULL;
 
-    return ipv6_addr_to_str(buf, (ipv6_addr_t *)&addr->addr, len);
+    if (buf->capacity - buf->used < IPV6_ADDR_MAX_STR_LEN) {
+        int err;
+
+        err = sol_buffer_expand(buf, IPV6_ADDR_MAX_STR_LEN);
```
barbieri: it should be expanded by the missing amount, that is `IPV6_ADDR_MAX_STR_LEN - (buf->capacity - buf->used)`, right?

minor, but fix it before pushing.

src/lib/datatypes/include/sol-buffer.h
```diff
@@ -182,6 +182,25 @@ enum sol_decode_case {
...
+ * uint8_t buf_storage[1024] = { 0 };
+ * struct sol_buffer buf = SOL_BUFFER_INIT_FLAGS(buf_storage, 1024, SOL_BUFFER_FLAGS_FIXED_CAPACITY);
+ * @endcode
+ *
+ * @param name_ The name of the struct sol_buffer variable
+ * @param size_ The capacity of the buffer
+ */
+#define SOL_BUFFER_DECLARE_STATIC(name_, size_) \
+    uint8_t name_ ## storage[(size_)] = { 0 }; \
+    struct sol_buffer (name_) = SOL_BUFFER_INIT_FLAGS(name_ ## storage, (size_), SOL_BUFFER_FLAGS_FIXED_CAPACITY)
```
barbieri: i think `(` in `(name_)` may cause problems in some compilers.

src/lib/parsers/include/sol-json.h
```diff
@@ -698,19 +697,19 @@ char *sol_json_escape_string(const char *str, char *buf, size_t len) SOL_ATTR_NO
  */
 #define SOL_JSON_ESCAPE_STRING(s) ({ \
         size_t str_len ## __COUNT__  = sol_json_calculate_escaped_string_len(s); \
-        sol_json_escape_string(s, (char *)alloca(str_len ## __COUNT__), str_len ## __COUNT__); \
+        SOL_BUFFER_DECLARE_STATIC(aux_buf ## __COUNT__, str_len ## __COUNT__); \
```
barbieri: before it used explicit call to `alloca()`, now we rely on newer C99 (?) standard to allocate stuff in the stack using a variable. No big deal, but we should keep an eye if this will work in other compilers.

@lpereira

src/lib/datatypes/include/sol-buffer.h
ibriano: On Wed, 24 Feb 2016 11:59:34 -0800, Gustavo Sverzut Barbieri wrote:
> > + *
> > + * @brief A helper macro to create a static allocated buffer with a fixed capacity.
> > + *
> > + * This macro will expand into the following code:
> > + * @code{.c}
> > + * // SOL_BUFFER_DECLARE_STATIC(buf, 1024);
> > + * uint8_t buf_storage[1024] = { 0 };
> > + * struct sol_buffer buf = SOL_BUFFER_INIT_FLAGS(buf_storage, 1024, SOL_BUFFER_FLAGS_FIXED_CAPACITY);
> > + * @endcode
> > + *
> > + * @param name_ The name of the struct sol_buffer variable
> > + * @param size_ The capacity of the buffer
> > + */
> > +#define SOL_BUFFER_DECLARE_STATIC(name_, size_) \
> > +    uint8_t name_ ## storage[(size_)] = { 0 }; \
> > +    struct sol_buffer (name_) = SOL_BUFFER_INIT_FLAGS(name_ ## storage, (size_), SOL_BUFFER_FLAGS_FIXED_CAPACITY)
> 
> i think `(` in `(name_)` may cause problems in some compilers.

It also makes no sense, as you can't have a variable name that's not
just an identifier.

src/lib/parsers/include/sol-json.h
barbieri: also... given that you now handle this expand inside `sol_json_escape_string()`, maybe this macro is not needed at all?

src/lib/parsers/sol-json.c
```diff
@@ -607,40 +607,52 @@ sol_json_calculate_escaped_string_len(const char *str)
...
+    size_t i, escaped_len;
+    int r;
 
     SOL_NULL_CHECK(str, NULL);
     SOL_NULL_CHECK(buf, NULL);
 
-    for (i = 0; *str && i < len; str++, i++) {
+    escaped_len = sol_json_calculate_escaped_string_len(str);
+
+    r = sol_buffer_expand(buf, escaped_len);
```
barbieri: you should only expand what is missing. Eventually there is enough capacity to fit the escaped string and nothing needs to be expanded.

src/lib/datatypes/include/sol-buffer.h
cabelitos: I will remove the (). Thank you

src/lib/parsers/include/sol-json.h
cabelitos: No one is using it, I will remove it.

src/lib/parsers/sol-json.c
cabelitos: I changed sol_buffer_expand() to expand only the necessary amount.
Thank you.

cabelitos:
#1475

PR #1475, opened 2016-02-24
Change some apis to use struct sol_buffer insteand of char *buf, size_t buflen - V3

cabelitos:
Fixed problems pointed by @barbieri 
Changes since v2:
* Removed SOL_JSON_ESCAPE_STRING macro
* sol_buffer_expand() will not only expand the mount necessary to fit the required bytes.


barbieri:
+1

cabelitos:
merged

PR #1476, opened 2016-02-24
Nodejs simple bindings v2

gabrielschulhof:
This replaces #1471 

Changes:

* Find the closest soletta npm package by crawling up the tree and looking for package.json files which define the package.


bindings/nodejs/tests/closestSoletta.js
```diff
@@ -0,0 +1,19 @@
...
+// Find the closest instance of soletta via its package.json
+function closestSoletta( startDir ) {
+	for ( dirToCheck = startDir; dirToCheck !== "/";
+			dirToCheck = path.resolve( dirToCheck, ".." ) ) {
+		try {
+			if ( JSON.parse( fs.readFileSync(
+					path.join( dirToCheck, "package.json" ), "utf8" ) ).name == "soletta" ) {
+				return dirToCheck;
+			}
+		} catch( errors ) {}
```
brunobottazzini: If an error occurs here we would get clueless of what is happening.

I think it worths putting, at least, a console.log warning us what could be happening

gabrielschulhof: An error is actually expected here and this is the correct way to handle it. An error indicates that a ```package.json``` file was not found at that location, so you must keep searching up the tree. You don't want, like
```
Failed to open /home/nix/iot/soletta/tests/tests/package.json
Failed to open /home/nix/iot/soletta/tests/package.json
```
and then the test works fine, because it finds ```package.json``` at the next location up the tree, ```/home/nix/iot/soletta```.

```console.log()``` cannot be used in tests, because tests run as child processes of ```suite.js``` which expects their stdout to be a sequence of well-formed JSON objects. Thus, ```console.error()``` is our only recourse for arbitrary output - but again, I wouldn't output anything in this case, because a failure is very much expected.

gabrielschulhof: Similarly, if ```JSON.parse()``` throws an exception, that's also very much expected because we do not accept a malformed ```package.json``` and we must keep moving up the tree looking for the one we want.

Ultimately, if soletta's package.json is not found in the tree, the tests will definitely fail - which is their job :)

brunobottazzini: It makes sense :)

I looked at the rest of the patch and everything looks good to me

:+1:

PR #1480, opened 2016-02-25
submodules: ignore dirty submodules

bdilly:
We shouldn't change submodules.

At the moment, they'll be showed as dirty,
listed as untracked content after a build, because
some of them will generate objects on these directories.


barbieri:
+1

PR #1481, opened 2016-02-25
Some changes to the wallclock node

cabelitos:
@barbieri

### DIFF
--- a/src/modules/flow/wallclock/wallclock.c
+++ b/src/modules/flow/wallclock/wallclock.c
@@ -34,10 +34,12 @@
 #include "sol-flow-internal.h"
 #include "sol-mainloop.h"
 #include "sol-vector.h"
+#include "sol-platform.h"
 
 #include <stdio.h>
 #include <stdlib.h>
 #include <time.h>
+#include <errno.h>
 
 #define SECONDS_IN_MINUTE (60)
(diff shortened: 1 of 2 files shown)


src/modules/flow/wallclock/wallclock.c
```diff
@@ -573,9 +676,58 @@ static void
...
+    r = sol_flow_packet_get_boolean(packet, &enabled);
+    SOL_INT_CHECK(r, < 0, r);
+
+    if (enabled && !mdata->registered) {
+        struct wallclock_timer *timer = &timers[mdata->type];
+
+        r = wallclock_add_client(node, mdata);
+        SOL_INT_CHECK(r, < 0, r);
+        return sol_flow_send_irange_packet(node, 0, &timer->val);
+    }
```
barbieri: check if it was registered before we try to remove the client? Similar logic you did for timeblock.

barbieri:
overall looks good, just the minor issue trying to remove a client that is already removed, it shouldn't crash, but would do an useless loop.

cabelitos:
@barbieri I will add:
```C
if (!enabled && mdata->registered)
....
```
Before trying to remove it.

barbieri:
ok

cabelitos:
merged

PR #1485, opened 2016-02-25
Create a node to set the platform target.

cabelitos:
@barbieri

### DIFF
--- a/src/modules/flow/platform/platform.c
+++ b/src/modules/flow/platform/platform.c
@@ -673,4 +673,17 @@ locale_apply_time_process(struct sol_flow_node *node, void *data, uint16_t port,
     return sol_platform_apply_locale(SOL_PLATFORM_LOCALE_TIME);
 }
 
+static int
+platform_target_process(struct sol_flow_node *node, void *data, uint16_t port,
+    uint16_t conn_id, const struct sol_flow_packet *packet)
+{
+    int r;
+    const char *target;
+
+    r = sol_flow_packet_get_string(packet, &target);
+    SOL_INT_CHECK(r, < 0, r);
+
+    return sol_platform_set_target(target);
+}
+
 #include "platform-gen.c"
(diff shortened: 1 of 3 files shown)


src/modules/flow/platform/platform.c
```diff
@@ -673,4 +673,17 @@ locale_apply_time_process(struct sol_flow_node *node, void *data, uint16_t port,
...
+platform_target_process(struct sol_flow_node *node, void *data, uint16_t port,
+    uint16_t conn_id, const struct sol_flow_packet *packet)
+{
+    int r;
+    const char *target;
+
+    r = sol_flow_packet_get_string(packet, &target);
+    SOL_INT_CHECK(r, < 0, r);
+
+    return sol_platform_set_target(target);
```
barbieri: i wonder if we should just return the error like you did or send an error packet. I think that set-target may fail if there is an ongoing request, in this case it would be better to explicitly send an error packet saying so.

the return-error is more like the case of packet_get_string() failing, something that is not up to the user to handle

src/modules/flow/platform/platform.json
```diff
@@ -36,6 +36,22 @@
...
+        {
+          "data_type": "string",
+          "description": "The new platform target to set which must be one of: default, rescue, emergency, poweroff, reboot or suspend.",
+          "methods": {
+            "process": "platform_target_process"
+          },
+          "name": "TARGET"
+        }
+      ],
+      "name": "platform/target",
```
barbieri: call it 'set-target' or make it provide an output port to list the current target ... not sure there is such thing.

Also provide a sample on its usage. A common pattern is good to handle: press-and-hold GPIO for 10s to reboot. If one releases early, the timer should be cancelled.

src/modules/flow/platform/platform.c
cabelitos: Ok then, I will send an error packet and return 0.

src/modules/flow/platform/platform.json
cabelitos: Currently there's no API to get the current target. Maybe we can create one if possible.
For now I will rename the node to set-target and create a sample for it.

cabelitos:
#1502

PR #1486, opened 2016-02-25
build: Don't build with fPIC static library

ceolin:


bdilly:
yeah, it certainly seems better.
+1

PR #1489, opened 2016-02-26
Properly parse ipv6

cabelitos:
@glima


glima:
From a quick read here, it seems the brackets are indeed mandatory. :+1: from me

cabelitos:
Merged!

PR #1490, opened 2016-02-26
check-fbp-bin: Fix building the javascript test

ibriano:
Given how JS nodes, converted to C by sol-fbp-generator, require that
duktape.c is built along, we also need to pass appropriate LDFLAGS,
namely in this case, -lm.


ceolin:
:+1:

PR #1491, opened 2016-02-26
common: Support setting log levels at compile time

ibriano:
In order to support changing log levels per domain on systems where
getenv() is not available, now we allow passing SOL_LOG_LEVEL and
SOL_LOG_LEVELS as compile time definitions.
They work the same way as the environment variables on Unix systems,
with the only caveat that SOL_LOG_LEVEL must always be defined as a
string.


src/lib/common/include/sol-log.h
```diff
@@ -359,13 +359,55 @@ extern struct sol_log_domain *sol_log_global_domain;
...
+ *
+ * Not to be used directly. Applications using #SOL_MAIN_DEFAULT can be built
+ * passing -DSOL_LOG_LEVELS=\"domain:level,...\" on @c CFLAGS, in which case
+ * this macro will initialize each domain's log level to the values specified
+ * in the macro.
+ */
 #ifdef SOL_LOG_ENABLED
 void sol_log_domain_init_level(struct sol_log_domain *domain);
+void sol_log_init_level_global(const char *str, size_t length);
+void sol_log_init_levels(const char *str, size_t length);
```
barbieri: maybe use the slice struct and macro? no big deal

barbieri:
+1

PR #1492, opened 2016-02-26
Sol-lwm2m: Do not send CoAP notifications to objects/instances that

cabelitos:
have already been notified.


cabelitos:
To be honest I don't like that it returns a bool as well. We could change that later.

cabelitos:
#1538

PR #1493, opened 2016-02-26
oic: Add implementation of an Ownership Tranfer Method: Just Works

otaviobp:
Use Just Works Ownership Transfer Method to pair with oic secure resources
and to stablish a secure communication between client and server.
Tested using soletta as client and server and compatible with iotivity 1.1.0 samples.

Note: There is an important issue not solved yet in this series. The oic-creds.json and
oic-security-context.json files are being stored in the tmp filesystem, but they should
be stored using a secure location.

### DIFF
--- a/src/lib/comms/sol-coap.c
+++ b/src/lib/comms/sol-coap.c
@@ -130,6 +130,18 @@ struct outgoing {
 
 static bool on_can_write(void *data, struct sol_socket *s);
 
+/* This is an internal API and shouldn't be marked with SOL_API. */
+struct sol_socket *sol_coap_server_get_socket(const struct sol_coap_server *server);
+
+struct sol_socket *
+sol_coap_server_get_socket(const struct sol_coap_server *server)
+{
+
+    SOL_NULL_CHECK(server, NULL);
+
+    return server->socket;
+}
+
 SOL_API uint8_t
 sol_coap_header_get_ver(const struct sol_coap_packet *pkt)
 {
--- a/src/lib/comms/sol-oic-security.h
+++ b/src/lib/comms/sol-oic-security.h
@@ -32,46 +32,29 @@
 
 #pragma once
 
-#include <stdbool.h>
-
-#include "sol-buffer.h"
-#include "sol-network.h"
-#include "sol-socket.h"
+#include "sol-coap.h"
+#include "sol-oic-client.h"
+#include "sol-oic-common.h"
 #include "sol-str-slice.h"
 
-#define DTLS_PSK_ID_LEN 16
-#define DTLS_PSK_KEY_LEN 16
+#define MACHINE_ID_LEN 16
 
-enum sol_socket_dtls_cipher {
-    SOL_SOCKET_DTLS_CIPHER_ECDH_ANON_AES128_CBC_SHA256,
-    SOL_SOCKET_DTLS_CIPHER_PSK_AES128_CCM8,
-    SOL_SOCKET_DTLS_CIPHER_ECDHE_ECDSA_AES128_CCM8,
-    SOL_SOCKET_DTLS_CIPHER_NULL_NULL_NULL,
-};
+struct sol_oic_security;
 
-struct sol_socket_dtls_credential_cb {
-    const void *data;
+struct sol_oic_security *sol_oic_server_security_add(
+    struct sol_coap_server *server, struct sol_coap_server *server_dtls);
+void sol_oic_server_security_del(struct sol_oic_security *security);
 
-    void *(*init)(const void *data);
-    void (*clear)(void *creds);
+struct sol_oic_security *sol_oic_client_security_add(
+    struct sol_coap_server *server, struct sol_coap_server *server_dtls);
+void sol_oic_client_security_del(struct sol_oic_security *security);
 
-    ssize_t (*get_id)(const void *creds, char *id, size_t id_len);
-    ssize_t (*get_psk)(const void *creds, struct sol_str_slice id,
-        char *psk, size_t psk_len);
-};
+bool sol_oic_security_get_is_paired(const struct sol_oic_security *security,
+    struct sol_str_slice device_id);
+int sol_oic_security_pair_request(struct sol_oic_security *security,
+    struct sol_oic_resource *resource, enum sol_oic_pairing_method pm,
+    void (*paired_cb)(void *data, enum sol_oic_security_pair_result result), void *data);
 
-struct sol_socket *sol_socket_dtls_wrap_socket(struct sol_socket *socket);
-
-int sol_socket_dtls_set_handshake_cipher(struct sol_socket *s,
-    enum sol_socket_dtls_cipher cipher);
-
-int sol_socket_dtls_set_anon_ecdh_enabled(struct sol_socket *s, bool setting);
-
-int sol_socket_dtls_prf_keyblock(struct sol_socket *s,
-    const struct sol_network_link_addr *addr, struct sol_str_slice label,
-    struct sol_str_slice random1, struct sol_str_slice random2,
-    struct sol_buffer *buffer);
-
-int sol_socket_dtls_set_credentials_callbacks(struct sol_socket *s,
-    const struct sol_socket_dtls_credential_cb *cb);
-int sol_socket_dtls_close(struct sol_socket *s, struct sol_network_link_addr *addr);
+bool sol_oic_set_token_and_mid(struct sol_coap_packet *pkt, int64_t *token);
+const uint8_t *get_machine_id(void);
+bool sol_oic_pkt_has_same_token(const struct sol_coap_packet *pkt, int64_t token);
--- a/src/lib/comms/sol-socket-dtls-impl-tinydtls.c
+++ b/src/lib/comms/sol-socket-dtls-impl-tinydtls.c
@@ -52,9 +52,6 @@
 
 static const uint32_t dtls_magic = 'D' << 24 | 't' << 16 | 'L' << 8 | 's';
 
-static bool encrypt_payload(void *data, struct sol_socket *wrapped);
-
-
 struct queue_item {
     struct sol_buffer buffer;
     struct sol_network_link_addr addr;
@@ -86,6 +83,8 @@ struct creds {
     struct sol_vector items;
     char *id;
 };
+
+static bool encrypt_payload(struct sol_socket_dtls *socket);
 
 static int
 from_sockaddr(const struct sockaddr *sockaddr, socklen_t socklen,
@@ -301,7 +300,7 @@ sol_socket_dtls_sendmsg(struct sol_socket *socket, const void *buf, size_t len,
     sol_buffer_init_flags(&item->buffer, buf_copy, len, SOL_BUFFER_FLAGS_FIXED_CAPACITY | SOL_BUFFER_FLAGS_NO_NUL_BYTE);
     item->buffer.used = len;
 
-    encrypt_payload(s, s->wrapped);
+    encrypt_payload(s);
 
     return (int)len;
 }
@@ -410,9 +409,8 @@ no_item:
  * encrypting the payload, it calls sol_socket_sendmsg() to finally pass it
  * to the wire.  */
 static bool
-encrypt_payload(void *data, struct sol_socket *wrapped)
+encrypt_payload(struct sol_socket_dtls *socket)
 {
-    struct sol_socket_dtls *socket = data;
     struct queue_item *item;
     session_t session;
     int r;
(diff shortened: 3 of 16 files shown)


lpereira:
The code is too large for me to review right now, but as I wrote the first draft of this, there are a few points I'd like to mention:

- There should be a timeout to disable the anonymous ECDH key if pairing hasn't completed. Nothing too long; I'm thinking about 30~45s. (Maybe more for other OXMs, such as the one requiring a PIN from the user, but for JW this is more than adequate.)
- If it's possible to know if a given request from a DTLS socket came by while encrypted using an ECDH key, only allow requests to go through if they're pairing-related. If not, lock down the server while a JW pairing procedure is in process. This is important, since someone might start the pairing process just to enable the anonymous ECDH key and then issue encrypted requests that should otherwise be protected, without actually pairing.

src/lib/comms/sol-oic-security.h
```diff
@@ -0,0 +1,60 @@
...
+void sol_oic_client_security_del(struct sol_oic_security *security);
+
+bool sol_oic_security_get_is_paired(const struct sol_oic_security *security,
+    struct sol_str_slice device_id);
+int sol_oic_security_pair_request(struct sol_oic_security *security,
+    struct sol_oic_resource *resource, enum sol_oic_pairing_method pm,
+    void (*paired_cb)(void *data, enum sol_oic_security_pair_result result), void *data);
+
+bool sol_oic_set_token_and_mid(struct sol_coap_packet *pkt, int64_t *token);
+const uint8_t *get_machine_id(void);
```
lpereira: Why is this exposed in a header?

src/lib/comms/sol-socket-dtls-impl-tinydtls.c
```diff
@@ -160,16 +146,21 @@ session_from_linkaddr(const struct sol_network_link_addr *addr,
 }
 
 static void
+clear_queue_item(struct queue_item *item)
+{
+    sol_util_secure_clear_memory(item->buffer.data, item->buffer.capacity);
```
lpereira: This has been done before sol_buffer had a flag to clear memory securely. Please set this flag in the buffer so this is done automatically. Note that this only works when sol_buffer does not own the memory due to realloc() possibly keeping the old data around when the pointer changes; you might need to fix that first (basically, if that flag is set, instead of calling realloc(), always allocate a new block, clear the previous one -- using the previous capacity information from the buffer struct -- and free the previous block).

src/lib/comms/sol-socket-dtls-impl-tinydtls.c
```diff
@@ -355,7 +343,11 @@ read_encrypted(void *data, struct sol_socket *wrapped)
     if (to_sockaddr(&cliaddr, &session.addr.sa, &session.size) < 0)
         return false;
 
-    return dtls_handle_message(socket->context, &session, buf, len) == 0;
+    err_code = dtls_handle_message(socket->context, &session, buf, len);
+    if (err_code != 0)
+        SOL_DBG("dtls_handle_message failed with error code = %d. "
```
lpereira: Is it possible to obtain a string from this error code?

src/lib/comms/sol-socket-dtls-impl-tinydtls.c
```diff
@@ -538,12 +527,42 @@ retransmit_timer_check(struct sol_socket_dtls *s)
     clock_time_t next_retransmit = 0;
 
     dtls_check_retransmit(s->context, &next_retransmit);
+    SOL_WRN("retransmit_timer_check %d", next_retransmit);
```
lpereira: This looks more of a SOL_DBG thing than a SOL_WRN kind of thing.

src/lib/comms/sol-socket-dtls-impl-tinydtls.c
```diff
@@ -854,8 +786,9 @@ sol_socket_dtls_set_handshake_cipher(struct sol_socket *s,
 {
     static const dtls_cipher_t conv_tbl[] = {
         [SOL_SOCKET_DTLS_CIPHER_ECDH_ANON_AES128_CBC_SHA256] = TLS_ECDH_anon_WITH_AES_128_CBC_SHA_256,
-        [SOL_SOCKET_DTLS_CIPHER_PSK_AES128_CCM8] = TLS_PSK_WITH_AES_128_CCM_8,
-        [SOL_SOCKET_DTLS_CIPHER_ECDHE_ECDSA_AES128_CCM8] = TLS_ECDHE_ECDSA_WITH_AES_128_CCM_8
+        [SOL_SOCKET_DTLS_CIPHER_PSK_AES128_CCM8] = TLS_NULL_WITH_NULL_NULL,
```
lpereira: This doesn't seem correct.

src/lib/comms/sol-socket-dtls-impl-tinydtls.c
```diff
@@ -907,3 +840,35 @@ sol_socket_dtls_prf_keyblock(struct sol_socket *s,
...
+
+    SOL_INT_CHECK(socket->dtls_magic, != dtls_magic, -EINVAL);
+
+    socket->credentials = cb;
+
+    return 0;
+}
+
+int
+sol_socket_dtls_close(struct sol_socket *s, struct sol_network_link_addr *addr)
```
lpereira: Shouldn't this be part of the del method?

src/lib/comms/sol-oic-security.h
otaviobp: This is used in security and in server. It is in a not exported header. Is this a problem?

src/lib/comms/sol-socket-dtls-impl-tinydtls.c
otaviobp: ok. I'll look for usages of sol_util_secure_clear_memory and use buffer

src/lib/comms/sol-socket-dtls-impl-tinydtls.c
otaviobp: maybe. I'll look for it

src/lib/comms/sol-socket-dtls-impl-tinydtls.c
otaviobp: I did it for tests and forgot it here. I'll remove this line

otaviobp: not remove the line, I'll remove the change.

src/lib/comms/sol-socket-dtls-impl-tinydtls.c
otaviobp: I am using this to close the dtls connection without deleting everything, so we can reuse the socket to start  a new dtls connection.

src/lib/comms/sol-oic-security.h
ibriano: On Tue, 01 Mar 2016 06:28:22 -0800, Otávio wrote:
> > +struct sol_oic_security *sol_oic_server_security_add(
> > +    struct sol_coap_server *server, struct sol_coap_server *server_dtls);
> > +void sol_oic_server_security_del(struct sol_oic_security *security);
> > +
> > +struct sol_oic_security *sol_oic_client_security_add(
> > +    struct sol_coap_server *server, struct sol_coap_server *server_dtls);
> > +void sol_oic_client_security_del(struct sol_oic_security *security);
> > +
> > +bool sol_oic_security_get_is_paired(const struct sol_oic_security *security,
> > +    struct sol_str_slice device_id);
> > +int sol_oic_security_pair_request(struct sol_oic_security *security,
> > +    struct sol_oic_resource *resource, enum sol_oic_pairing_method pm,
> > +    void (*paired_cb)(void *data, enum sol_oic_security_pair_result result), void *data);
> > +
> > +bool sol_oic_set_token_and_mid(struct sol_coap_packet *pkt, int64_t *token);
> > +const uint8_t *get_machine_id(void);
> 
> This is used in security and in server. It is in a not exported header. Is this a problem?

How does this differ from platform functions to get the machine id?

lpereira:
Since I can't comment in sol-oic-security.c due to GitHub limitations:
  * In parse_pstat_json(), the check at the end should be `if (fields & (PF_ISOP | PF_DEVICEID | PF_CH | PF_CM | PF_OM | PF_SM) == (PF_ISOP | PF_DEVICEID | PF_CH | PF_CM | PF_OM | PF_SM))`, or, better yet, `if ((fields & PF_ALL_FIELDS) == PF_ALL_FIELDS)`.
  * There are memory leaks in security_load_default_context().
  * What happens if security_load_context() fails to parse one but not the other key in the JSON file?
  * In put_doxm_owned_cb(), the handshake cipher that's selected is correct?
  * Please avoid using `//comments`. Use `/* Comments. */` instead. Notice the proper capitalization, punctuation, and spacing.

src/lib/comms/sol-oic-security.h
lpereira: Every exported function, even if not in a public header, should be properly prefixed.

This function is different from the platform function because the format is different (it's binary, rather than ASCII). It calls the platform function and then convert it to a format that can then be represented (on demand) as Base64, as required by the protocol.

src/lib/comms/sol-socket-dtls-impl-tinydtls.c
lpereira: Maybe this should be part of the sol_socket interface then?

src/lib/comms/sol-oic-security.h
otaviobp: sol_platform_get_machine_id is encoded as hex and get_machine_id is a 16 byte array with decoded bytes from platform id.

ibriano: On Tue, 01 Mar 2016 06:36:37 -0800, Otávio wrote:
> > +struct sol_oic_security *sol_oic_server_security_add(
> > +    struct sol_coap_server *server, struct sol_coap_server *server_dtls);
> > +void sol_oic_server_security_del(struct sol_oic_security *security);
> > +
> > +struct sol_oic_security *sol_oic_client_security_add(
> > +    struct sol_coap_server *server, struct sol_coap_server *server_dtls);
> > +void sol_oic_client_security_del(struct sol_oic_security *security);
> > +
> > +bool sol_oic_security_get_is_paired(const struct sol_oic_security *security,
> > +    struct sol_str_slice device_id);
> > +int sol_oic_security_pair_request(struct sol_oic_security *security,
> > +    struct sol_oic_resource *resource, enum sol_oic_pairing_method pm,
> > +    void (*paired_cb)(void *data, enum sol_oic_security_pair_result result), void *data);
> > +
> > +bool sol_oic_set_token_and_mid(struct sol_coap_packet *pkt, int64_t *token);
> > +const uint8_t *get_machine_id(void);
> 
> sol_platform_get_machine_id is encoded as hex and get_machine_id is a 16 byte array with decoded bytes from platform id.

In which case, I'd rather have that function added next to the other. I
can imagine it being useful for other scenarios too.

src/lib/comms/sol-socket-dtls-impl-tinydtls.c
lpereira: Or, better yet, just close the socket and open a new one. It's cheap, and this will help destroy any data that could be potentially left over from the time the socket was being used for the initial exchange.

otaviobp:
Q - What happens if security_load_context() fails to parse one but not the other key in the JSON file?
A - If parser fails there is something wrong with the file (corruption?). I prefer to abort than trying to load the other key. In this case it is returning an error and caller is clearing the security instance.
What is missing here is that if one key is loaded and the other is not present in file, we should load the default values. I'll fix that.

otaviobp:
About the usage of TLS_NULL_WITH_NULL_NULL, there was a mistake there. In iotivity code there is a comment saying that tinydtls would select the correct cipher if NULL cipher was selected. But that is not the case. So it looks that iotivity 1.0.1 is using an insecure cipher for secure connections. This has changed in master branch. I'll adapt soletta to work with the correct cipher.

src/lib/comms/sol-coap.c
```diff
@@ -130,6 +130,18 @@ struct outgoing {
 
 static bool on_can_write(void *data, struct sol_socket *s);
 
+/* This is an internal API and shouldn't be marked with SOL_API. */
+struct sol_socket *sol_coap_server_get_socket(const struct sol_coap_server *server);
+
+struct sol_socket *
+sol_coap_server_get_socket(const struct sol_coap_server *server)
```
vcgomes: Why is this function needed?

lpereira: It's necessary to obtain the sol_socket_dtls to derivate the key from the initial handshake. (It's in sol-oic-security.c, part of this PR, which is not shown in the GitHub UI due to size limitations.)

vcgomes: Ah, sorry for the noise then. Should we call github "gitbug"?

otaviobp:
v2: #1527

PR #1494, opened 2016-02-27
Lazy comms

barbieri:
This is an *initial version* for broader review and testing.

@ceolin please check it extensively as you wrote most of that and can know if something would break.

I'm seeing some uninitialized errors in valgrind, but I didn't have time to  check if those appeared before my commit, right now they look unrelated but I'll investigate more.


cabelitos:
Series looks good

barbieri:
fixed the `SOL_NULL_CHECK(network...` and pushed.

PR #1495, opened 2016-02-27
Single node

barbieri:
this is an *initial version*  for broader review, it needs more testing and adapting to the new 'enabled' port from #1481.

I tried some local tests by adapting that sample, like using array ports, ports with connections and disconnections, they seem to work.

BUT some nodes only monitor `connect()` and doesn't do the proper behavior on `disconnect()`, so they are not friendly to dynamic connect/disconnect. Grep 'connect'  and then 'disconnect' in `src/modules/flow` to see where they don't match. We must fix those ASAP.

@bdilly @cabelitos and @cmarcelo please review.


barbieri:
see fixes in #1505

PR #1496, opened 2016-02-29
Mem fixes

barbieri:
wow, this was huge... started by me going back to master to valgrind http-client sample to understand if my #1494 introduced regressions.

I've found that new helpers introduced by @cabelitos in sol-buffer were missing some trailing null-bytes, but then I reviewed the whole sol-buffer and found many other issues, most were harmless since they were adding one more byte than was needed, but some were actually bugs by failing to copy or account that byte.

By checking that I also found that some code were using `sol_util_fill_buffer()` but were not checking if the requested amount was read, in the case of less bytes read it would proceed with garbage. This was the case with @ceolin unix-socket and @edersondisouza persistence. Likely the original versions didn't had problems, but when the sol-util-file helpers were added the details were "lost in translation".

That said, please @cabelitos, @edersondisouza and @ceolin review and test this PR extensively using valgrind, try to break it... if you do, introduce more tests so we avoid to introduce more bugs in the future. I didn't add new tests for null-byte, so it could be a nice addition from you.

### DIFF
--- a/src/lib/comms/sol-http-common.c
+++ b/src/lib/comms/sol-http-common.c
@@ -494,8 +494,10 @@ sol_http_create_uri(struct sol_buffer *buf, const struct sol_http_url url,
         SOL_INT_CHECK_GOTO(r, < 0, err_exit);
     }
 
-    r = sol_buffer_ensure_nul_byte(buf);
-    SOL_INT_CHECK_GOTO(r, < 0, err_exit);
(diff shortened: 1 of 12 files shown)


barbieri:
oops. did a mistake in the ~1 commit and forgot to compile locally... my battery is draining. I've fixed it already, need to re-trigger the CI

cabelitos:
lgtm

barbieri:
fixes in #1503

PR #1497, opened 2016-02-29
Sol-buffer: Use SOL_BUFFER_FLAGS_MEMORY_NOT_OWNED at SOL_BUFFER_DECLA…

cabelitos:
…RE_STATIC()

Otherwise if one tries to sol_buffer_fini(&buf), soletta would crash.

### DIFF
--- a/src/lib/datatypes/include/sol-buffer.h
+++ b/src/lib/datatypes/include/sol-buffer.h
@@ -198,7 +198,7 @@ enum sol_decode_case {
  */
 #define SOL_BUFFER_DECLARE_STATIC(name_, size_) \
     uint8_t name_ ## storage[(size_)] = { 0 }; \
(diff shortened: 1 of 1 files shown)


cabelitos:
I just wanted to init things..

barbieri:
answering @otaviobp, it is a good practice to `sol_buffer_fini()` stuff. I see that @cabelitos did most of his patch without doing so, it's okay since memory-not-owned does nothing in the fini, but if one converts that to dynamic, then we start to leak.

otaviobp:
Sorry if I wasn't very clear. I was saying about zeroing the "name_ ## storage" variable. This is not added in this patch, but I saw that in this review.

cabelitos:
Can I merge this?

otaviobp:
yep

cabelitos:
merged

PR #1499, opened 2016-02-29
Fixes

edersondisouza:
This PR should fix some tests that break with PR #1496 
Fixes here fix code that was broke anyway, PR #1496 only exposed them.


barbieri:
+1

edersondisouza:
Integrated. Tks!

PR #1500, opened 2016-02-29
Sol-monitors: Fix infinite loop while iterating over a sol-monitor ca…

cabelitos:
…llbacks.

The following code would lead soletta to an infinite loop:

static void
my_monitor_cb(void *data)
{
    struct sol_monitors *ms = data;

    sol_monitors_append(my_monitor_cb, ms);
}

// C code...

SOL_MONITORS_WALK(sol_monitor, cb, i)
	cb(sol_monitor);

//More C code...

This happens because sol_monitors_append() does not care if
the sol_monitor is walking. In order to prevent an infinite
loop, __SOL_MONITORS_WALK() should save how many monitors
are registered before the walk starts and walk until that number is reached.

### DIFF
--- a/src/lib/datatypes/include/sol-vector.h
+++ b/src/lib/datatypes/include/sol-vector.h
@@ -239,6 +239,20 @@ sol_vector_take_data(struct sol_vector *v)
         idx++)
 
 /**
+ * @def SOL_VECTOR_FOREACH_IDX_UNTIL(vector, itrvar, idx, until)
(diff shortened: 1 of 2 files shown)


gabrielschulhof:
Maybe we should add the code from the original report as a unit test to make sure that the synchronous add/remove continues to work. Otherwise,

:+1:

barbieri:
+1

and as @gabrielschulhof said, add this as a unit test

cabelitos:
#1504

PR #1502, opened 2016-02-29
Create a node to set the platform target. v2

cabelitos:
Changes since V1:
* Added a sample.
* Rename node to set-target
* Send error packet if sol_platform_set_target() fails


barbieri:
+1

cabelitos:
Merged.

PR #1503, opened 2016-02-29
Mem fixes v2

barbieri:
v2 since (#1496):
 - fixed typo "Thi"
 - fixed sol-iio so the next partial read accounts for only what is missing


barbieri:
damn, the trim is still wrong. I'll upload a new version with that... or push directly if that's the only issue that is broken.

barbieri:
see #1506

PR #1504, opened 2016-02-29
Sol-monitors: Fix infinite loop while iterating over a sol-monitor callbacks V2.

cabelitos:
Changes since v1:
* Added a test

### DIFF
--- a/src/test/test-monitors.c
+++ b/src/test/test-monitors.c
@@ -354,5 +354,39 @@ custom_entry_type(void)
     sol_monitors_clear(ms);
 }
 
+DEFINE_TEST(infinite_loop_test);
+
+static void
+infinite_loop_cb(const void *data)
+{
+    static uint8_t counter = 0;
+    struct sol_monitors *ms = (struct sol_monitors *)data;
+    void *v;
+
+    counter++;
+
+    if (counter != 1)
+        ASSERT(1 == 2);
+
+    v = sol_monitors_append(ms, infinite_loop_cb, ms);
+    ASSERT(v);
+}
+
+static void
+infinite_loop_test(void)
+{
+    struct sol_monitors ms;
+    void *v;
+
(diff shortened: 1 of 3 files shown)


src/test/test-monitors.c
```diff
@@ -354,5 +354,39 @@ custom_entry_type(void)
...
+infinite_loop_cb(const void *data)
+{
+    static uint8_t counter = 0;
+    struct sol_monitors *ms = (struct sol_monitors *)data;
+    void *v;
+
+    counter++;
+
+    if (counter != 1)
+        ASSERT(1 == 2);
```
barbieri: better to use `ASSERT_INT` (we have a macro to asser 2 integers are the same)

src/test/test-monitors.c
```diff
@@ -354,5 +354,39 @@ custom_entry_type(void)
     sol_monitors_clear(ms);
 }
 
+DEFINE_TEST(infinite_loop_test);
+
+static void
+infinite_loop_cb(const void *data)
+{
+    static uint8_t counter = 0;
+    struct sol_monitors *ms = (struct sol_monitors *)data;
```
barbieri: no need to cast... and the function should be `(void *data)`, no const in there.

barbieri:
just minor issues, you can push directly after fix

src/test/test-monitors.c
cabelitos: You're right. I could have written ASSERT_INT_EQ(counter, 1)

src/test/test-monitors.c
cabelitos: Actually sol_monitors_cb_t requires a const void * as argument. If I change here to void *, I will have to cast infinite_loop_cb when calling sol_monitors_append(). Either way I will need to cast something.

cabelitos:
merged.
Thanks for the review.

PR #1505, opened 2016-02-29
Single node v2

barbieri:
Changes since v1 (#1495):
 - rebased to new soletta.h changes
 - uses the new wallclock `ENABLED` ports, making the sample complete
 - sample: created options were leaking, delete them
 - isolate the api version define in `#ifndef SOL_NO_API_VERSION`
 - disconnect returns error from `sol_vector_del_element()`
 - sample `find-type.c` now explain why check version is needed in that case


barbieri:
oh crap the style check... :-/

barbieri:
Fixes in #1513

PR #1506, opened 2016-02-29
Mem fixes v3

barbieri:
now for good, fixes the `sol_buffer_trim()` usage.


barbieri:
pending #1499 to fix the tests, that one should go before this PR

PR #1507, opened 2016-03-01
Nodejs platform monitors

gabrielschulhof:
Main loop integration and platform monitors, which need it.


bindings/nodejs/src/functions/sol-platform-monitors.cc
```diff
@@ -0,0 +1,208 @@
...
+#include "../structures/js-handle.h"
+
+using namespace v8;
+
+class SolPlatformMonitor: public JSHandle<SolPlatformMonitor> {
+public:
+	static const char *jsClassName() { return "SolPlatformMonitor"; }
+};
+
+#define ADD_MONITOR(name, marshaller) \
```
barbieri: maybe instead of a macro you use a function that receives the pointer to the `sol_platform_add_XXX_monitor` to use?

barbieri:
I still think that the approach to integrate the mainloop by hijack is not good :-/

I'm okay to let this in as is if it works for you, but we should look to revisit it later, we can offer more hooks in "sol_mainloop_implementaiton"  if they are needed, like to spawn child processes and threads.

gabrielschulhof:
@barbieri 
I agree. The fundamental problems I have encountered with trying to implement the soletta main loop via libuv are these:

0. libuv does not support polling for hangup (though that may be [changing](https://github.com/libuv/libuv/pull/691)), which is needed for at least the hostname platform monitor. Polling for error is no substitute, because I need to have at least read or write in order for libuv to poll for anything, including errors, and there's always data available for reading in ```/proc/sys/kernel/hostname```, so I can't poll for read.
0. libuv does not support adding child process watches for arbitrary pids. It only allows watching pids which were spawned using ```uv_spawn()```, so we'd have to keep our ```waitpid()``` around.
0. node.js quits when the libuv main loop runs out of active and referenced handles. Thus, even if we were to solve the above two problems we would still have to be careful which handles we keep and which ones we dereference. For example, soletta creates 2(IIRC) handles as part of ```sol_init()``` which we would have to dereference in order to retain the expected behaviour whereby node.js quits when all the handles are gone. When I tried implementing the soletta main loop via libuv, I set a flag during ```sol_init()``` to indicate that all incoming handles be dereferenced, and unset it after ```sol_init()``` completed, to indicate that subsequent handles be kept as they were. This is also pretty hackish. Additionally, the fact that the expectation in soletta is that ```sol_quit()``` will stop the main loop even if there are active and referenced handles means that on the soletta side the code (whether ours or somebody else's non-node.js soletta-based code, such as maybe a module) may not take care to remove all handles it adds to the main loop, causing node.js to hang.

Thus, I believe that, for the time being, this approach to main loop integration is our only option. In the long run we can certainly try to contribute to libuv in order to eliminate the shortcomings I mention above. After that, we can definitely implement the soletta main loop via libuv.

bindings/nodejs/src/functions/sol-platform-monitors.cc
gabrielschulhof: I'm concerned about the fact that C++ is pretty pedantic about function signatures, and the various marshallers and C APIs all have different signatures. If I were to use a function, I'd have to do some serious casting (more like type bludgeoning :)).

gabrielschulhof:
@barbieri The other approach I have tried is to use a second thread wherein the soletta main loop has full reign. There are no fundamental problems with that approach, however, passing information between the two threads becomes extremely complex - not intractably, but potentially not worth the myriad bugs which can be introduced by the absence of a proper thread-safe queue implementation.

gabrielschulhof:
... not to mention the potential memory usage spike when, on the soletta thread, say, a hardware input is producing a large amount of data, whereas on the node.js side I have to use an idle function to process the data.

gabrielschulhof:
@barbieri one thing we could do in soletta is to provide a non-blocking operation of the main loop, as well as a backend fd - which is exactly what libuv provides. Then we could turn the whole thing upside-down, and have the libuv main loop pump the soletta main loop whenever it needs pumping. This approach has the added benefit that we need not worry about what handles are being added to the soletta main loop, because we can simply dereference the libuv handle encapsulating soletta's backend fd. That would allow node.js to terminate "naturally", and we could hook into the shutdown process and call ```sol_quit()```:
```JS
process.on( "exit", soletta.sol_quit );
```

barbieri:
just a note that I didn't forget these comments, I'll try to address them this week. My plan is:

 - functions such as the core of `worker-thread` and `fork-run` should be overloadable in `sol_mainloop_implementation`, this is because currently their code will do `sigprocmask()` and that is harmful. As you said, to spawn processes in libuv one should call `uv_spawn()`. This will address soletta trying to handle signals at all.
 - the `HUP` handling is kinda harmless since the node.js is not expected to run as `micro-linux` (PID1), but as systemd. You can either update to newer systemd or disable hostname testing until libuv supports that.
 - the wrappers can keep track of pending handles and cancel them on our `shutdown()` or `quit()` wrappers. Eventually the `quit()` wrapper may add an idler that will run to clean them at soletta-side and enable libuv to quit. At that point we can also flag "quit" is ongoing and adding new mainloop items will fail (return NULL).

The last item is very easy and quick to do. The first one will demand me more time :-/

PR #1509, opened 2016-03-01
Persistence fixes

edersondisouza:
While testing #1496 I've found some problems with persistence, especially efivars. So, time to send fixes. I've also set up tests here on my setup to avoid problems in future, but I'd be happier if there's a way to automate that...


edersondisouza:
If no other comment appear, I'll merge - with due fixes - it at end of day.

edersondisouza:
Integrated with fixes.

PR #1510, opened 2016-03-01
util - helpers to play with endianness

ceolin:


src/shared/include/sol-util.h
```diff
@@ -752,6 +752,161 @@ long int sol_util_strtol(const char *nptr, char **endptr, ssize_t len, int base)
 unsigned long int sol_util_strtoul(const char *nptr, char **endptr, ssize_t len, int base);
 
 /**
+ * @brief Swaps the bytes of a 16 bytes unsigned int
+ */
+#define uint16_bytes_swap(val) \
```
barbieri: if it goes in a header, namespace it with `sol_`

vcgomes:
I would rename these functions to something like: `sol_util_get_{be,le}{16,32,64}(const void *ptr)` and `sol_util_put_{be,le}{16,32,64}(uint{16,32,64}, void *ptr)`

another alternative is use to the same nomeclature as the kernel:
`sol_util_cpu_to_be16(uint16)` and `sol_util_be16_to_cpu(uint16_t)`

for the conversions, this is a good reference:
http://commandcenter.blogspot.com/2012/04/byte-order-fallacy.html

src/shared/include/sol-util.h
ceolin: doing it

ceolin:
@vcgomes i'll follow the kernel nomeclature, but i'll also add sol_util_cpu_to_le16/sol_util_le16_to_cpu too

ceolin:
https://github.com/solettaproject/soletta/pull/1515

PR #1511, opened 2016-03-01
check-style: works with uncrustify greater than 0.60

ceolin:


edersondisouza:
lgtm
Do we need to change semaphore as well?

otaviobp:
+1

bdilly:
looks like it was merged already

PR #1512, opened 2016-03-01
Sol-platform: Remove the '\"' and '\'' characters from the locale v3

cabelitos:
Changes since v2:

* Added commit 5a29ff1

### DIFF
--- a/src/lib/common/sol-platform-linux-common.c
+++ b/src/lib/common/sol-platform-linux-common.c
@@ -970,12 +970,12 @@ sol_platform_impl_load_locales(char **locale_cache)
     }
 
     line = NULL;
-    r = -EINVAL;
     while (fscanf(f, "%m[^\n]\n", &line) != EOF) {
         sep = strchr(line, '=');
         if (!sep) {
             SOL_WRN("The locale entry: %s does not have the separator '='",
                 line);
+            r = -EINVAL;
             goto err_val;
         }
 
@@ -986,7 +986,11 @@ sol_platform_impl_load_locales(char **locale_cache)
 
         category = system_locale_to_sol_locale(key);
         if (category != SOL_PLATFORM_LOCALE_UNKNOWN) {
-            locale_cache[category] = sol_str_slice_to_string(value);
+            struct sol_buffer buf;
+
+            r = sol_util_escape_quotes(value, &buf);
+            SOL_INT_CHECK_GOTO(r, < 0, err_val);
+            locale_cache[category] = sol_buffer_steal_or_copy(&buf, NULL);
             SOL_NULL_CHECK_GOTO(locale_cache[category], err_nomem);
         }
         free(line);
--- a/src/shared/include/sol-util.h
+++ b/src/shared/include/sol-util.h
@@ -752,6 +752,17 @@ long int sol_util_strtol(const char *nptr, char **endptr, ssize_t len, int base)
 unsigned long int sol_util_strtoul(const char *nptr, char **endptr, ssize_t len, int base);
 
 /**
+ * @brief Escape a string removing quotes from it.
+ *
+ * This function will also remove any ''' and '"' from the slice.
+ *
+ * @param slice The slice to be escaped
+ * @param buf - The buffer to hold the escaped string - It will be initialized by this function.
+ * @return 0 on success, negative errno otherwise.
+ */
+int sol_util_escape_quotes(const struct sol_str_slice slice, struct sol_buffer *buf);
+
+/**
  * @}
  */
 
--- a/src/shared/sol-util.c
+++ b/src/shared/sol-util.c
@@ -905,3 +905,79 @@ sol_util_uuid_str_valid(const char *str)
 
     return true;
 }
+
+SOL_API int
+sol_util_escape_quotes(const struct sol_str_slice slice,
+    struct sol_buffer *buf)
+{
+    bool is_escaped = false;
+    size_t i, last_append;
+    int r;
+
+    SOL_NULL_CHECK(buf, -EINVAL);
+
+    sol_buffer_init(buf);
+
+    if (!slice.len)
+        return 0;
+
+    last_append = 0;
+
+    for (i = 0; i < slice.len; i++) {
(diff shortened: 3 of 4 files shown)


cabelitos:
@barbieri What do you think?

src/lib/common/sol-platform-linux-common.c
```diff
@@ -986,7 +986,11 @@ sol_platform_impl_load_locales(char **locale_cache)
 
         category = system_locale_to_sol_locale(key);
         if (category != SOL_PLATFORM_LOCALE_UNKNOWN) {
-            locale_cache[category] = sol_str_slice_to_string(value);
+            struct sol_buffer buf;
+
+            r = sol_util_escape_quotes(value, &buf);
+            SOL_INT_CHECK_GOTO(r, < 0, err_val);
+            locale_cache[category] = sol_buffer_steal_or_copy(&buf, NULL);
```
barbieri: although you are doing steal-or-copy, make sure you call `sol_buffer_fini()` to keep the pattern. I'm worried we're skipping fini because we know there is no memory left in it to be freed, but this is bad development pattern since people may read our code and use it as example and do mistake in theirs.

This is something we should also review in that batch from you where you moved some APIs to buffer. Just to make it consistent.

src/shared/include/sol-util.h
```diff
@@ -752,6 +752,17 @@ long int sol_util_strtol(const char *nptr, char **endptr, ssize_t len, int base)
...
 /**
+ * @brief Escape a string removing quotes from it.
+ *
+ * This function will also remove any ''' and '"' from the slice.
+ *
+ * @param slice The slice to be escaped
+ * @param buf - The buffer to hold the escaped string - It will be initialized by this function.
+ * @return 0 on success, negative errno otherwise.
+ */
+int sol_util_escape_quotes(const struct sol_str_slice slice, struct sol_buffer *buf);
```
barbieri: if it's removing then it's not "escape", it is "unescape".

src/shared/sol-util.c
```diff
@@ -905,3 +905,79 @@ sol_util_uuid_str_valid(const char *str)
...
+
+    sol_buffer_init(buf);
+
+    if (!slice.len)
+        return 0;
+
+    last_append = 0;
+
+    for (i = 0; i < slice.len; i++) {
+        if (!is_escaped && (slice.data[i] == '"' || slice.data[i] == '\'')) {
```
barbieri: this is not right. If we have quotes, then we should adapt the indexes and pointers so if there is no escape, we can fallback to the `if (!last_append)` case you handle post the loop is finished, that is: no copy, just return a slice of the input buffer without the quotes.

Given the usage I'd handle the following input as tests (please add a test case for those):
 - `x` -> `x`, no copy
 - ` x` -> ` x`, no copy (leading space, keep it)
 - `x ` -> `x `, no copy (trailing space, keep it)
 - `"x"` -> `x`, no copy
 - ` "x"` -> `x`, no copy (leading space, ignore it)
 - `"x" ` -> `x`, no copy (trailing space, ignore it)
 - `"x'` -> error
 - `'x"` -> error
 - `"x` -> error
 - `'x` -> error
 - `x"` -> error
 - `x'` -> error
 - `x"y"` -> `xy` (or error, or copy verbatim, your pick, but document it).
 - `x'y'` -> `xy` (or error, or copy verbatim, your pick, but document it).
 - `"x\"y"` -> `x"y`, copy/unescape
 - `"x\'y"` -> `x'y`, copy/unescape

Also add test to check if it was NOT copied, we need to check if we are handling the optimized cases right.

anyway, as you can see we need to change this loop

src/shared/sol-util.c
```diff
@@ -905,3 +905,79 @@ sol_util_uuid_str_valid(const char *str)
...
+    if (is_escaped) {
+        SOL_WRN("Invalid string format, missing character to be escapted. String: %.*s",
+            SOL_STR_SLICE_PRINT(slice));
+        r = -EINVAL;
+        goto err_exit;
+    }
+
+    if (!last_append) {
+        sol_buffer_init_flags(buf, (char *)slice.data, slice.len,
+            SOL_BUFFER_FLAGS_MEMORY_NOT_OWNED);
```
barbieri: I guess we should set the flag saying we have no nul-byte. (if you based your code in json and it was like this, please fix that as well)

src/shared/sol-util.c
```diff
@@ -905,3 +905,79 @@ sol_util_uuid_str_valid(const char *str)
 
     return true;
 }
+
+SOL_API int
+sol_util_escape_quotes(const struct sol_str_slice slice,
```
barbieri: I think we should change the return to `ssize_t` and return the index in slice to the last used char, so we can use it to process the information further. This is particularly of help if we use the `x"y"` to only get `x`, then user would get the pointer to the first quote and may be able to handle it.

barbieri: and in this case also test the return code in the unit test

src/lib/common/sol-platform-linux-common.c
cabelitos: Sounds fair.

src/shared/include/sol-util.h
cabelitos: typo :/

src/shared/sol-util.c
cabelitos: Ok.

src/shared/sol-util.c
cabelitos: @barbieri 
In this case x'y', I prefer to return xy, instead of just returning x or an error.
Is that ok for you?

src/shared/sol-util.c
cabelitos: I will keep as int, because in the case that you described I will transform x"y" into xy.

src/shared/sol-util.c
cabelitos: I have the patch ready, waiting for your reply so I can send it.

cabelitos:
#1539

PR #1513, opened 2016-03-01
Single node v3

barbieri:
Changes since v2 (#1505):
 - fixed typos reported by @edersondisouza 
 - changed find_ports to be more correct in the handling of the array ports, also printing the error messages via `SOL_DBG()` and added a test, that was missing;
 - renamed and documented the `get_int32_packet()` to make clear it is also logging... you don't nee the 28 lines of code to get a value as reported by Ederson :-) 
 - fixed a bug with `sol_flow_get_node_type()`,  I was incorrectly giving it the full type name, but it expects the module name, so `"wallclock"` instead of `"wallclock/minute"`. It was working since wallclock is a builtin, but should be fixed nonetheless.


barbieri:
@gabrielschulhof, @zolkis and @tripzero: this PR implements the base you should use to wrap node types outside of a flow. There are 2 new samples in `src/samples/flow/c-api`, one shows how to load a type from a string and do some introspection, the other shows how to create the single node (outside the flow), feed it packets and get the outgoing packets in a callback.

barbieri:
merged.

PR #1514, opened 2016-03-01
Io tools

mbelluzzo:
Adding some tools that are helping test our platforms a little bit. It will also help other teams in their tests because they are used with similar tools from another library and need to check that both have the same behavior.

There is room for additional tools and maybe a sol-io <protocol> <options> to consolidate them. It's an interesting entry point for new comers to get in touch with our API and submit a first pullreq, upon merge of this PR I can open task pointing for the missing protocols.

### DIFF
--- a/Kconfig
+++ b/Kconfig
@@ -213,6 +213,7 @@ depends on FEATURE_RUNNABLE_PROGRAMS
 source "src/bin/sol-fbp-generator/Kconfig"
 source "src/bin/sol-fbp-runner/Kconfig"
 source "src/bin/sol-fbp-to-dot/Kconfig"
+source "src/bin/sol-tools/Kconfig"
 endmenu
 
 menu "Test suite"
--- a/src/bin/sol-tools/sol-aio.c
+++ b/src/bin/sol-tools/sol-aio.c
@@ -1,0 +1,125 @@
+/*
+ * This file is part of the Soletta Project
+ *
+ * Copyright (C) 2016 Intel Corporation. All rights reserved.
+ *
+ * Redistribution and use in source and binary forms, with or without
+ * modification, are permitted provided that the following conditions
+ * are met:
+ *
+ *   * Redistributions of source code must retain the above copyright
+ *     notice, this list of conditions and the following disclaimer.
+ *   * Redistributions in binary form must reproduce the above copyright
+ *     notice, this list of conditions and the following disclaimer in
+ *     the documentation and/or other materials provided with the
+ *     distribution.
+ *   * Neither the name of Intel Corporation nor the names of its
+ *     contributors may be used to endorse or promote products derived
+ *     from this software without specific prior written permission.
+ *
+ * THIS SOFTWARE IS PROVIDED BY THE COPYRIGHT HOLDERS AND CONTRIBUTORS
+ * "AS IS" AND ANY EXPRESS OR IMPLIED WARRANTIES, INCLUDING, BUT NOT
+ * LIMITED TO, THE IMPLIED WARRANTIES OF MERCHANTABILITY AND FITNESS FOR
+ * A PARTICULAR PURPOSE ARE DISCLAIMED. IN NO EVENT SHALL THE COPYRIGHT
+ * OWNER OR CONTRIBUTORS BE LIABLE FOR ANY DIRECT, INDIRECT, INCIDENTAL,
+ * SPECIAL, EXEMPLARY, OR CONSEQUENTIAL DAMAGES (INCLUDING, BUT NOT
+ * LIMITED TO, PROCUREMENT OF SUBSTITUTE GOODS OR SERVICES; LOSS OF USE,
+ * DATA, OR PROFITS; OR BUSINESS INTERRUPTION) HOWEVER CAUSED AND ON ANY
+ * THEORY OF LIABILITY, WHETHER IN CONTRACT, STRICT LIABILITY, OR TORT
+ * (INCLUDING NEGLIGENCE OR OTHERWISE) ARISING IN ANY WAY OUT OF THE USE
+ * OF THIS SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
+ */
+
+#include <signal.h>
+#include "sol-aio.h"
+#include "sol-log.h"
+#include "sol-mainloop.h"
+
+#include <stdbool.h>
+#include <stdio.h>
+#include <string.h>
+
+struct sol_aio *aio = NULL;
+struct sol_aio_pending *pending = NULL;
+struct sol_timeout *timer = NULL;
+
+static void
+sigint_handler(int signo)
+{
+    if (signo == SIGINT)
+        sol_quit();
+}
+
+static void
+read_cb(void *cb_data, struct sol_aio *_aio, int32_t ret)
+{
+    pending = NULL;
+
+    if (ret < 0)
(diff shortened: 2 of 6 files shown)


Kconfig
```diff
@@ -213,6 +213,7 @@ depends on FEATURE_RUNNABLE_PROGRAMS
 source "src/bin/sol-fbp-generator/Kconfig"
 source "src/bin/sol-fbp-runner/Kconfig"
 source "src/bin/sol-fbp-to-dot/Kconfig"
+source "src/bin/sol-tools/Kconfig"
```
barbieri: sol-io-tools, one tool per directory so it matches the other binaries.

src/bin/sol-tools/sol-aio.c
```diff
@@ -0,0 +1,125 @@
...
+
+struct sol_aio *aio = NULL;
+struct sol_aio_pending *pending = NULL;
+struct sol_timeout *timer = NULL;
+
+static void
+sigint_handler(int signo)
+{
+    if (signo == SIGINT)
+        sol_quit();
```
barbieri: this should be automatic using the mainloop

src/bin/sol-tools/sol-aio.c
```diff
@@ -0,0 +1,125 @@
...
+ * DATA, OR PROFITS; OR BUSINESS INTERRUPTION) HOWEVER CAUSED AND ON ANY
+ * THEORY OF LIABILITY, WHETHER IN CONTRACT, STRICT LIABILITY, OR TORT
+ * (INCLUDING NEGLIGENCE OR OTHERWISE) ARISING IN ANY WAY OUT OF THE USE
+ * OF THIS SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
+ */
+
+#include <signal.h>
+#include "sol-aio.h"
+#include "sol-log.h"
+#include "sol-mainloop.h"
```
barbieri: after @ibriano's commit we should use `soletta.h`

src/bin/sol-tools/sol-aio.c
```diff
@@ -0,0 +1,125 @@
...
+        sol_quit();
+}
+
+static void
+read_cb(void *cb_data, struct sol_aio *_aio, int32_t ret)
+{
+    pending = NULL;
+
+    if (ret < 0)
+        fprintf(stderr, "%s\n", strerror(-ret));
```
barbieri: `sol_util_strerrora()`. also prefix error messages with `"ERROR:"`

barbieri: actually "error and what failed, in this case reading from aio

src/bin/sol-tools/sol-aio.c
```diff
@@ -0,0 +1,125 @@
...
+
+static bool
+on_timeout(void *data)
+{
+    if (sol_aio_busy(aio))
+        return true;
+
+    pending = sol_aio_get_value(aio, read_cb, NULL);
+    if (!pending) {
+        fprintf(stderr, "Failed to request read operation.");
```
barbieri: prefix with ERROR, say which aio failed, add missing "\n"

src/bin/sol-tools/sol-aio.c
```diff
@@ -0,0 +1,125 @@
...
+static void
+usage(const char *program)
+{
+     fprintf(stdout,
+        "Usage: %s [device] [pin]]\n",
+        program);
+}
+
+int
+main(int argc, char *argv[])
```
barbieri: don't go main direcly, instead use the MACROS, from the startup function you process things. If there is no need to wait you can call `sol_quit()` directly from that function so it behaves like procedural programs.

src/bin/sol-tools/sol-gpio.c
```diff
@@ -0,0 +1,153 @@
...
+    if (signo == SIGINT)
+        sol_quit();
+}
+
+static void
+get_cmd(uint32_t pin)
+{
+    int ret;
+
+    SOL_SET_API_VERSION(config.api_version = SOL_GPIO_CONFIG_API_VERSION; )
```
barbieri: put the `;` outside of the parenthesis, this way works but will cause indendation errors on some editors.

src/bin/sol-tools/sol-gpio.c
```diff
@@ -0,0 +1,153 @@
...
+    SOL_NULL_CHECK(gpio);
+
+    sol_gpio_close(gpio);
+}
+
+static void
+usage(const char *program)
+{
+     fprintf(stdout,
+        "Usage: %s [[get | monitor] [pin] | set [pin] [value]]\n",
```
barbieri: expand this to multiple example lines, one per each subcommand.

src/bin/sol-tools/sol-aio.c
mbelluzzo: Thats funny, while I as typing this I thought: "Shit, we should have a single include". Is it in already?

Kconfig
mbelluzzo: "sol-io-tools" and "one tools per directory" doesn't compute. Or you mean inside "sol-io-tools"? Anyway this bunch o single filed folders will just become noise. And my plan is to have a aggregator binary for the IO tools, like to run sol-gpio one could also run

sol-io gpio <gpio options>
sol-io aio <aio options>

And so on.

src/bin/sol-tools/sol-aio.c
mbelluzzo: Nice!

src/bin/sol-tools/sol-aio.c
mbelluzzo: "Which aio" will always be the one that the guy typed in the command line, I don't think it is relevant in this case.

Kconfig
barbieri: your call... if you want 2 binaries, then we have `src/bin/sol-gpio` and `src/bin/sol-aio`... or we have a single binary with multiple subcommands.

src/bin/sol-tools/sol-aio.c
barbieri: it is in, but it is not a single include, it is the entry point (what used to be sol-mainloop.h). It is cleaner... however it breaks the pattern of all our files called "sol-*.h"

In any way, you didn't notice it because it is where the macros you did not use are defined :-) Once you add the macros you'll need to use that file, then you can remove the mainloop.h

src/bin/sol-tools/sol-aio.c
barbieri: better safe than sorry. You'll get a bug report that will have a single line pasted... then you'll need to ask the command line he executed, etc :-/

src/bin/sol-tools/sol-aio.c
mbelluzzo: When I was doing this I checked the SOL_MAIN and SOL_MAIN_DEFAULT and end-up being a worse code so I switch back to int main on purpose. I think it was because I did the gpio version first and needed to avoid the mainloop in some cases and then I just copied in here. But for aio that is always "monitor" mode, the macro should fit.

src/bin/sol-tools/sol-gpio.c
mbelluzzo: :+1:

src/bin/sol-tools/sol-gpio.c
mbelluzzo: :+1:

mbelluzzo:
Closing in favor of #1537

PR #1518, opened 2016-03-02
oic: Use correct response code for packets with content

otaviobp:
Correct code used in responses with payload content is
SOL_COAP_RSPCODE_CONTENT and not SOL_COAP_RSPCODE_OK


otaviobp:
merged

PR #1520, opened 2016-03-03
Nodejs platform monitors v2

gabrielschulhof:
Replaces #1507.

* Add invasive tests which need to be run separately, but which test the functionality of the main loop integration and some of the platform monitors
* Use ```errno``` values for hijacking
* Use ```uint16_t``` refcounter
* Reduce casts in C code


bindings/nodejs/src/hijack.cc
```diff
@@ -0,0 +1,93 @@
...
+#include "hijack.h"
+
+using namespace v8;
+
+static uint16_t hijack_refcount = 0;
+
+bool hijack_ref() {
+    int result = 0;
+
+    SOL_DBG("Entering");
```
brunobottazzini: Is this "Entering" message needed ? If yes, could you specify more what it is going on ?

bindings/nodejs/src/hijack.cc
```diff
@@ -0,0 +1,93 @@
...
+error:
+    Nan::ThrowError((std::string("Hijack main loop: ") +
+        strerror(-result)).c_str());
+    return false;
+}
+
+bool hijack_unref() {
+    int result = 0;
+
+    SOL_DBG("Entering\n");
```
brunobottazzini: The \n is not needed. The macro SOL_DBG puts it to us.  The same for this ¨Entering" message

bindings/nodejs/src/sol-uv-integration.c
```diff
@@ -0,0 +1,212 @@
...
+
+    SOL_DBG("Entering with state %s", RESOLVE_MAINLOOP_STATE(mainloopState));
+    if (mainloopState == MAINLOOP_RELEASED ||
+        mainloopState == MAINLOOP_RELEASING_STARTED) {
+        return returnValue;
+    }
+
+    // hijack_main_loop() was called, but the idler has not yet run
+    if (mainloopState == MAINLOOP_HIJACKING_STARTED) {
+        SOL_DBG("idler has not yet run, so stopping it");
```
brunobottazzini: s/yet run/run yet/

It should be in the end: 
"Idler has run yet, so stopping it"

bindings/nodejs/tests/invasive-tests/Main Loop Integration.js
```diff
@@ -0,0 +1,102 @@
...
+		} );
+	},
+
+	function testHostnameMonitor( callback ) {
+		hostnameMonitor = soletta.sol_platform_add_hostname_monitor( function( newHostname ) {
+			utils.assert( "strictEqual", newHostname, initialHostname,
+				"The new hostname is as expected" );
+			callback();
+		} );
+		exec( "sudo hostname " + initialHostname, function( error ) {
```
brunobottazzini: Do you have to be root, to run this command?

bindings/nodejs/src/hijack.cc
gabrielschulhof: Dang! I new I missed one :)

bindings/nodejs/tests/invasive-tests/Main Loop Integration.js
gabrielschulhof:
```
hostname: you must be root to change the host name
```

bindings/nodejs/src/sol-uv-integration.c
edersondisouza: On 03-03-2016 10:34, Bruno Bottazzini wrote:
> In bindings/nodejs/src/sol-uv-integration.c
> <https://github.com/solettaproject/soletta/pull/1520#discussion_r54879434>:
>
>> +}
>> +
>> +int
>> +release_main_loop()
>> +{
>> +    int returnValue = 0;
>> +
>> +    SOL_DBG("Entering with state %s", RESOLVE_MAINLOOP_STATE(mainloopState));
>> +    if (mainloopState == MAINLOOP_RELEASED ||
>> +        mainloopState == MAINLOOP_RELEASING_STARTED) {
>> +        return returnValue;
>> +    }
>> +
>> +    // hijack_main_loop() was called, but the idler has not yet run
>> +    if (mainloopState == MAINLOOP_HIJACKING_STARTED) {
>> +        SOL_DBG("idler has not yet run, so stopping it");
>
> s/yet run/run yet/
>
> It should be in the end:
> "Idler has run yet, so stopping it"

?
This sound weird... With the change you suggests, message would have a 
different meaning from current message or code comment...

brunobottazzini: oh yeah i forgot to add the "not"... lol
Sorry...
"Idler has not run yet, so stopping it"

gabrielschulhof: I will correct it as "Idler has **not** run yet, so stopping it" :)

brunobottazzini: :+1: thanks

bindings/nodejs/tests/invasive-tests/Main Loop Integration.js
brunobottazzini: yeah you are right

bindings/nodejs/src/hijack.cc
gabrielschulhof: The unconditionally printed "Entering" message is necessary for getting a complete picture as to the behaviour of the main loop hijacking. All I really need to see at that point is the name of the function and that it's entering, so I can't really think of anything else to put in that message.

bindings/nodejs/src/sol-uv-integration.c
```diff
@@ -0,0 +1,212 @@
...
+    return returnValue;
+}
+
+static bool
+uv_loop_source_get_next_timeout(void *data,
+    struct timespec *timeout)
+{
+    int t = uv_backend_timeout(data);
+
+    SOL_DBG("t = %d", t);
```
brunobottazzini: The var "t" looks too generic. Why not: "timeout = %D ms"

Also, If this function returns -1, do you have to make all this math with it ? Or, can you just return -1 ?

bindings/nodejs/src/hijack.cc
brunobottazzini: What about: "starting hijack reference" ? Or would this be pleonasm?

gabrielschulhof: Yeah, since the name of the function, ```hijack_ref()```, does appear in the debug statement.

bindings/nodejs/src/sol-uv-integration.c
gabrielschulhof: I can change the function to not do the math if it's going to return false anyway.

brunobottazzini: Yes just returns false I guess it is worth it

bindings/nodejs/src/hijack.cc
gabrielschulhof: "commencing preparations for potentially incrementing the main loop hijacking reference counter" :D

PR #1524, opened 2016-03-03
Nodejs platform monitors v3

gabrielschulhof:
Replaces #1520


bindings/nodejs/src/functions/sol-platform-monitors.cc
```diff
@@ -0,0 +1,294 @@
...
+
+class SolPlatformMonitor: public JSHandle<SolPlatformMonitor> {
+public:
+	static const char *jsClassName() { return "SolPlatformMonitor"; }
+};
+
+#define ADD_MONITOR(name, marshaller) \
+	do { \
+    	VALIDATE_ARGUMENT_COUNT(info, 1); \
+    	VALIDATE_ARGUMENT_TYPE(info, 0, IsFunction); \
```
brunobottazzini: Just silly question here, where do these macros come from ?

package.json
```diff
@@ -29,6 +29,7 @@
     "async": "^1.5.2",
     "glob": "^6.0.4",
     "lodash": "^4.3.0",
+    "optimist": "^0.6.1",
```
brunobottazzini: I had a quick looked at this nodejs module and the author says it was deprecated. He said that he is using minimist.
Wouldn't it be better using a non deprecated lib ?

Reference: https://github.com/substack/node-optimist

bindings/nodejs/src/functions/sol-platform-monitors.cc
gabrielschulhof: Basically, all platform monitors except the service monitor have the same body. There's no point copying the same code over, like, eight times. I can't use a static function either, because each service monitor has a different signature, and I want to avoid casting them all to a common signature as C++ would require me to do. Besides, the ```info``` parameter is a variable whose type can vary from node version to node version, so I can't really pass it to a function. ```NAN_METHOD(some_binding)``` declares a function for me that accepts this parameter, but it never tells me its type.

package.json
gabrielschulhof: Thanks for checking!

bindings/nodejs/src/functions/sol-platform-monitors.cc
brunobottazzini: Good to know thanks

PR #1526, opened 2016-03-03
Server "no change" fixes

barbieri:
This PR was born in order to make the last commit with a sample work. The idea is that to persist or synchronize two different resources we need to cross-link and if the nodes propagate all packets, including when the state did not change, then we enter infinite loop.


data/scripts/sol-oic-gen.py
```diff
@@ -324,48 +334,60 @@ def get_field_enum_client_c(id, struct_name, name, prop):
...
-        id += 1
+            fields.append(type_to_fn[prop['type']](id, prop_name, prop))
     return '\n'.join(fields)
 
 def generate_object_from_repr_vec_fn_common_c(name, props):
     fields_init = []
     for field_name, field_props in props.items():
         if 'enum' in field_props:
-            fields_init.append('.%s = state->%s,' % (field_name, field_name))
+            fields_init.append('        .%(name)s = state->%(name)s,' % {"name": field_name})
```
lpereira: No need to indent the generated code. If `indent` is installed, the whole file will be indented automatically so it can be manually inspected. This simplifies the code generator.

src/samples/flow/oic-and-http-light/Kconfig
```diff
@@ -0,0 +1,4 @@
+config FLOW_OIC_AND_HTTP_LIGHT_SAMPLE
+        bool
+	depends on FLOW_FBP_GENERATOR_SAMPLES && FLOW_NODE_TYPE_OIC && FLOW_NODE_TYPE_HTTP_CLIENT && FLOW_NODE_TYPE_HTTP_SERVER && FLOW_NODE_TYPE_CONSOLE && FLOW_NODE_TYPE_TIMER && FLOW_NODE_TYPE_APP && FLOW_NODE_TYPE_CONSTANT && FLOW_NODE_TYPE_CONVERTER && FLOW_NODE_TYPE_PLATFORM
+	default y
```
brunobottazzini: The bool here is not aligned

data/scripts/sol-oic-gen.py
```diff
@@ -1560,7 +1595,42 @@ def master_c_as_string(generated, oic_gen_c, oic_gen_h):
     return client_connect(resource, device_id);
 }
 
-#define RETURN_ERROR(errcode) do { goto out; } while(0)
+static inline bool
+check_updated_string(const char *a, const char *b)
+{
+    if (a && b)
+        return strcmp(a, b) != 0;
+    else if ((a && !b) || (!a && b))
```
lpereira: return (a && b) ? strcmp(a, b) : a || b;

barbieri: I don't like this kind of constructs, so will keep the long version :-)

barbieri:
Fixes in #1551

PR #1527, opened 2016-03-03
oic: Add implementation of an Ownership Tranfer Method: Just Works - v2

otaviobp:
Changelog from v1:
 - minor reported issues fixed (memory leaks, logs, style, etc)
 - Add timeout to disable anonymous ECDH key if pairing hasn't completed.
 - Don't accept requests that are not related to pairing if anonymous ECDH key is enabled.
 - get_machine_id exported in platform as sol_platform_get_machine_id_bytes
 - Use buffer clear feature instead of calling sol_util_secure_clear_memory when possible.
 - Load default doxm if only pstat is loaded and vice-versa.

Problems not fixed so far:
 - TLS_NULL_WITH_NULL_NULL is still being used as Iotivity 1.0.1. Iotivity master is not using this cipher anymore. Working on a fix for that.
 - I didn't fixed comments about sol_socket_dtls_close because in a way it is related to the cipher problem.

Replaces #1493

### DIFF
--- a/src/lib/common/include/sol-platform.h
+++ b/src/lib/common/include/sol-platform.h
@@ -87,9 +87,10 @@ extern "C" {
 const char *sol_platform_get_board_name(void);
 
 /**
- * @brief Retrieves, in @a id, the machine-id present in the file system.
+ * @brief Retrieves the machine-id present in the file system as a string.
  *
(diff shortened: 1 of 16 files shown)


otaviobp:
merging only commits with positive review.

PR #1531, opened 2016-03-03
coap: change it to handle payload in sol_buffers.

glima:
This is the 1st iteration of trying to use less memory on CoAP buffers.
sol_coap will now, on packet creation, use a sol_buffer to handle all
the payload, and not allocate a blob with COAP_UDP_MTU bytes right away
anymore. This way, the user is able to append on that buffer to make the
user payload, so that less data hangs around at least on the sending
side. All places where a minimum amount of sending bytes could be
assured now do so.

We still have to think on how to proceed on some of the OIC/cbor paths,
since the involve users callbacks appending cbor data to a packet's
buffer (and issuing the callback *twice* just to have the cbor
phony-size-calculating calls to take place would be bad). At the
receiving side, we still allocate COAP_UDP_MTU bytes to the buffer. We
can make this better from here as well. On Linux, at least, we have the
MSG_TRUNC recvmsg() flag.

Also, the next iteration of this work will be to change de
sol_coap_packet_new() to receive a possible user-allocated buffer to use
as payload (at the receiving side this could be used as well). So, on
small OSs the direct network buffers could be passed around -- we'll
surely evolve from here.

----

This was tested extensively, but of couse more testing is welcome. @otaviobp, would you help by exercising this on all your tests too? I will now check how do we perform memory-wise with the lwm2m RIOT's sample app.

### DIFF
--- a/src/lib/comms/coap.c
+++ b/src/lib/comms/coap.c
@@ -97,51 +97,53 @@ decode_delta(int num, const uint8_t *buf, int16_t buflen, uint16_t *decoded)
 }
 
 int
-coap_parse_option(const struct sol_coap_packet *pkt, struct option_context *context,
+coap_parse_option(struct option_context *context,
     uint8_t **value, uint16_t *vlen)
 {
     uint16_t delta, len;
     int r;
 
-    if (context->buflen < 1)
+    if (context->buf->used - context->pos < 1)
         return 0;
 
     /* This indicates that options have ended */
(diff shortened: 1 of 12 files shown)


glima:
Doxygen has been fixed locally as well.

glima:
All issues fixed locally (not bothering with a v2 yet, they are too small)

glima:
Sending a v2.

PR #1532, opened 2016-03-04
Enable Soletta running over Zephyr's nanokernel

ibriano:
Also wake up the mainloop when timeouts/idlers are added.

### DIFF
--- a/data/jsons/dependencies.json
+++ b/data/jsons/dependencies.json
@@ -273,10 +273,15 @@
       ]
     },
     {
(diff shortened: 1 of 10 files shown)


glima:
Overall, looking good.

glima:
Style issue pending

vcgomes:
Looks good.

For reference, after discussing it offline, it became clear that my example of adding an idler from an interrupt context can not be expected to work, waiting for the mainloop lock to become available is a "no no" in an interrupt context.

PR #1534, opened 2016-03-04
Flow: Use the correct expression to calculate the option padding.

cabelitos:
The old one could lead to unaligned memory access.


barbieri:
+1

cabelitos:
merged

PR #1537, opened 2016-03-04
Io tools v2

mbelluzzo:
Changes from v1:
sol_main_default in aio
fix the issues pointed by k-s

### DIFF
--- a/src/bin/sol-gpio/sol-gpio.c
+++ b/src/bin/sol-gpio/sol-gpio.c
@@ -1,0 +1,146 @@
+/*
+ * This file is part of the Soletta Project
+ *
+ * Copyright (C) 2016 Intel Corporation. All rights reserved.
+ *
+ * Redistribution and use in source and binary forms, with or without
+ * modification, are permitted provided that the following conditions
+ * are met:
+ *
+ *   * Redistributions of source code must retain the above copyright
+ *     notice, this list of conditions and the following disclaimer.
+ *   * Redistributions in binary form must reproduce the above copyright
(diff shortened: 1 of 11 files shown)


src/bin/sol-gpio/sol-gpio.c
```diff
@@ -0,0 +1,146 @@
...
+    fprintf(stdout,
+        "Usage:\n"
+        "   %s set [pin] [value]\n"
+        "   %s get [pin]]\n"
+        "   %s monitor [pin]\n",
+        program, program, program);
+}
+
+int
+main(int argc, char *argv[])
```
barbieri: no main(), use the macro.

src/modules/pin-mux/intel-edison-rev-c/intel-edison-rev-c.c
```diff
@@ -399,39 +414,68 @@ static struct mux_pin_map pin_map[] = {
 };
 
 // =============================================================================
+static bool ardu_breakout = false;
+
+static int
+_mux_init(void)
+{
+    //Try to detect the Arduino Breakout
+    if (sol_util_write_file("/sys/class/gpio/export", "214") >= 0) {
```
barbieri: shouldn't we make this another board? The problem is if we can export 214 or not (the board-detect can't write, just read). can't we have another way to do it?

src/bin/sol-gpio/sol-gpio.c
mbelluzzo: our macros always run the mainloop and for this tools we want to avoid that (except monitor mode).

PR #1538, opened 2016-03-04
LWM2M and Coap changes v2

cabelitos:
Changes since v1:

* sol_coap_server_register_resource return an int now

### DIFF
--- a/src/samples/coap/simple-server.c
+++ b/src/samples/coap/simple-server.c
@@ -237,7 +237,7 @@ main(int argc, char *argv[])
         return -1;
     }
 
-    if (!sol_coap_server_register_resource(server, &light, NULL)) {
+    if (sol_coap_server_register_resource(server, &light, NULL) < 0) {
         SOL_WRN("Could not register resource for the light");
         return -1;
     }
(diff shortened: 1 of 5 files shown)


src/samples/coap/simple-server.c
```diff
@@ -237,7 +237,7 @@ main(int argc, char *argv[])
         return -1;
     }
 
-    if (!sol_coap_server_register_resource(server, &light, NULL)) {
+    if (sol_coap_server_register_resource(server, &light, NULL) < 0) {
         SOL_WRN("Could not register resource for the light");
```
brunobottazzini: I know this is not in the patch but wouldn't this be better ? 

s/Could not register resource for the light/Could not register light resource/g

brunobottazzini:
LGTM :+1:

src/samples/coap/simple-server.c
cabelitos: I can change that. NP

glima:
+1

cabelitos:
Merging this one.

PR #1539, opened 2016-03-04
Sol-platform: Remove the '\"' and '\'' characters from the locale v4

cabelitos:
Changes since v3:
* Properly unescape the locale string


barbieri:
Ederson, your example should be possible, yes. As well as `"I'm good"`, given that you have the outter quotes.

@cabelitos would you check that?

cabelitos:
#1575

PR #1542, opened 2016-03-07
oic-samples: Some small improvements.

otaviobp:


glima:
OK, for me.

cabelitos:
+1

otaviobp:
Answering ceolin. Before this patch I used NULL resource for scanning and core.light for everything else. With this patch we use the argument value, if informed, for everything, even scanning. I think this is a better approach to make easier to test scanning with filters.
If I have a default value, it would be hard to test the sample using NULL resource type.

otaviobp:
merged and closed.

PR #1544, opened 2016-03-07
coap-tests: link with sol-socket.o too.

glima:
At least on Fedora, we get complaints of missing sol_socket_setsockopt()
without this.


glima:
I'll have to spend some more time on this, it's very wrong indeed

glima:
v2 coming.

PR #1546, opened 2016-03-07
comms:lwm2m: Fix return value

ceolin:
sol_lwm2m_notify_observers should return negative errno in case of
error.


brunobottazzini:
+2

ceolin:
merged.

PR #1548, opened 2016-03-07
Io tools v3

mbelluzzo:
Changes from v2:
- Add missing return
- Add mission \n
- delete the timer
- wont use main macro for gpio, I dont want to run the mainloop

### DIFF
--- a/src/bin/sol-gpio/sol-gpio.c
+++ b/src/bin/sol-gpio/sol-gpio.c
@@ -1,0 +1,146 @@
+/*
+ * This file is part of the Soletta Project
+ *
+ * Copyright (C) 2016 Intel Corporation. All rights reserved.
+ *
+ * Redistribution and use in source and binary forms, with or without
+ * modification, are permitted provided that the following conditions
+ * are met:
+ *
+ *   * Redistributions of source code must retain the above copyright
+ *     notice, this list of conditions and the following disclaimer.
+ *   * Redistributions in binary form must reproduce the above copyright
+ *     notice, this list of conditions and the following disclaimer in
+ *     the documentation and/or other materials provided with the
+ *     distribution.
+ *   * Neither the name of Intel Corporation nor the names of its
+ *     contributors may be used to endorse or promote products derived
+ *     from this software without specific prior written permission.
+ *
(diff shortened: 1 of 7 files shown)


src/bin/sol-gpio/sol-gpio.c
```diff
@@ -0,0 +1,146 @@
...
+    fprintf(stdout,
+        "Usage:\n"
+        "   %s set [pin] [value]\n"
+        "   %s get [pin]]\n"
+        "   %s monitor [pin]\n",
+        program, program, program);
+}
+
+int
+main(int argc, char *argv[])
```
barbieri: it doesn't matter the comment, the macro is to be used.

If you don't want to run the mainloop just call `sol_quit()` from `startup()`.

ibriano: On Mon, 07 Mar 2016 13:12:26 -0800, Gustavo Sverzut Barbieri wrote:
> > +    sol_gpio_close(gpio);
> > +}
> > +
> > +static void
> > +usage(const char *program)
> > +{
> > +    fprintf(stdout,
> > +        "Usage:\n"
> > +        "   %s set [pin] [value]\n"
> > +        "   %s get [pin]]\n"
> > +        "   %s monitor [pin]\n",
> > +        program, program, program);
> > +}
> > +
> > +int
> > +main(int argc, char *argv[])
> 
> it doesn't matter the comment, the macro is to be used.
> 
> If you don't want to run the mainloop just call `sol_quit()` from `startup()`.

That doesn't work, if you call sol_quit() before the loop is running, it
complains and does nothing else.
Also, from what I remember, the main loop implementations will check if
they have to quit and maybe do nothing of what the tool does (no idea
what that is, didn't check the code). It'd need an idler to call quit
from it.

barbieri: `startup()` is called from the mainloop, so it works

mbelluzzo:
wow great, implement a startup function that just call sol_quit() just to use a macro ... great!

barbieri:
You have the monitor in there, so instead of jumping over it, you quit. Not much different and yet provided us with uniformity and allows the macro to be changed without us having to remember to fix this example

barbieri:
ping? Fix what I said and merge plz

bdilly:
merged with proposed changes

PR #1549, opened 2016-03-07
glima	coap: change it to handle payload in sol_buffers [v3]

glima:
Changes since #1540:
- finally removed the submodule change
- suggestions by @vcgomes done
- new patch handling exact buffer size at the receiving side for coap (linux & riot)

----

The Zephyr part of the second patch is locally written and should be fine, for the record.

### DIFF
--- a/src/lib/comms/coap.c
+++ b/src/lib/comms/coap.c
@@ -97,51 +97,53 @@ decode_delta(int num, const uint8_t *buf, int16_t buflen, uint16_t *decoded)
 }
 
 int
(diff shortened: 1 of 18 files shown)


glima:
All above were addressed. There's a plan to fix the sol_oic_packet_cbor* API WRT buffer-awareness in the sequence. With the code above, a previous instant-crash sample execution on RIOT with Soletta + LWM2M runs now for minutes without issues.

vcgomes:
Looks good. :+1:

glima:
Cool, merging.

PR #1550, opened 2016-03-07
coap-tests: no reason to link against sol-socket.o [v2]

glima:
sol-coap.o is there because that API is not exported on the final lib.


cabelitos:
+1

glima:
Thanks, merging

PR #1551, opened 2016-03-07
Server no change fixes v2

barbieri:
Changes since v1 (#1526):
 - fixed typo in Makefile so binaries are going to be generated (thanks @cabelitos)
 - fixed typo in FBP options, now generator works (thanks @cabelitos)
 - fixed Kconfig alignment (thanks @brunobottazzini)
 - improved documentations to say when packets are emitted or not (thanks @glima)
 - using `sol_drange_equal()` and `sol_irange_equal()` to be consistent (thanks @edersondisouza)
 - fixed http-server to do not emit packets if it did not changed
 - improve handling of HTTP errors, first check status, then `used`, otherwise we may not report error codes such as not-modified since they come without a payload.


barbieri:
fixes in #1573

PR #1568, opened 2016-03-08
Nodejs oic bindings

gabrielschulhof:
Also cleans up whitespace in Node.js bindings and adds invalidation of JS objects that wrap native pointers.

### DIFF
--- a/bindings/nodejs/generate-main.awk
+++ b/bindings/nodejs/generate-main.awk
@@ -1,0 +1,91 @@
+# This file is part of the Soletta Project
+#
+# Copyright (C) 2015 Intel Corporation. All rights reserved.
+#
+# Redistribution and use in source and binary forms, with or without
+# modification, are permitted provided that the following conditions
+# are met:
+#
+#   * Redistributions of source code must retain the above copyright
+#     notice, this list of conditions and the following disclaimer.
+#   * Redistributions in binary form must reproduce the above copyright
+#     notice, this list of conditions and the following disclaimer in
+#     the documentation and/or other materials provided with the
+#     distribution.
+#   * Neither the name of Intel Corporation nor the names of its
+#     contributors may be used to endorse or promote products derived
+#     from this software without specific prior written permission.
+#
+# THIS SOFTWARE IS PROVIDED BY THE COPYRIGHT HOLDERS AND CONTRIBUTORS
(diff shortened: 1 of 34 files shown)


gabrielschulhof:
I've modified the semaphore to build OIC, but I've kept the ```sed``` that removes DTLS from the configuration.

barbieri:
OIC is under big changes to cope with different transports and some optimizations.

@ceolin, @glima and @otaviobp would you hint @gabrielschulhof on what may impact him?

gabrielschulhof:
```
sol_oic_client_find_resource(client, NULL, "", resourceFound, NULL);
```
would be nice. Aside from that, the contents of ```sol_network_link_addr``` will probably fluctuate, and I'll have to keep up.

gabrielschulhof:
If ```sol_oic_client_find_resource()``` accepted a ```NULL``` pointer for the ```struct sol_network_link_addr *``` parameter, I could treat ```struct sol_network_link_addr *``` pointers as opaque and simply wrap them in JS objects when I get them from soletta.

gabrielschulhof:
That's how ```iotivity-node``` works.

gabrielschulhof:
... in that the destination can be ```NULL```, I mean.

gabrielschulhof:
Another idea: If ```struct sol_network_link_addr``` were a shallow structure, meaning it contained no pointers to other structures, I could also copy it by value and maintain the copies myself, like so:

```
struct sol_network_link_addr *destination = malloc(sizeof(
struct sol_network_link_addr));
*destination = *pointer_from_soletta;
```

Then, when the contents of the structure settled down, I could open it up and expose its properties to JS.

ceolin:
@barbieri  @gabrielschulhof I'm changing **COAP** API separating the transport from the coap logic. It should not affect OIC API that much.

gabrielschulhof:
This PR provides node.js bindings for ```sol_oic_client_*()``` and ```sol_oic_server*()``` as well as ```sol_network_addr_to_str()```.

bindings/nodejs/generate-main.awk
```diff
@@ -0,0 +1,91 @@
+# This file is part of the Soletta Project
+#
+# Copyright (C) 2015 Intel Corporation. All rights reserved.
```
zqzhang: `s/2015/2016/g` for all new added files

PR #1573, opened 2016-03-08
Server no change fixes v3

barbieri:
Changes since v2 (#1551):
 - fixed `make check-fbp` as spotted by @edersondisouza 
 - fixed `Kconfig` to depend on persistence


edersondisouza:
lol this doxygen error...

barbieri:
the doxygen error is weird, it was interpreting `user@host` trying to get the `@host` as a command... I've changed it, although doesn't happen in my machine (maybe newer doxygen?)

change the http-client to do both GET and POST and merged.

PR #1576, opened 2016-03-09
coap: Limit packet size to COAP_UDP_MTU when reading from socket

otaviobp:
When calling sol_socket_recvmsg with empty buffer, socket will return
the size of packet read. This can be larger than COAP_UDP_MTU when not
using a real udp socket (tinydtls socket wrapper, for example).

This patch also fixes sol_socket_dtls_recvmsg behavior when calling it
with empty buffer.

### DIFF
--- a/src/lib/comms/sol-coap.c
+++ b/src/lib/comms/sol-coap.c
@@ -1430,7 +1430,7 @@ on_can_read(void *data, struct sol_socket *s)
      * have to change on those cases. */
 
     /* calculate exact needed size */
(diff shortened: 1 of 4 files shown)


otaviobp:
@glima Can you test that? It is related to the pr you merged. There was a problem with dtls that I didn't realize at the time

otaviobp:
I thought that we would fail if we have a packet that is larger than COAP_UDP_MTU, that's why I did this patch. My bad.
I'll close and send another patch, only fixing the dtls socket behavior.

PR #1577, opened 2016-03-09
HTTP-Server: Add option to set allowed HTTP methods.

cabelitos:

### DIFF
--- a/src/modules/flow/http-server/http-server.c
+++ b/src/modules/flow/http-server/http-server.c
@@ -46,6 +46,7 @@
 #include "sol-util-internal.h"
 #include "sol-vector.h"
 #include "sol-flow-internal.h"
+#include "sol-str-table.h"
 
 #define HTTP_HEADER_ACCEPT "Accept"
 #define HTTP_HEADER_CONTENT_TYPE "Content-Type"
@@ -71,6 +72,7 @@ struct http_data {
     struct server_data *sdata;
     char *path;
     char *basename;
+    enum sol_http_method *allowed_methods;
 };
 
 struct http_server_node_type {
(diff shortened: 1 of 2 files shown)


src/modules/flow/http-server/http-server.c
```diff
@@ -292,6 +316,66 @@ common_close(struct sol_flow_node *node, void *data)
...
+        SOL_WRN("Allowed HTTP methods is empty");
+        return -EINVAL;
+    }
+
+    tokens = sol_str_slice_split(sol_str_slice_from_str(allowed_methods),
+        "|", 0);
+    SOL_INT_CHECK_GOTO(tokens.len, < 1, err_exit);
+
+    *methods_array = calloc(tokens.len + 1, sizeof(enum sol_http_method));
+    SOL_NULL_CHECK_GOTO(*methods_array, err_nomem_exit);
```
barbieri: argh... I don't like the split and this array, adds memory allocations for no good reason. Parse using strchr and slices, then convert to enum and then use a bitmask. Somehow like:

```c
uint8_t allowed = 0;
while (*allowed_methods) {
    const char *ptr = strchr(allowed_methods, '|');
    struct sol_str_slice slice;

    if (!ptr) ptr = allowed_methods + strlen(allowed_methods);
    slice.data = allowed_methods;
    slice.len = ptr - allowed_methods;

    if (sol_str_slice_eq(slice, "GET"))
        allowed |= (1 << SOL_HTTP_METHOD_GET);
    else if (sol_str_slice_eq(slice, "POST"))
        allowed |= (1 << SOL_HTTP_METHOD_POST);
    else SOL_WRN("Unsupported allowed_method=%.*s", SOL_STR_SLICE_PRINT(slice));

    if (*ptr == '|')
       allowed_methods = endptr + 1;
   else break;
}
```

Then checks are:

```c
if (mdata->allowed & (1 << method)) { ...
```

cabelitos: Fair enought.

cabelitos: Indeed, your solution is much better. I will change the code. thanks.

cabelitos:
#1580

PR #1578, opened 2016-03-09
dtls: Don't crash when calling recvmsg to calculate packet size

otaviobp:
sol_socket_dtls_recvmsg was crashing if buf or cliaddr were NULL and
this is used to calculate the packet size.

@glima and @vcgomes: This replaces my fix in commit #1576


otaviobp:
merged

PR #1579, opened 2016-03-09
coap: finally get rid of hardcoded packet sizes at sending side.

glima:
We're abusing tinycbor's API, but at least we get to incrementally bump
a CoAP packet's buffer size (that should not happen frequently, since
the growth is exponential and it starts with a guessed typical OIC/cbor
payload size).

----------------

@otaviobp: help with testing still welcome :)


otaviobp:
lgtm and it didn't break any oic tests

PR #1580, opened 2016-03-09
HTTP-Server: Add option to set allowed HTTP methods. v2

cabelitos:
Chances since v1:
* Added commit 9ee5ff7
* Changed how allowed_methods is stored.


cabelitos:
merged

PR #1587, opened 2016-03-10
A few build system simplifications

ibriano:
Nothing much


barbieri:
+1

while at it I'd ask you to review some rules we generate files to do in a ".new", `cmp` and only replace if changed. This way we avoid some rebuilds we do. `sol_config.h`, `sol_definitions.h`, `.config` are cases I recall had problems

ibriano:
On Sat, 12 Mar 2016 07:02:15 -0800, Gustavo Sverzut Barbieri wrote:
> +1
> 
> while at it I'd ask you to review some rules we generate files to do in a ".new", `cmp` and only replace if changed. This way we avoid some rebuilds we do. `sol_config.h`, `sol_definitions.h`, `.config` are cases I recall had problems

Good idea, I will do that as part of the other patch set I'm working on.

PR #1588, opened 2016-03-10
coap: finally get rid of hardcoded packet sizes at sending side [v2]

glima:
We're abusing tinycbor's API, but at least we get to incrementally bump
a CoAP packet's buffer size (that should not happen frequently, since
the growth is exponential and it starts with a guessed typical OIC/cbor
payload size).

I chose to change struct sol_oic_repr_field field names so that a macro
could handle both scalar field and key cases of cbor encoding.

----

Changes since #1579:
 - enlarge_buffer() code got slightly smaller, with the switch() being possibly evaluated more than once
 - struct sol_oic_repr_field had to have fields renamed to ease my life

@vcgomes: better now?


glima:
v3 coming

PR #1592, opened 2016-03-11
coap: Fix wrong unref

ceolin:
It was unref the wrong variable and this could cause double free and
leakings.


cabelitos:
+1

ceolin:
thanks guys

PR #1593, opened 2016-03-11
HTTP Clien/Server Add missing node types

cabelitos:


src/modules/flow/http-client/http-client.c
```diff
@@ -712,6 +712,186 @@ float_post_process(struct sol_flow_node *node, void *data, uint16_t port, uint16
...
+{
+    int r;
+    struct sol_rgb rgb;
+
+    SOL_BUFFER_DECLARE_STATIC(red, DOUBLE_STRING_LEN);
+    SOL_BUFFER_DECLARE_STATIC(green, DOUBLE_STRING_LEN);
+    SOL_BUFFER_DECLARE_STATIC(blue, DOUBLE_STRING_LEN);
+    SOL_BUFFER_DECLARE_STATIC(red_max, DOUBLE_STRING_LEN);
+    SOL_BUFFER_DECLARE_STATIC(green_max, DOUBLE_STRING_LEN);
+    SOL_BUFFER_DECLARE_STATIC(blue_max, DOUBLE_STRING_LEN);
```
barbieri: why double-string len if we're formatting an u32? the size is smaller AFAIR

src/modules/flow/http-server/http-server.c
```diff
@@ -713,6 +714,167 @@ float_process_cb(struct http_data *mdata, const struct sol_flow_packet *packet)
...
+static void
+rgb_send_packet_cb(struct http_data *mdata, struct sol_flow_node *node)
+{
+    sol_flow_send_rgb_packet(node,
+        SOL_FLOW_NODE_TYPE_HTTP_SERVER_RGB__OUT__OUT, &mdata->value.rgb);
+}
+
+static int
+rgb_response_cb(struct http_data *mdata, struct sol_buffer *content, bool json)
+{
```
barbieri: shouldn't you check the `json`? what you provide looks like the json return. I've noticed that other types, when complex, return the main value, something that doesn't exist for RGB... so I'm bit unsure on how to proceed, eventually we may force this return and change the Content-type? In this case we'd need to return something to notify a different type is to be used.

Another option is to return a text-like representation of color, like %g,%g,%g, that is, normalized 0-1 value of color components (red / red_max...).

It's basically your call, but document the behavior in the node description.

src/modules/flow/http-server/http-server.c
```diff
@@ -713,6 +714,167 @@ float_process_cb(struct http_data *mdata, const struct sol_flow_packet *packet)
...
+    r = sol_json_double_to_str(mdata->value.dir_vector.z, &z);
+    SOL_INT_CHECK(r, < 0, r);
+
+    r = sol_json_double_to_str(mdata->value.dir_vector.min, &min);
+    SOL_INT_CHECK(r, < 0, r);
+
+    r = sol_json_double_to_str(mdata->value.dir_vector.max, &max);
+    SOL_INT_CHECK(r, < 0, r);
+
+    return sol_buffer_append_printf(content,
```
barbieri: like I said for `rgb`, we should see what to do with `json`.

BTW, I've just noticed that you create 6 buffers, format them to string, then create the final json... which is a waste! The `content` is already a buffer, you can progressively encode to it and save intermediate buffers and copies.

barbieri:
the patch looks good, with minor details that are easy to fix.

I'd just ask you to watch my #1573 since it will be merged and you need to change your `http-server.c` to consider unchanged contents.

also, I've just noticed `persistence` module lacks RGB and direction-vector as well :-) Can come in this PR or in different, your call.

src/modules/flow/http-client/http-client.c
cabelitos: I'm sorry. Copy & pasta.

src/modules/flow/http-server/http-server.c
cabelitos: My bad. I will change the code to use one buffer.

cabelitos:
Thank you for the review.
@barbieri I will properly merge with #1573. 

I will create a different PR for the persistence nodes !

src/modules/flow/http-server/http-server.c
cabelitos: I don't think enforcing the content type is a good idea. As we discussed I will return the color type in hexadecimal if the content type is not json.

cabelitos:
#1607

PR #1594, opened 2016-03-11
coap: finally get rid of hardcoded packet sizes at sending side [v3]

glima:
Changes since #1588:
- buffer realloc loop changed to look nicer/simpler


glima:
Cool! Merging :)

PR #1595, opened 2016-03-11
[WIP] initial add of python binding directory layout

tripzero:
This is the basic infrastructure required to start creating python bindings.  Right now, the Makefile rule works and generates a stub python "binding" c file (mainloop.c).

What isn't working yet is mainfile.o does not seem to be building into a shared library.  I haven't discovered how that is done yet.  Any hints would be appreciated.


.gitignore
```diff
@@ -24,3 +24,4 @@ config.log*
 /node_modules/
 /bindings/nodejs/generated/main.cc
 /bindings/nodejs/generated/main.h
+/bindings/python/generated/mainloop.c
```
barbieri: maybe ignore all inside /bindings/python/generated?

also, try to build these generated inside /build? like `/build/stage/python`?

binding.gyp
```diff
@@ -59,6 +59,34 @@
 					},
 					"libraries": [ '<(SOLETTA_LIBS)' ],
 					"dependencies": [ "collectbindings" ]
+				},
+				{
+					"target_name" : "generatecythonbindings",
```
barbieri: do you really want to use node.js build system? Any benefits or just following what they did? For us it would be nicer to integrate in the host buildsystem (kconfig/makefiles). Gabriel used gyp to benefit from npm/node support.

bindings/python/Makefile
```diff
@@ -0,0 +1,9 @@
+include $(top_srcdir)/bindings/python/bindings.mk
+
+#$(error "python cflags? $(LIBPYTHON3_CFLAGS)")
+
+obj-$(USE_PYTHON) += soletta.mod
+
+obj-soletta-$(USE_PYTHON) := $(PYTHON_BINDING_OBJS)
+obj-soletta-$(USE_PYTHON)-extra-ldflags = $(SOLETTA_LIBS) $(LIBPYTHON3_LDFLAGS)
+obj-soletta-$(USE_PYTHON)-extra-cflags = $(SOLETTA_CFLAGS) $(LIBPYTHON3_CFLAGS)
```
barbieri: missing newline

bindings/python/mainloop.pyx
```diff
@@ -0,0 +1,13 @@
+
+cdef extern from "sol-mainloop.h":
+	int sol_init()
+	int sol_run()
+
+class SolettaMainloop():
+	def __init__(self):
+		sol_init()
+
```
barbieri: then provide a destructor that calls `sol_shutdown()`

It would be nice to also call `sol_args_set()` based on what you get from python, this way we can have some modules such as FBP's `app/argc-argv` to work.

data/scripts/dependency-resolver.py
```diff
@@ -281,6 +281,7 @@ def handle_exec_check(args, conf, context):
     if success:
         dir_path = os.path.dirname(os.path.realpath("%s" % path))
         context.add_cond_makefile_var("%s_DIR" % dep_sym, dir_path)
+        context.add_cond_makefile_var("%s_PATH" % dep_sym, dir_path)
```
barbieri: is this right? why not use the `_DIR` then?

bindings/python/mainloop.pyx
```diff
@@ -0,0 +1,13 @@
...
+	def run(self):
+		sol_run()
```
barbieri: You should release the GIL before calling `sol_run()`, see http://docs.cython.org/src/userguide/external_C_code.html#releasing-the-gil

Then from C callbacks that would go back to Python, you must declare `with gil`. IOW, if you wrap `sol_timeout_add()`, the callback you give to that function is `cdef ... with gil`, so cython will emit preamble to acquire the interpreter lock before proceeding. If you follow Node.js and want to wrap `sol_platform_add_state_monitor()`, it is the same... or GPIO's `sol_gpio_config::in::cb`.

barbieri:
this is a great start, would be nice to fix the GIL issues and get something that does an useful operation to test, I'd say go for gpio bindings :-)

binding.gyp
tripzero: I was just following what they did... but I'm not sure those lines are being used.  I'll double check.

bindings/python/mainloop.pyx
tripzero: Agreed.  These are just "stubs" for now while I get cython integrated with the build system.  I wouldn't consider anything here an actual API at this point.

data/scripts/dependency-resolver.py
tripzero: I figured since we are using to "dir" and "path" ambiguously, we should have both _DIR and _PATH...

binding.gyp
tripzero: It's not needed.  I'll remove this change.

tripzero:
@barbieri any thoughts on why soletta.so isn't building?  How do I make sure bindings/python/Makefile doesn't do anything until AFTER the cython c-files are generated?

data/scripts/dependency-resolver.py
barbieri: ok

Kconfig
```diff
@@ -227,4 +227,12 @@ config USE_NODEJS
 	default n
 	help
 		Enable Node.js bindings
+
+config USE_PYTHON
+	bool "Python bindings"
+	depends on HAVE_LIBPYTHON3
```
barbieri: && HAVE_CYTHON3

bindings/python/mainloop.pyx
```diff
@@ -0,0 +1,17 @@
...
+	int sol_init()
+	int sol_run()
+	void sol_shutdown()
+
+class Mainloop():
+	def __init__(self):
+		sol_init()
+
+	def __del__(self):
+		sol_shutdown()
```
barbieri: good. But now I'm thinking if we should  do these in class or automatically. Likely automatically would make more sense, since calling soletta methods before `init()` is not possible. In EFL we did it automatically, see: https://git.enlightenment.org/bindings/python/python-efl.git/tree/efl/ecore/efl.ecore.pyx

```py
def init():
    sol_init()

def shutdown():
    sol_shutdown()

init()
atexit.register(shutdown)
```

But you can hold on that. I need to think how to integrate with `asyncio`. One option is to have soletta to be a guest mainloop to that one, like I'm proposing to node.js. Another is to make `asyncio` use soletta.

My main concern with the second idea is that soletta itself will do `sigprocmask` and handle signals, that may conflict with python's own. So we'll need a way to cooperate both. If we can forward soletta ML primitives to `asyncio`, then we can, **before** calling `sol_init()` call `sol_mainloop_set_implementation()` with a set of `cdef` functions that wraps and forwards soletta to asyncio and back.

bindings/python/mainloop.pyx
```diff
@@ -0,0 +1,17 @@
...
+
+	def run(self):
+			return sol_run()
```
barbieri: Maybe take the `startup` and `shutdown` functions so we simulate the behavior as https://github.com/solettaproject/soletta/blob/master/src/lib/common/include/soletta.h#L105?

The idea is that we run user code inside the main loop. This guarantees `sol_init()` is called and mainloop is running. As said above, auto-calling `sol_init()` on module load may be a solution that solves that issue.

tools/build/Makefile.rules
```diff
@@ -405,3 +405,11 @@ $(KCONFIG_AUTOHEADER): $(DEPENDENCY_CACHE) $(KCONFIG_CONFIG) $(obj)/conf
 	$(Q)echo "     "GEN"   "$@
 	$(Q)mkdir -p $(KCONFIG_INCLUDE)config $(KCONFIG_INCLUDE)generated
 	$(Q)$(obj)/conf -s --silentoldconfig Kconfig
+
+include $(top_srcdir)bindings/python/bindings.mk
```
barbieri: it's better to avoid this different makefile, follow the pattern used by other soletta native parts, just define a bindings/python/Makefile and we'll create a rule to `.c -> .pyx` and it will be done automatically. We have similar rules to generate `.c -> .json` inside this file, we can base on that.

barbieri:
@dorileo can you help @tripzero with the integration? You can commit his files to repo directly, it's very small, almost a stub. Then he can proceed without bothering with the main loop.

For Python purposes I believe we'll end with different sub-modules, like the following **installed** structure:
```
/usr/lib/python3.5/site-packages/soletta/__init__.py # entry point, will load c_soletta.so that is the built from .pyx -> .c -> .so, includes mainloop, platform...
/usr/lib/python3.5/site-packages/soletta/flow.so # fbp stuff, what is inside src/lib/flow built from .pyx
/usr/lib/python3.5/site-packages/soletta/io.so # what is inside src/lib/io, built from .pyx
```

The use of `__init__.py` is to do some general purpose entry point, then you hide the actual stuff generated in the bindings as `c_soletta.so`. Our `__init__.py` will do things like:

```py
from c_soletta import * # more more specific symbols
```

bindings/python/mainloop.pyx
tripzero: I was leaning towards implementing asyncio's EventLoopPolicy interface and call into soletta mainloop functions.  I think this is your latter suggestion.  We'd implement "run_forever" like this:

```python

class SolettaEventLoopPolicy(asyncio.BaseEventLoopPolicy):
def __init__(self):
   self.mainloop = soletta.MainLoop

def run_forever(self):
    self.mainloop.run()

def call_later(self, ...):
   self.mainloop.add_timeout(...)

File: bindings/python/mainloop.pyx (hunk 9 above)
Comment (tripzero) [inline bindings/python/mainloop.pyx]: Agreed.

**Comment by tripzero** [comment]
> For Python purposes I believe we'll end with different sub-modules, like the following installed structure:

Yes.  This makes sense.

File: bindings/python/mainloop.pyx (hunk 8 above)
Comment (barbieri) [inline bindings/python/mainloop.pyx]: There is `sol_mainloop_set_implementation()`, then when soletta calls `sol_timeout_add()`, you can forward to Python. When Python dispatches, it will call you and you call back C. AFAIU this is similar to what you do in Qt.

The benefit of this approach is that you keep a single entity to manage the process signals, set `sigprocmask()`and spawns threads and forks, each may need some special treatment, as in Soletta we clear mask before proceeding. I recall having issues with that in EFL due the same case of double-management.


that is, in Cython we do something like:

```py

cdef void wrap_sol_init():
    pass

cdef void wrap_sol_shutdown():
   pass

cdef void wrap_sol_quit():
    asyncio.get_event_loop().stop()

cdef void wrap_sol_run():
    asyncio.get_event_loop().run_forever()

cdef wrap_timeout_cb(args):
    cdef object wrap_data = args # may need some cython casts
    handle = wrap_data.pop()
    cdef bool (*cb)(void *data) = wrap_data[0] # may need some cython casts
    cdef void *data = wrap_data[1] # may need some cython casts
    cdef float timeout_s = wrap_data[2]
    if cb(data): # renew
        handle = asyncio.get_event_loop().call_later(timeout_s, wrap_timeout_cb, wrap_data)
        wrap_data.append(handle)

cdef void *wrap_sol_timeout_add(uint32_t timeout_ms, bool (*cb)(void *data), const void *data):
    wrap_data = [<long>cb, <long>data, timeout_ms / 1000.]
    handle = asyncio.get_event_loop().call_later(timeout_ms / 1000., wrap_timeout_cb, wrap_data)
    wrap_data.append(handle)
    return wrap_data # eventually you may need a way to incref it here.

cdef void wrap_sol_timeout_del(void *wrap_handle):
    cdef list wrap_data = wrap_handle # may need some cython casts
    if len(wrap_data) > 3:
        handle = wrap_data.pop()
        handle.cancel() # eventually you may need to decref it here

static const struct sol_mainlooop_implementation _py_asyncio_impl = {
    .init = wrap_sol_init,
    .shutdown = wrap_sol_shutdown,
    .run = wrap_sol_run,
    .quit = wrap_sol_quit,
    .timeout_add = wrap_sol_timeout_add,
    .timeout_del = wrap_sol_timeout_del,
...
}; # I don't recall the syntax!

sol_mainloop_set_implementation(&_py_asyncio_impl) # to be called on module load
```

As said you can use EFL as an example, although it changed a lot since I've stopped contributing, it still looks good to be use as example. They **don't do exactly this mainloop approach!** just use it as example of C<->Cython usage.

A good way to test this is, after the `sol_mainloop_set_implementation()` is called you create a test timer in C/soletta and see if it is fired from asyncio (print or gdb/backtrace):
```py
cdef bool my_test_cb(void *data):
    print("running my test...")
    return 1
sol_timeout_add(1000, my_test_cb, NULL)

```

tools/build/Makefile.rules
tripzero: I'm trying to centralize the place where I will have to modify variables in the future.  If I add a cython source file, absent this central include, I'm going to have to edit Makefile.rules and bindings/python/Makefile with mostly the same variables.

tripzero:
@dorileo any hints on build integration?  Right now I have an issue with the rules in Makefile.rules not running before bindings/python/Makefile tries to compile.

barbieri:
he is on vacation :-(

tripzero:
@barbieri any new comments on the new build integration and bindings?

bindings/python/mainloop.pyx
```diff
@@ -0,0 +1,66 @@
...
+
+cdef void * wrap_source_get_data(args):
+	return NULL
+
+cdef sol_mainloop_implementation _py_asyncio_impl
+
+_py_asyncio_impl.run = wrap_sol_run
+
+
+"""_py_asyncio_impl.api_version = 1
```
barbieri: =1 will will cause you problems, you should use the define. I guess you can treat it as a regular C symbol, it will be a passtru.

setup.py
```diff
@@ -0,0 +1,17 @@
+"setup.py"
+
+from distutils.core import setup
+from distutils.extension import Extension
+from Cython.Build import cythonize
+
+extensions = [
+    Extension("soletta", ["bindings/python/mainloop.pyx"],
+        include_dirs = ["./build/soletta_sysroot/usr/include/soletta"],
```
barbieri: this is not right, you can't assume prefix = `/usr`, for instance I never install soletta to my host system, instead I configure it with `/tmp/test-soletta` and define the `$PATH`, `$LD_LIBRARY_PATH` and `$PYTHONPATH` to make it usable. :-)

I'm not sure we even need to provide this `setup.py` in the top directory. If you want to allow "building from easy_install" or the likes, we need to do a different approach with it calling Soletta's build (kconfig/kbuild).

barbieri:
Hi @tripzero, sorry about the delay... it seems github doesn't send notifications on PR updates and I was clueless you did update it :-)

did a review, it's almost there! Just minor details, I think the build system can be done fully inside our makefile instead of using python's disutils. But it's not a must, you can just fix it to take the PREFIX from command line or envvar, pass those from our Makefile and keep going for a while.

setup.py
barbieri: see my overall comment with a suggestion for your prefix problem. You can pass it using the command line (get with `sys.argv` or argparse) or environment variable (get with `os.environ`). then you define those on Makefile before calling this script.

barbieri:
sorry, I didn't mean to close it... pressed the wrong button.

dorileo:
I can take a look at it soon.

tripzero:
@barbieri this last push has it installing where I think it should be using envvars defined by solettas buid system.

"import soletta" now works

Now to get the mainloop code all written...

barbieri:
ok, just be aware you committed lots of weird files into tools/build :-)

next time I'd also recommend close the PR and open a new one, so you get everyone notified. GH doesn't notify people of new commits in this PR

tripzero:
closing.  New WIP PR coming

PR #1597, opened 2016-03-12
coap: Separate transport from coap logic

ceolin:
This is probably the most straightforward way to separate the  coap from the transport. 
It defines a basic interface that coap will use to send and received data.

Now it look likes another abstraction over socket but it will allows using bluetooth as transport. The fact of each transport has its header and its own api will allows the oic implementation uses bluetooth as the transport easily.

Once we have mapped all the necessity of different transports we can improve this api.

### DIFF
--- a/src/lib/comms/Makefile
+++ b/src/lib/comms/Makefile
@@ -33,7 +33,9 @@ obj-networking-$(PLATFORM_LINUX) += \
 
 obj-networking-$(COAP) += \
     coap.o \
-    sol-coap.o
+    sol-coap.o \
+    sol-coap-transport.o \
+    sol-coap-transport-ip.o
 
 obj-networking-$(OIC) += \
     sol-oic-cbor.o \
@@ -101,7 +103,9 @@ headers-$(NETWORK) += \
     include/sol-network.h
 
 headers-$(COAP) += \
-    include/sol-coap.h
+    include/sol-coap.h \
+    include/sol-coap-transport.h \
+    include/sol-coap-transport-ip.h
 
 headers-$(OIC) += \
     include/sol-oic-common.h \
--- a/src/lib/comms/include/sol-coap-transport-ip.h
+++ b/src/lib/comms/include/sol-coap-transport-ip.h
@@ -1,0 +1,100 @@
+/*
+ * This file is part of the Soletta Project
+ *
+ * Copyright (C) 2016 Intel Corporation. All rights reserved.
+ *
+ * Redistribution and use in source and binary forms, with or without
+ * modification, are permitted provided that the following conditions
+ * are met:
+ *
+ *   * Redistributions of source code must retain the above copyright
+ *     notice, this list of conditions and the following disclaimer.
+ *   * Redistributions in binary form must reproduce the above copyright
+ *     notice, this list of conditions and the following disclaimer in
+ *     the documentation and/or other materials provided with the
+ *     distribution.
+ *   * Neither the name of Intel Corporation nor the names of its
+ *     contributors may be used to endorse or promote products derived
+ *     from this software without specific prior written permission.
+ *
+ * THIS SOFTWARE IS PROVIDED BY THE COPYRIGHT HOLDERS AND CONTRIBUTORS
+ * "AS IS" AND ANY EXPRESS OR IMPLIED WARRANTIES, INCLUDING, BUT NOT
+ * LIMITED TO, THE IMPLIED WARRANTIES OF MERCHANTABILITY AND FITNESS FOR
+ * A PARTICULAR PURPOSE ARE DISCLAIMED. IN NO EVENT SHALL THE COPYRIGHT
+ * OWNER OR CONTRIBUTORS BE LIABLE FOR ANY DIRECT, INDIRECT, INCIDENTAL,
+ * SPECIAL, EXEMPLARY, OR CONSEQUENTIAL DAMAGES (INCLUDING, BUT NOT
+ * LIMITED TO, PROCUREMENT OF SUBSTITUTE GOODS OR SERVICES; LOSS OF USE,
+ * DATA, OR PROFITS; OR BUSINESS INTERRUPTION) HOWEVER CAUSED AND ON ANY
+ * THEORY OF LIABILITY, WHETHER IN CONTRACT, STRICT LIABILITY, OR TORT
+ * (INCLUDING NEGLIGENCE OR OTHERWISE) ARISING IN ANY WAY OUT OF THE USE
+ * OF THIS SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
+ */
+
+#pragma once
+
+#include <sol-coap-transport.h>
+#include <sol-network.h>
+
+#ifdef __cplusplus
+extern "C" {
+#endif
+
+/**
+ * @file
+ * @brief Routines to use CoAP protocol IP transport.
+ */
+
+/**
+ * @defgroup CoAP_Transport_IP CoAP_Transport_IP
+ * @ingroup CoAP_Transport
+ *
+ * @{
+ */
+
+/**
+ * @brief Creates a new CoAP transport over UDP.
+ *
+ * This function will create a new CoAP transport using UDP socket.
+ *
+ * @param addr The address to be used (where the socket will be bound).
+ * @return A CoAP transport handle or @c NULL on error
+ *
+ * @see sol_coap_server_new()
+ * @see sol_coap_transport_ip_secure_new()
+ * @see sol_coap_transport_del()
+ */
+struct sol_coap_transport *sol_coap_transport_ip_new(const struct sol_network_link_addr *addr);
+
+/**
+ * @brief Creates a new CoAP transport over UDP using a DTLS socket.
+ *
+ * This function will create a new CoAP transport using UDP socket and all the
+ * traffic will be encrypted.
+ *
+ * @param addr The address to be used (where the socket will be bound).
+ * @return A CoAP transport handle or @c NULL on error
+ *
+ * @see sol_coap_server_new()
+ * @see sol_coap_transport_ip_new()
+ * @see sol_coap_transport_del()
+ */
+struct sol_coap_transport *sol_coap_transport_ip_secure_new(const struct sol_network_link_addr *addr);
+
+/**
+ * @brief Deletes a CoAP transport created with @c sol_coap_transport_ip_secure_new() or
+ * @c sol_coap_transport_ip_new()
+ *
+ * @param transport The transport to be deleted.
+ *
+ * @see sol_coap_transport_ip_secure_new()
+ * @see sol_coap_transport_ip_new()
+ */
+void sol_coap_transport_ip_del(struct sol_coap_transport *transport);
+
+/**
+ * @}
+ */
+
+#ifdef __cplusplus
+}
+#endif
--- a/src/lib/comms/include/sol-coap-transport.h
+++ b/src/lib/comms/include/sol-coap-transport.h
@@ -1,0 +1,197 @@
+/*
+ * This file is part of the Soletta Project
+ *
+ * Copyright (C) 2016 Intel Corporation. All rights reserved.
+ *
+ * Redistribution and use in source and binary forms, with or without
+ * modification, are permitted provided that the following conditions
+ * are met:
+ *
+ *   * Redistributions of source code must retain the above copyright
+ *     notice, this list of conditions and the following disclaimer.
+ *   * Redistributions in binary form must reproduce the above copyright
+ *     notice, this list of conditions and the following disclaimer in
+ *     the documentation and/or other materials provided with the
+ *     distribution.
+ *   * Neither the name of Intel Corporation nor the names of its
+ *     contributors may be used to endorse or promote products derived
+ *     from this software without specific prior written permission.
+ *
+ * THIS SOFTWARE IS PROVIDED BY THE COPYRIGHT HOLDERS AND CONTRIBUTORS
+ * "AS IS" AND ANY EXPRESS OR IMPLIED WARRANTIES, INCLUDING, BUT NOT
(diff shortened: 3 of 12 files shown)


src/lib/comms/include/sol-coap-transport-ip.h
```diff
@@ -0,0 +1,100 @@
...
+
+/**
+ * @defgroup CoAP_Transport_IP CoAP_Transport_IP
+ * @ingroup CoAP_Transport
+ *
+ * @{
+ */
+
+/**
+ * @brief Creates a new CoAP transport over UDP.
```
barbieri: great description, but it doesn't match the function name :-)

I'd rather call it `sol_coap_transport_udp_new()` or `sol_coap_transport_new_udp()` (I guess this one matches more the name schema)

src/lib/comms/include/sol-coap-transport-ip.h
```diff
@@ -0,0 +1,100 @@
...
+/**
+ * @brief Deletes a CoAP transport created with @c sol_coap_transport_ip_secure_new() or
+ * @c sol_coap_transport_ip_new()
+ *
+ * @param transport The transport to be deleted.
+ *
+ * @see sol_coap_transport_ip_secure_new()
+ * @see sol_coap_transport_ip_new()
+ */
+void sol_coap_transport_ip_del(struct sol_coap_transport *transport);
```
barbieri: why do you need a `sol_coap_transport_ip_del()` instead of a generic `sol_coap_transport_del()`?

src/lib/comms/include/sol-coap-transport.h
```diff
@@ -0,0 +1,197 @@
...
+     * @return 0 on success, -errno on error.
+     */
+    int (*recvmsg)(struct sol_coap_transport *transport,
+        void *buf, size_t len, struct sol_network_link_addr *addr);
+
+    /**
+     * @brief Set a callback to be called when is possible to send data.
+     *
+     * @param transport The pointer to this struct (self).
+     * @param on_can_write The callback to be called.
```
barbieri: document callback behavior, like return false/true.

src/lib/comms/include/sol-coap-transport.h
```diff
@@ -0,0 +1,197 @@
...
+ *
+ * @param transport The pointer to this struct (self).
+ * @param buf The buffer that will be used to store the received data.
+ * @param len The size of the given data's buffer.
+ * @param addr The address from where the data came.
+ * @return 0 on success, -errno on error.
+ *
+ * @see sol_coap_transport_set_on_read()
+ */
+int sol_coap_transport_recvmsg(struct sol_coap_transport *transport, void *buf, size_t len,
```
barbieri: return looks strange, you should match `sendmsg()` that returns `ssize_t` and thus the amount received, not 0 on success.

src/lib/comms/include/sol-coap-transport.h
```diff
@@ -0,0 +1,197 @@
...
+ * struct. It checks the parameters and the API version of the transport.
+ *
+ * @param transport The pointer to this struct (self).
+ * @param on_can_write The callback to be called.
+ * @param user_data The data that will be given when the callback be called.
+ * @return 0 on success, -errno on error.
+ *
+ * @see sol_coap_transport_sendmsg()
+ */
+int sol_coap_transport_set_on_write(struct sol_coap_transport *transport,
```
barbieri: messed up return, see my comment for `recvmsg`

src/lib/comms/include/sol-coap-transport.h
```diff
@@ -0,0 +1,197 @@
...
+ *
+ * @param transport The pointer to this struct (self).
+ * @param buf The data that is wanted to be sent.
+ * @param len The size of the given data's buffer.
+ * @param addr The address that will be sent.
+ * @return 0 on success, -errno on error.
+ *
+ * @see sol_coap_transport_set_on_write()
+ */
+int sol_coap_transport_sendmsg(struct sol_coap_transport *transport, const void *buf, size_t len,
```
barbieri: messed up return, see my comment for recvmsg

src/lib/comms/include/sol-coap-transport-ip.h
```diff
@@ -0,0 +1,100 @@
...
+ * traffic will be encrypted.
+ *
+ * @param addr The address to be used (where the socket will be bound).
+ * @return A CoAP transport handle or @c NULL on error
+ *
+ * @see sol_coap_server_new()
+ * @see sol_coap_transport_ip_new()
+ * @see sol_coap_transport_del()
+ */
+struct sol_coap_transport *sol_coap_transport_ip_secure_new(const struct sol_network_link_addr *addr);
```
barbieri: looking at your internal `sol_coap_transport_ip_new_full()` the parameter set is simple. I'd then keep a single `sol_coap_transport_new_udp(const struct sol_network_link_addr *addr, bool use_dtls)` and that's it, inside you choose the actual socket type based on that boolean.

src/lib/comms/sol-coap-transport.c
```diff
@@ -0,0 +1,100 @@
...
+#define SOL_COAP_TRANSPORT_API_CHECK(transport, expected, ...)
+#endif
+
+SOL_API int
+sol_coap_transport_sendmsg(struct sol_coap_transport *transport, const void *buf, size_t len,
+    const struct sol_network_link_addr *addr)
+{
+    SOL_NULL_CHECK(transport, -EINVAL);
+    SOL_COAP_TRANSPORT_API_CHECK(transport, SOL_COAP_TRANSPORT_API_VERSION, -EINVAL);
+    SOL_NULL_CHECK(transport->sendmsg, -EINVAL);
```
barbieri: no reason to check sendmsg here, rather do at creation time

src/lib/comms/sol-coap-transport.c
```diff
@@ -0,0 +1,100 @@
...
+    return transport->sendmsg(transport, buf, len, addr);
+}
+
+SOL_API int
+sol_coap_transport_recvmsg(struct sol_coap_transport *transport, void *buf, size_t len,
+    struct sol_network_link_addr *addr)
+{
+    SOL_NULL_CHECK(transport, -EINVAL);
+    SOL_COAP_TRANSPORT_API_CHECK(transport, SOL_COAP_TRANSPORT_API_VERSION, -EINVAL);
+    SOL_NULL_CHECK(transport->recvmsg, -EINVAL);
```
barbieri: no reason to check here, do at creation time

src/lib/comms/sol-lwm2m.c
```diff
@@ -163,6 +165,7 @@ struct lifetime_ctx {
 
 struct sol_lwm2m_server {
     struct sol_coap_server *coap;
+    struct sol_coap_transport *transport;
```
barbieri: `sol_coap_server` already keeps a reference to transport, no need to duplicate it here, just add a getter.

src/lib/comms/Makefile
```diff
@@ -33,7 +33,9 @@ obj-networking-$(PLATFORM_LINUX) += \
 
 obj-networking-$(COAP) += \
     coap.o \
-    sol-coap.o
+    sol-coap.o \
+    sol-coap-transport.o \
+    sol-coap-transport-ip.o
```
barbieri: with your change we should make `ip` (or actually udp) conditional to that, not `$(COAP)`, then it would be more like:

```
ifeq (y,$(COAP))
obj-networking-$(UDP) += sol-coap-transport-udp.o
endif
```

We'd need platforms to declare if they have UDP or not, currently they say "networking", but I guess we should be more explicit. What do you think, @ibriano?

Also, I'm assuming udp here and in your file, but if CoAP over TCP "just works", then you can keep that file with such name and call it general. However, my impression is that for TCP you'd need some different code, since there is no framing, and for coap purposes very likely we can implement different methods to say if the message needs a confirmation or not (I see this is not in the code yet, but with time we should optimize that case for protocols that guarantee delivery)

src/lib/comms/Makefile
```diff
@@ -101,7 +103,9 @@ headers-$(NETWORK) += \
     include/sol-network.h
 
 headers-$(COAP) += \
-    include/sol-coap.h
+    include/sol-coap.h \
+    include/sol-coap-transport.h \
+    include/sol-coap-transport-ip.h
```
barbieri: as per above, you should only install this header if there is udp/ip

src/lib/comms/include/sol-coap-transport.h
```diff
@@ -0,0 +1,197 @@
...
+/**
+ * @brief Struct that represents CoAP transport.
+ *
+ * It declares a required interface that CoAP uses to communicate.
+ * One can create a new transport filling this struct with the
+ * proper methods.
+ *
+ * @see sol_coap_server_new()
+ */
+struct sol_coap_transport {
```
barbieri: this should be called `struct sol_coap_transport_type` and it should be declared as `static const` in the providers, this part is the "class", it won't change per-instance.

`struct sol_coap_transport` should be much simpler:
```c
struct sol_coap_transport {
    const stuct sol_coap_transport_type *type;
};
```

src/lib/comms/include/sol-coap-transport.h
```diff
@@ -0,0 +1,197 @@
...
+ * @param user_data The data that will be given when the callback be called.
+ * @return 0 on success, -errno on error.
+ *
+ * @see sol_coap_transport_recvmsg()
+ */
+int sol_coap_transport_set_on_read(struct sol_coap_transport *transport,
+    bool (*on_can_read)(void *data, struct sol_coap_transport *transport),
+    const void *user_data);
+
+/**
```
barbieri: you should provide `sol_coap_transport_del()` here.

src/lib/comms/include/sol-coap.h
```diff
@@ -480,29 +480,11 @@ int sol_coap_header_set_id(struct sol_coap_packet *pkt, uint16_t id);
...
- * and errno will be set to indicate the reason.
- *
- * @param addr The address where the server will listen on.
- *
- * @return A new server instance, or NULL in case of failure.
- *
- * @see sol_coap_server_new()
  */
-struct sol_coap_server *sol_coap_secure_server_new(const struct sol_network_link_addr *addr);
+struct sol_coap_server *sol_coap_server_new(struct sol_coap_transport *transport);
```
barbieri: document this function will take ownership of the transport, then there is no need to delete it. Even in the case of failure, so the below construct can be used:

```c
server = sol_coap_server(sol_coap_transport_new_udp(addr, false));
SOL_NULL_CHECK(server, -EINVAL);
```

of course static-inline helper `sol_coap_server_new_udp(addr, use_dtls)` to do the pair would be nice :-)

src/lib/comms/sol-coap-transport-ip.c
```diff
@@ -0,0 +1,310 @@
...
+    } on_can_write;
+};
+
+static bool
+transport_ip_on_can_read(void *data, struct sol_socket *s)
+{
+    struct sol_coap_transport_ip *tip = data;
+
+    if (tip->on_can_read.cb)
+        return tip->on_can_read.cb((void *)tip->on_can_read.cb_data, data);
```
barbieri: to avoid confusion and allow compiler to alert on changes, instead of passing the "data" as transport, make it explicit `&tip->base`

src/lib/comms/sol-coap-transport-ip.c
```diff
@@ -0,0 +1,310 @@
...
+
+    return true;
```
barbieri: is this right? do we want to keep the fd watcher if there is no callback?

src/lib/comms/sol-coap.c
```diff
@@ -1506,8 +1501,6 @@ sol_coap_server_destroy(struct sol_coap_server *server)
     struct outgoing *o;
     uint16_t i;
 
-    sol_socket_del(server->socket);
```
barbieri: change to take the ownership of the transport, then delete it here.

src/lib/comms/include/sol-coap-transport.h
barbieri: btw, coap doesn't return false.. then there is no need to return a boolean.

src/lib/comms/include/sol-coap-transport.h
barbieri: Well, after reviewing it couple of times I think if we change slightly the interface we can simplify it a lot.

The only difference you need in transport is actually a setup and then it's a regular socket-based operation, then if you move all of what you call transport to a way to specify your own `sol_socket`, you're good to go. Example:

```c
static struct sol_socket_type coap_socket_type; // impl should be renamed to type, remove new().
static void (*orig_socket_del)(struct sol_socket *socket);

struct sol_socket *sol_coap_new_udp_socket(const struct sol_network_link_addr *addr, bool use_dtls)
{
     struct sol_socket *socket = sol_socket_new(...);
     if (!coap_socket_type.del) {
         memcpy(&coap_socket_type, socket->type, sizeof(coap_socket_type));
         coap_socket_type.del = coap_socket_del; // overload del() so we can unregister network events
         socket->type = &coap_socket_type;
     }
     // bind
     // join mcast
     // monitor network events
     return socket;
}
```

What is missing:
 - install `sol-socket.h`
 - move `struct sol_socket_impl` parts other than `new()` to it, the `get_impl()` can then be a `sol_socket_impl_new()`

Then, when we create a BLE/Iotivity like @brunobottazzini did, we do not hack an `AF_FAMILY`, we just create a new socket manually:

```c
static const struct sol_socket_type oic_ble_socket_type = {
    .sendmsg = oic_ble_sendmsg,
    .recvmsg = oic_ble_recvmsg,
...
};

struct oic_ble_socket {
    struct sol_socket base;
};

static struct sol_socket *oic_ble_socket_new(...) {
   socket = calloc(1, sizeof(struct oic_ble_socket));
   socket->base.type = &oic_ble_socket_type;
   // ble stuff here
   return &socket->base;
}
```

ceolin:
Changes on https://github.com/solettaproject/soletta/pull/1773

PR #1600, opened 2016-03-14
Add api to set a link down

ceolin:


ceolin:
merged with suggestions

PR #1605, opened 2016-03-14
scripts/gen-html: use module names as category id

bdilly:
Instead of using numbers, so category (and entry)
ids will be fixed, not changing when new
categories are added.

We have a few wrong links spread around our wiki
already.

@gpaes ptal


gpaes:
LGTM.
Once this gets merged we should go through the wiki and fix the broken links. I can help with that.

bdilly:
nice, merged
thanks

PR #1606, opened 2016-03-14
doc: Add Soletta first tutorial

edersondisouza:
This tutorial should be a step by step guide through
`samples/flow/oic-and-http-light`.

### DIFF
--- a/doc/tutorials/ostro-oic-tutorial.md
+++ b/doc/tutorials/ostro-oic-tutorial.md
@@ -1,0 +1,1030 @@
+# Overview
+
+In this tutorial, we'll show how to develop a simple application using
+Soletta. Although simple, this application will show some important
+features of Soletta, like OIC and HTTP communication, besides GPIO
+interaction.
+Initially, we are going to run the application on desktop [1], but
+later we're going to show how to run it on devices, such as [MinnowBoard
+MAX](http://wiki.minnowboard.org/MinnowBoard_Wiki_Home).
+
+## Step 0
+
+Soletta must be installed on developer's computer. One can use a package
+provide for his or her distro, or [build Soletta](https://github.com/solettaproject/soletta/wiki/How-to-start).
+
+## Step 1
+
+### Flow Based Programming overview
+
+Soletta provides two API [3] for developers. A C API and a flow based
+API. In this tutorial, we're going to show the flow based API.
+The flow based API uses [Flow Based Programming](https://en.wikipedia.org/wiki/Flow-based_programming)
+concepts. It is a higher level API than its C counterpart, and it's
+much easier to use when expressing relationships among application
+components.
+Basically, a flow based application contains 'nodes' which represents
+entities - like sensors or actuators, as well as operations like
+concatenating string. These nodes have input or output 'ports' that
+send or receive 'packets' of information. These ports can be connected
+to express the flow of information among components. For instance:
+
+```
+button(gpio/reader:pin=1) OUT -> IN light(gpio/writer:pin=2)
+```
+
+Here, `button` is a node of node type [4] `gpio/reader` and light is
+another node, of node type `gpio/writer`. A `gpio/reader` node
+has an output port `OUT` - which sends boolean packets - that
+is connected to input port `IN` of the `gpio/writer` - which
+in turn, receives boolean packets as well. So, in this sample,
+we use `button` output to feed `light`. The `pin=N` on those
+nodes is a 'node option'. Node options are used to setup nodes,
+so each node, in this sample, is aware of which GPIO pin they
+represent.
+Note that subsequent uses of a node don't take its setup information
+again, so, if we wanted to debug `button` output to console,
+we would do:
+
+```
+button OUT -> IN _(console)
+```
+
+For more information about Soletta flow programming approach,
+you can see [Soletta README](https://github.com/solettaproject/soletta#flow-based-programming).
+For more information about Soletta available node types,
+see Soletta node type [Cheat Sheet](http://solettaproject.github.io/docs/nodetypes/).
+
+### Writing a **really** simple application
+
+The first step to write an application is to define what it shall do.
+Let's imagine that we want to control a device that has two features:
+'state' and 'power'. One can think of a light, for instance. It's either
+'ON' or 'OFF' - the state, and has different light intensities - the power.
+We can use two accumulators for this task: one from 0 to 1, to store
+'state', and another from 0 to 100 to store 'power'.
+
+```
+power_value(int/accumulator:setup_value=min:0|max:100|step:1,initial_value=50)
+state_value_as_int(int/accumulator:setup_value=min:0|max:1)
+```
+
+As `int/accumulator` node type deals with integers, and 'state' is more
+naturally represented as boolean, we will need some 'converters' to change
+from/to integer/boolean:
+
+```
+state_value_from_boolean(converter/boolean-to-int) OUT -> SET state_value_as_int
+state_value_as_int OUT -> IN state_value(converter/int-to-boolean)
+```
+
+The first converter, `state_value_from_boolean`, will receive boolean
+packets and converter to integer and deliver the result to
+`state_value_as_int`. The second one does the opposite: it receives the
+output from `state_value_as_int` and converts to boolean.
+
+### Running the application
+
+If we write those to a file named 'controller.fbp', we can run it by:
+
+```
+$ sol-fbp-runner controller.fbp
+```
+
+The result is that nothing will appear. As Soletta applications are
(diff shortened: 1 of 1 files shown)


doc/tutorials/ostro-oic-tutorial.md
```diff
@@ -0,0 +1,1030 @@
...
+Initially, we are going to run the application on desktop [1], but
+later we're going to show how to run it on devices, such as [MinnowBoard
+MAX](http://wiki.minnowboard.org/MinnowBoard_Wiki_Home).
+
+## Step 0
+
+Soletta must be installed on developer's computer. One can use a package
+provide for his or her distro, or [build Soletta](https://github.com/solettaproject/soletta/wiki/How-to-start).
+
+## Step 1
```
barbieri: for step-1 I'd start with ideation of the solution. You can describe it in text and then add a graphical representation about it, close to FBP, but no mention of FBP at that point.

Then, when you get into actual implementation you mention FBP was chosen since it is closer to that way of thinking

doc/tutorials/ostro-oic-tutorial.md
```diff
@@ -0,0 +1,1030 @@
...
+
+### Flow Based Programming overview
+
+Soletta provides two API [3] for developers. A C API and a flow based
```
barbieri: don't use this kind of "[3]" for footnotes. If markdown lacks such, then just add within parenthesis

doc/tutorials/ostro-oic-tutorial.md
```diff
@@ -0,0 +1,1030 @@
...
+concatenating string. These nodes have input or output 'ports' that
+send or receive 'packets' of information. These ports can be connected
+to express the flow of information among components. For instance:
+
+```
+button(gpio/reader:pin=1) OUT -> IN light(gpio/writer:pin=2)
+```
+
+Here, `button` is a node of node type [4] `gpio/reader` and light is
+another node, of node type `gpio/writer`. A `gpio/reader` node
```
barbieri: you can compare `button` to an instance of `gpio/reader` class, this will help those coming from OOP.

doc/tutorials/ostro-oic-tutorial.md
```diff
@@ -0,0 +1,1030 @@
...
+has an output port `OUT` - which sends boolean packets - that
+is connected to input port `IN` of the `gpio/writer` - which
+in turn, receives boolean packets as well. So, in this sample,
+we use `button` output to feed `light`. The `pin=N` on those
+nodes is a 'node option'. Node options are used to setup nodes,
+so each node, in this sample, is aware of which GPIO pin they
+represent.
```
barbieri: we should link to http://solettaproject.github.io/docs/nodetypes/ to explain to user where those are documented.

doc/tutorials/ostro-oic-tutorial.md
```diff
@@ -0,0 +1,1030 @@
...
+Note that subsequent uses of a node don't take its setup information
+again, so, if we wanted to debug `button` output to console,
+we would do:
+
+```
+button OUT -> IN _(console)
```
barbieri: document `_` is about anonymous nodes. Or give it a name.

I'd document that and also explain a bit about console properties, like prefix/suffix options and that the prefix defaults to the node name, in tha case of it being anonymous it will be auto-generated based on line and column.

doc/tutorials/ostro-oic-tutorial.md
```diff
@@ -0,0 +1,1030 @@
...
+Let's imagine that we want to control a device that has two features:
+'state' and 'power'. One can think of a light, for instance. It's either
+'ON' or 'OFF' - the state, and has different light intensities - the power.
+We can use two accumulators for this task: one from 0 to 1, to store
+'state', and another from 0 to 100 to store 'power'.
+
+```
+power_value(int/accumulator:setup_value=min:0|max:100|step:1,initial_value=50)
+state_value_as_int(int/accumulator:setup_value=min:0|max:1)
+```
```
barbieri: well, this part is kinda cumbersome I've to admit. Maybe start with the simpler part that is the server itself. You can start by defining the server using http, use curl as controller. You can use a tool most people are familar with to query and set the value. Then you elaborate on the server and keep the controller like that. Follow with an http controller and then add oic, with oic controller. The learning curve won't be as steep.

doc/tutorials/ostro-oic-tutorial.md
```diff
@@ -0,0 +1,1030 @@
...
+
+Those binary codes correspond to 'j', 'k' and 'a', respectively.
+It's time to connect them to their accumulators. As buttons outputs
+a `true` packet when pressed and `false` when released, we need to
+filter its output: only when they are pressed matters.
+We achieve this with node type `boolean/filter` - connect only its
+TRUE output port to get only true packets.
+
+```
+increase_button OUT -> IN _(boolean/filter) TRUE -> INC power_value
```
barbieri: nice, I always forget about `boolean/filter`, could you update the oic-and-http-light to reflect this as well?

doc/tutorials/ostro-oic-tutorial.md
```diff
@@ -0,0 +1,1030 @@
...
+# Uses command line argument as device id
+cmdline_arguments OUT -> DEVICE_ID oic
+
+# Cross link OIC and controller
+oic OUT_STATE -> STATE controller
+oic OUT_POWER -> POWER controller
+
+controller STATE -> IN_STATE oic
+controller POWER -> IN_POWER oic
+```
```
barbieri: explain the cross-link properties here, I've found that people don't get that it will be stable and not loop infinitely.

doc/tutorials/ostro-oic-tutorial.md
```diff
@@ -0,0 +1,1030 @@
...
+
+### Running the server
+
+To really test our OIC controller, we need to run Soletta
+`light-server.fbp` sample, using the device id we are going to pass
+to the application. To do so, you can run, on Soletta dir:
+
+```
+SOL_MACHINE_ID=2f3089a1dbfb43d38cab64383bdf9380 sol-fbp-runner src/samples/flow/oic/light-server.fbp
```
barbieri: not good to require the `SOL_MACHINE_ID`, we'd have to explain that detail. it's better to immediately change `light-server` to print it and then ask user to use that value.

doc/tutorials/ostro-oic-tutorial.md
```diff
@@ -0,0 +1,1030 @@
...
+```
+$ sol-fbp-runner controller-oic.fbp 2f3089a1dbfb43d38cab64383bdf9380
+Server state is: false (boolean)
+Server power level is: 76 (integer range)
+Power value changed: 76 (integer range)
+State changed: true (boolean)
+Power value changed: 77 (integer range)
+```
+
+**Note:** You may see some warnings related to OIC as well, as it tries
```
barbieri: hum... I didn't have these, maybe we should move them to `INF`?

doc/tutorials/ostro-oic-tutorial.md
```diff
@@ -0,0 +1,1030 @@
...
+OUTPORT=power.OUT:OUT_POWER
+OUTPORT=state.OUT:OUT_STATE
+```
+
+If you run this file, you can even access current state on browser, by
+typing `http://localhost:8002/power`, for instance.
+
+### An HTTP controller to communicate with our HTTP server
+
+Now let's controll our HTTP server. Our OIC controller connected
```
barbieri: typo, controlL

doc/tutorials/ostro-oic-tutorial.md
```diff
@@ -0,0 +1,1030 @@
...
+'state' values. On a `http-light-server.fbp` write:
+
+```
+power(http-server/int:path="/power",port=8002)
+state(http-server/boolean:path="/state",port=8002)
+
+_(constant/string:value="http-light-server running on port 8002, paths: /state and /power") OUT -> IN _(console:prefix="note: ")
+```
+
+First two lines create two HTTP servers, one for 'power' and another for
```
barbieri: clarify soletta will create an internal server, only one per port and then register paths as needed. Users didn't understand such properties and were asking themselves if Apache or what was being started... expecting bugs due 2 servers being used.

They were also expecting the FBP to halt in "loop" when server was created... pretty much like Python or JS servers where you have a trailing "run-forever" at the bottom. Then explain these will not block.

doc/tutorials/ostro-oic-tutorial.md
edersondisouza: On Github, it's possible to workaround the missing link - I'll do that.

edersondisouza:
#1645

PR #1607, opened 2016-03-14
HTTP Client/Server Add missing node types v2

cabelitos:
Changes since v1:
* Fixed problems pointed by @barbieri

### DIFF
--- a/src/modules/flow/http-client/http-client.c
+++ b/src/modules/flow/http-client/http-client.c
@@ -847,6 +847,143 @@ rgb_post_process(struct sol_flow_node *node, void *data, uint16_t port,
 }
 
 /*
+ * --------------------------------- direction vector node  -----------------------------
+ */
+static int
+direction_vector_process_token(struct sol_flow_node *node,
+    struct sol_json_token *key, struct sol_json_token *value)
+{
+    struct sol_direction_vector dir_vector = { 0 };
+    enum sol_json_loop_reason reason;
+    struct sol_json_scanner sub_scanner;
+    struct sol_json_token sub_key, sub_value, token;
+
+    sol_json_scanner_init(&sub_scanner, value->start,
+        value->end - value->start);
+    SOL_JSON_SCANNER_OBJECT_LOOP (&sub_scanner, &token, &sub_key,
+        &sub_value, reason) {
+        if (SOL_JSON_TOKEN_STR_LITERAL_EQ(&sub_key, "x")) {
+            if (sol_json_token_get_double(&sub_value, &dir_vector.x) < 0)
+                return -EINVAL;
+        } else if (SOL_JSON_TOKEN_STR_LITERAL_EQ(&sub_key, "y")) {
+            if (sol_json_token_get_double(&sub_value, &dir_vector.y) < 0)
+                return -EINVAL;
+        } else if (SOL_JSON_TOKEN_STR_LITERAL_EQ(&sub_key, "z")) {
+            if (sol_json_token_get_double(&sub_value, &dir_vector.z) < 0)
+                return -EINVAL;
+        } else if (SOL_JSON_TOKEN_STR_LITERAL_EQ(&sub_key, "min")) {
+            if (sol_json_token_get_double(&sub_value, &dir_vector.min) < 0)
+                return -EINVAL;
+        } else if (SOL_JSON_TOKEN_STR_LITERAL_EQ(&sub_key, "max")) {
+            if (sol_json_token_get_double(&sub_value, &dir_vector.max) < 0)
+                return -EINVAL;
+        }
+    }
+
+    return sol_flow_send_direction_vector_packet(node,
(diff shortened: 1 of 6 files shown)


src/modules/flow/http-client/http-client.c
```diff
@@ -714,6 +714,276 @@ float_post_process(struct sol_flow_node *node, void *data, uint16_t port, uint16
...
+        } else if (SOL_JSON_TOKEN_STR_LITERAL_EQ(&sub_key, "blue")) {
+            if (sol_json_token_get_uint32(&sub_value, &rgb.blue) < 0)
+                return -EINVAL;
+        } else if (SOL_JSON_TOKEN_STR_LITERAL_EQ(&sub_key, "red_max")) {
+            if (sol_json_token_get_uint32(&sub_value, &rgb.red_max) < 0)
+                return -EINVAL;
+        } else if (SOL_JSON_TOKEN_STR_LITERAL_EQ(&sub_key, "green_max")) {
+            if (sol_json_token_get_uint32(&sub_value, &rgb.green_max) < 0)
+                return -EINVAL;
+        } else if (SOL_JSON_TOKEN_STR_LITERAL_EQ(&sub_key, "blue_max")) {
```
barbieri: hum.... I think we should declare `rgb` with better defaults for `red_max` and the likes, otherwise if we do not produce them in the JSON the resulting rgb will be correct. Maybe check if components are `<=` their maximum values before emitting the packet.

src/modules/flow/http-client/http-client.c
```diff
@@ -714,6 +714,276 @@ float_post_process(struct sol_flow_node *node, void *data, uint16_t port, uint16
...
+            if (sol_json_token_get_double(&sub_value, &dir_vector.z) < 0)
+                return -EINVAL;
+        } else if (SOL_JSON_TOKEN_STR_LITERAL_EQ(&sub_key, "min")) {
+            if (sol_json_token_get_double(&sub_value, &dir_vector.min) < 0)
+                return -EINVAL;
+        } else if (SOL_JSON_TOKEN_STR_LITERAL_EQ(&sub_key, "max")) {
+            if (sol_json_token_get_double(&sub_value, &dir_vector.max) < 0)
+                return -EINVAL;
+        }
+    }
```
barbieri: same comment I did for rgb, we should have better `min` and `max` defaults, check them before dispatching the packet.

src/modules/flow/http-server/http-server.c
```diff
@@ -789,6 +793,208 @@ float_process_cb(struct http_data *mdata, const struct sol_flow_packet *packet)
...
+        return sol_buffer_append_printf(content,
+            "{\"red\":%" PRIu32 ",\"green\":%" PRIu32 ",\"blue\":%" PRIu32
+            ",\"red_max\":%" PRIu32 ",\"green_max\":%" PRIu32
+            ",\"blue_max\":%" PRIu32 "}",
+            mdata->value.rgb.red, mdata->value.rgb.green, mdata->value.rgb.blue,
+            mdata->value.rgb.red_max, mdata->value.rgb.green_max,
+            mdata->value.rgb.blue_max);
+    }
+
+    return sol_buffer_append_printf(content, "#%X%X%X", mdata->value.rgb.red,
```
barbieri: %02x, otherwise if number is <= 0xf it will take a single digit. Like 0 will produce "#000" instead of "#000000".

src/modules/flow/http-server/http-server.c
```diff
@@ -789,6 +793,208 @@ float_process_cb(struct http_data *mdata, const struct sol_flow_packet *packet)
...
+    }
+
+    //Format (X;Y;Z)
+    r = sol_buffer_append_slice(content, sol_str_slice_from_str("("));
+    SOL_INT_CHECK(r, < 0, r);
+
+    r = sol_json_double_to_str(mdata->value.dir_vector.x, content);
+    SOL_INT_CHECK(r, < 0, r);
+
+    r = sol_buffer_append_slice(content, sol_str_slice_from_str(";"));
```
barbieri: we have append_char.

src/modules/flow/http-client/http-client.c
cabelitos: I will just set the max values to 255 then. What do you think?

barbieri:
looks good, fix the minor issues and push directly

src/modules/flow/http-client/http-client.c
barbieri: reasonable. Set to a good max and after the loop you validate it.

src/modules/flow/http-server/http-server.c
cabelitos: Ok.

src/modules/flow/http-server/http-server.c
cabelitos: Fair.

cabelitos:
merged

PR #1608, opened 2016-03-14
Nodejs oic bindings v2

gabrielschulhof:
This PR replaces #1568 

* Rebased
* Added license checking for .js files
* Made whitespaces uniform
* Corrected copyright dates

### DIFF
--- a/bindings/nodejs/src/functions/oic-client-discovery.cc
+++ b/bindings/nodejs/src/functions/oic-client-discovery.cc
@@ -1,0 +1,100 @@
+/*
+ * This file is part of the Soletta Project
+ *
+ * Copyright (C) 2015 Intel Corporation. All rights reserved.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+#include <string.h>
+#include <nan.h>
+#include <sol-oic-client.h>
+
+#include "../common.h"
+#include "../structures/network.h"
+#include "../structures/oic-client.h"
+
+using namespace v8;
+
+static bool resourceFound(struct sol_oic_client *client,
+    struct sol_oic_resource *resource, void *data) {
+    Nan::HandleScope scope;
+
+    // If Soletta tells us there are no more resources, we detach this callback
+    // no matter what the JS callback returns
+    bool keepDiscovering = !!resource;
+    OicCallbackData *callbackData = (OicCallbackData *)data;
+
+    // Call the JS callback
+    Local<Value> arguments[2] = {
+        Nan::New(*(callbackData->jsClient)),
+        Nan::Null()
+    };
+    if (resource) {
+        arguments[1] = SolOicClientResource::New(resource);
+    }
+    Local<Value> jsResult = callbackData->callback->Call(2, arguments);
+
+    // Determine whether we should keep discovering
+    if (!jsResult->IsBoolean()) {
+        Nan::ThrowTypeError(
+            "Resource discovery callback return value is not boolean");
+    } else {
+        keepDiscovering = keepDiscovering &&
+            Nan::To<bool>(jsResult).FromJust();
+    }
+
+    // Tear down if discovery is done
+    if (!keepDiscovering) {
+        delete callbackData;
+    }
+
+    return keepDiscovering;
+}
+
+NAN_METHOD(bind_sol_oic_client_find_resource) {
+    VALIDATE_ARGUMENT_COUNT(info, 4);
+    VALIDATE_ARGUMENT_TYPE(info, 0, IsObject);
+    VALIDATE_ARGUMENT_TYPE(info, 1, IsObject);
+    VALIDATE_ARGUMENT_TYPE(info, 2, IsString);
+    VALIDATE_ARGUMENT_TYPE(info, 3, IsFunction);
+
+    struct sol_network_link_addr theAddress;
+    if (!c_sol_network_link_addr(Nan::To<Object>(info[1]).ToLocalChecked(),
+        &theAddress)) {
+        return;
+    }
+
+    Local<Object> jsClient = Nan::To<Object>(info[0]).ToLocalChecked();
+    struct sol_oic_client *client = (struct sol_oic_client *)
+        SolOicClient::Resolve(jsClient);
+    if (!client) {
+        return;
+    }
+
+    OicCallbackData *callbackData =
+        OicCallbackData::New(jsClient, Local<Function>::Cast(info[3]));
+    if (!callbackData) {
+        return;
+    }
+
+    bool result = sol_oic_client_find_resource((struct sol_oic_client *)client,
+        &theAddress, (const char *)*String::Utf8Value(info[2]), resourceFound,
+        callbackData);
+
+    if (!result) {
+        delete callbackData;
+    }
+
+    info.GetReturnValue().Set(Nan::New(result));
+}
--- a/bindings/nodejs/src/functions/oic-server.cc
+++ b/bindings/nodejs/src/functions/oic-server.cc
@@ -1,0 +1,262 @@
+/*
+ * This file is part of the Soletta Project
+ *
+ * Copyright (C) 2015 Intel Corporation. All rights reserved.
+ *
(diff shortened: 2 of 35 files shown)


gabrielschulhof:
This PR binds ```sol_oic_client*()```, ```sol_oic_server*()```, ```sol_oic_notify_observers()```, and ```sol_network_link_addr_from_str()```

poussa:
@barbieri @bdilly @ceolin Why are we not getting any feedback on this? We need to make progress!

otaviobp:
The commit "Check license on JS files and accommodate spaces in file names" is ok to me, but do we really need spaces in file names? I wouldn't like to have this.

otaviobp:
PR looks good to me. I'll just wait for @ceolin ack to merge it.

gabrielschulhof:
@otaviobp it's kinda nice to use the file name as the name of the test, because it makes it easy to do something like
```bash
node bindings/nodejs/tests/suite --testList="OIC Polled Observation,Simple Platform APIs.js"
```
and the test suite doesn't have to load all the tests to find out which one bears the name "OIC Polled Observation".

otaviobp:
but why not:
node bindings/nodejs/tests/suite --testList="OICPolledObservation,SimplePlatformAPIs.js"
or:
node bindings/nodejs/tests/suite --testList="OIC_Polled_Observation,Simple_Platform_APIs.js"

Just a note, I am not opposing to merge this pr because of the whitespaces. I am just trying to understand why you are using them.

bindings/nodejs/src/functions/oic-client-discovery.cc
```diff
@@ -0,0 +1,100 @@
...
+    bool result = sol_oic_client_find_resource((struct sol_oic_client *)client,
+        &theAddress, (const char *)*String::Utf8Value(info[2]), resourceFound,
+        callbackData);
+
+    if (!result) {
+        delete callbackData;
+    }
+
+    info.GetReturnValue().Set(Nan::New(result));
+}
```
ceolin: i'm not familiar with js bindings but i guess this method should return a bool just like the C functions does, don't you think ?

bindings/nodejs/src/functions/oic-server.cc
```diff
@@ -0,0 +1,262 @@
...
+        to.member.len = strlen(to.member.data); \
+    } while(0)
+
+static bool c_sol_oic_resource_type(Local<Object> js,
+    struct sol_oic_resource_type *definition) {
+    const char *error = 0;
+    struct sol_oic_resource_type local = {
+#ifdef SOL_OIC_RESOURCE_TYPE_API_VERSION
+        .api_version = SOL_OIC_RESOURCE_TYPE_API_VERSION,
+#endif /* def SOL_OIC_RESOURCE_TYPE_API_VERSION */
```
ceolin: use `SOL_SET_API_VERSION` to avoid this `ifdef`

bindings/nodejs/src/functions/oic-server.cc
```diff
@@ -0,0 +1,262 @@
...
+path_failed:
+    free((void *)(local.interface.data));
+interface_failed:
+    free((void *)(local.resource_type.data));
+resource_type_failed:
+    Nan::ThrowError(error);
+    return false;
+}
+
+NAN_METHOD(bind_sol_oic_server_add_resource) {
```
ceolin: i'm afraid not understanding it. What is the return of this method ?

bindings/nodejs/src/functions/oic-server.cc
```diff
@@ -0,0 +1,262 @@
...
+        jsPayload = new Nan::Persistent<Object>(
+            Nan::To<Object>(info[1]).ToLocalChecked());
+    }
+
+    info.GetReturnValue().Set(Nan::New(
+        sol_oic_notify_observers(resourceInfo->resource,
+            (jsPayload ? oic_map_writer_callback : 0), jsPayload)));
+
+    if (jsPayload) {
+        jsPayload->Reset();
```
ceolin: is it necessary ? shouldn't the destructor be responsilbe to clear all the necessary data ?

bindings/nodejs/src/structures/oic-client.cc
```diff
@@ -0,0 +1,128 @@
...
+    hijackRefWasSuccessful = hijack_ref();
+    return hijackRefWasSuccessful;
+}
+
+OicCallbackData *OicCallbackData::New(Local<Object> jsClient,
+    Local<Function> jsCallback) {
+    OicCallbackData *data = new OicCallbackData;
+    if (!data) {
+        Nan::ThrowError("Failed to allocate OicCallbackData");
+    } else if (!data->init(jsClient, jsCallback)) {
```
ceolin: is it expected the failure of the init call ? or should throw an exception also here ?

bindings/nodejs/src/structures/oic-map.cc
```diff
@@ -0,0 +1,172 @@
...
+
+    if (value->IsInt32()) {
+        field.key = name;
+        field.type = SOL_OIC_REPR_TYPE_INT;
+        field.v_uint = value->Int32Value();
+    }
+    else
+    if (value->IsUint32()) {
+        field.key = name;
+        field.type = SOL_OIC_REPR_TYPE_TEXT_STRING;
```
ceolin: i guess the right here is `SOL_OIC_REPR_TYPE_UINT`

ceolin:
the serie looks good to me, just small comments (some of them questions)

bindings/nodejs/src/functions/oic-client-discovery.cc
gabrielschulhof: The signature of a JS binding returns void, but accepts a parameter, ```info```, which contains a return value which, in turn, can be set. This way, if you need to throw a JS exception, you can just throw it and return. That plays the role of the JS jump that happens when an exception is thrown. The effect of the above code is that ```result``` becomes the return value of the binding.

bindings/nodejs/src/functions/oic-server.cc
gabrielschulhof: ```NAN_METHOD(x)``` roughly translates as ```void x(something info)```, where ```info``` has these:
* ```info[0], info[1], ... info[n]``` are JS values (```v8::Local<v8::Value>```) that were passed to the binding on the JS side
*```info.GetReturnValue()``` gives you a slot into which you can place the ```v8::Local<v8::Value>``` which should become the return value of the binding on the JS side.

bindings/nodejs/src/functions/oic-server.cc
gabrielschulhof: Yes, it is necessary, unfortunately. You need to make sure the ```Nan::Persistent<v8::Object>``` points to nothing at all before you delete it. Nan claims that in future versions the ```Reset()``` will be included in the destructor but, alas, we are not yet living that future.

bindings/nodejs/src/structures/oic-client.cc
gabrielschulhof: I have a policy of throwing an exception at the point of failure - not later. Basically:
```
bool someOtherFunction() {
  if (uh-oh) {
    Nan::ThrowError("someOtherFunction: FAIL");
    return false;
  }
  return true;
}

bool someFunction() {
  if (uh-oh) {
    Nan::ThrowError("someFunction: FAIL");
    return false;
  }
  if (!someOtherFunction()) {
    return false;
  }
  return true;
}

NAN_METHOD(something) {
  if (uh-oh) {
    Nan::ThrowError("FAIL");
    return;
  }
  if (!someFunction()) {
    return;
  }
```
So, in this case, ```->init()``` will throw the exception and return ```false```. The ```false``` will propagate all the way back to the binding, which will simply return. This whole thing mimics the way JS tears down the stack when something throws an exception.

gabrielschulhof:
@otaviobp I use the file name as the title of the test, so
```
OIC Polled Observation
✓ @       352: Client 1: Resource found
✓ @      2427: Client 2: Resource found
✓ @     10366: Client 1: payload is as expected
...
```
looks a lot better than
```
OICPolledObservation
✓ @       352: Client 1: Resource found
✓ @      2427: Client 2: Resource found
✓ @     10366: Client 1: payload is as expected
...
```
or
```
OIC_Polled_Observation
✓ @       352: Client 1: Resource found
✓ @      2427: Client 2: Resource found
✓ @     10366: Client 1: payload is as expected
...
```
Besides, it doesn't hurt for our scripts to be space-in-the-filename-safe.

bindings/nodejs/src/functions/oic-server.cc
gabrielschulhof: https://github.com/nodejs/nan/blob/master/doc/persistent.md#nannoncopyablepersistenttraits--v8noncopyablepersistenttraits

barbieri:
Sorry @poussa and @gabrielschulhof. This week we had some events like presentation at a local university and it blocked our engineers.

In any case, you guys have full powers to timeout-commit these. Particularly the bindings part is up to you.

If we see any issues later, we can request changes via new PRs. Don't worry :-)

PR #1612, opened 2016-03-15
run-fbp-tests: Add test PASS/FAIL info to be compatible to ptest log format definition

otcyanglei:
According to package test (ptest) log format definition, the test case result recorded in log as
        PASS: <case name>
        FAIL: <case name>
    The patch prints more output to be compatible to ptest log format when execute Soletta FBP tests
    The Soletta test result mapping to ptest output logic is:
        Soletta test pass -> ptest output PASS
        Soletta test fail -> ptest output FAIL
        Soletta test skip by dependence test fail -> ptest output FAIL
        Soletta test skip by user specified       -> ptest output SKIP

Signed-off-by: Lei Yang <lei.a.yang@intel.com>

### DIFF
--- a/tools/run-fbp-tests
+++ b/tools/run-fbp-tests
@@ -241,6 +241,7 @@ def run_tests(args):
         if curr.should_skip(skips):
             logging.debug("SKIPPED running %s: %s" % (t, curr.skip_msg))
             status["skipped"] += 1
+            logging.info("SKIP: %s" % path["file"])
             continue
 
         if not args.compiled:
@@ -269,6 +270,7 @@ def run_tests(args):
                 logging.warning("Dependence test %s for test %s failed. Skipping it",
                         curr.depends, curr.test_file)
                 status["skipped"] += 1
+                logging.info("SKIP: %s" % path["file"])
                 continue
 
             out, code = sh(cmd, cwd=test_cwd)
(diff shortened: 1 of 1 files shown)


tools/run-fbp-tests
```diff
@@ -269,6 +270,7 @@ def run_tests(args):
                 logging.warning("Dependence test %s for test %s failed. Skipping it",
                         curr.depends, curr.test_file)
                 status["skipped"] += 1
+                logging.info("FAIL: %s" % path["file"])
```
barbieri: I've read your comment on the other PR (to ".bb"), but I'll reply it here: this is not FAIL, it's `SKIP` or just omit it. You didn't test, so there is no way to know if this failed or not.

otcyanglei: @barbieri That's OK. The case should be blocked by fail case. However, by reading ptest result definition (http://www.yoctoproject.org/docs/2.0/dev-manual/dev-manual.html#testing-packages-with-ptest), it seems BLOCK is not in the list. I can amend the patch and log the result as SKIP. Thanks.

otcyanglei:
I re-submit the patch according to feedback. And it passed the CI check too.
Any more feedback? Or could anyone help review and merge the patch? 
Thanks.

otcyanglei:
Can anyone help review and merge the patch? Thanks!

bdilly:
thanks @otcyanglei 
merged

PR #1616, opened 2016-03-15
HTTP Server/Client make use of the HTTP Last Modified header

cabelitos:


cabelitos:
@ceolin Could you take a look?

src/lib/comms/sol-http-server-impl-microhttpd.c
```diff
@@ -181,8 +182,31 @@ sol_http_request_get_method(const struct sol_http_request *request)
...
+static bool
+set_last_modified_header(struct MHD_Response *response, time_t last_modified)
+{
+    struct tm result;
+    char buf[128];
+    size_t r;
+
+    SOL_NULL_CHECK(gmtime_r(&last_modified, &result), false);
+
+    r = strftime(buf, sizeof(buf), "%a, %d %b %Y %H:%M:%S GMT", &result);
```
barbieri: hum... I've missed this in our other bits, but if we `setlocale()` this will be wrong. Do we have a locale-less version of `strftime()` or one we could explicitly specify the locale and then we use "C"? It's not a blocker for this patch, but we should review this in the future.

eventually microhttp provides code to do that.

src/modules/flow/http-client/http-client.c
```diff
@@ -311,9 +344,19 @@ common_request_finished(void *data,
     struct http_data *mdata = sol_flow_node_get_private_data(node);
     const struct http_client_node_type *type;
 
-    if (check_response(mdata, node, connection, response) < 0)
+    ret = check_response(mdata, node, connection, response);
+    if (ret < 0) {
+        SOL_WRN("Invalid HTTP response");
```
barbieri: show response to help debug?

barbieri:
overall looks good.

src/lib/comms/sol-http-server-impl-microhttpd.c
cabelitos: I just found out that we have strftime_l() I will use that

ibriano: On Tue, 15 Mar 2016 11:12:51 -0700, Guilherme Íscaro wrote:
> > @@ -181,8 +182,31 @@ sol_http_request_get_method(const struct sol_http_request *request)
> >      return request->method;
> >  }
> >  
> > +static bool
> > +set_last_modified_header(struct MHD_Response *response, time_t last_modified)
> > +{
> > +    struct tm result;
> > +    char buf[128];
> > +    size_t r;
> > +
> > +    SOL_NULL_CHECK(gmtime_r(&last_modified, &result), false);
> > +
> > +    r = strftime(buf, sizeof(buf), "%a, %d %b %Y %H:%M:%S GMT", &result);
> 
> I just found out that we have strftime_l() I will use that

Also, %H:%M:%S == %T

cabelitos:
I will create a different PR that will use strftime_l(). I'm merging this one.

cabelitos:
merged

PR #1627, opened 2016-03-15
build: make it possible to declare OIC specs inside node dirs [v2]

glima:
So that it's easier for a new node, not related to our flow/oic, to get
in Soletta's tree and also benefit from our auto-generation (think of a
static build with a new OIC-capable node).

-----------

Changes since v1:
- make samples not affected anymore (even with the file move)
- all complaints addressed


barbieri:
-1 we shouldn't delete the resource json

src/modules/flow/oic/Makefile
```diff
@@ -1,3 +1,3 @@
 obj-$(FLOW_NODE_TYPE_OIC) += oic.mod
-obj-oic-$(FLOW_NODE_TYPE_OIC) := oic.json oic.o
+obj-oic-$(FLOW_NODE_TYPE_OIC) := core.light.json oic.json oic.o
```
ibriano: I guess the deletion of core.light.json is a mistake, but even without that, I don't like adding it here. The obj-*-y variable refers to things that will be generated. The flow modules have the .json there because that's the resulting json with the node types descriptions after passing through the node-gen script. It's not the source .json

tools/build/Makefile.rules
```diff
@@ -355,14 +360,29 @@ $(SOL_LIB_SO): $(builtin-objs)
...
 	$(Q)echo "     "GEN"   "$$@
 	$(Q)$(MKDIR) -p $(dir $($(1)-hdr))
 	$(Q)$(MKDIR) -p $(dir $($(1)-src))
 	$(Q)$(MKDIR) -p $(dir $($(1)-desc))
-	$(Q)$(PYTHON) $(NODE_TYPE_GEN_SCRIPT) --prefix=sol-flow-node-type --namespace=$($(1)-namespace) \
-		--genspec-schema=$(NODE_TYPE_SCHEMA) $(1) $($(1)-hdr) $($(1)-src) $($(1)-desc)
+	$(Q)$(PYTHON) $(NODE_TYPE_GEN_SCRIPT) \
+		--prefix=sol-flow-node-type --namespace=$($(1)-namespace) \
+		--genspec-schema=$(NODE_TYPE_SCHEMA) $(1) $($(1)-hdr) \
+		$($(1)-src) $($(1)-desc)
```
ibriano: More than ever, when dealing with our Makefiles, do NOT mix style changes with functional ones.

src/modules/flow/oic/Makefile
glima: It was by mistake, sorry, it had to be git added back here. I'll go with a different entry there specific for OIC templates, for now, then. I'll maybe wait for you big-phat-build-system patch to get in before mine gets (let's see if I figure out the damn samples bug soon, also).

tools/build/Makefile.rules
glima: OK, I won't anymore.

glima:
@ibriano, better now?

tools/build/Makefile.rules
```diff
@@ -369,6 +374,19 @@ $(SOL_LIB_SO): $(builtin-objs)
 	$(Q)$(LN) -fs $(notdir $(@).$(VERSION)) $(@)
 
 # generators
+define make-oic-gen
+$(addprefix $(addprefix $(1), $(notdir $(abspath $(1)))), .json) $(addprefix $(addprefix $(1), $(notdir $(abspath $(1)))), .c) $(addprefix $(addprefix $(1), $(notdir $(abspath $(1)))), -gen.c) $(addprefix $(addprefix $(1), $(notdir $(abspath $(1)))), .h): $(FLOW_OIC_GEN_SCRIPT)
```
ibriano: Why not process these things once, when you are adding the directory to oic-gens and then use those created variables here directly? You are replicating too much stuff below and it gets too error prone.

glima: Mmm, it could be. Tuning never ends...

glima:
I'll do that in a subsequent patch, landing this for now.

PR #1636, opened 2016-03-16
Bring us up to date with RIOT's API

ibriano:
And make tinycbor and OIC build for it.


tools/build/Kconfig.riot
```diff
@@ -9,5 +9,6 @@ config RIOT
     select FEATURE_WORKER_THREADS
     select FEATURE_NETWORK
     select FEATURE_COAP
+    select FEATURE_OIC
```
glima: This will require the FEATURE_IPV4 or something like that afterwards, but it's OK to get in nevertheless.

glima:
And it builds fine. :+1:

PR #1642, opened 2016-03-17
General improvements - V2

cabelitos:
Changes since v1:
* Fixing typos
*  Platform node - Register service monitor if service name changes.
* Sol-bus: always call bus_initialized function.


src/shared/sol-util.c
```diff
@@ -1063,3 +1063,40 @@ sol_util_double_equal(double var0, double var1)
...
+        buf->used += used;
+        return used;
+    }
+#endif
+
+    /**
+       Even with SOL_ATTR_STRFTIME() GCC still warns that the format parameter was not checked.
+       This is a known GCC bug: https://gcc.gnu.org/bugzilla/show_bug.cgi?id=39438
+       This cast was suggested in order to make the compiler happy.
+     */
```
glima: Weird, and it does not happen in the ``strftime_l()`` path?

src/modules/flow/platform/platform.json
```diff
@@ -608,8 +616,9 @@
         "members": [
           {
             "data_type": "string",
-            "description": "the platform service name",
-            "name": "service_name"
```
glima: Nice. There are other nodes with redundant names as options, like ``output_value``, just as a tip.

glima:
The bus patch deserves an extra +1, the rest should be fine.

src/shared/sol-util.c
cabelitos: Nope

src/lib/common/sol-platform-impl-systemd.c
```diff
@@ -53,9 +53,12 @@ struct ctx {
     struct sol_bus_client *timedate;
     struct sol_bus_client *hostname;
     struct sol_ptr_vector services;
+    bool locale_monitor_registered : 1;
+    bool timedate_monitor_registered : 1;
+    bool hostname_monitor_registered : 1;
 };
 
-static struct ctx _ctx;
+static struct ctx _ctx = { 0 };
```
vcgomes: This explicit initialization is unneeded, the `static` guarantees that it will be zero initialized.

src/lib/common/sol-platform-impl-systemd.c
```diff
@@ -743,29 +774,27 @@ int
...
-    bus = sol_bus_get(NULL);
-    SOL_NULL_CHECK(bus, -ENOTCONN);
-
-    _ctx.locale = sol_bus_client_new(bus, "org.freedesktop.locale1");
-    SOL_NULL_CHECK_GOTO(_ctx.locale, error);
+    bus = _get_sd_bus(_ctx.locale, _localed_bus_initialized);
+    SOL_NULL_CHECK(bus, -EINVAL);
 
-    return sol_bus_map_cached_properties(_ctx.locale, "/org/freedesktop/locale1",
+    r = sol_bus_map_cached_properties(_ctx.locale, "/org/freedesktop/locale1",
```
vcgomes: [off-topic] I wish I changed this API to return a handle, that I would pass to `_unmap_cached_properties()` :-/

vcgomes:
:+1: (just a minor improvement (optional)) on the `sol-bus` patch.

src/lib/common/sol-platform-impl-systemd.c
cabelitos: Fair.

cabelitos:
merged

PR #1644, opened 2016-03-17

sragavan:
changes.

Signed-off-by: Srinivasa Ragavan <srinivasa.ragavan.venkateswaran@intel.com>


poussa:
@barbieri @bdilly This is the first HW IO JS API. Who should review this and the other HW IO APIs once they appear?

bindings/nodejs/lib/gpio.js
```diff
@@ -0,0 +1,121 @@
...
+
+ var soletta = require( 'bindings' )( 'soletta' ),
+    _ = require( 'lodash' );
+
+var edges = {
+    "none": soletta.sol_gpio_edge.SOL_GPIO_EDGE_NONE,
+    "rising": soletta.sol_gpio_edge.SOL_GPIO_EDGE_RISING,
+    "falling": soletta.sol_gpio_edge.SOL_GPIO_EDGE_FALLING,
+    "any": soletta.sol_gpio_edge.SOL_GPIO_EDGE_BOTH,
+}
```
barbieri: I've talked to guys before that my preferred way is to do string<->enum mappings in the C header files, then we make it uniform and can be used by C application developers (i.e.: debug), bindings other than yours.

That said, you can add a issue so we add the function, or you can add it right away. 

We can follow s pattern like <enum_name>_to_string() and from_string(). Then your code generator can use it.

bindings/nodejs/lib/gpio.js
```diff
@@ -0,0 +1,121 @@
...
+    return new Promise( function( fulfill, reject ) {
+        var pin = init.pin;
+        var dir = 0;
+        var drive_mode = 0;
+        var config = null;
+
+        if ( init.pullup )
+            drive_mode = soletta.sol_gpio_drive.SOL_GPIO_DRIVE_PULL_UP;
+        else
+            drive_mode = soletta.sol_gpio_drive.SOL_GPIO_DRIVE_PULL_DOWN;
```
barbieri: It's not a Boolean but an enum with 3 states, I'm not sure but there might be a reason, see http://solettaproject.github.io/docs/c-api/group__GPIO.html#ga9b8be46f5c047916a5cb7ecc51cf9ec6

@bdilly @ibriano or @mbelluzzo can you confirm and either change to Boolean or improve the C documentation?

bindings/nodejs/lib/gpio.js
```diff
@@ -0,0 +1,121 @@
...
+    "falling": soletta.sol_gpio_edge.SOL_GPIO_EDGE_FALLING,
+    "any": soletta.sol_gpio_edge.SOL_GPIO_EDGE_BOTH,
+}
+
+function pin_read_cb( pin, value ) {
+    _pins[ pin ].dispatchEvent( "change", {
+                        type: "change",
+                        value: value
+    } );
+}
```
barbieri: Why not make this function enclosed in the context it is used, then you'd have enough context to dispatch the event, not requiring to use the array.

bindings/nodejs/lib/gpio.js
```diff
@@ -0,0 +1,121 @@
...
+
+        } else {
+            config = {
+                dir: soletta.sol_gpio_direction.SOL_GPIO_DIR_OUT,
+                active_low: init.activeLow,
+                drive_mode: drive_mode,
+            }
+        }
+
+        fulfill( GPIOPin( soletta.sol_gpio_open( pin, config ) ) );
```
barbieri: Like declare a local pin_read_cb here that would have access to GPIOin() results... Then no need to keep the array bookkeeping. Or give your callback some extra data, like the C api.

bindings/nodejs/lib/gpio.js
```diff
@@ -0,0 +1,121 @@
...
+    onchange: null,
+
+    read: function() {
+        return new Promise( _.bind( function( fulfill, reject ) {
+
+            fulfill( soletta.sol_gpio_read( this._pin ) );
+        }, this ) );
+    },
+
+    write: function( value ) {
```
barbieri: Is it usual to have both in/out in the same prototype? In Python class I'd create a base class with close, a subclass with Read and a different with Write

bindings/nodejs/src/functions/gpio.cc
```diff
@@ -0,0 +1,113 @@
...
+NAN_METHOD(bind_sol_gpio_open) {
+    VALIDATE_ARGUMENT_COUNT(info, 2);
+    VALIDATE_ARGUMENT_TYPE_OR_NULL(info, 0, IsUint32);
+    VALIDATE_ARGUMENT_TYPE(info, 1, IsObject);
+
+    uint32_t pin;
+    sol_gpio_config config;
+    sol_gpio *gpio = NULL;
+
+    pin = info[0]->Uint32Value();
```
barbieri: Number is not required if you use "open_by_label()" function, then you can see how to map that to JS, but some boards expose a pin name like "A1", this depends on the pin muxer on some boards.

This reminds me of @mbelluzzo: do we use the kernel exported names as well? Recall that pinctl can provide us names? We should use pinmux or that!

bindings/nodejs/src/structures/handles.cc
```diff
@@ -0,0 +1,42 @@
...
+    if (!fillCArrayFromJSArray(((unsigned char *)&local), sizeof(cType),    \
+                             jsHandle)) {   \
+        return false;   \
+    }   \
+    *destination = local;   \
+    return true;
+
+Local<Array> js_sol_gpio(sol_gpio *handle) {
+    return jsArrayFromBytes(((unsigned char *)(&handle)),
+                            sizeof(sol_gpio *));
```
barbieri: In Python bindings I usually go with "unsigned long" as it would fit the pointer and is cheaper in resources.

bindings/nodejs/src/structures/handles.h
```diff
@@ -0,0 +1,31 @@
...
+ * limitations under the License.
+ */
+
+#ifndef __SOLETTA_JS_HANDLES_H__
+#define __SOLETTA_JS_HANDLES_H__
+
+#include <v8.h>
+extern "C" {
+#include <sol-gpio.h>
+}
```
barbieri: We must do this properly in our C side. If not let us know.

bindings/nodejs/src/structures/sol-js-gpio.cc
```diff
@@ -0,0 +1,87 @@
...
+    Local<Value> arguments[2] = {
+        js_sol_gpio(gpio),
+        Nan::New(value)
+    };
+    callback->Call(2, arguments);
+}
+
+bool c_sol_gpio_config(v8::Local<v8::Object> jsGPIOConfig, sol_gpio_config *p_config) {
+    sol_gpio_config config = {
+        SOL_GPIO_CONFIG_API_VERSION,
```
barbieri: While I don't expect someone using JS to disable Ali version to save memory, the good practice is to isolate in ifdef or use the SOL_SET_API() macro (I'm on my phone, so the name may be slightly different)

bindings/nodejs/src/structures/sol-js-gpio.cc
```diff
@@ -0,0 +1,87 @@
...
+
+        if (read_cb->IsFunction()) {
+            Nan::Callback *callback =
+                new Nan::Callback(Local<Function>::Cast(read_cb));
+            config.in.cb = sol_gpio_read_callback;
+            config.in.user_data = callback;
+        }
+    }
+
+    memcpy(p_config, &config, sizeof(sol_gpio_config));
```
barbieri: Since you already return the true/false, why not just work on the given pointer? What's good in doing the copy?

bindings/nodejs/lib/gpio.js
sragavan: @barbieri  the JS spec mentions that pullup is boolean. 
https://github.com/zolkis/soletta/blob/master/doc/js-spec/gpio.md 

cc @zolkis

ibriano: On Mon, 21 Mar 2016 11:42:21 -0700, Srinivasa Ragavan wrote:
> > +    } );
> > +}
> > +
> > +var _pins = new Array();
> > +
> > +exports.open = function( init ) {
> > +    return new Promise( function( fulfill, reject ) {
> > +        var pin = init.pin;
> > +        var dir = 0;
> > +        var drive_mode = 0;
> > +        var config = null;
> > +
> > +        if ( init.pullup )
> > +            drive_mode = soletta.sol_gpio_drive.SOL_GPIO_DRIVE_PULL_UP;
> > +        else
> > +            drive_mode = soletta.sol_gpio_drive.SOL_GPIO_DRIVE_PULL_DOWN;
> 
> @barbieri  the JS spec mentions that pullup is boolean. 
> https://github.com/zolkis/soletta/blob/master/doc/js-spec/gpio.md 
> 

At the C level, it can be either UP, DOWN or NONE. I don't know enough
about the electronics level to tell if those values make sense for both
input and output GPIOs or if what makes sense depends on the direction.

bindings/nodejs/lib/gpio.js
sragavan: Hmm, Im not sure if that is possible. Because while we define the callback, the object isn't created and it is just an array for config that we create which we pass later.  @barbieri  @nagineni just correct me if Im wrong.

Do you mean:

fulfill( GPIOPin( soletta.sol_gpio_open( pin, { 
                                 active_low: init.activeLow,
                                 poll_timeout: init.poll,
                                 drive_mode: drive_mode,
                                 trigger_mode: edge,
                                 callback: function(value) {
                                                      this.dispatchEvent( "change", {
                                                      type: "change",
                                                      value: value });
                                  }
} ) ) );


Im not sure, if that 'this' works as expected. Ill give a check.

bindings/nodejs/lib/gpio.js
sragavan: I wrote iotivity-node(js) using the similar pattern. I guess, it is normal, unless it breaks the way other modules are written.

bindings/nodejs/src/functions/gpio.cc
sragavan: https://github.com/zolkis/soletta/blob/master/doc/js-spec/gpio.md

@barbieri the GPIO js spec doesn't mention the use case of label. For AIO it exposes the _label which is a different impl altogether.

cc @zolkis

bindings/nodejs/src/structures/handles.cc
sragavan: tbh, Im following what we did on iotivity-node 

@gabrielschulhof had some ideas here. Ill check with him for sure.

bindings/nodejs/src/structures/sol-js-gpio.cc
sragavan: Ill fix this.

bindings/nodejs/src/structures/sol-js-gpio.cc
sragavan: Ill fix this.

bindings/nodejs/src/structures/handles.h
sragavan: Ill fix this.

bindings/nodejs/lib/gpio.js
sragavan: Ill fix this.

bindings/nodejs/lib/gpio.js
barbieri: another option would be to set after-fact, like this:

```js
var some_mutable_object = [];
g = GPIOPin(pin, ..., callback: function (value) { some_mutable_object[0].dispatch(...);...);
some_mutable_object.push(g);
fulfill(g);
```

bindings/nodejs/src/functions/gpio.cc
zolkis: No problem we can change ```pin``` to ```DOMString``` from integer in the spec. I wonder why didn't I do that from start.

bindings/nodejs/lib/gpio.js
zolkis: At HW level it can be either up or down, but at SW level it may be unknown. In JS that may translate to ```undefined```.

sragavan:
@barbieri New patch, fixes the comments mentioned above and rebased to the latest version. Please review and let me know. Thanks!

bindings/nodejs/src/structures/handles.cc
gabrielschulhof: I don't like assuming that the size of a pointer is in any way related to the size of an integer. In JS this is especially true since the maximum integer size is 32 bits.

sragavan:
@gabrielschulhof Ill close this and open a fresh one. Thanks for the info.

bindings/nodejs/src/structures/handles.cc
barbieri: that 's why long exists... or to be more standard "uintptr_t".

PR #1652, opened 2016-03-18
nodejs: Add NODE_GYP_FLAGS to specify additional flags to node-gyp

nagineni:
Additional flags like --arch and --proxy needs to passed to node-gyp
when building JS bindings for different targets inside Yocto. So, adding
NODE_GYP_FLAGS to specify those additional flags.


nagineni:
Better fix proposal in PR #1653. Closing this.

PR #1653, opened 2016-03-18
nodejs: Set NODE_GYP only if it doesn't have a value already

nagineni:
This helps us to specify additional flags that are needed for building JS bindings for different targets using node-gyp.


gabrielschulhof:
:+1:

cabelitos:
+1

PR #1680, opened 2016-03-21
Fixes - Invalid memory access

ceolin:
These two modules were writing in read only memory

### DIFF
--- a/src/modules/flow/update/update.c
+++ b/src/modules/flow/update/update.c
@@ -30,20 +30,15 @@
 
 struct update_data {
     struct sol_update_handle *handle;
-};
-
-struct update_node_type {
-    struct sol_flow_node_type base;
     uint16_t progress_port;
 };
 
 static int
 check_open(struct sol_flow_node *node, void *data, const struct sol_flow_node_options *options)
 {
-    struct update_node_type *type;
+    struct update_data *mdata = data;
 
-    type = (struct update_node_type *)sol_flow_node_get_type(node);
-    type->progress_port = SOL_FLOW_NODE_TYPE_UPDATE_CHECK__OUT__PROGRESS;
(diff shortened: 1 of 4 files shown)


edersondisouza:
Hmmm... So content returned by `sol_flow_get_node_type` was read only? Does this PR fix all uses? Is this information, about content returned by `sol_flow_get_node_type` being read only documented?

src/modules/flow/update/update.c
```diff
@@ -30,20 +30,15 @@
 
 struct update_data {
     struct sol_update_handle *handle;
-};
-
-struct update_node_type {
```
cabelitos: I do not agree with this one. You can still leave the progress_port at the update_node_type and do the following in the json spec. Check below.

src/modules/flow/update/update.json
```diff
@@ -40,12 +40,6 @@
         "close": "common_close",
         "open": "check_open"
       },
-      "node_type": {
-        "access": [
-          "base"
-        ],
-        "data_type": "struct update_node_type"
```
cabelitos: Add:
"extra_methods": {
 "progress_port" : "SOL_FLOW_NODE_TYPE_UPDATE_INSTALL__OUT__PROGRESS"
}

cabelitos:
The first commit looks ok

src/modules/flow/update/update.json
ceolin: this is a hack. you are playing with something that clearly was not created to do. I really liked !!

ceolin:
@edersondisouza i found only these two wrong usages. I'll check the documentation and fix if necessary.

src/modules/flow/update/update.json
cabelitos: For me this was created to init the type fields and that's what you doing.

ceolin:
fixes on https://github.com/solettaproject/soletta/pull/1681

PR #1681, opened 2016-03-21
Fixes v2 - Invalid memory access 

ceolin:
from v1:
 - use `extra_methods` to set the progress port in update nodes

### DIFF
--- a/src/modules/flow/oauth/oauth.c
+++ b/src/modules/flow/oauth/oauth.c
@@ -50,30 +50,29 @@ struct v1_request_data {
     char *timestamp, *nonce, *callback_url;
 };
 
-struct oauth_node_type {
-    struct sol_flow_node_type base;
+static struct {
     struct sol_http_server *server;
-    uint16_t server_ref;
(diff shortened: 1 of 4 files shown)


glima:
:+1:  for the 1st patch.

cabelitos:
+1

src/modules/flow/oauth/oauth.c
```diff
@@ -50,30 +50,29 @@ struct v1_request_data {
...
+static struct {
     struct sol_http_server *server;
-    uint16_t server_ref;
-};
+    uint16_t ref;
+} oauth;
 
 static int
-server_ref(struct oauth_node_type *oauth)
+ref(void)
```
vcgomes: [minor] I would continue with the `server_ref()` name, at least to maintain consistency with `server_unref()`

vcgomes:
typo: s/udpate/update' on the subject of the second patch.

ceolin:
merged with the suggestions

PR #1682, opened 2016-03-21
Http sse v2

ceolin:
from v1:
 - Fixes on sample
 - Change variables name on `sol_buffer_append_buffer`

### DIFF
--- a/src/lib/comms/include/sol-http-server.h
+++ b/src/lib/comms/include/sol-http-server.h
@@ -68,6 +68,16 @@ struct sol_http_server;
 struct sol_http_request;
 
 /**
+ * @struct sol_http_response_sse
+ * @brief Opaque handler for response using server-sent-event.
+ *
+ * This response is create with @c sol_http_server_send_response_sse
+ * and data can be given using @c sol_http_server_reponse_sse_send_data,
+ * to delete it use @c sol_http_server_reponse_ss2_del.
+ */
+struct sol_http_response_sse;
+
+/**
  * @brief Creates a HTTP server, binding on all interfaces
  * in the specified @a port.
  *
@@ -206,6 +216,50 @@ int sol_http_server_set_last_modified(struct sol_http_server *server, const char
  * @return @c 0 on success, error code (always negative) otherwise.
  */
 int sol_http_server_send_response(struct sol_http_request *request, struct sol_http_response *response);
+
+/**
+ * @brief Send the response and keep connection alive to request given in the
+ * callback registered on @c sol_http_server_register_handler.
+ *
+ * After this call, the caller will be resposible to close the connection calling
+ * @c sol_http_response_sse_del. @note all the necessary headers are set by this function.
+ *
+ * @param request The request given on the callback of
+ * @c sol_http_server_register_handler.
+ * @param cb Callback called when the connection is closed from the client side.
+ * @param cb_data The data pointer to be passed to @a cb.
+ *
+ * @return @c sol_http_response_sse on success, @c NULL otherwise.
+ *
+ * @see sol_http_response_sse_del
+ * @see sol_http_response_sse_send_data
+ */
+struct sol_http_response_sse *sol_http_server_send_response_sse(struct sol_http_request *request,
+    void (*cb)(void *data, const struct sol_http_response_sse *sse),
+    const void *cb_data);
+
+/**
+ * @brief Send data for the sse response.
+ *
+ * @return @c 0 on success, error code (always negative) otherwise.
+ *
+ * @see sol_http_response_sse_del
+ * @see sol_http_server_send_response_sse
+ */
+int sol_http_response_sse_send_data(struct sol_http_response_sse *sse, struct sol_buffer *buffer);
+
+/**
+ * @brief Delete the sse response.
+ *
+ * This function deletes the sse response and its resources and closes
+ * the connection.
+ *
+ * @return @c 0 on success, error code (always negative) otherwise.
+ *
+ * @see sol_http_server_send_response_sse
+ * @see sol_http_response_sse_send_data
+ */
+int sol_http_response_sse_del(struct sol_http_response_sse *sse);
 
 /**
  * @brief Gets the URL from a given request.
--- a/src/lib/comms/sol-http-server-impl-microhttpd.c
+++ b/src/lib/comms/sol-http-server-impl-microhttpd.c
@@ -28,6 +28,7 @@
 #include <sys/types.h>
 #include <unistd.h>
 
+#include "sol-buffer.h"
 #include "sol-http-server.h"
 #include "sol-log.h"
 #include "sol-mainloop.h"
@@ -68,6 +69,15 @@ struct sol_http_request {
     time_t if_since_modified;
     time_t last_modified;
     bool is_multipart;
+};
+
+struct sol_http_response_sse {
+    struct MHD_Connection *connection;
+    void (*cb)(void *data, const struct sol_http_response_sse *sse);
+    const void *cb_data;
+    struct sol_buffer buffer;
+    bool delete_me;
+    bool suspended;
 };
 
 struct static_dir {
@@ -206,17 +216,12 @@ set_last_modified_header(struct MHD_Response *response, time_t last_modified)
 }
 
 static struct MHD_Response *
-build_mhd_response(const struct sol_http_response *response, time_t last_modified)
+build_mhd_response_params(struct MHD_Response *r, const struct sol_http_response *response,
+    time_t last_modified)
 {
     struct sol_buffer buf;
     uint16_t idx;
-    struct MHD_Response *r;
     struct sol_http_param_value *value;
-
-    r = MHD_create_response_from_buffer(response->content.used, response->content.data,
-        MHD_RESPMEM_MUST_COPY);
-    if (!r)
-        return NULL;
 
     sol_buffer_init(&buf);
     SOL_HTTP_PARAMS_FOREACH_IDX (&response->param, value, idx) {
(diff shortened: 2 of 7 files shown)


src/lib/comms/include/sol-http-server.h
```diff
@@ -68,6 +68,16 @@ struct sol_http_server;
...
 
 /**
+ * @struct sol_http_response_sse
+ * @brief Opaque handler for response using server-sent-event.
+ *
+ * This response is create with @c sol_http_server_send_response_sse
+ * and data can be given using @c sol_http_server_reponse_sse_send_data,
+ * to delete it use @c sol_http_server_reponse_ss2_del.
+ */
+struct sol_http_response_sse;
```
barbieri: I do not like to name this so specifically after `sse`. Usually dynamic pages, a big bulk of internet content, is not "one go" responses, but rather they are created interactively. Old PHP, Perl or so called CGI-BIN were like that, you started to reply with "print" to stdout (which was redirected to socket), then you did some DB query, then you printed each line, and so on. Contrast this with the API we support at the moment, where you must create a fixed content and send it in one go.

If you have fixed amount of memory, like 5Mb, you cannot serve a big content using the "fixed" response, since you'd end without memory after 5Mb. However with dynamic responses you progressively dispatch chunks and 5Mb will be enough to deliver a 100Mb contents extracted from DB (see above).

Then we can call this "dynamic" or "progressive". @lpereira any suggestion of a good name?

`sol_http_response_progressive`
`sol_http_response_dynamic`

maybe progressive represents best.

src/lib/comms/include/sol-http-server.h
```diff
@@ -208,6 +218,50 @@ int sol_http_server_set_last_modified(struct sol_http_server *server, const char
...
+ * @c sol_http_server_register_handler.
+ * @param cb Callback called when the connection is closed from the client side.
+ * @param cb_data The data pointer to be passed to @a cb.
+ *
+ * @return @c sol_http_response_sse on success, @c NULL otherwise.
+ *
+ * @see sol_http_response_sse_del
+ * @see sol_http_response_sse_send_data
+ */
+struct sol_http_response_sse *sol_http_server_send_response_sse(struct sol_http_request *request,
```
barbieri: as above I'd not call it "see", but rather take a `sol_http_response` and return a progressive handle to it. Like:
```c
struct sol_http_response_progressive *sol_http_server_send_progressive_response(struct sol_http_request *request, struct sol_http_response *response, void (*on_close)(void *data.. ));
```

This would not set any SSE specific header, these must be done by the user. You can offer a helper to set the standard http headers used by SSE applications as a separate function (static inline in the header)

Extra comments:
 - why is `response` non-const in send_response?
 - rename `cb` to `on_close`.

src/lib/comms/include/sol-http-server.h
```diff
@@ -208,6 +218,50 @@ int sol_http_server_set_last_modified(struct sol_http_server *server, const char
...
+ *
+ * This function deletes the sse response and its resources and closes
+ * the connection.
+ *
+ * @return @c 0 on success, error code (always negative) otherwise.
+ *
+ * @see sol_http_server_send_response_sse
+ * @see sol_http_response_sse_send_data
+ */
+int sol_http_response_sse_del(struct sol_http_response_sse *sse);
```
barbieri: are we using return on delete functions?

src/lib/comms/include/sol-http-server.h
barbieri: also, cross-link with `@see` doxygen command the two ways to send a response (progressive x fixed).

src/lib/comms/sol-http-server-impl-microhttpd.c
```diff
@@ -270,6 +275,100 @@ build_mhd_response(const struct sol_http_response *response, time_t last_modifie
...
+    if (sse->delete_me)
+        return MHD_CONTENT_READER_END_OF_STREAM;
+
+    if (!sse->buffer.used) {
+        MHD_suspend_connection(sse->connection);
+        sse->suspended = true;
+        return 0;
+    }
+
+    len = sse->buffer.used > size ? size : sse->buffer.used;
```
barbieri: why not use the min macro?

src/lib/comms/sol-http-server-impl-microhttpd.c
```diff
@@ -270,6 +275,100 @@ build_mhd_response(const struct sol_http_response *response, time_t last_modifie
...
+    memcpy(buf, sse->buffer.data, len);
```
barbieri: I'd say we need to investigate what MHD does internally... since if they ask you for a fixed size buffer to be copied, very likely we're doing double-buffering. One option would be to disable MHD and keep only our buffer, with MHD trying to write based on its full length and then informing the result of `write(2)` syscall so we can update.

check the stack when this function is called, should give you some hint on how to achieve that.

src/lib/comms/sol-http-server-impl-microhttpd.c
```diff
@@ -270,6 +275,100 @@ build_mhd_response(const struct sol_http_response *response, time_t last_modifie
...
+
+    r = sol_buffer_remove_data(&sse->buffer, len, 0);
+    SOL_INT_CHECK(r, < 0, MHD_CONTENT_READER_END_WITH_ERROR);
+
+    return len;
+}
+
+static struct MHD_Response *
+build_mhd_response_sse(struct sol_http_response_sse *sse)
```
barbieri: make this a static inline helper in the header that given a sol_http_response you add the headers to it using our methods.

src/lib/comms/sol-http-server-impl-microhttpd.c
```diff
@@ -959,6 +1058,81 @@ sol_http_server_send_response(struct sol_http_request *request, struct sol_http_
...
+    const void *cb_data)
+{
+    int ret;
+    struct sol_http_response_sse *sse;
+    struct MHD_Response *mhd_response;
+
+    SOL_NULL_CHECK(request, NULL);
+    SOL_NULL_CHECK(request->connection, NULL);
+
+    sse = malloc(sizeof(*sse));
```
barbieri: use calloc to avoid problems, like here we are not setting "suspended" and valgrind should bitch about it :-)

src/lib/comms/sol-http-server-impl-microhttpd.c
```diff
@@ -959,6 +1058,81 @@ sol_http_server_send_response(struct sol_http_request *request, struct sol_http_
...
+
+err:
+    free(sse);
+    return NULL;
+}
+
+SOL_API int
+sol_http_response_sse_del(struct sol_http_response_sse *sse)
+{
+    SOL_NULL_CHECK(sse, -EINVAL);
```
barbieri: maybe also check `delete_me` to avoid "double-free"? It won't fail and hint at valgrind as the memory may still be alive, so if we can detect the `delete_me` was already set, we can report to user.

src/lib/comms/include/sol-http-server.h
```diff
@@ -208,6 +218,50 @@ int sol_http_server_set_last_modified(struct sol_http_server *server, const char
...
+
+/**
+ * @brief Send data for the sse response.
+ *
+ * @return @c 0 on success, error code (always negative) otherwise.
+ *
+ * @see sol_http_response_sse_del
+ * @see sol_http_server_send_response_sse
+ */
+int sol_http_response_sse_send_data(struct sol_http_response_sse *sse, struct sol_buffer *buffer);
```
barbieri: people usually call these functions `feed`, it's shorter and usually implies a progressive action. I've used that in crypto's message digest.

barbieri: also `const` the buffer, document the parameters and make sure to say the buffer will not be referenced, only its data will be copied. In this sense I'm thinking if we should use a buffer or a slice, all we need is the pointer and size, there is a nice `get_slice()` function for buffer.

src/samples/http/server-sse.c
```diff
@@ -0,0 +1,210 @@
...
+    if (flags & (SOL_FD_FLAGS_ERR | SOL_FD_FLAGS_HUP)) {
+        fprintf(stderr, "ERROR: Something wrong happened with file descriptor: %d\n", fd);
+        goto err;
+    }
+
+    if (flags & SOL_FD_FLAGS_IN) {
+        int err;
+
+        value.used = 0;
+        err = sol_buffer_insert_bytes(&value, 0,
```
barbieri: why send this to the intermediate buffer? Just feed it to the http response.

src/samples/http/server-sse.c
```diff
@@ -0,0 +1,210 @@
...
+            (uint8_t *)"data: ", sizeof("data: ") - 1);
+        if (err < 0) {
+            fprintf(stderr, "ERROR: failed to append the data prefix: %s\n",
+                sol_util_strerrora(-err));
+            goto err;
+        }
+        /* this will loop trying to read as much data as possible to buffer. */
+        err = sol_util_load_file_fd_buffer(fd, &value);
```
barbieri: this will fail if you get `\n` in the input... One option is to loop using `strchr` to find the lines, then send multiple `data:`, remove the `\n` or encode it somehow. It seems that SSE supports sending `\n` if you start the next line with `data:`

src/samples/http/server-sse.c
```diff
@@ -0,0 +1,210 @@
...
+        if (err < 0) {
+            fprintf(stderr, "ERROR: failed to read from stdin: %s\n",
+                sol_util_strerrora(-err));
+            goto err;
+        }
+
+        if (value.used == 0) {
+            /* no data usually means ^D on the terminal, quit the application */
+            printf("no data on stdin, quitting.\n");
+            sol_quit();
```
barbieri: this won't work, if the server did not send all the data, the mainloop will exit and client won't receive it all.

You should call `sol_quit()` from your callback to response "on_close". And here is a problem with your implementation: when user explicitly called `sse_response_del()`, you do not call his callback, then there is no way to know when the data was fully sent to client. Actually, since you mark `delete_me`, it will exit even if not all data was sent. This whole thing must be reworked so we can abort (close immediately) or close after flush/done.

src/samples/http/server-sse.c
```diff
@@ -0,0 +1,210 @@
...
+        } else {
+            err = sol_buffer_append_bytes(&value, (uint8_t *)"\n\n", sizeof("\n\n") - 1);
```
barbieri: ugly, use the slice and macros

src/samples/http/server-sse.c
```diff
@@ -0,0 +1,210 @@
...
+            if (err < 0) {
+                fprintf(stderr, "ERROR: Could not append data to buffer: %s\n",
+                    sol_util_strerrora(-err));
+                goto err;
+            }
+        }
+    }
+
+    SOL_PTR_VECTOR_FOREACH_IDX (&responses, sse, i)
+        sol_http_response_sse_send_data(sse, &value);
```
barbieri: no, this is far from real world usage scenarios, you want to send data when it's available. No need to keep this vector and buffers... which are leaking

barbieri:
reviewed, I think the SSE should go from name and this should be made a "progressive" response, which people can use to do SSE.

Also convert the http-server flow nodes to do it, should be a nice test. We can then do the http-client in another PR. The inspector could be also in a different PR, since it will be more work.

src/lib/comms/include/sol-http-server.h
lpereira: In Golang, there's the `Flusher` interface that the `http.ResponseWriter` object might implement. This interface has only one method, `Flush()`, which sends the currently-buffered response to the socket. I like this name, it's similar to other streaming APIs, such as `fflush()`.

src/lib/comms/sol-http-server-impl-microhttpd.c
```diff
@@ -270,6 +275,100 @@ build_mhd_response(const struct sol_http_response *response, time_t last_modifie
...
+        SOL_WRN("Could not add the header: Connection");
+        goto err;
+    }
+
+    if (MHD_add_response_header(r, MHD_HTTP_HEADER_CACHE_CONTROL, "no-cache") == MHD_NO) {
+        SOL_WRN("Could not add the header: Cache-Control");
+        goto err;
+    }
+
+    if (MHD_add_response_header(r, MHD_HTTP_HEADER_ACCESS_CONTROL_ALLOW_ORIGIN, "*") == MHD_NO) {
```
lpereira: `*` is a dangerous default for CORS. Take a look at some [best practices](https://www.owasp.org/index.php/HTML5_Security_Cheat_Sheet#Cross_Origin_Resource_Sharing) here.

src/lib/comms/sol-http-server-impl-microhttpd.c
ceolin: I forgot about this

src/lib/comms/sol-http-server-impl-microhttpd.c
ceolin: ack

src/lib/comms/sol-http-server-impl-microhttpd.c
ceolin: i've added `suspended` after i've written it and forgot to change. I will do it.

src/samples/http/server-sse.c
ceolin: no reason, it can be done directly

src/samples/http/server-sse.c
ceolin: explained personally, this code is to broadcast the data

src/lib/comms/include/sol-http-server.h
ceolin: I will rename to `on_del` because microhttpd calls the internal callback when the socket is deleted and not closed. Also the function exposed to the user is `sol_http_response_progressive_del`

src/lib/comms/sol-http-server-impl-microhttpd.c
ceolin: as we have talked, this buffer is given by microhttpd and there is no way (now) to we pass our own buffer, so the copy is necessary.

src/samples/http/server-sse.c
ceolin: actually there is a reason, i have to broadcast this buffer to all connections, so I decided to do this in the end.

src/samples/http/server-sse.c
ceolin: ok

src/lib/comms/sol-http-server-impl-microhttpd.c
ceolin: removed, thanks

src/samples/http/server-sse.c
ceolin: fixed.

src/lib/comms/sol-http-server-impl-microhttpd.c
ceolin: we do not have such macro ...

src/samples/http/server-sse.c
ceolin: this is a sample, i guess this will bring more complexity than necessary to understand it. The normal usage of this will be run from terminal in canonical mode, it meas that the data will come when return is pressed (no \n).  To check the \n properly, it's necessary split the string into tokens then prepend `data: ` and append `\n\n`.

ceolin:
https://github.com/solettaproject/soletta/pull/1733

Issue #1660, opened 2016-03-18
API Changes: Sol-Buffer


cabelitos:
#### Task Description

In order to keep our API consistent with soletta's style, some changes are required in Sol-Buffer.

- [ ] Fix typo in SOL_DECODE_UPERCASE  -  it should be UPPERCASE.
- [ ] In `sol_buffer_remove_data()` swap the arguments order  (`size` and `offset`)

#### Dependencies
None.

kaspersky:
#1692

PR #1692, opened 2016-03-22
buffer: api change

kaspersky:
Fix issue #1660

### DIFF
--- a/src/lib/datatypes/include/sol-buffer.h
+++ b/src/lib/datatypes/include/sol-buffer.h
@@ -894,7 +894,7 @@ int sol_buffer_ensure_nul_byte(struct sol_buffer *buf);
  * @note the buffer keeps its capacity after this function, it means,
  * the data is not released. If that is wanted, one should call @ref sol_buffer_trim
  */
-int sol_buffer_remove_data(struct sol_buffer *buf, size_t size, size_t offset);
+int sol_buffer_remove_data(struct sol_buffer *buf, size_t offset, size_t size);
 
 /**
  * @}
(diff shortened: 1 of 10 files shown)


cabelitos:
+1

src/lib/datatypes/include/sol-buffer.h
```diff
@@ -894,7 +894,7 @@ int sol_buffer_ensure_nul_byte(struct sol_buffer *buf);
  * @note the buffer keeps its capacity after this function, it means,
  * the data is not released. If that is wanted, one should call @ref sol_buffer_trim
  */
-int sol_buffer_remove_data(struct sol_buffer *buf, size_t size, size_t offset);
+int sol_buffer_remove_data(struct sol_buffer *buf, size_t offset, size_t size);
```
otaviobp: Please, change the order of size and offset in comments too. Doxygen works fine, but it is easier to read in .h files if the order of parameters is correct.

src/test/test-buffer.c
```diff
@@ -781,23 +781,23 @@ test_remove_data(void)
     ASSERT_INT_EQ(buf.used, strlen("ABCDEFGHI"));
     ASSERT_STR_EQ(buf.data, "ABCDEFGHI");
 
-    err = sol_buffer_remove_data(&buf, strlen("ABC"), 0);
+    err = sol_buffer_remove_data(&buf, 0, strlen("ABC"));
     ASSERT_INT_EQ(err, 0);
     ASSERT_INT_EQ(buf.used, strlen("DEFGHI"));
 
-    err = sol_buffer_remove_data(&buf, strlen("GHI"), 3);
+    err = sol_buffer_remove_data(&buf, 0, strlen("GHI"));
```
otaviobp: why are you using 0 instead of 3?

src/lib/datatypes/include/sol-buffer.h
kaspersky: Will do.

src/test/test-buffer.c
kaspersky: Mistake, I will fix it.

Issue #1661, opened 2016-03-18
Add tests for sol_list


cabelitos:
#### Task Description

Currently sol_list does not have a test suit. One should be created.

- [ ] Write tests for sol_list - It should cover the whole API.
- [ ] Bonus task: Fix `SOL_LIST_INIT` Macro, since it causes segfault. Right know no one is using it, however we should provide a macro that works properly

#### Dependencies
None.

amandawrcoelho:
@ceolin we can close this task

PR #1709, opened 2016-03-23
Sol-list: Add test suite

kaspersky:
#1661


kaspersky:
The SOL_LIST_INIT segfault comment was too vague, so since sol-list resembles linux kernel lists implementation very much I made it similar (there is now a usage of the macro in the list test).

PR #1710, opened 2016-03-23
Sol-list: Add test suite

kaspersky:
#1661

### DIFF
--- a/src/lib/datatypes/include/sol-list.h
+++ b/src/lib/datatypes/include/sol-list.h
@@ -66,9 +66,15 @@ struct sol_list {
 
 /**
  * @def SOL_LIST_INIT
- * @brief Helper macro to initialize a @c sol_list structure.
+ * @brief Helper macro to initialize a @c sol_list structure. Example:
+ *
+ * @code
+ * struct sol_list mylist = SOL_LIST_INIT(mylist);
+ * @endcode
+ *
+ * @param name List to be initialized
  */
-#define SOL_LIST_INIT { NULL, NULL }
+#define SOL_LIST_INIT(name) { &(name), &(name) }
 
 /**
  * @def SOL_LIST_GET_CONTAINER(list, type, member)
--- a/src/test/test-list.c
+++ b/src/test/test-list.c
@@ -1,0 +1,108 @@
+/*
+ * This file is part of the Soletta Project
+ *
+ * Copyright (C) 2015 Intel Corporation. All rights reserved.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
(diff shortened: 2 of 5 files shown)


kaspersky:
The SOL_LIST_INIT segfault comment was too vague, so since sol-list resembles linux kernel lists implementation very much I made it similar (there is now a usage of the macro in the list test).

cabelitos:
lgtm

barbieri:
@cabelitos I guess he doesn't have commit rights, so would you rebase/push?

src/lib/datatypes/include/sol-list.h
```diff
@@ -66,9 +66,15 @@ struct sol_list {
...
+ * @brief Helper macro to initialize a @c sol_list structure. Example:
+ *
+ * @code
+ * struct sol_list mylist = SOL_LIST_INIT(mylist);
+ * @endcode
+ *
+ * @param name List to be initialized
  */
-#define SOL_LIST_INIT { NULL, NULL }
+#define SOL_LIST_INIT(name) { &(name), &(name) }
```
ceolin: This is partially right. This is a kind of inlist, it means that this struct should be used in other struct to provide lists. There is no utiliity in use  it directly, as you can see the struct has two fields `prev` and `next` that with this macro will point to the same address. The usage of this is:

```
struct delayed_packet {
    struct sol_list list;
    struct sol_flow_packet *packet;
    uint16_t source_idx;
    uint16_t source_port_idx;
};
```

It said, i guess it is better initialize with the value given by the user and not get the address.

src/test/test-list.c
```diff
@@ -0,0 +1,108 @@
...
+        ASSERT_INT_EQ(entry->data, i++);
+
+        sol_list_remove(it);
+        free(entry);
+    }
+
+    ASSERT(sol_list_is_empty(&list2));
+}
+
+TEST_MAIN();
```
ceolin: :+1:  for test. Nice job

src/lib/datatypes/include/sol-list.h
kaspersky: @ceolin, I don't fully understand what you mean. As you can see in the test, the usage is the same as the code you provided. Following your example, you need a sentinel (list variable) to access the list API:

```c
struct sol_list delayed_packets = SOL_LIST_INIT(delayed_packets);

sol_list_add(&delayed_packets, &delayed_packate.list);
```

The sentinel must have ```next``` and ```prev``` pointers set to its address (that is the convention for an empty list). The limitation of ```SOL_LIST_INIT``` is that you have to pass a sol_list struct, and not a pointer to it, in the last case you have ```sol_list_init```. If you mean to define the macro as ```{ (name), (name) }```, then you'll have to have something like ```*my_list = SOL_LIST_INIT(my_list);``` which looks strange.

cabelitos: @ceolin I did not understand as well. @kaspersky approach looks fine to me

ceolin: ok, go ahead with this

ceolin:
merged :)

Issue #1673, opened 2016-03-18
API Changes: Sol-util


cabelitos:
#### Task Description
- [x] Rename `SOL_UTIL_ARRAY_SIZE()`  to `sol_util_array_size()`
- [x] Rename `sol_uint16_bytes_swap()` to `sol_util_uint16_bytes_swap()`
- [x] Rename `sol_uint32_bytes_swap()` to `sol_util_uint32_bytes_swap()`
- [x] Rename `sol_uint64_bytes_swap()` to `sol_util_uint64_bytes_swap()`
- [x] Rename `SOL_NSEC_PER_SEC`  to `SOL_UTIL_NSEC_PER_SEC`
- [x] Rename `SOL_MSEC_PER_SEC`  to `SOL_UTIL_MSEC_PER_SEC`
- [x] Rename `SOL_USEC_PER_SEC`  to `SOL_UTIL_USEC_PER_SEC`
- [x] Rename `SOL_NSEC_PER_MSEC`  to `SOL_UTIL_NSEC_PER_MSEC`
- [x] Rename `SOL_NSEC_PER_USEC` to `SOL_UTIL_NSEC_PER_USEC`
- [x] Rename `sol_util_secure_clear_memory()` to `sol_util_clear_memory_secure()`
- [x] Rename `sol_util_strtodn()` to `sol_util_strtod_n()`
- [x] Rename `sol_util_strtol()` to `sol_util_strtol_n()`
- [x] Rename `sol_util_strtoul()` to `sol_util_strtoul_n()`
- [x] Rename `sol_util_timespec_sum()` to `sol_util_timespec_add()`
- [x] Rename `sol_util_uuid_str_valid()` to `sol_util_uuid_str_is_valid()`

#### Dependencies
None.

kaspersky:
@cabelitos , can you detail a bit the last rename?

cabelitos:
@kaspersky Sorry, it was a typo. Fixed now

PR #1711, opened 2016-03-23
Sol-util: API changes

kaspersky:
#1673


otaviobp:
+1

bdilly:
It seems ok, but I believe that at this point you'll need to rebase and fix a few new occurrences of the old API

kaspersky:
#2011

PR #1712, opened 2016-03-23
fbp-generator: Check if symbol is a valid string.

cabelitos:
As the node symbol and the options symbol will be used in the generated C code,
it must be valid string in order to compile it in C.

### DIFF
--- a/src/bin/sol-fbp-generator/type-store.c
+++ b/src/bin/sol-fbp-generator/type-store.c
@@ -699,6 +699,23 @@ read_options(struct decoder *d, struct sol_vector *options)
 }
 
 static bool
+is_valid_symbol(const char *symbol)
+{
+    size_t i;
+
+    for (i = 0; symbol[i]; i++) {
+        if ((i == 0 && isdigit((int)symbol[i])) ||
+            (symbol[i] != '_' && !isalnum((int)symbol[i]))) {
+            fprintf(stderr, "Symbol '%s' with invalid character '%c' - pos: '%zu'\n",
+                    symbol, symbol[i], i);
+            return false;
+        }
+    }
+
+    return true;
+}
+
+static bool
 read_type(struct decoder *d, struct type_description *desc)
 {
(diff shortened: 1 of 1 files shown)


src/bin/sol-fbp-generator/type-store.c
```diff
@@ -796,6 +817,7 @@ type_store_read_from_json(struct type_store *store, struct sol_str_slice input)
         if (!read_type(&decoder, desc)) {
             type_description_fini(desc);
             sol_vector_del_last(&store->types);
+            return false;
```
edersondisouza: A wild `return false;` appears.
1. Delete
2. Leave it there, it was forgotten to start with!

src/bin/sol-fbp-generator/type-store.c
```diff
@@ -699,6 +699,23 @@ read_options(struct decoder *d, struct sol_vector *options)
...
 
 static bool
+is_valid_symbol(const char *symbol)
+{
+    size_t i;
+
+    for (i = 0; symbol[i]; i++) {
+        if ((i == 0 && isdigit((int)symbol[i])) ||
+            (symbol[i] != '_' && !isalnum((int)symbol[i]))) {
+            fprintf(stderr, "Symbol '%s' with invalid character '%c' - pos: '%zu'\n",
```
barbieri: prefix with `ERROR: `? or some other we're using in the other error cases of this file?

cabelitos: Fair. I will add the `Error:` prefix

src/bin/sol-fbp-generator/type-store.c
cabelitos: @barbieri what you think about leaving this `return false` here?

barbieri: ok for me, that means we'll stop early.

cabelitos:
Merged

PR #1728, opened 2016-03-23
sol-fbp-to-dot: Fix handling of ports with the same name

bdilly:
When a node have input and output ports with the same
name, graph was wrong. It was using input ports
to connect as output.

So prefixing input ports ids with "IN" and output with "OUT"
should be enough.

I'm not touching labels.

@otaviobp ptal . After you changed oic generator
some samples were broken, as you may see at:

http://solettaproject.github.io/docs/nodetypes/#category_oic

Thanks


otaviobp:
+1.
Sorry, I couldn't note that this was incorrect after applying my patch. thanks for fixing.

bdilly:
np
merged
thanks

PR #1730, opened 2016-03-24
vector: fix suboptimal sol_ptr_vector_del_element implementation

kaspersky:
Signed-off-by: Vladimir Cernov <cernov.vladimir@gmail.com>

### DIFF
--- a/src/lib/datatypes/sol-vector.c
+++ b/src/lib/datatypes/sol-vector.c
@@ -349,18 +349,20 @@ sol_vector_del_element(struct sol_vector *v, const void *elem)
 SOL_API int
 sol_ptr_vector_del_element(struct sol_ptr_vector *pv, const void *elem)
 {
-    int r;
-    uint16_t i, removed = 0;
-    void *cur;
+    uint16_t i, removed = 0, offset = 0;
+    void **v = pv->base.data;
 
-    SOL_PTR_VECTOR_FOREACH_REVERSE_IDX (pv, cur, i)
-        if (cur == elem) {
-            r = sol_ptr_vector_del(pv, i);
-            SOL_INT_CHECK(r, < 0, r);
+    for (i = 0; i < pv->base.len; i++)
+        if (v[i] == elem)
             removed++;
-        }
+        else
+            v[offset++] = v[i];
 
     if (!removed)
         return -ENOENT;
+
+    pv->base.len -= removed;
+    sol_vector_shrink(&pv->base);
+
     return 0;
 }
--- a/src/lib/datatypes/include/sol-vector.h
+++ b/src/lib/datatypes/include/sol-vector.h
@@ -540,7 +540,7 @@ sol_ptr_vector_del(struct sol_ptr_vector *pv, uint16_t i)
  *
  * @return @c 0 on success, error code (always negative) otherwise
  *
- * @remark Time complexity: number of @c elem occurences * vector size
+ * @remark Time complexity: linear in vector size
  *
  * @see sol_ptr_vector_del()
  * @see sol_ptr_vector_remove()


kaspersky:
When evaluating functions complexity in sol-vector I discovered a few suboptimal implementations, one routine is sol_ptr_vector_del_element. The old version had very bad performance when there are many occurences in the first half of the vector, in that case the second case was moved copied over and over on each element occurence removal. The new versions moves each element in the vector at most once.

I tested with the example below:

```c
#include <stdio.h>
#include <soletta/sol-vector.h>

int main()
{
    int i, k;
    int sum1 = 0, sum2 = 0;

    for (k = 0; k < 10; k++) {
        struct sol_ptr_vector v = SOL_PTR_VECTOR_INIT;
        void *data, *data2;
        for (i = 0; i < 20000; i++) {
            sol_ptr_vector_append(&v, data);
            sol_ptr_vector_append(&v, data2);
            sol_ptr_vector_append(&v, data2);
        }

        sum1 += v.base.len;
        sol_ptr_vector_del_element(&v, data);
        sum2 += v.base.len;
    }

    printf("%d %d\n", sum1, sum2);

    return 0;
}
```

New version:
```
user@umachine:~$ time ./vtest
600000 400000

real	0m0.152s
user	0m0.020s
sys	0m0.004s
```

Old version:
```
user@umachine:~$ time ./vtest
600000 400000

real	0m5.361s
user	0m1.680s
sys	0m0.004s
```

ibriano:
Neat

src/lib/datatypes/sol-vector.c
```diff
@@ -349,18 +349,20 @@ sol_vector_del_element(struct sol_vector *v, const void *elem)
...
-            SOL_INT_CHECK(r, < 0, r);
+    uint16_t i, removed = 0, offset = 0;
+    void **v = pv->base.data;
+
+    for (i = 0; i < pv->base.len; i++)
+        if (v[i] == elem)
             removed++;
-        }
+        else
+            v[offset++] = v[i];
```
edersondisouza: This makes me uncomfortable... Lot's attributions are made even if no data is going to be removed. Wouldn't be valid doing so only after `removed > 0`? But not sure actually, as branch maybe more costly. @lpereira?

kaspersky: I also thought of this. Performance depends on where first occurence in vector, sometimes may be close to beginning, sometimes not. Profiling is required to compare branch vs. say half attributions on average performance. I wasn't too worried because vector elements are pointers.

lpereira: The branch here is the same as it was before, so this part shouldn't change anything. However, for the case where many elements would be removed, `sol_vector_del()` will call `memmove()` and `sol_vector_shrink()` for each removed element, which isn't optimal; the new version is better in this case.

BTW, there's no need to have `offset` and `removed`: just increment `offset` if the pointer you're looking for isn't in the vector, and then calculate `removed = len - offset`.

edersondisouza: On 24-03-2016 13:49, Leandro A. F. Pereira wrote:
> In src/lib/datatypes/sol-vector.c
> <https://github.com/solettaproject/soletta/pull/1730#discussion_r57348113>:
>
>>              removed++;
>> -        }
>> +        else
>> +            v[offset++] = v[i];
>
> The branch here is the same as it was before, so this part shouldn't
> change anything. However, for the case where many elements would be
> removed, |sol_vector_del()| will call |memmove()| and
> |sol_vector_shrink()| for each removed element, which isn't optimal; the
> new version is better in this case.

Definitely new code is better =D
I was asking if adding a branch here to avoid `v[offset++] = v[i]` while 
no element was deleted would be worth.

barbieri: branch is always cheaper than memory write, which may invalidate caches and so on.

I did graphics blitting functions before, it did a huge difference if you'd check "alpha == 0" and just skip the bit instead of copying the value to itself.

While at that, what used to help me a lot in those graphics primitives were loop unrolling, that would reduce the impact of the `i < pv->base.len` and `i++`. I used to do 16-entries block and then finish with the single entry for for the remaining parts.

kaspersky: @barbieri, I'll do some profiling to confirm that. I'll also apply @lpereira suggestion on eliminating ```removed``` variable and then make the pull request v2.

kaspersky:
#2012

PR #1732, opened 2016-03-24
Coap mem improvements [v7]

glima:
Some memory usage revamps.

-----------

Changes since v1:
- bad mem access fixed (ptr vector instead of vector for contexts)

Changes since v2:
- early continue on a func, to diminish nesting.

Changes since v3:
- Fixed the way new tokens were written to the original packet
- Kept a way to send notifications *with* confirmation, if the user wants to

Changes since v4:
- @vcgomes's approach to notifications packet reuse was taken
- fixed keeping context for ack/reset packets
- we don't keep context for noncon packts anymore

Changes since v5:
- new patch dealing with NONCON but with reply messages.

Changes since v6:
- new approach by @vcgomes on pending timeouts, with fixes

@edersondisouza's code sending a lot of con packets was his timers on neighbour discovery/scan. Those are being acked at least, so we're fine. What was not escalating in any way were the non-con messages on both directions which would blow the memory usage very quickly.

### DIFF
--- a/src/lib/comms/sol-coap.c
+++ b/src/lib/comms/sol-coap.c
@@ -49,7 +49,7 @@ SOL_LOG_INTERNAL_DECLARE(_sol_coap_log_domain, "coap");
  */
 #define ACK_TIMEOUT_MS 2345
 #define MAX_RETRANSMIT 4
-#define NONCON_PKT_TIMEOUT_MS (ACK_TIMEOUT_MS << MAX_RETRANSMIT)
+#define MAX_PKT_TIMEOUT_MS (ACK_TIMEOUT_MS << MAX_RETRANSMIT)
 
 #ifndef SOL_NO_API_VERSION
 #define COAP_RESOURCE_CHECK_API(...) \
@@ -93,11 +93,13 @@ struct resource_observer {
 };
 
 struct pending_reply {
+    struct sol_coap_server *server;
+    struct sol_timeout *timeout;
     bool (*cb)(struct sol_coap_server *server, struct sol_coap_packet *req,
         const struct sol_network_link_addr *cliaddr, void *data);
     const void *data;
-    bool observing;
     char *path;
+    bool observing;
     uint16_t id;
     uint8_t tkl;
     uint8_t token[0];
@@ -625,28 +627,32 @@ pending_reply_free(struct pending_reply *reply)
 
     if (reply->observing)
         free(reply->path);
+
+    if (reply->timeout)
+        sol_timeout_del(reply->timeout);
+
     free(reply);
 }
 
+/* This is mostly for !CON packets, which do not go to the outgoing
+ * list but also keep a context of their own, for response handling */
 static bool
-call_reply_timeout_cb(struct sol_coap_server *server, struct sol_coap_packet *pkt)
+pending_timeout_cb(void *data)
 {
-    uint16_t i, id;
-    struct pending_reply *reply;
+    struct pending_reply *reply = data;
+    struct sol_coap_server *server = reply->server;
 
-    sol_coap_header_get_id(pkt, &id);
+    if (reply->cb(server, NULL, NULL, (void *)reply->data))
+        return true;
 
-    SOL_PTR_VECTOR_FOREACH_REVERSE_IDX (&server->pending, reply, i) {
-        if (reply->observing || reply->id != id)
-            continue;
+    sol_ptr_vector_remove(&server->pending, reply);
 
-        if (reply->cb(server, NULL, NULL, (void *)reply->data))
-            return false;
-        sol_ptr_vector_del(&server->pending, i);
-        pending_reply_free(reply);
-    }
+    /* Timeout will be free'd when 'false' is returned */
+    reply->timeout = NULL;
 
-    return true;
+    pending_reply_free(reply);
+
+    return false;
 }
 
 static bool
@@ -656,30 +662,22 @@ timeout_expired(struct sol_coap_server *server, struct outgoing *outgoing)
     int timeout;
     uint16_t i, id;
     uint8_t type;
-    int max_retransmit;
     bool expired = false;
 
     sol_coap_header_get_type(outgoing->pkt, &type);
-    if (type == SOL_COAP_TYPE_CON) {
-        max_retransmit = MAX_RETRANSMIT;
-        timeout = ACK_TIMEOUT_MS << outgoing->counter++;
-    } else {
-        max_retransmit = 1;
-        timeout = NONCON_PKT_TIMEOUT_MS;
-        outgoing->counter++;
-    }
+    /* no re-transmissions for !CON packets, we just keep a
+     * pending_reply for a while */
+    if (type != SOL_COAP_TYPE_CON)
+        return false;
 
-    if (outgoing->counter > max_retransmit) {
+    timeout = ACK_TIMEOUT_MS << outgoing->counter++;
+
+    if (outgoing->counter > MAX_RETRANSMIT) {
         SOL_PTR_VECTOR_FOREACH_REVERSE_IDX (&server->outgoing, o, i) {
             if (o == outgoing) {
                 sol_coap_header_get_id(outgoing->pkt, &id);
                 SOL_DBG("packet id %d dropped, after %d transmissions",
                     id, outgoing->counter);
-
-                if (!call_reply_timeout_cb(server, outgoing->pkt)) {
-                    expired = true;
-                    break;
-                }
 
                 sol_ptr_vector_del(&server->outgoing, i);
                 outgoing_free(o);
@@ -747,8 +745,7 @@ on_can_write(void *data, struct sol_socket *s)
     struct sol_coap_server *server = data;
     struct outgoing *outgoing;
     struct sol_buffer buf;
-    int err;
-    int idx;
+    int err, idx;
 
     if (sol_ptr_vector_get_len(&server->outgoing) == 0)
         return false;
@@ -771,8 +768,10 @@ on_can_write(void *data, struct sol_socket *s)
     if (err == -EAGAIN)
         return true;
 
-    SOL_DBG("CoAP packet sent (payload of %zu bytes, "
-        "buffer holding it with %zu bytes)",
+    SOL_DBG("CoAP packet sent (outgoing_len=%d, pending_len=%d)"
+        " -- payload of %zu bytes, buffer holding it with %zu bytes",
+        sol_ptr_vector_get_len(&server->outgoing),
+        sol_ptr_vector_get_len(&server->pending),
         outgoing->pkt->buf.used, outgoing->pkt->buf.capacity);
     sol_coap_packet_debug(outgoing->pkt);
     if (err < 0) {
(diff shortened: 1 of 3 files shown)


cabelitos:
@glima I guess your changes broke LWM2M test

glima:
I'm on this now, sorry

src/lib/comms/sol-coap.c
```diff
@@ -682,13 +690,74 @@ timeout_expired(struct sol_coap_server *server, struct outgoing *outgoing)
...
+        r = sol_ptr_vector_find_first(&outgoing->notified, o);
+        if (r < 0) {
+            uint8_t *token, tkl;
+
+            r = sol_ptr_vector_append(&outgoing->notified, o);
+            SOL_INT_CHECK(r, < 0, r);
+
+            token = sol_coap_header_get_token(outgoing->pkt, &tkl);
+            if (tkl > outgoing->tkl) {
+                uint8_t bytes[tkl - o->tkl];
```
vcgomes: Uninitialised memory? 

Does the compiler in semaphore support this "dynamic" array (I forgot the proper name)? I remember this being relatively "new".

src/lib/comms/sol-coap.c
```diff
@@ -104,12 +104,19 @@ struct pending_reply {
 };
 
 struct outgoing {
+    struct sol_ptr_vector notified; /* Only for notification packets */
+    struct resource_context *context; /* Only set for notification packets */
```
vcgomes: Wouldn't it be more direct if the context kept the reference to the notification packet?

But I still need to think about this a little more.

src/lib/comms/sol-coap.c
```diff
@@ -682,13 +690,74 @@ timeout_expired(struct sol_coap_server *server, struct outgoing *outgoing)
...
+notification_pkt_iterate(struct sol_coap_server *server,
+    struct outgoing *outgoing)
+{
+    int r, i;
+    struct resource_observer *o;
+
+    SOL_PTR_VECTOR_FOREACH_IDX (&outgoing->context->observers, o, i) {
+        r = sol_ptr_vector_find_first(&outgoing->notified, o);
+        if (r < 0) {
+            uint8_t *token, tkl;
```
vcgomes: Wouldn't an early `continue` when the that packet was already notified make the code more readable?

src/lib/comms/sol-coap.c
ibriano: On Thu, 24 Mar 2016 13:41:40 -0700, Vinicius Gomes wrote:
> > +    struct outgoing *outgoing)
> > +{
> > +    int r, i;
> > +    struct resource_observer *o;
> > +
> > +    SOL_PTR_VECTOR_FOREACH_IDX (&outgoing->context->observers, o, i) {
> > +        r = sol_ptr_vector_find_first(&outgoing->notified, o);
> > +        if (r < 0) {
> > +            uint8_t *token, tkl;
> > +
> > +            r = sol_ptr_vector_append(&outgoing->notified, o);
> > +            SOL_INT_CHECK(r, < 0, r);
> > +
> > +            token = sol_coap_header_get_token(outgoing->pkt, &tkl);
> > +            if (tkl > outgoing->tkl) {
> > +                uint8_t bytes[tkl - o->tkl];
> 
> Uninitialised memory? 
> 
> Does the compiler in semaphore support this "dynamic" array (I forgot the proper name)? I remember this being relatively "new".

It's a C99 thing, so it'd better support it.

src/lib/comms/sol-coap.c
glima: Ack.

src/lib/comms/sol-coap.c
glima: Not sure either. I guess I'll wait too.

vcgomes:
There are valid use cases for sending notifications with confirmable packets, for example, think of a light dimmer, while the user is moving the lever, you may want to send non-confirmable notifications (as there is little loss if some notifications are lost), but after the user stops moving the lever, you may want to send a confirmable notification.

glima:
@vcgomes, fair. I can call ``enqueue_packet() with the ``with_reply`` arg set to ``pkt_type == CON`` and I think we're fine with your use case. I'll get to that tomorrow, on a newer version.

src/lib/comms/sol-coap.c
glima: Now I see that the token setter does buffer *insert* slice, not *set*. Thanks for pointing out, fixing.

glima:
@vcgomes, when you can, new review is welcome

vcgomes:
I see valor in tackling this problem, but am I thinking about an intermediate solution, that could be simpler, see what you think:

the basics are: maintain the current logic for the `outgoing` packets, just add a `header` field of type `sol_coap_packet`, and the `pkt` field becomes just a reference for the `payload` packet; add a `payload` packet to `struct resource_context` which is set when the user calls `_new_notification_packet()`.

Assembling the two parts into a data structure that can be trivially sent is the only thing that I didn't find a good way to solve.

The advantages that I see are that the code complexity doesn't increase too much, the current structures doesn't increase too much either; but, I know that, the gains in memory consumption could be smaller indeed. Thinking a little more, perhaps in the case that there are a small number of observers it shouldn't be too bad, considering that the current structures didn't change much.

glima:
Your solution still does not handle ``sol_coap_cancel_send_packet()`` properly for notification packets. Moreover, I don't think it's that complicated what we got in this patch -- if we want to address all gotchas from the CoAP RFC (doing tricks to be lean on memory for small devices too) we have a long way to go in complexity, yet.

vcgomes:
As the `pkt` field in the `outgoing` would be a reference to the `payload` packet (what was returned by `_notification_new()`), I am not seeing problems with handling `sol_coap_cancel_send_packet()`, am I missing something?

I will give it a quick try, just to have something to show. Perhaps it ends being more complicated than what you are proposing.

vcgomes:
@glima please take a look at https://github.com/solettaproject/soletta/pull/1764, this is what I had in mind, more or less. In the end, I didn't need to keep the reference to the payload in each `resource_context`. 

Not tested.

glima:
@vcgomes, I like, but I still want to ditch the packet from ``outgoing`` on ``NONCON`` scenarios. I'll work on top of yours to get that now, thanks a lot.

glima:
To be fair, let's test this new schema on devices 1st, it may be good enough.

glima:
@vcgomes, what do you think now? This reduces our memory usage massively. All confirmable packets still doing their thing, only non-confirmable and ack/reset ones being discarded right away after sent. The spec allows for that, we're not being aliens.

ibriano:
On Tue, 29 Mar 2016 14:33:07 -0700, Gustavo Lima Chaves wrote:
> @vcgomes, what do you think now? This reduces our memory usage massively. All confirmable packets still doing their thing, only non-confirmable and ack/reset ones being discarded right away after sent. The spec allows for that, we're not being aliens.

I'm ok with that.

src/lib/comms/sol-coap.c
```diff
@@ -684,7 +684,8 @@ timeout_expired(struct sol_coap_server *server, struct outgoing *outgoing)
     return expired;
 }
 
-static struct sol_buffer *prepare_buffer(struct outgoing *outgoing)
+static struct sol_buffer *
+prepare_buffer(struct outgoing *outgoing)
```
vcgomes: I will suggest to squash this into the previous commit.

src/lib/comms/sol-coap.c
```diff
@@ -715,7 +716,7 @@ static struct sol_buffer *prepare_buffer(struct outgoing *outgoing)
     SOL_INT_CHECK(r, < 0, NULL);
 
     r = sol_buffer_insert_bytes(&buffer, new_offset,
-        (uint8_t *) payload->buf.data + old_offset, payload->buf.used - old_offset);
+        (uint8_t *)payload->buf.data + old_offset, payload->buf.used - old_offset);
```
vcgomes: Squash this too, please.

vcgomes:
There's a problem when the user sent a NONCON packet passing a reply callback, in this case, the callback will not be called.

glima:
This novel never ends, damn. Do we add a boolean flag for these cases? Anything better on your mind?

src/lib/comms/sol-coap.c
glima: OK.

vcgomes:
What I am thinking is that I mixed some concepts when writing this, and this "mess" is a consequence of that. The concepts that I mixed are `outgoing->timeout` (which represent the timeout for retransmissions) and when I should tell the user that no reply is forthcoming.

What I propose is adding an timeout for each pending reply, independent of the retransmission timeout, that will trigger after a time such that I am sure that no reply is coming, let's say 45 or so seconds. This is an arbitrary time, as this is a `sol-coap` concept.

In this particular case, for NONCON packets with replies, as they don't need retransmissions, I don't need to keep the associated `outgoing` struct around, just the `pending_reply` struct.

With this, I guess the whole logic for keeping or not the `outgoing` packets around could be simplified a little.

What do you guys think?

glima:
The idea looks good, @vcgomes. After finishing a review here I'll revisit this. We indeed need NONCONs with replies, as in ``sol_oic_client_find_resource()`` (tentative broadcasts expecting replies).

glima:
I'm only afraid of the fact that the sample on my hands uses the ports doing scan/discovery too much (every 5s or something like that) and a big timeout would still bite us in the ass. I'll check the whole thing again with care.

src/lib/comms/sol-coap.c
```diff
@@ -678,13 +684,52 @@ timeout_expired(struct sol_coap_server *server, struct outgoing *outgoing)
...
+    static struct sol_buffer buffer;
+    struct sol_coap_packet *header = outgoing->header;
+    struct sol_coap_packet *payload = outgoing->pkt;
+    uint16_t new_size, new_offset, old_offset;
+    uint8_t new_tkl, old_tkl;
+    int r;
+
+    if (!header) {
+        buffer = SOL_BUFFER_INIT_CONST(payload->buf.data, payload->buf.used);
+        return &buffer;
```
barbieri: see my comments to #1764

vcgomes:
Just updated https://github.com/solettaproject/soletta/pull/1764 with the comments from @barbieri

src/lib/comms/sol-coap.c
```diff
@@ -784,6 +788,23 @@ on_can_write(void *data, struct sol_socket *s)
...
+     * these are not received by the server. Thus, a server can always
+     * pretend that a RESET message rejecting a non-confirmable
+     * notification was lost. If a server does this, it could
+     * accelerate cancellation by sending the following notifications
+     * to that client in confirmable messages".
+     *
+     * By taking this shortcut we reduce memory usage A LOT and are
+     * able to run on very small devices with no memory issues.
+     */
+    if (type == SOL_COAP_TYPE_ACK || type == SOL_COAP_TYPE_RESET
```
vcgomes: Sorry for only thinking about this now, the correct check would be `!outgoing->timeout`, i.e. for everything that I don't need to retransmit and was just sent successfully.

vcgomes:
I updated #1764 with a commit with what I had in mind, see if it makes sense: https://github.com/solettaproject/soletta/pull/1764/commits/491690b9684965b04a382357c91109b5e8707207

glima:
@vcgomes: just tested and it seems we're fine. Please tell if the commit arrangement is fine as well.

vcgomes:
:+1: I'm liking it.

glima:
Cool, will merge :)

PR #1733, opened 2016-03-24
Http sse v3

ceolin:
from v2:

 - Change `SSE` to `progressive`
 - New helper function to set the necessary headers
 - Improvements in the sample
 - Add option to delete a function imediatally or wait until its buffer is empty
 - Always notfiy the user when the functions is deleted

### DIFF
--- a/src/lib/comms/include/sol-http-server.h
+++ b/src/lib/comms/include/sol-http-server.h
@@ -68,6 +68,16 @@ struct sol_http_server;
 struct sol_http_request;
 
 /**
+ * @struct sol_http_response_progressive
+ * @brief Opaque handler for response using server-sent-event.
+ *
+ * This response is create with @c sol_http_server_send_response_progressive
+ * and data can be given using @c sol_http_server_reponse_progressive_feed,
+ * to delete it use @c sol_http_server_reponse_ss2_del.
+ */
+struct sol_http_response_progressive;
+
+/**
  * @brief Creates a HTTP server, binding on all interfaces
  * in the specified @a port.
  *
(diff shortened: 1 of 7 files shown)


src/lib/comms/include/sol-http-server.h
```diff
@@ -68,6 +68,16 @@ struct sol_http_server;
 struct sol_http_request;
 
 /**
+ * @struct sol_http_response_progressive
+ * @brief Opaque handler for response using server-sent-event.
+ *
+ * This response is create with @c sol_http_server_send_response_progressive
+ * and data can be given using @c sol_http_server_reponse_progressive_feed,
+ * to delete it use @c sol_http_server_reponse_ss2_del.
```
edersondisouza: `sol_http_server_reponse_ss2_del`?

src/lib/comms/include/sol-http-server.h
```diff
@@ -204,10 +214,97 @@ int sol_http_server_set_last_modified(struct sol_http_server *server, const char
...
+ *
+ * @param response The response to set the headers
+ *
+ * @return @c 0 on success, error code (always negative) otherwise.
+ */
+static inline int
+sol_http_response_set_progressive_headers(struct sol_http_response *response)
+{
+    if (!response)
+        return -EINVAL;
```
edersondisouza: Why not `SOL_NULL_CHECK`? Is it expected to be common flow to call this with no `response`?

edersondisouza:
For sample commit, I'd be happy if there was some README explaining how it works and what to expect. I didn't test, but I think that, besides running it on a terminal, I'd also need to run a browser and open the same port, right? This is more complex than just running it on terminal, so I guess a README would be worth.

src/lib/comms/include/sol-http-server.h
ceolin: my bad :) I will fix it

src/lib/comms/include/sol-http-server.h
ceolin: there is no guarantee log domain is properly set

ceolin:
@edersondisouza there is a ridicolous help in the proper sample and the user can reach it using `-h` or `--help`. Do you think it's better to improve it or to add a README in the directory ?

src/lib/datatypes/include/sol-buffer.h
```diff
@@ -420,6 +420,17 @@ int sol_buffer_append_slice(struct sol_buffer *buf, const struct sol_str_slice s
 int sol_buffer_append_bytes(struct sol_buffer *buf, const uint8_t *bytes, size_t size);
 
 /**
+ * @brief Appends the buffer @c from in the end of @c buf,
+ * reallocating if necessary.
```
glima: I'd do "s/Appends the/Appends the contents of the/", "s/in/at" and s/reallocating if/reallocating @c if/".

src/lib/comms/include/sol-http-server.h
```diff
@@ -204,10 +214,97 @@ int sol_http_server_set_last_modified(struct sol_http_server *server, const char
...
+/**
+ * @brief Delete the progressive response.
+ *
+ * This function deletes the progressive response and its resources and closes
+ * the connection.
+ *
+ * @param progressive The progressive response created with
+ * @c sol_http_server_send_response_progressive
+ * @param graceful_del If @c true all pending data will be sent before
+ * the connection be closed.
```
glima: Heheh, maybe deferred_del or something like that?

src/samples/http/server-sse.c
```diff
@@ -0,0 +1,280 @@
...
+        }
+        /* this will loop trying to read as much data as possible to buffer. */
+        err = sol_util_load_file_fd_buffer(fd, &value);
+        if (err < 0) {
+            fprintf(stderr, "ERROR: failed to read from stdin: %s\n",
+                sol_util_strerrora(-err));
+            goto err;
+        }
+
+        if (value.used == 0) {
```
glima: After line 86 will this ever apply?

ceolin: you are right

src/lib/comms/include/sol-http-server.h
ceolin: this is the same name used by glib

ceolin:
fixes in https://github.com/solettaproject/soletta/pull/1740

PR #1735, opened 2016-03-24
[WIP] Use OIC Data Models

bdilly:
It still fails on some points when building.

I wasn't able to check all issues yet.

### DIFF
--- a/src/modules/flow/oic/Makefile
+++ b/src/modules/flow/oic/Makefile
@@ -1,4 +1,5 @@
 obj-$(FLOW_NODE_TYPE_OIC) += oic.mod
 obj-oic-$(FLOW_NODE_TYPE_OIC) := oic.json oic.o
-obj-oic-$(FLOW_NODE_TYPE_OIC)-oic := core.light.json
+obj-oic-$(FLOW_NODE_TYPE_OIC)-oic := y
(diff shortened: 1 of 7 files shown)


otaviobp:
looks good to me so far

src/modules/flow/oic/Makefile
```diff
@@ -1,4 +1,5 @@
 obj-$(FLOW_NODE_TYPE_OIC) += oic.mod
 obj-oic-$(FLOW_NODE_TYPE_OIC) := oic.json oic.o
-obj-oic-$(FLOW_NODE_TYPE_OIC)-oic := core.light.json
+obj-oic-$(FLOW_NODE_TYPE_OIC)-oic := y
+obj-oic-$(FLOW_NODE_TYPE_OIC)-oic-dirs := $(OIC_DATA_MODELS)
```
glima: We can live with just one of both (new) variables, no?

glima: OK, that would involve changes in the makefile rule. I can tweak that later.

bdilly: If there is any way to represent the current dir all of them could be used on oic-dirs.

But I believe just using

obj-oic-$(FLOW_NODE_TYPE_OIC)-oic := y

would be more frequent than including extra dirs anyway.

bdilly:
New version pushed: #1781

PR #1737, opened 2016-03-28
oic: Add support for secure OIC connections using pre shared certs

otaviobp:
As we had some securities concerns in PR #1527, sending for review only the security communication implementation, without the just works pairing.

### DIFF
--- a/src/lib/common/include/sol-platform.h
+++ b/src/lib/common/include/sol-platform.h
@@ -73,9 +73,10 @@ extern "C" {
 const char *sol_platform_get_board_name(void);
 
 /**
- * @brief Retrieves, in @a id, the machine-id present in the file system.
+ * @brief Retrieves the machine-id present in the file system as a string.
  *
- * The returned string is assured to be a valid, 16 bytes-long (128 bits) UUID.
+ * The returned string is assured to be a valid null terminated string with
+ * the 16 bytes-long (128 bits) UUID encoded as hexadecimal ASC characters.
  *
  * @note: If the environment variable SOL_MACHINE_ID is set and is
  * properly formatted as a UUID, its value is returned by this call.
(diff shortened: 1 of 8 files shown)


src/lib/common/include/sol-platform.h
```diff
@@ -73,9 +73,10 @@ extern "C" {
 const char *sol_platform_get_board_name(void);
 
 /**
- * @brief Retrieves, in @a id, the machine-id present in the file system.
+ * @brief Retrieves the machine-id present in the file system as a string.
  *
- * The returned string is assured to be a valid, 16 bytes-long (128 bits) UUID.
+ * The returned string is assured to be a valid null terminated string with
+ * the 16 bytes-long (128 bits) UUID encoded as hexadecimal ASC characters.
```
glima: ASCII?

src/lib/common/include/sol-platform.h
```diff
@@ -86,6 +87,20 @@ const char *sol_platform_get_board_name(void);
 const char *sol_platform_get_machine_id(void);
 
 /**
+ * @brief Retrieves the machine-id present in the file system as a byte array.
+ *
+ * The returned byte array is assured to be a 16 bytes-long (128 bits) array
+ * with the machine UUID.
+ *
+ * @note: If the environment variable SOL_MACHINE_ID is set and is
+ * properly formatted as a UUID byte array, its value is returned by this call.
```
glima: The env var will be a string, more specifically, no?

src/lib/common/include/sol-platform.h
```diff
@@ -73,9 +73,10 @@ extern "C" {
 const char *sol_platform_get_board_name(void);
 
 /**
- * @brief Retrieves, in @a id, the machine-id present in the file system.
+ * @brief Retrieves the machine-id present in the file system as a string.
  *
```
zolkis: It is not clear what is machine-id. OIC defines platform ID and device ID. Only device ID is UUID. Platform ID is opaque (can be UUID, VIN, etc).

otaviobp: This patch is in this pull req because this function will be used by OIC implementation, but it is not directly related to OIC.
Sol-platform has functions related to soletta platform. Machine-ID is an UUID that defines a machine, i.e. a device. This is used by OIC as the device id, but it is used in other Soletta modules too.

src/lib/common/include/sol-platform.h
otaviobp: Yes. It is not clear, I'll improve that comment

otaviobp:
merged

src/lib/comms/sol-oic-security.c
```diff
@@ -0,0 +1,394 @@
...
+#include "sol-json.h"
+#include "sol-log.h"
+#include "sol-oic-cbor.h"
+#include "sol-oic-security.h"
+#include "sol-oic-server.h"
+#include "sol-platform.h"
+#include "sol-socket-dtls.h"
+#include "sol-socket.h"
+#include "sol-str-slice.h"
+#include "sol-util-file.h"
```
glima: This files compiles under ``FEATURE_FILESYSTEM`` only. Please make sure Soletta won't break if that feature is not available, @otaviobp.

PR #1738, opened 2016-03-28
buffer: Don't use implicit flags when setting secure clear

otaviobp:
Supports in buffer to use SOL_BUFFER_FLAGS_CLEAR_MEMORY for memory owned
by buffer. This was, SOL_BUFFER_FLAGS_CLEAR_MEMORY wont enforce
SOL_BUFFER_FLAGS_FIXED_CAPACITY flag.

### DIFF
--- a/src/lib/datatypes/sol-buffer.c
+++ b/src/lib/datatypes/sol-buffer.c
@@ -30,6 +30,23 @@ nul_byte_size(const struct sol_buffer *buf)
     return (buf->flags & SOL_BUFFER_FLAGS_NO_NUL_BYTE) ? 0 : 1;
 }
 
+static void *
+secure_realloc(struct sol_buffer *buf, size_t new_size)
+{
+    char *new_data = NULL;
+
+    if (!new_size)
+        goto end;
+
+    new_data = malloc(new_size);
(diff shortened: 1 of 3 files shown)


src/lib/datatypes/sol-buffer.c
```diff
@@ -30,6 +30,23 @@ nul_byte_size(const struct sol_buffer *buf)
...
 
+static void *
+secure_realloc(struct sol_buffer *buf, size_t new_size)
+{
+    char *new_data = NULL;
+
+    if (!new_size)
+        goto end;
+
+    new_data = malloc(new_size);
```
ceolin: check the return of `malloc`

otaviobp: Caller is checking that. I am freeing and clearing memory even on malloc errors.

ibriano: You are copying into this new_data right after malloc. If it failed, boom?

ceolin: yep, that is the point of the check

otaviobp: ok. my bad. I moved some code and didn't realize I needed to check that.

otaviobp:
merged with fixes

PR #1740, opened 2016-03-28
HTTP sse v4

ceolin:
from v3:
 - Fix doc
 - Better usage description
 - Handle Ctrl-d properly

### DIFF
--- a/src/lib/comms/include/sol-http-server.h
+++ b/src/lib/comms/include/sol-http-server.h
@@ -68,6 +68,16 @@ struct sol_http_server;
 struct sol_http_request;
 
 /**
+ * @struct sol_http_response_progressive
+ * @brief Opaque handler for response using server-sent-event.
+ *
+ * This response is create with @c sol_http_server_send_response_progressive
+ * and data can be given using @c sol_http_response_progressive_feed,
+ * to delete it use @c sol_http_response_progressive_del.
+ */
+struct sol_http_response_progressive;
+
+/**
  * @brief Creates a HTTP server, binding on all interfaces
  * in the specified @a port.
  *
@@ -204,8 +214,97 @@ int sol_http_server_set_last_modified(struct sol_http_server *server, const char
  * and parameters (e.g http headers)
  *
  * @return @c 0 on success, error code (always negative) otherwise.
+ *
+ * @see sol_http_server_send_response_progressive
  */
 int sol_http_server_send_response(struct sol_http_request *request, struct sol_http_response *response);
+
+/**
+ * @brief Set the necessary headers to allow server sent events
+ *
+ * @param response The response to set the headers
+ *
+ * @return @c 0 on success, error code (always negative) otherwise.
+ */
+static inline int
+sol_http_response_set_progressive_headers(struct sol_http_response *response)
+{
+    if (!response)
+        return -EINVAL;
+
+    if (!sol_http_param_add(&response->param,
+        SOL_HTTP_REQUEST_PARAM_HEADER("Content-Type", "text/event-stream"))) {
+        return -1;
+    }
+
+    if (!sol_http_param_add(&response->param,
+        SOL_HTTP_REQUEST_PARAM_HEADER("Connection", "keep-alive"))) {
+        return -1;
+    }
+
+    if (!sol_http_param_add(&response->param,
+        SOL_HTTP_REQUEST_PARAM_HEADER("Cache-Control", "no-cache"))) {
+        return -1;
+    }
+
+    return 0;
+}
+
+/**
+ * @brief Send the response and keep connection alive to request given in the
+ * callback registered on @c sol_http_server_register_handler.
+ *
+ * After this call, the caller will be resposible to close the connection calling
+ * @c sol_http_response_progressive_del. @note all the necessary headers are set by this function.
+ *
+ * @param request The request given on the callback of
+ * @c sol_http_server_register_handler.
+ * @param response The response for the request containing the data
+ * and parameters (e.g http headers)
+ * @param on_close Callback called when the connection is closed from the client side.
+ * @param cb_data The data pointer to be passed to @a cb.
+ *
+ * @return @c sol_http_response_progressive on success, @c NULL otherwise.
+ *
+ * @see sol_http_server_send_response
+ * @see sol_http_response_progressive_del
+ * @see sol_http_response_progressive_feed
+ */
+struct sol_http_response_progressive *sol_http_server_send_response_progressive(struct sol_http_request *request,
+    const struct sol_http_response *response,
+    void (*on_close)(void *data, const struct sol_http_response_progressive *progressive),
+    const void *cb_data);
+
+/**
+ * @brief Send data for the progressive response.
+ *
+ * @param progressive The progressive response created with
+ * @c sol_http_server_send_response_progressive
+ * @param data The data to be sent.
+ *
+ * @return @c 0 on success, error code (always negative) otherwise.
+ *
+ * @see sol_http_response_progressive_del
+ * @see sol_http_server_send_response_progressive
+ */
+int sol_http_response_progressive_feed(struct sol_http_response_progressive *progressive,
+    const struct sol_str_slice data);
+
+/**
+ * @brief Delete the progressive response.
+ *
+ * This function deletes the progressive response and its resources and closes
+ * the connection.
+ *
+ * @param progressive The progressive response created with
+ * @c sol_http_server_send_response_progressive
+ * @param graceful_del If @c true all pending data will be sent before
+ * the connection be closed.
+ *
+ * @see sol_http_server_send_response_progressive
+ * @see sol_http_response_progressive_feed
+ */
+void sol_http_response_progressive_del(struct sol_http_response_progressive *progressive, bool graceful_del);
 
 /**
  * @brief Gets the URL from a given request.
--- a/src/lib/comms/sol-http-server-impl-microhttpd.c
+++ b/src/lib/comms/sol-http-server-impl-microhttpd.c
@@ -28,6 +28,7 @@
 #include <sys/types.h>
 #include <unistd.h>
 
+#include "sol-buffer.h"
 #include "sol-http-server.h"
 #include "sol-log.h"
 #include "sol-mainloop.h"
@@ -68,6 +69,15 @@ struct sol_http_request {
     time_t if_since_modified;
     time_t last_modified;
     bool is_multipart;
+};
+
+struct sol_http_response_progressive {
+    struct MHD_Connection *connection;
+    void (*on_del)(void *data, const struct sol_http_response_progressive *progressive);
+    const void *cb_data;
+    struct sol_buffer buffer;
+    bool delete_me, graceful_del;
+    bool suspended;
 };
 
 struct static_dir {
@@ -207,17 +217,12 @@ set_last_modified_header(struct MHD_Response *response, time_t last_modified)
 }
 
 static struct MHD_Response *
-build_mhd_response(const struct sol_http_response *response, time_t last_modified)
+build_mhd_response_params(struct MHD_Response *r,
+    const struct sol_http_response *response)
 {
     struct sol_buffer buf;
     uint16_t idx;
-    struct MHD_Response *r;
     struct sol_http_param_value *value;
-
-    r = MHD_create_response_from_buffer(response->content.used, response->content.data,
-        MHD_RESPMEM_MUST_COPY);
-    if (!r)
-        return NULL;
 
     sol_buffer_init(&buf);
     SOL_HTTP_PARAMS_FOREACH_IDX (&response->param, value, idx) {
@@ -259,9 +264,6 @@ build_mhd_response(const struct sol_http_response *response, time_t last_modifie
         }
     }
 
-    if (!set_last_modified_header(r, last_modified))
-        goto err;
-
     sol_buffer_fini(&buf);
     return r;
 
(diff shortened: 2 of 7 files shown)


src/lib/comms/include/sol-http-server.h
```diff
@@ -204,10 +214,99 @@ int sol_http_server_set_last_modified(struct sol_http_server *server, const char
...
+ */
+static inline int
+sol_http_response_set_progressive_headers(struct sol_http_response *response)
+{
+    if (!response)
+        return -EINVAL;
+
+    if (!sol_http_param_add(&response->param,
+        SOL_HTTP_REQUEST_PARAM_HEADER("Content-Type", "text/event-stream"))) {
+        return -1;
```
barbieri: also, I'm thinking since these are all the same pattern if you should do a loop:

```c
const struct { const char *k, const char *v } *itr, headers[] = { // or make it of struct sol_http_param
   { "Content-type", "text/event-stream" },
   { "Connection", "keep-alive" },
   { "Cache-Control", "no-cache" },
   { NULL, NULL }
};
for (itr = headers; itr->k != NULL; itr++) {
    if (!sol_http_param_add(&response->param, SOL_HTTP_REQUEST_PARAM_HEADER(itr->k, itr->v))) { ... }
}
```

You can even make this a helper function in sol-http.h (or server.h) and use it everywhere, then you take a null-terminated array and loop-set it. Or make a va_args version that is null terminated. This pattern is very common and I've seen it in many places.

src/lib/comms/include/sol-http-server.h
```diff
@@ -204,10 +214,99 @@ int sol_http_server_set_last_modified(struct sol_http_server *server, const char
...
+ * @param on_close Callback called when the connection is closed from the client side.
+ * @param cb_data The data pointer to be passed to @a cb.
+ *
+ * @return @c sol_http_response_progressive on success, @c NULL otherwise.
+ *
+ * @see sol_http_server_send_response
+ * @see sol_http_response_progressive_del
+ * @see sol_http_response_progressive_feed
+ */
+struct sol_http_response_progressive *sol_http_server_send_response_progressive(struct sol_http_request *request,
```
barbieri: this name looks like the efl-pattern, check with others if we should swap progressive and response, making it more natural English `sol_http_progressive_response`.

@bdilly @glima @ibriano ?

src/lib/comms/include/sol-http-server.h
```diff
@@ -204,10 +214,99 @@ int sol_http_server_set_last_modified(struct sol_http_server *server, const char
...
+ * callback registered on @c sol_http_server_register_handler.
+ *
+ * After this call, the caller will be resposible to close the connection calling
+ * @c sol_http_response_progressive_del. @note all the necessary headers are set by this function.
+ *
+ * @param request The request given on the callback of
+ * @c sol_http_server_register_handler.
+ * @param response The response for the request containing the data
+ * and parameters (e.g http headers)
+ * @param on_close Callback called when the connection is closed from the client side.
```
barbieri: is this only about "client side"?

src/lib/comms/include/sol-http-server.h
```diff
@@ -204,10 +214,99 @@ int sol_http_server_set_last_modified(struct sol_http_server *server, const char
...
+/**
+ * @brief Delete the progressive response.
+ *
+ * This function deletes the progressive response and its resources and closes
+ * the connection.
+ *
+ * @param progressive The progressive response created with
+ * @c sol_http_server_send_response_progressive
+ * @param graceful_del If @c true all pending data will be sent before
+ * the connection be closed.
```
barbieri: do we call `on_close` after this? if yes, then when? IF no, then say it here.

src/lib/comms/sol-http-server-impl-microhttpd.c
```diff
@@ -70,6 +71,15 @@ struct sol_http_request {
     bool is_multipart;
 };
 
+struct sol_http_response_progressive {
+    struct MHD_Connection *connection;
+    void (*on_del)(void *data, const struct sol_http_response_progressive *progressive);
+    const void *cb_data;
```
barbieri: maybe `struct { void (*cb)(...), const void *data } on_del;`?

src/lib/comms/sol-http-server-impl-microhttpd.c
```diff
@@ -70,6 +71,15 @@ struct sol_http_request {
...
+    struct sol_buffer buffer;
+    bool delete_me, graceful_del;
+    bool suspended;
```
barbieri: why delete_me and graceful_del are on the same line, suspended isn't ?

src/lib/comms/sol-http-server-impl-microhttpd.c
```diff
@@ -271,6 +273,82 @@ build_mhd_response(const struct sol_http_response *response, time_t last_modifie
...
+    return len;
+}
+
+static struct MHD_Response *
+build_mhd_response_progressive(const struct sol_http_response *response,
+    struct sol_http_response_progressive *progressive)
+{
+    struct MHD_Response *r;
+
+    r = MHD_create_response_from_callback(MHD_SIZE_UNKNOWN, 128,
```
barbieri: I'm wondering if 128 is a good number. Maybe go with `LINE_MAX` or page size (4096)?

src/lib/comms/sol-http-server-impl-microhttpd.c
```diff
@@ -960,6 +1038,80 @@ sol_http_server_send_response(struct sol_http_request *request, struct sol_http_
...
+    SOL_NULL_CHECK(request, NULL);
+    SOL_NULL_CHECK(request->connection, NULL);
+    SOL_NULL_CHECK(response, NULL);
+
+    progressive = calloc(1, sizeof(*progressive));
+    SOL_NULL_CHECK(progressive, NULL);
+
+    sol_buffer_init(&progressive->buffer);
+    progressive->connection = request->connection;
+    progressive->delete_me = false;
```
barbieri: why set delete_me to false but not the other booleans? either set all or none.

src/lib/comms/sol-http-server-impl-microhttpd.c
```diff
@@ -960,6 +1038,80 @@ sol_http_server_send_response(struct sol_http_request *request, struct sol_http_
...
+
+    MHD_resume_connection(request->connection);
+
+    mhd_response = build_mhd_response_progressive(response, progressive);
+    SOL_NULL_CHECK_GOTO(mhd_response, err);
+
+    ret = MHD_queue_response(request->connection, SOL_HTTP_STATUS_OK, mhd_response);
```
barbieri: shouldn't we get the status from the `response`? This may be used to deliver everything, not just successful content.

src/lib/comms/sol-http-server-impl-microhttpd.c
```diff
@@ -960,6 +1038,80 @@ sol_http_server_send_response(struct sol_http_request *request, struct sol_http_
...
+
+    SOL_INT_CHECK_GOTO(ret, != MHD_YES, err);
+
+    progressive->on_del = on_del;
+    progressive->cb_data = cb_data;
+
+    return progressive;
+
+err:
+    free(progressive);
```
barbieri: you should `sol_buffer_fini(&progressive->buffer);` to be safe. It won't do a thing if buffer is unused.

And you should also handle the connection... you did `MHD_resume_connection()`, do you need to abort it?

src/lib/comms/sol-http-server-impl-microhttpd.c
```diff
@@ -960,6 +1038,80 @@ sol_http_server_send_response(struct sol_http_request *request, struct sol_http_
...
+sol_http_response_progressive_del(struct sol_http_response_progressive *progressive, bool graceful_del)
+{
+    SOL_NULL_CHECK(progressive);
+    SOL_EXP_CHECK(progressive->delete_me == true);
+
+    if (progressive->suspended)
+        MHD_resume_connection(progressive->connection);
+
+    progressive->graceful_del = graceful_del;
+    progressive->delete_me = true;
```
barbieri: IMO it's safer to always set these before calling other libraries. For some reason if `MHD_resume_connection()` calls you directly, you'd miss these values.

src/lib/datatypes/include/sol-buffer.h
```diff
@@ -420,6 +420,17 @@ int sol_buffer_append_slice(struct sol_buffer *buf, const struct sol_str_slice s
...
 /**
+ * @brief Appends the contents of the buffer @c from in the end of @c buf,
+ * reallocating @c if necessary.
+ *
+ * @param dst Destination buffer
+ * @param src The buffer from where the data will be copied
+ *
+ * @return @c 0 on success, error code (always negative) otherwise
+ */
+int sol_buffer_append_buffer(struct sol_buffer *dst, const struct sol_buffer *src);
```
barbieri: do we still need this?

src/samples/http/server-sse.c
```diff
@@ -0,0 +1,295 @@
...
+static int port = 8080;
+
+static bool
+on_stdin(void *data, int fd, uint32_t flags)
+{
+    uint16_t i;
+    struct sol_http_response_progressive *sse;
+    static const struct sol_str_slice begin = {
+        .len = sizeof("data: ") - 1,
+        .data = (const char *)"data: "
```
barbieri: why not use the macro? it does the same.

src/samples/http/server-sse.c
```diff
@@ -0,0 +1,295 @@
...
+        if (err < 0) {
+            fprintf(stderr, "ERROR: failed to read from stdin: %s\n",
+                sol_util_strerrora(-err));
+            goto err;
+        }
+
+        if (value.used == 0) {
+            /* no data usually means ^D on the terminal, quit the application */
+            printf("no data on stdin, quitting.\n");
+            sol_quit();
```
barbieri: we should close the progressive connection and wait for it to gracefully exit before we do quit, otherwise we may lose data.

src/samples/http/server-sse.c
```diff
@@ -0,0 +1,295 @@
...
+        } else {
+            err = sol_buffer_insert_slice(&value, 0, begin);
+            if (err < 0) {
+                fprintf(stderr, "ERROR: failed to append the data prefix: %s\n",
```
barbieri: s/append/prepend/g

src/samples/http/server-sse.c
```diff
@@ -0,0 +1,295 @@
...
+                fprintf(stderr, "ERROR: Could not append data to buffer: %s\n",
+                    sol_util_strerrora(-err));
+                goto err;
+            }
+        }
+    }
+
+    SOL_PTR_VECTOR_FOREACH_IDX (&responses, sse, i)
+        sol_http_response_progressive_feed(sse, sol_buffer_get_slice(&value));
+
```
barbieri: did you valgrind this? we should `sol_buffer_fini(&value)`, no? You're likely missing the leak since for some unknown reason you used "value" as static global, then you must run valgrind and ask it to show all kinds of leak (this is under "still reachable").

My preference is to keep buffers local, so leaks will be easily spotted.

src/samples/http/server-sse.c
```diff
@@ -0,0 +1,295 @@
...
+        return ret;
+
+    sse = sol_http_server_send_response_progressive(request, &response, delete_cb, NULL);
+    sol_http_params_clear(&response.param);
+    if (!sse)
+        return -1;
+
+    ret = sol_ptr_vector_append(&responses, sse);
+    if (ret < 0) {
+        sol_http_response_progressive_del(sse, false);
```
barbieri: I'm not sure this part will not cause a double free, try to force this error and confirm with valgrind.

`sol_http_response_progressive_del()` will just mark, but will keep the `sse` alive, but then this function returns an error and `request` will be deleted. Please confirm with a forced error if that works as expected.

src/samples/http/server-sse.c
```diff
@@ -0,0 +1,295 @@
...
+        .response_code = SOL_HTTP_STATUS_OK,
+    };
+
+    r = sol_http_request_get_interface_address(request, &addr);
+    SOL_INT_CHECK(r, < 0, r);
+
+    SOL_NULL_CHECK(sol_network_link_addr_to_str(&addr, &buf), -1);
+
+    sol_buffer_init(&response.content);
+    r = sol_buffer_append_printf(&response.content, HTML_FILE,
```
barbieri: add a comment to say that SSE needs url to match so you need to do this. also test with a remote browser.

src/samples/http/server-sse.c
```diff
@@ -0,0 +1,295 @@
...
+    }
+
+    if (sol_http_server_register_handler(server, "/events",
+        request_events_cb, NULL) < 0) {
+        fprintf(stderr, "ERROR: Failed to register the handler\n");
+        goto err;
+    }
+
+    printf("HTTP server at port %d.\nDefault reply set to '%.*s'\n",
+        port, SOL_STR_SLICE_PRINT(sol_buffer_get_slice(&value)));
```
barbieri: this "value" here will always be empty since no options or arguments set it.

Please change this to explain to the user he should type something to send to connected servers.

src/samples/http/server-sse.c
barbieri: my bad, there is a `sol_buffer_fini()` in shutdown function. But as I see there is no reason why this buffer should be global.

src/samples/http/server-sse.c
ceolin: i've tested

src/samples/http/server-sse.c
ceolin: nope, the macro is:
```
(struct sol_str_slice){.len = (_len), .data = (_s) }
```
This construction makes the semaphoreci's gcc complains about using not const data to initialize ..

src/lib/datatypes/include/sol-buffer.h
ceolin: we are no using anymore, but this function is still good isn't ?

src/lib/comms/sol-http-server-impl-microhttpd.c
ceolin: ok, i change it

src/lib/comms/sol-http-server-impl-microhttpd.c
ceolin: damn it, it is a left over from when it was using malloc

src/lib/comms/sol-http-server-impl-microhttpd.c
ceolin: the same problem here, fixing

src/lib/comms/sol-http-server-impl-microhttpd.c
ceolin: this is not a patter that we are following is ?

ceolin: i have just seen it when we have more than one callback with the same signature

src/lib/comms/include/sol-http-server.h
ceolin: no, i forgot change the documentation

src/lib/comms/include/sol-http-server.h
```diff
@@ -204,10 +214,99 @@ int sol_http_server_set_last_modified(struct sol_http_server *server, const char
...
 
 /**
+ * @brief Set the necessary headers to allow server sent events
+ *
+ * @param response The response to set the headers
+ *
+ * @return @c 0 on success, error code (always negative) otherwise.
+ */
+static inline int
+sol_http_response_set_progressive_headers(struct sol_http_response *response)
```
ceolin: i'm not sure. These headers are necessary to keep the connection open, we can not use the progressive response without those headers, maybe the cache-control is not necessary.

The name of this function is also something that i was in doubt.

ceolin:
fixes on https://github.com/solettaproject/soletta/pull/1794

src/lib/datatypes/include/sol-buffer.h
barbieri: ok, still good... but I'd make all of these single-line helpers static inlines... so we don't have to maintain many symbols in our ABI.

src/samples/http/server-sse.c
barbieri: what about `SOL_STR_SLICE_LITERAL()`?

PR #1743, opened 2016-03-28
Add network support for Zephyr

ibriano:
Implemented up to sol-socket, tested with a basic CoAP sample, it still
has some issues making two boards talk to each other, but they don't
seem to come from our side.
In any case, getting this upstream will help to avoid rebasing
endlessly every time I come back to it.


src/lib/comms/sol-network-impl-zephyr.c
```diff
@@ -0,0 +1,192 @@
...
+
+#include "sol-log.h"
+#include "sol-network.h"
+#include "sol-util.h"
+#include "sol-vector.h"
+
+static struct sol_vector links = SOL_VECTOR_INIT(struct sol_network_link);
+
+static inline char
+to_hex(uint8_t num)
```
vcgomes: Unused function?

vcgomes:
I just did a quick look, looking good. If that's alright, I would like to take another closer look later today.

src/lib/comms/sol-network-impl-zephyr.c
ibriano: On Mon, 28 Mar 2016 11:32:29 -0700, Vinicius Gomes wrote:
> > + */
> > +
> > +#include <errno.h>
> > +
> > +#include <net/net_core.h>
> > +#include <contiki/ip/uiplib.h>
> > +
> > +#include "sol-log.h"
> > +#include "sol-network.h"
> > +#include "sol-util.h"
> > +#include "sol-vector.h"
> > +
> > +static struct sol_vector links = SOL_VECTOR_INIT(struct sol_network_link);
> > +
> > +static inline char
> > +to_hex(uint8_t num)
> 
> Unused function?

Interesting, I can vividly remember deleting it.

src/lib/comms/sol-socket-impl-zephyr.c
```diff
@@ -0,0 +1,383 @@
...
+    }
+
+    memcpy(buf, ip_buf_appdata(netbuf), buflen);
+
+    ip_buf_unref(netbuf);
+
+    return buflen;
+}
+
+#include <ip/simple-udp.h>
```
edersondisouza: Is there any reason for this `include` to be here? (Compared to be on file header...)

ibriano: On Mon, 28 Mar 2016 11:45:38 -0700, Ederson de Souza wrote:
> > +        buflen = len;
> > +
> > +    if (cliaddr) {
> > +        cliaddr->family = SOL_NETWORK_FAMILY_INET6;
> > +        cliaddr->port = uip_ntohs(NET_BUF_UDP(netbuf)->srcport);
> > +        memcpy(cliaddr->addr.in6, &NET_BUF_IP(netbuf)->srcipaddr, sizeof(cliaddr->addr.in6));
> > +    }
> > +
> > +    memcpy(buf, ip_buf_appdata(netbuf), buflen);
> > +
> > +    ip_buf_unref(netbuf);
> > +
> > +    return buflen;
> > +}
> > +
> > +#include <ip/simple-udp.h>
> 
> Is there any reason for this `include` to be here? (Compared to be on file header...)

Because it's a horrible hack, but unfortunately we need it to be able to
get the udp_conn pointer from the net_context.
If it were possible, I'd also have an #uninclude after the function that
uses it.

src/lib/comms/sol-socket-impl-zephyr.c
```diff
@@ -0,0 +1,383 @@
...
+
+    bindaddr.family = sol_network_sol_to_af(addr->family);
+    memcpy(&bindaddr.in6_addr, addr->addr.in6, sizeof(bindaddr.in6_addr));
+
+    ctx = net_context_get(IPPROTO_UDP, NULL, 0, &bindaddr, addr->port);
+    SOL_NULL_CHECK(ctx, -ENOBUFS);
+
+    socket->unicast_context = ctx;
+
+    sol_ptr_vector_append(&bound_sockets, socket);
```
edersondisouza: Can't this fail?

src/lib/comms/sol-socket-impl-zephyr.c
edersondisouza: Is it worth comment about it, you know, to avoid hacks to become norm, and norm to become patterns?

src/lib/comms/sol-socket-impl-zephyr.c
ibriano: On Mon, 28 Mar 2016 11:51:11 -0700, Ederson de Souza wrote:
> > +
> > +    if (addr->family != SOL_NETWORK_FAMILY_INET6)
> > +        return -EAFNOSUPPORT;
> > +
> > +    if (socket->unicast_context)
> > +        return -EALREADY;
> > +
> > +    bindaddr.family = sol_network_sol_to_af(addr->family);
> > +    memcpy(&bindaddr.in6_addr, addr->addr.in6, sizeof(bindaddr.in6_addr));
> > +
> > +    ctx = net_context_get(IPPROTO_UDP, NULL, 0, &bindaddr, addr->port);
> > +    SOL_NULL_CHECK(ctx, -ENOBUFS);
> > +
> > +    socket->unicast_context = ctx;
> > +
> > +    sol_ptr_vector_append(&bound_sockets, socket);
> 
> Can't this fail?

Yes. Also, I left so much debug in there...

src/lib/comms/sol-socket-impl-zephyr.c
```diff
@@ -0,0 +1,383 @@
...
+
+static void
+sol_socket_zephyr_del(struct sol_socket *s)
+{
+    struct sol_socket_zephyr *socket = (struct sol_socket_zephyr *)s;
+    struct net_context *ctx;
+    struct net_buf *buf;
+    uint16_t idx;
+
+    sol_ptr_vector_del_element(&bound_sockets, socket);
```
glima: Maybe if all bound sockets go away we can kill the fiber as well.

src/lib/comms/sol-socket-impl-zephyr.c
```diff
@@ -0,0 +1,383 @@
...
+        .data = s
+    };
+
+    nano_fiber_sem_take(&s->lock, TICKS_UNLIMITED);
+
+    sol_ptr_vector_append(&s->bufs, buf);
+
+    if (!s->read_available) {
+        sol_mainloop_event_post(&me);
+        s->read_available = true;
```
glima: This is never checked elsewhere.

ibriano: On Mon, 28 Mar 2016 13:59:05 -0700, Gustavo Lima Chaves wrote:
> > +
> > +static void
> > +socket_signal_mainloop(struct sol_socket_zephyr *s, struct net_buf *buf)
> > +{
> > +    struct mainloop_event me = {
> > +        .cb = socket_read_available,
> > +        .data = s
> > +    };
> > +
> > +    nano_fiber_sem_take(&s->lock, TICKS_UNLIMITED);
> > +
> > +    sol_ptr_vector_append(&s->bufs, buf);
> > +
> > +    if (!s->read_available) {
> > +        sol_mainloop_event_post(&me);
> > +        s->read_available = true;
> 
> This is never checked elsewhere.

It's checked right in that if.

ibriano: On Mon, 28 Mar 2016 13:59:05 -0700, Gustavo Lima Chaves wrote:
> > +
> > +static void
> > +socket_signal_mainloop(struct sol_socket_zephyr *s, struct net_buf *buf)
> > +{
> > +    struct mainloop_event me = {
> > +        .cb = socket_read_available,
> > +        .data = s
> > +    };
> > +
> > +    nano_fiber_sem_take(&s->lock, TICKS_UNLIMITED);
> > +
> > +    sol_ptr_vector_append(&s->bufs, buf);
> > +
> > +    if (!s->read_available) {
> > +        sol_mainloop_event_post(&me);
> > +        s->read_available = true;
> 
> This is never checked elsewhere.

The problem is that I'm not setting it back to false when read.

ibriano:
Pushed new version. Fixes galore!

 - Added a new commit to remove the whole sol_socket_*_get_impl(), now it's a single name.
 - Removed debug and unused function leftovers.
 - Bunch of fixes on sol-socket-impl-zephyr:
   * Use sol-list for for the list of bound_sockets.
   * Checked that appending to ptr-vectors may fail, which was, because their init() was missing.
   * More better protections around things shared by task/fiber
   * Receive fiber kills itself if there are no more sockets to listen on.

src/lib/comms/sol-socket-impl-zephyr.c
```diff
@@ -0,0 +1,438 @@
...
+        if (itr == &bound_sockets)
+            itr = itr->next;
+        s = SOL_LIST_GET_CONTAINER(itr, struct sol_socket_zephyr, bound_list);
+        nano_fiber_sem_give(&bound_sockets_lock);
+
+        buf = net_receive(s->unicast_context, 100);
+        if (buf)
+            socket_signal_mainloop(s, buf);
+        SOL_PTR_VECTOR_FOREACH_IDX (&s->mcast_contexts, ctx, i) {
+            buf = net_receive(ctx, 100);
```
vcgomes: Now I feel your pain. And having a fiber per context would be even more painful. I just feel that 100 ticks is too fine grained.

ibriano: On Tue, 29 Mar 2016 13:54:25 -0700, Vinicius Gomes wrote:
> > +        struct net_buf *buf;
> > +        uint16_t i;
> > +
> > +        nano_fiber_sem_take(&bound_sockets_lock, TICKS_UNLIMITED);
> > +        if (sol_list_is_empty(&bound_sockets))
> > +            break;
> > +        if (itr == &bound_sockets)
> > +            itr = itr->next;
> > +        s = SOL_LIST_GET_CONTAINER(itr, struct sol_socket_zephyr, bound_list);
> > +        nano_fiber_sem_give(&bound_sockets_lock);
> > +
> > +        buf = net_receive(s->unicast_context, 100);
> > +        if (buf)
> > +            socket_signal_mainloop(s, buf);
> > +        SOL_PTR_VECTOR_FOREACH_IDX (&s->mcast_contexts, ctx, i) {
> > +            buf = net_receive(ctx, 100);
> 
> Now I feel your pain. And having a fiber per context would be even more painful. I just feel that 100 ticks is too fine grained.

I have absolutely no idea what would be a reasonable number to put
there. Even more so considering that this should in no way be how this
should be done.
I'm open to suggestions.

vcgomes: For now I am more worried about wakeups, did you think about having a fiber per context?

ibriano: On Tue, 29 Mar 2016 14:26:01 -0700, Vinicius Gomes wrote:
> > +        struct net_buf *buf;
> > +        uint16_t i;
> > +
> > +        nano_fiber_sem_take(&bound_sockets_lock, TICKS_UNLIMITED);
> > +        if (sol_list_is_empty(&bound_sockets))
> > +            break;
> > +        if (itr == &bound_sockets)
> > +            itr = itr->next;
> > +        s = SOL_LIST_GET_CONTAINER(itr, struct sol_socket_zephyr, bound_list);
> > +        nano_fiber_sem_give(&bound_sockets_lock);
> > +
> > +        buf = net_receive(s->unicast_context, 100);
> > +        if (buf)
> > +            socket_signal_mainloop(s, buf);
> > +        SOL_PTR_VECTOR_FOREACH_IDX (&s->mcast_contexts, ctx, i) {
> > +            buf = net_receive(ctx, 100);
> 
> For now I am more worried about wakeups, did you think about having a fiber per context?

I can try that. I don't expect to have many contexts here anyway.

src/lib/comms/sol-socket-impl-zephyr.c
```diff
@@ -0,0 +1,438 @@
...
+    tuple = net_context_get_tuple(socket->unicast_context);
+    SOL_NULL_CHECK(tuple, -EINVAL);
+
+    bindaddr.family = sol_network_sol_to_af(group->family);
+    memcpy(&bindaddr.in6_addr, group->addr.in6, sizeof(bindaddr.in6_addr));
+
+    ctx = net_context_get(IPPROTO_UDP, NULL, 0, &bindaddr, tuple->local_port);
+    SOL_NULL_CHECK(ctx, -ENOBUFS);
+
+    ret = sol_ptr_vector_append(&socket->mcast_contexts, ctx);
```
glima: We don't lock on the mcast sockets list too?

ibriano: On Wed, 30 Mar 2016 07:12:52 -0700, Gustavo Lima Chaves wrote:
> > +    struct net_addr bindaddr;
> > +    int ret;
> > +
> > +    if (group->family != SOL_NETWORK_FAMILY_INET6)
> > +        return -EAFNOSUPPORT;
> > +
> > +    tuple = net_context_get_tuple(socket->unicast_context);
> > +    SOL_NULL_CHECK(tuple, -EINVAL);
> > +
> > +    bindaddr.family = sol_network_sol_to_af(group->family);
> > +    memcpy(&bindaddr.in6_addr, group->addr.in6, sizeof(bindaddr.in6_addr));
> > +
> > +    ctx = net_context_get(IPPROTO_UDP, NULL, 0, &bindaddr, tuple->local_port);
> > +    SOL_NULL_CHECK(ctx, -ENOBUFS);
> > +
> > +    ret = sol_ptr_vector_append(&socket->mcast_contexts, ctx);
> 
> We don't lock on the mcast sockets list too?

Oh for fucks sake... yes, we should, and it's going to be a beautiful
mess.

src/lib/comms/sol-network-impl-zephyr.c
```diff
@@ -0,0 +1,186 @@
...
+}
+
+SOL_API const struct sol_network_link_addr *
+sol_network_link_addr_from_str(struct sol_network_link_addr *addr, const char *buf)
+{
+    SOL_NULL_CHECK(addr, NULL);
+    SOL_NULL_CHECK(buf, NULL);
+
+    if (addr->family != SOL_NETWORK_FAMILY_INET6)
+        return NULL;
```
glima: Maybe set errno, as you do elsewhere.

ibriano: On Wed, 30 Mar 2016 07:24:25 -0700, Gustavo Lima Chaves wrote:
> > +        sep = true;
> > +    }
> > +    if (skipping)
> > +        sol_buffer_append_char(buf, ':');
> > +
> > +    return p;
> > +}
> > +
> > +SOL_API const struct sol_network_link_addr *
> > +sol_network_link_addr_from_str(struct sol_network_link_addr *addr, const char *buf)
> > +{
> > +    SOL_NULL_CHECK(addr, NULL);
> > +    SOL_NULL_CHECK(buf, NULL);
> > +
> > +    if (addr->family != SOL_NETWORK_FAMILY_INET6)
> > +        return NULL;
> 
> Maybe set errno, as you do elsewhere.

Maybe we should, I don't know, but I just looked at the other
implementations and we are not setting errno in those, so consistency
mandates that I get to be lazy on this one.

ibriano:
Another update:
 - Use a fiber per net_context. This uses more resources, but now everything works a lot better. Large packets may still be dropped, but there's nothing we can do about it.
 - Since we'll have multiple fibers now, reduce the stack size for them. Might be able to reduce more, but I didn't test what would be a good minimal number.

src/lib/comms/sol-socket-impl-zephyr.c
```diff
@@ -0,0 +1,423 @@
...
+unsupported_family:
+    errno = EAFNOSUPPORT;
+    return NULL;
+}
+
+static void
+socket_net_context_del(struct sol_socket_net_context *ctx)
+{
+    if (ctx->fiber)
+        fiber_fiber_delayed_start_cancel(ctx->fiber);
```
vcgomes: Please add a comment explaining why using this function here is ... less than ideal.

src/lib/comms/sol-socket-impl-zephyr.c
```diff
@@ -0,0 +1,423 @@
...
+
+    struct {
+        bool (*cb)(void *data, struct sol_socket *s);
+        const void *data;
+    } read, write;
+
+    struct sol_timeout *write_timeout;
+
+    struct sol_socket_net_context *unicast_context;
+    struct sol_ptr_vector mcast_contexts;
```
vcgomes: [optional] I would have only a `struct sol_ptr_vector contexts` field, and make sure that the first context is the unicast context.

ibriano: On Wed, 30 Mar 2016 14:42:29 -0700, Vinicius Gomes wrote:
> > +#include "sol-network-util.h"
> > +
> > +struct sol_socket_net_context;
> > +
> > +struct sol_socket_zephyr {
> > +    struct sol_socket base;
> > +
> > +    struct {
> > +        bool (*cb)(void *data, struct sol_socket *s);
> > +        const void *data;
> > +    } read, write;
> > +
> > +    struct sol_timeout *write_timeout;
> > +
> > +    struct sol_socket_net_context *unicast_context;
> > +    struct sol_ptr_vector mcast_contexts;
> 
> [optional] I would have only a `struct sol_ptr_vector contexts` field, and make sure that the first context is the unicast context.

I thought of that, then thought "What if the user joins groups before
binding?", but really, that would break anyway.

vcgomes: You could add a check for an empty `contexts` in `_join_group()`

ibriano: So, in the end we decided to leave it as is to be coherent with how Linux and RIOT work regarding order of bind() and join_group(), right?

vcgomes: Yeah, I think that we should keep it consistent with the rest, erroring out if we try to join groups in a unbound socket.

But I think the suggestion I gave is orthogonal to that point, and it is really optional, so up to you.

ibriano: On Thu, 31 Mar 2016 10:04:53 -0700, Vinicius Gomes wrote:
> > +#include "sol-network-util.h"
> > +
> > +struct sol_socket_net_context;
> > +
> > +struct sol_socket_zephyr {
> > +    struct sol_socket base;
> > +
> > +    struct {
> > +        bool (*cb)(void *data, struct sol_socket *s);
> > +        const void *data;
> > +    } read, write;
> > +
> > +    struct sol_timeout *write_timeout;
> > +
> > +    struct sol_socket_net_context *unicast_context;
> > +    struct sol_ptr_vector mcast_contexts;
> 
> Yeah, I think that we should keep it consistent with the rest, erroring out if we try to join groups in a unbound socket.
> 
> But I think the suggestion I gave is orthogonal to that point, and it is really optional, so up to you.

https://www.youtube.com/watch?v=dqM5L5JQseI

vcgomes: Fair enough. Using "Lazy" as argument was a low blow.

ibriano:
Merged so master works again with latest Zephyr. Small fixes can come later.

PR #1752, opened 2016-03-28
Add http server/client json and blob nodes

cabelitos:
@ceolin

### DIFF
--- a/src/modules/flow/http-client/http-client.c
+++ b/src/modules/flow/http-client/http-client.c
@@ -556,7 +556,7 @@ string_process_token(struct sol_flow_node *node,
     result = strndup(value->start + 1, value->end - value->start - 2);
     if (result) {
         return sol_flow_send_string_take_packet(node,
-            SOL_FLOW_NODE_TYPE_HTTP_CLIENT_BOOLEAN__OUT__OUT, result);
+            SOL_FLOW_NODE_TYPE_HTTP_CLIENT_STRING__OUT__OUT, result);
     }
 
     return -ENOMEM;
@@ -574,7 +574,7 @@ string_process_data(struct sol_flow_node *node,
 
     if (result) {
         return sol_flow_send_string_take_packet(node,
-            SOL_FLOW_NODE_TYPE_HTTP_CLIENT_BOOLEAN__OUT__OUT, result);
+            SOL_FLOW_NODE_TYPE_HTTP_CLIENT_STRING__OUT__OUT, result);
     }
 
     return -ENOMEM;
--- a/src/modules/flow/http-client/http-client.c
+++ b/src/modules/flow/http-client/http-client.c
@@ -40,7 +40,6 @@ struct http_data {
     enum sol_http_method method;
     struct sol_http_params url_params;
     char *url;
-    char *content_type;
     char *accept;
     char *last_modified;
     bool machine_id;
@@ -219,7 +218,6 @@ common_close(struct sol_flow_node *node, void *data)
         type->close_node(node, data);
 
     free(mdata->url);
-    free(mdata->content_type);
     free(mdata->accept);
     free(mdata->last_modified);
     SOL_PTR_VECTOR_FOREACH_IDX (&mdata->pending_conns, connection, i)
@@ -237,8 +235,9 @@ common_open(struct sol_flow_node *node, void *data, const struct sol_flow_node_o
         (struct sol_flow_node_type_http_client_boolean_options *)options;
 
     mdata->machine_id = opts->machine_id;
-    sol_ptr_vector_init(&mdata->pending_conns);
+    mdata->strict = opts->strict;
 
+    sol_ptr_vector_init(&mdata->pending_conns);
     sol_http_params_init(&mdata->url_params);
 
     if (opts->url && opts->url[0] != '\0') {
@@ -247,7 +246,16 @@ common_open(struct sol_flow_node *node, void *data, const struct sol_flow_node_o
         get_key(mdata);
     }
 
+    if (opts->accept) {
+        mdata->accept = strdup(opts->accept);
+        SOL_NULL_CHECK_GOTO(mdata->accept, err_accept);
+    }
+
     return 0;
+
+err_accept:
+    free(mdata->url);
+    return -ENOMEM;
 }
 
 static int
@@ -358,7 +366,17 @@ common_request_finished(void *data,
     SOL_INT_CHECK_GOTO(ret, < 0, err);
 
     type = (const struct http_client_node_type *)sol_flow_node_get_type(node);
-    if (streq(response->content_type, "application/json")) {
+
+    if (mdata->strict && mdata->accept && response->content_type &&
+        !streq(response->content_type, mdata->accept)) {
+        sol_flow_send_error_packet(node, EINVAL,
+            "Response has different content type. Received: %s - Desired: %s",
+            response->content_type, mdata->accept);
+        return;
+    }
+
+    if (streq(response->content_type, "application/json") &&
+        type->process_token) {
         struct sol_json_scanner scanner;
         struct sol_json_token token, key, value;
         enum sol_json_loop_reason reason;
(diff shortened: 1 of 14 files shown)


src/modules/flow/http-server/http-server.c
```diff
@@ -255,40 +248,84 @@ common_response_cb(void *data, struct sol_http_request *request)
...
+    SOL_INT_CHECK_GOTO(r, < 0, err_exit);
 
     if (send_json) {
-        r = sol_buffer_append_char(&response.content, '}');
-        SOL_INT_CHECK_GOTO(r, < 0, end);
+        r = sol_buffer_append_char(&response->content, '}');
+        SOL_INT_CHECK_GOTO(r, < 0, err_exit);
+    }
+
+    err_r = sol_http_param_add(&response->param, SOL_HTTP_REQUEST_PARAM_HEADER(
```
glima: Ugh, the code in this function even tries to return ``-EINVAL``, it's crying to have the return type changed.

src/modules/flow/http-server/http-server.json
```diff
@@ -498,6 +504,143 @@
       ],
       "private_data_type": "http_data",
       "url": "http://solettaproject.org/doc/latest/node_types/http-server/direction-vector.html"
+    },
+    {
+      "category": "output/network",
+      "description": "HTTP Server for Blob.",
```
glima: We got to explain (much) better what our nodes do here :/

src/modules/flow/http-server/http-server.json
```diff
@@ -498,6 +504,143 @@
...
+          {
+            "data_type": "string",
+            "default": "/blob",
+            "description": "The http path where it will be served",
+            "name": "path"
+          },
+          {
+            "data_type": "int",
+            "default": -1,
+            "description": "The port used to bind the server. If a negative value, default port will be used.",
```
glima: bind the server TO.

Also, which default port is that?

src/modules/flow/http-server/http-server.json
edersondisouza: On 29-03-2016 11:15, Gustavo Lima Chaves wrote:
> In src/modules/flow/http-server/http-server.json
> <https://github.com/solettaproject/soletta/pull/1752#discussion_r57729943>:
>
>> @@ -498,6 +504,143 @@
>>        ],
>>        "private_data_type": "http_data",
>>        "url": "http://solettaproject.org/doc/latest/node_types/http-server/direction-vector.html"
>> +    },
>> +    {
>> +      "category": "output/network",
>> +      "description": "HTTP Server for Blob.",
>
> We got to explain (much) better what our nodes do here :/

Http Server for Bob. If you're not Bob, move over =P

cabelitos: hahhaaah

src/modules/flow/http-client/http-client.c
```diff
@@ -358,7 +366,17 @@ common_request_finished(void *data,
...
+    if (mdata->strict && mdata->accept && response->content_type &&
+        !streq(response->content_type, mdata->accept)) {
+        sol_flow_send_error_packet(node, EINVAL,
+            "Response has different content type. Received: %s - Desired: %s",
+            response->content_type, mdata->accept);
+        return;
+    }
+
+    if (streq(response->content_type, "application/json") &&
+        type->process_token) {
```
ceolin: why are you checking if `type->process_token` is not NULL and is not checking for `type->process_data` in the `else` statement ?

Also, is it expected the response be processed by `type->process_data` in case of this method not be defined ?

src/modules/flow/http-client/http-client.c
```diff
@@ -399,8 +422,8 @@ common_get_process(struct sol_flow_node *node, void *data, uint16_t port, uint16
     }
 
     sol_http_params_init(&params);
-    if (!sol_http_param_add(&params,
-        SOL_HTTP_REQUEST_PARAM_HEADER("Accept", "application/json"))) {
+    if (mdata->accept && !sol_http_param_add(&params,
+        SOL_HTTP_REQUEST_PARAM_HEADER("Accept", mdata->accept))) {
```
ceolin: `mdata->accept` is an option optional. This will always fail if this option was not given.
```
if (mdata->accept) {
    if(!sol_httt_param_add(...)) {
        ...
    }
}
```

ceolin: forgot, i missunderstood. This is right

src/modules/flow/http-server/http-server.c
```diff
@@ -255,40 +248,84 @@ common_response_cb(void *data, struct sol_http_request *request)
...
+
+    method = sol_http_request_get_method(request);
+    response.url = sol_http_request_get_url(request);
+
+    if (!is_method_allowed(mdata->allowed_methods, method)) {
+        SOL_INF("HTTP Method not allowed. Method: %d", (int)method);
+        response.response_code = SOL_HTTP_STATUS_FORBIDDEN;
+        r = sol_http_server_send_response(request, &response);
+        if (r < 0)
+            SOL_WRN("Could not send the forbidden response");
```
ceolin: should send an error packet here as in the default error treatment (see label end)

src/modules/flow/http-server/http-server.c
```diff
@@ -994,6 +1032,164 @@ direction_vector_open(struct sol_flow_node *node, void *data,
...
+
+        SOL_HTTP_PARAMS_FOREACH_IDX (sol_http_request_get_params(request),
+            param, i) {
+            if (param->type != SOL_HTTP_PARAM_POST_DATA)
+                continue;
+
+            blob_mem = malloc(param->value.data.value.len);
+            SOL_NULL_CHECK(blob_mem, -ENOMEM);
+            memcpy(blob_mem, param->value.data.value.data,
+                param->value.data.value.len);
```
ceolin: `sol_util_memdup` does the malloc + memcpy, use it.

src/modules/flow/http-server/http-server.c
```diff
@@ -994,6 +1032,164 @@ direction_vector_open(struct sol_flow_node *node, void *data,
...
+    }
+
+    r = sol_buffer_append_bytes(&response->content,
+        (uint8_t *)mdata->value.blob->mem, mdata->value.blob->size);
+    SOL_INT_CHECK(r, < 0, r);
+
+    if (!sol_http_param_add(&response->param, SOL_HTTP_REQUEST_PARAM_HEADER(
+        HTTP_HEADER_CONTENT_TYPE, HTTP_HEADER_CONTENT_TYPE_JSON))) {
+        SOL_WRN("Could not set the Content-Type:application/json");
+        return -ENOMEM;
```
ceolin: `response.content` will leak. You need to call `sol_buffer_fini`

src/modules/flow/http-server/http-server.c
```diff
@@ -994,6 +1032,164 @@ direction_vector_open(struct sol_flow_node *node, void *data,
...
+    }
+
+    if (!mdata->value.blob) {
+        response->response_code = SOL_HTTP_STATUS_NOT_FOUND;
+        return 0;
+    }
+
+    r = sol_buffer_append_bytes(&response->content,
+        (uint8_t *)mdata->value.blob->mem, mdata->value.blob->size);
+    SOL_INT_CHECK(r, < 0, r);
```
ceolin: i guess is better send a response with `INTERNAL_ERROR` code

src/modules/flow/http-server/http-server.c
ceolin: the same about the response with error here.

ceolin: ... the error treatment is done by who call this function. Forget these comments

src/modules/flow/http-server/http-server.c
```diff
@@ -994,6 +1032,164 @@ direction_vector_open(struct sol_flow_node *node, void *data,
...
+            mdata->value.blob);
+    } else if (sol_json_is_valid_type(&array_scanner,
+        SOL_JSON_TYPE_ARRAY_START)) {
+        sol_flow_send_json_array_packet(node,
+            SOL_FLOW_NODE_TYPE_HTTP_SERVER_JSON__OUT__ARRAY,
+            mdata->value.blob);
+    } else {
+        sol_flow_send_error_packet(node, EINVAL,
+            "Invalid json format - It will not be sent");
+        SOL_WRN("Invalid json format - It will not be sent");
```
ceolin: i guess the warning message is not necessary. If the error packet is not handled its message is printed.

In this case the message will be printed twice

src/modules/flow/http-server/http-server.json
ceolin: is defined by the user in compile time

src/modules/flow/http-server/http-server.c
cabelitos: :]

src/modules/flow/http-server/http-server.c
cabelitos: fair

src/modules/flow/http-server/http-server.c
cabelitos: We have an easy task for that :]

src/modules/flow/http-server/http-server.c
cabelitos: It will send it. check common_response_cb()

src/modules/flow/http-server/http-server.c
cabelitos: Fair

cabelitos:
#1793

PR #1764, opened 2016-03-29
[Draft] sol-coap: Fix copying the payload for every notification

vcgomes:
Instead of copying the payload for every notification packet, we keep a
reference for the payload, and have a copy of the header (that will
change according to each observer).

### DIFF
--- a/src/lib/comms/sol-coap.c
+++ b/src/lib/comms/sol-coap.c
@@ -93,11 +93,13 @@ struct resource_observer {
 };
 
 struct pending_reply {
+    struct sol_coap_server *server;
+    struct sol_timeout *timeout;
     bool (*cb)(struct sol_coap_server *server, struct sol_coap_packet *req,
         const struct sol_network_link_addr *cliaddr, void *data);
     const void *data;
-    bool observing;
     char *path;
+    bool observing;
     uint16_t id;
     uint8_t tkl;
     uint8_t token[0];
@@ -625,28 +627,58 @@ pending_reply_free(struct pending_reply *reply)
 
     if (reply->observing)
         free(reply->path);
+
+    if (reply->timeout)
+        sol_timeout_del(reply->timeout);
+
     free(reply);
 }
 
 static bool
-call_reply_timeout_cb(struct sol_coap_server *server, struct sol_coap_packet *pkt)
+user_wants_to_retry_by_id(struct sol_coap_server *server, uint16_t id)
 {
-    uint16_t i, id;
+    uint16_t i;
     struct pending_reply *reply;
 
-    sol_coap_header_get_id(pkt, &id);
-
     SOL_PTR_VECTOR_FOREACH_REVERSE_IDX (&server->pending, reply, i) {
-        if (reply->observing || reply->id != id)
+        if (reply->id != id)
             continue;
 
         if (reply->cb(server, NULL, NULL, (void *)reply->data))
-            return false;
+            return true;
         sol_ptr_vector_del(&server->pending, i);
         pending_reply_free(reply);
     }
 
-    return true;
+    return false;
+}
+
+static bool
+user_wants_to_retry(struct sol_coap_server *server, struct sol_coap_packet *pkt)
+{
+    uint16_t id;
+
+    sol_coap_header_get_id(pkt, &id);
+
+    return user_wants_to_retry_by_id(server, id);
+}
+
+/* This is mostly for !CON packets, which do not go to the outgoing
+ * list but also keep a context of their own, for response handling */
+static bool
+pending_timeout_cb(void *data)
+{
+    struct pending_reply *pending_reply = data;
+    struct sol_coap_server *server = pending_reply->server;
+
+    /* The pending context will be kept, as the user wishes */
+    if (user_wants_to_retry_by_id(server, pending_reply->id))
+        return true;
+
+    /* If such ID was found, the struct is now properly deleted as
+     * well */
+    pending_reply->timeout = NULL;
+    return false;
 }
 
 static bool
@@ -663,11 +695,10 @@ timeout_expired(struct sol_coap_server *server, struct outgoing *outgoing)
     if (type == SOL_COAP_TYPE_CON) {
         max_retransmit = MAX_RETRANSMIT;
         timeout = ACK_TIMEOUT_MS << outgoing->counter++;
-    } else {
-        max_retransmit = 1;
-        timeout = NONCON_PKT_TIMEOUT_MS;
-        outgoing->counter++;
-    }
+    } else
+        /* no re-transmissions for !CON packets, we just keep a
+         * pending_reply for a while */
+        return expired;
 
     if (outgoing->counter > max_retransmit) {
         SOL_PTR_VECTOR_FOREACH_REVERSE_IDX (&server->outgoing, o, i) {
@@ -676,7 +707,7 @@ timeout_expired(struct sol_coap_server *server, struct outgoing *outgoing)
                 SOL_DBG("packet id %d dropped, after %d transmissions",
                     id, outgoing->counter);
 
-                if (!call_reply_timeout_cb(server, outgoing->pkt)) {
+                if (user_wants_to_retry(server, outgoing->pkt)) {
                     expired = true;
                     break;
                 }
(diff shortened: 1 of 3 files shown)


src/lib/comms/sol-coap.c
```diff
@@ -678,11 +684,49 @@ timeout_expired(struct sol_coap_server *server, struct outgoing *outgoing)
...
+    static struct sol_buffer buffer;
+    struct sol_coap_packet *header = outgoing->header;
+    struct sol_coap_packet *payload = outgoing->pkt;
+    uint16_t new_size, new_offset, old_offset;
+    uint8_t new_tkl, old_tkl;
+    int r;
+
+    if (!header) {
+        buffer = SOL_BUFFER_INIT_CONST(payload->buf.data, payload->buf.used);
+        return &buffer;
```
barbieri: okay this is a PoC/Draft, but why return a pointer to the local buffer (even if static)? This smells problems, maybe just return the buffer (not a pointer to)?

src/lib/comms/sol-coap.c
```diff
@@ -678,11 +684,49 @@ timeout_expired(struct sol_coap_server *server, struct outgoing *outgoing)
...
+
+    sol_buffer_init(&buffer);
+
+    r = sol_buffer_ensure(&buffer, new_size);
+    SOL_INT_CHECK(r, < 0, NULL);
+
+    new_offset = sizeof(struct coap_header) + new_tkl;
+    old_offset = sizeof(struct coap_header) + old_tkl;
+
+    r = sol_buffer_insert_bytes(&buffer, 0, header->buf.data, new_offset);
```
barbieri: since buffer is empty you can append. insert at the end will be the same, but at least you can skip the index in the call.

src/lib/comms/sol-coap.c
```diff
@@ -678,11 +684,49 @@ timeout_expired(struct sol_coap_server *server, struct outgoing *outgoing)
...
+    SOL_INT_CHECK(r, < 0, NULL);
+
+    r = sol_buffer_insert_bytes(&buffer, new_offset,
```
barbieri: same here, just append.

src/lib/comms/sol-coap.c
```diff
@@ -678,11 +684,49 @@ timeout_expired(struct sol_coap_server *server, struct outgoing *outgoing)
...
+    }
+
+    (void)sol_coap_header_get_token(payload, &old_tkl);
+    (void)sol_coap_header_get_token(header, &new_tkl);
+
+    new_size = payload->buf.used + (new_tkl - old_tkl);
+
+    sol_buffer_init(&buffer);
+
+    r = sol_buffer_ensure(&buffer, new_size);
```
barbieri: be aware this will use our rounding policy. If you know a fixed size, then you can malloc yourself and set the buffer to a fixed capacity.

src/lib/comms/sol-coap.c
```diff
@@ -678,11 +684,49 @@ timeout_expired(struct sol_coap_server *server, struct outgoing *outgoing)
...
+    SOL_INT_CHECK(r, < 0, NULL);
+
+    new_offset = sizeof(struct coap_header) + new_tkl;
+    old_offset = sizeof(struct coap_header) + old_tkl;
+
+    r = sol_buffer_insert_bytes(&buffer, 0, header->buf.data, new_offset);
+    SOL_INT_CHECK(r, < 0, NULL);
+
+    r = sol_buffer_insert_bytes(&buffer, new_offset,
+        (uint8_t *) payload->buf.data + old_offset, payload->buf.used - old_offset);
```
barbieri: `sol_buffer_get_at(&payload->buf, old_offset)`

check the `sol-buffer.h`, you need a refresh man!

src/lib/comms/sol-coap.c
```diff
@@ -696,9 +740,14 @@ on_can_write(void *data, struct sol_socket *s)
     if (!outgoing)
         return false;
 
-    err = sol_socket_sendmsg(s, outgoing->pkt->buf.data,
-        outgoing->pkt->buf.used, &outgoing->cliaddr);
+    buf = prepare_buffer(outgoing);
```
barbieri: as per above, you can either return a buffer, instead of a pointer to a static, or receive a buffer in `prepare_buffer()`, then you give it your local struct to be filled.

glima:
It seems we have an issue with this -- packets are arriving with 0 size for some reason :( Now investigating where the error is...

src/lib/comms/sol-coap.c
```diff
@@ -690,11 +696,55 @@ timeout_expired(struct sol_coap_server *server, struct outgoing *outgoing)
...
+    struct sol_coap_packet *header = outgoing->header;
+    struct sol_coap_packet *payload = outgoing->pkt;
+    uint16_t new_size, new_offset, old_offset;
+    uint8_t new_tkl, old_tkl;
+    uint8_t *buf_data;
+    int r;
+
+    if (!header) {
+        sol_buffer_init_flags(buffer, payload->buf.data, payload->buf.used,
+            SOL_BUFFER_FLAGS_MEMORY_NOT_OWNED);
```
glima: This will *not* set the *used* portion, so we're doomed on line 765.

glima:
I'm pushing my other PR as is so that you can see beforehand (even with this due fix)

src/lib/comms/sol-coap.c
vcgomes: Ugh. Sorry guys. My bad.

src/lib/comms/sol-coap.c
```diff
@@ -690,11 +696,55 @@ timeout_expired(struct sol_coap_server *server, struct outgoing *outgoing)
...
+
+    (void)sol_coap_header_get_token(payload, &old_tkl);
+    (void)sol_coap_header_get_token(header, &new_tkl);
+
+    new_size = payload->buf.used + (new_tkl - old_tkl);
+
+    buf_data = malloc(new_size);
+    SOL_NULL_CHECK(buf_data, -ENOMEM);
+
+    sol_buffer_init_flags(buffer, buf_data, new_size, SOL_BUFFER_FLAGS_DEFAULT);
```
vcgomes: Yeah, makes sense. Will fix.

src/lib/comms/sol-coap.c
```diff
@@ -763,6 +731,8 @@ prepare_buffer(struct outgoing *outgoing, struct sol_buffer *buffer)
         payload->buf.used - old_offset);
     SOL_INT_CHECK_GOTO(r, < 0, error);
 
+    buffer->used = new_size;
```
barbieri: this shouldn't be needed as both appends will increment the size. Just be aware that the fixed-capacity doesn't prevent trailing null-bytes. I'm not sure append_bytes are appending them, but they should. To prevent you can use the `NO_NUL_BYTE` flag.

vcgomes: Just pushed an updated version, with those fixes. 

@glima, just added a few more fixes too, see if it makes sense.

src/lib/comms/sol-coap.c
```diff
@@ -828,11 +795,12 @@ on_can_write(void *data, struct sol_socket *s)
...
-     * By taking this shortcut we reduce memory usage A LOT and are
-     * able to run on very small devices with no memory issues.
+     * If 'timeout' is NULL, it means that the packet doesn't need to
+     * be retransmitted. By taking this shortcut we reduce memory
+     * usage A LOT and are able to run on very small devices with no
+     * memory issues.
      */
-    if (type == SOL_COAP_TYPE_ACK || type == SOL_COAP_TYPE_RESET
-        || type == SOL_COAP_TYPE_NONCON) {
+    if (!outgoing->timeout) {
```
glima: What about ``sol_coap_packet_send_notification()`` with CON packets??

src/lib/comms/sol-coap.c
```diff
@@ -1542,8 +1506,17 @@ respond_packet(struct sol_coap_server *server, struct sol_coap_packet *req,
...
-            } else if (!reply->observing)
+            } else if (!reply->observing) {
                 remove_outgoing = false;
+            } else {
+                /*
+                  This means that the user wishes to continue observing that resource,
+                  so we don't need to keep the reply timeout around.
+                */
+                sol_timeout_del(reply->timeout);
+                reply->timeout = NULL;
```
glima: And this pending reply lives till the end of the program??

src/lib/comms/sol-coap.c
```diff
@@ -634,50 +634,22 @@ pending_reply_free(struct pending_reply *reply)
...
 {
-    struct pending_reply *pending_reply = data;
-    struct sol_coap_server *server = pending_reply->server;
+    struct pending_reply *reply = data;
+    struct sol_coap_server *server = reply->server;
 
-    /* The pending context will be kept, as the user wishes */
-    if (user_wants_to_retry_by_id(server, pending_reply->id))
-        return true;
+    reply->cb(server, NULL, NULL, (void *)reply->data);
```
glima: I was respecting our documentation, which had a (for me, fair) contract about this function's return value:

```
 * After an internal timeout is reached, @a reply_cb will be called
 * with @c NULL @c req and @c cliaddr. The same behavior is expected
 * for its return: if @c true, @a server will issue a new timeout and
 * continue waiting responses until it ends, otherwise @a server will
 * terminate response waiting.
```

vcgomes:
This was integrated as part of https://github.com/solettaproject/soletta/pull/1732. Closing.

PR #1766, opened 2016-03-29
pinmux: Change Edison Arduino breakout detection

mbelluzzo:
Some OS didn't like "double export". It wasn't an issue in the original image.


brunobottazzini:
LGTM. 

Fixed the problems we were having

mbelluzzo:
merged thanks!

PR #1767, opened 2016-03-29

sragavan:
changes.

### DIFF
--- a/bindings/nodejs/lib/gpio.js
+++ b/bindings/nodejs/lib/gpio.js
@@ -1,0 +1,116 @@
+/*
+ * This file is part of the Soletta Project
+ *
+ * Copyright (C) 2015 Intel Corporation. All rights reserved.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+ var soletta = require( 'bindings' )( 'soletta' ),
+    _ = require( 'lodash' );
+
+exports.open = function( init ) {
+    return new Promise( function( fulfill, reject ) {
+        var pin = init.pin;
+        var dir = 0;
+        var drive_mode = 0;
+        var config = null;
+        var gpiopin;
+        var callback_data = [];
+        var edge = "any";
+
+        if ( init.pullup )
+            drive_mode = soletta.sol_gpio_drive.SOL_GPIO_DRIVE_PULL_UP;
+        else if ( init.pullup === false )
+            drive_mode = soletta.sol_gpio_drive.SOL_GPIO_DRIVE_PULL_DOWN;
+        else
+            drive_mode = soletta.sol_gpio_drive.SOL_GPIO_DRIVE_NONE;
+
+        if ( init.edge )
+            edge = init.edge;
+
+        if ( init.direction == "in" ) {
+            config = {
+                dir: soletta.sol_gpio_direction.SOL_GPIO_DIR_IN,
+                active_low: init.activeLow,
+                poll_timeout: init.poll,
+                drive_mode: drive_mode,
+                trigger_mode: edge,
+                callback: function( pin, value ) {
+                    callback_data[0].dispatchEvent( "change", {
+                        type: "change",
+                        value: value
+                    } );
+                },
+            }
+
+        } else {
+            config = {
+                dir: soletta.sol_gpio_direction.SOL_GPIO_DIR_OUT,
+                active_low: init.activeLow,
+                drive_mode: drive_mode,
+            }
+        }
+
+        gpiopin = GPIOPin( soletta.sol_gpio_open( pin, config ) );
+        callback_data.push( gpiopin );
+        fulfill( gpiopin );
+    });
+
+}
+
+var GPIOPin = function( pin ) {
+    if ( !this._isGPIOPin )
+        return new GPIOPin( pin );
+    this._pin = pin;
+}
+
+require( "util" ).inherits( GPIOPin, require( "events" ).EventEmitter );
+
+_.extend( GPIOPin.prototype, {
+    _isGPIOPin: true,
+    onchange: null,
+
+    read: function() {
+        return new Promise( _.bind( function( fulfill, reject ) {
+
+            fulfill( soletta.sol_gpio_read( this._pin ) );
+        }, this ) );
+    },
+
+    write: function( value ) {
+        return new Promise( _.bind( function( fulfill, reject ) {
+            fulfill( soletta.sol_gpio_write( this._pin, value ) );
+        }, this ) );
+    },
+
+    close: function() {
+        return new Promise( _.bind( function( fulfill, reject ) {
+            fulfill( soletta.sol_gpio_close( this._pin) );
+        }, this ) );
+    },
+
+    addEventListener: GPIOPin.prototype.addListener,
+
+    removeEventListener: GPIOPin.prototype.removeListener,
+
+    dispatchEvent: function( event, request ) {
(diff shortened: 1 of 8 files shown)


sragavan:
@gabrielschulhof Can you take this in?

bindings/nodejs/src/structures/handles.cc
```diff
@@ -0,0 +1,42 @@
+/*
```
gabrielschulhof: @nagineni @sragavan This is pretty similar to ```bindings/nodes/src/structures/js-handle.cc```. Can you not reuse that?

bindings/nodejs/src/functions/gpio.cc
```diff
@@ -0,0 +1,111 @@
...
+    VALIDATE_ARGUMENT_TYPE_OR_NULL(info, 0, IsUint32);
+    VALIDATE_ARGUMENT_TYPE(info, 1, IsObject);
+
+    uint32_t pin;
+    sol_gpio_config config;
+    sol_gpio *gpio = NULL;
+
+    pin = info[0]->Uint32Value();
+    if (!c_sol_gpio_config(info[1]->ToObject(), &config)) {
+        printf("Unable to extract sol_gpio_config\n");
```
gabrielschulhof: Please use ```Nan::ThrowError(const char *message)```, not ```printf()```!

bindings/nodejs/src/functions/gpio.cc
```diff
@@ -0,0 +1,111 @@
...
+        return;
+    }
+
+    Nan::Callback *callback = (Nan::Callback *)config.in.user_data;
+    gpio = sol_gpio_open(pin, &config);
+    if (gpio) {
+        info.GetReturnValue().Set(js_sol_gpio(gpio));
+        if (callback) {
+            annotation[gpio] = callback;
+            hijack_ref();
```
gabrielschulhof: Please handle the case where hijack_ref() returns non-zero, i.e. hijacking the main loop has failed!

bindings/nodejs/lib/gpio.js
```diff
@@ -0,0 +1,112 @@
+/*
```
gabrielschulhof: The location of this file is actually kinda problematic. If you want to use soletta GPIO, you now have to ```require( "soletta/bindings/nodejs/lib/gpio" );``` in your project.

Ideally this should be ```require( "soletta/gpio" );```, but that means you would have to move gpio.js to the repository root, which might pollute it.

bindings/nodejs/src/common.h
```diff
@@ -95,3 +95,12 @@
     VALIDATE_VALUE_TYPE(memberName, typecheck, message "." #memberName, \
     failReturn); \
     destination.memberName = (destinationType)memberName->accessor();
+
+#define VALIDATE_AND_ASSIGN_PTR(destination, memberName, destinationType, \
```
gabrielschulhof: You don't need this macro. You can just pass ```(*destination)``` to the above macro.

bindings/nodejs/src/structures/sol-js-gpio.cc
```diff
@@ -0,0 +1,95 @@
...
+        js_sol_gpio(gpio),
+        Nan::New(value)
+    };
+    callback->Call(2, arguments);
+}
+
+bool c_sol_gpio_config(v8::Local<v8::Object> jsGPIOConfig, sol_gpio_config *config) {
+    SOL_SET_API_VERSION(config->api_version = SOL_GPIO_CONFIG_API_VERSION; )
+
+    VALIDATE_AND_ASSIGN_PTR(config, dir, sol_gpio_direction, IsUint32,
```
gabrielschulhof: You can replace this with ```VALIDATE_AND_ASSIGN((*config), ...);``` and avoid the need for creating a new macro. See my [comment](https://github.com/solettaproject/soletta/pull/1767/files#r57879864) at the definition of ```VALIDATE_AND_ASSIGN_PTR()```.

bindings/nodejs/src/functions/gpio.cc
```diff
@@ -0,0 +1,111 @@
...
+    VALIDATE_ARGUMENT_COUNT(info, 1);
+    VALIDATE_ARGUMENT_TYPE(info, 0, IsArray);
+
+    sol_gpio *gpio = NULL;
+
+    if (!c_sol_gpio(Local<Array>::Cast(info[0]), &gpio)) {
+        return;
+    }
+
+    Nan::Callback *callback = annotation[gpio];
```
gabrielschulhof: There is exactly one ```Nan::Callback *``` for each ```sol_gpio *```, so you can allocate a structure
```C
struct {
  sol_gpio *gpio;
  Nan::Callback *callback;
}
```

upon ```bind_sol_gpio_open()```, store the pointer in a JSHandle subclass which you pass back to JS, then receive it back in ```bind_sol_gpio_close()``` and dispose of it. Then you don't need the annotation.

bindings/nodejs/lib/gpio.js
barbieri: take the same approach as we do with `build/stage` or `build/soletta_sysroot`. Then you always copy it and use from there, then it will look like installed version.

Sounds reasonable?

bindings/nodejs/src/structures/sol-js-gpio.cc
```diff
@@ -0,0 +1,95 @@
...
+ */
+
+#include <nan.h>
+#include "sol-js-gpio.h"
+#include "../common.h"
+#include "handles.h"
+
+using namespace v8;
+
+static sol_gpio_edge sol_edge_from_string(Local<Value> jsString) {
```
barbieri: as I said this would be better done in `sol-gpio.h`, that is in the C API. You can also do the other way around, like enum->string. In this sense, for the string->enum, you can use the `sol_str_table` struct. For the enum->string, you can use a simple array with `[value] = string` and validate `if (value >= sizeof(array)) return error;`

Examples:
 - https://github.com/solettaproject/soletta/blob/master/src/lib/common/sol-log.c#L95 string->enum;
 - https://github.com/solettaproject/soletta/blob/master/src/lib/common/sol-log.c#L357 enum->string;

you can name all these functions in the pattern `const char *<enum_name>_to_str(enum)` and `enum <enum_name>_from_str(const char *)`, like `sol_gpio_edge_to_str()` and `sol_gpio_edge_from_str()`

bindings/nodejs/lib/gpio.js
gabrielschulhof: @barbieri The problem is that npm packages do not have an installation step. The repo *is* the installed package. ```npm install``` merely modifies the repo in place to prepare it for use. We could copy the files to the root of the repo as part of ```npm install```. I guess we can keep the file where it is for now, and I'll experiment with locations when I do the OIC JS API.

barbieri: ok, I remember that. But AFAIR when we talked during our F2F in Finland the "pure npm install" method wasn't to be addressed now. That is, evoking `npm install soletta` wouldn't be supported. The usage of "npm" was to be used inside our build to ease the process of creating the bindings (guest build system, not host build system).

In order to refresh our discussion, my arguments is that Soletta build is fairly complex and unless the user have all the required dependencies for all modules, then things would be silently disabled and resulting build would be very different from one to another. Like if the user happens to have "gtk-dev", then the "gtk" flow module will be there, otherwise it's disabled. And the list is long, see `data/json/dependencies.json`

In this approach, you should install to the system, as the build system is going away. We, the soletta developers, are the only folks that will ever see the build environment. I agree this is not the "usual workflow" for nodejs people, but given the above complexities it will never be the usual workflow in any way. Then for the specific case of soletta-node, the expected workflow for our users is to get this pre-installed via Ostro or ClearLinux. At this point if user did a package.json for npm that requires "soletta", it will just continue since the soletta is already there :-)

Then we agreed people would build Soletta and from it get all bindings, like Python, NodeJS. Hopefully we'll update our https://github.com/solettaproject/soletta-packaging to produce split packages with them, so if you're a Fedora user you can `dnf install soletta-nodejs` and it would pull in all the dependencies, placing them in the proper location such as `/usr/lib/node_modules` and so on.

Once things are more complete we can revisit the pure `npm install soletta` case, then we must also cover a way to force all dependencies to be available. Eventually that can happen with a different repository with this "host npm" that will fetch/include the actual soletta source with nodejs bindings, then it can have all the freedom to reshuffle things.

Makes sense?

sragavan:
Ill submit a new one.

PR #1768, opened 2016-03-29
generator and flow: Add indirection when getting node type

edersondisouza:
All nodes with internal nodes were broken on generator: a typical
construction of such nodes to 'bootstrap' them is, on their `init_type`
functions replace current type definition with a new one, so they
typically had code like `*current = type;` to replace type definitions.
However, this construction was useless when generating code (instead of
running via sol-fbp-runner): generated code end up using previous
definition instead of new one created on `init_type`. Fixed by adding
one more level of indirection on `sol_flow_get_node_type`.
While fixing current code, also noticed that current nodes that use
internal nodes were failing to initialise internal node types - fixed
it. Worth mentioning that this wasn't necessary when using
`sol-fbp-runner` as it took care of initialising types - even internals
- that it found via json definitions.

### DIFF
--- a/src/bin/sol-fbp-generator/main.c
+++ b/src/bin/sol-fbp-generator/main.c
@@ -1655,15 +1655,15 @@ generate(struct sol_vector *fbp_data_vector)
         "static bool\n"
         "initialize_types(void)\n"
         "{\n"
-        "    const struct sol_flow_node_type *t;\n"
+        "    const struct sol_flow_node_type **t;\n"
         "    int i = 0;\n\n");
     SOL_VECTOR_FOREACH_IDX (&ctx->types_to_initialize, type, i) {
         out(
             "    if (sol_flow_get_node_type(\"%.*s\", %.*s, &t) < 0)\n"
             "        return false;\n"
-            "    if (t->init_type)\n"
-            "        t->init_type();\n"
-            "    external_types[i++] = t;\n",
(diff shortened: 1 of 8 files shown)


glima:
Have you had the time to run a static build of this with your custom nodes (or any hand-made composite node)? It *seems* that we get a run-time crash at ``connect_nodes()`` -- the packet type there is not initialized yet. I'll investigate it more tomorrow, just a heads up.

edersondisouza:
On 29-03-2016 20:08, Gustavo Lima Chaves wrote:
> Have you had the time to run a static build of this with your custom
> nodes (or any hand-made composite node)? It /seems/ that we get a
> run-time crash at |connect_nodes()| -- the packet type there is not
> initialized yet. I'll investigate it more tomorrow, just a heads up.
>
> —
> <https://github.com/solettaproject/soletta/pull/1768#issuecomment-203150618>
>

What do you mean by 'any hand-made composite node'? Remember that nodes 
that use internal nodes must also `init_type` such nodes. So there are 
changes required on factory-monitor demo, as well. I'll PR that, I've 
forgotten it, sorry.

edersondisouza:
Anyone?

glima:
I did not hit this crash anymore, indeed (but others, not from Soletta). We can proceed with this, sorry for the delay.

edersondisouza:
Integrated, tks!

PR #1769, opened 2016-03-30
oic: Oic security module fails in insecure devices

otaviobp:
Oic security module should be disabled when running in a Soletta without
dtls support.


bdilly:
it seems ok

otaviobp:
merged

PR #1772, opened 2016-03-30
Add the correct dependencies for samples

cabelitos:


edersondisouza:
Looks OK. But I didn't double check dependencies =D

cabelitos:
I updated the PR. Fixing a type at 2fb4221 and add more missing dependencies.

cabelitos:
@edersondisouza I created a script to do that :]

bdilly:
@cabelitos where is this script?

I believe you could push it to our repo so other may use it in the future

edersondisouza:
On 30-03-2016 14:46, Bruno Dilly wrote:
> @cabelitos <https://github.com/cabelitos> where is this script?
>
> I believe you could push it to our repo so other may use it in the future

A script that checks it? Could even be part of tests that we run on CI?

cabelitos:
@bdilly @edersondisouza I will add it to the CI. soon I will create a PR with it.

cabelitos:
#1788

PR #1774, opened 2016-03-30
Coverity fixes

cabelitos:


src/modules/flow/persistence/persistence.c
```diff
@@ -676,10 +676,10 @@ persist_string_open(struct sol_flow_node *node,
...
         SOL_FLOW_NODE_TYPE_PERSISTENCE_STRING_OPTIONS_API_VERSION, -EINVAL);
 
-    if (opts->default_value) {
-        mdata->default_value = strdup((char *)opts->default_value);
-        SOL_NULL_CHECK(mdata->default_value, -ENOMEM);
-    }
+    SOL_NULL_CHECK_MSG(opts->default_value, -EINVAL,
+        "The default value can't be NULL");
+    mdata->default_value = strdup(opts->default_value);
+    SOL_NULL_CHECK(mdata->default_value, -ENOMEM);
```
edersondisouza: OK. But as the only way to get `NULL` here is by it being set to `persistence.json` - which is ours - maybe could you simplify and just get rid of all null checks? Or is this too outrageous?

bdilly:
@edersondisouza is right

It's the way I would prefer to go

src/modules/flow/persistence/persistence.c
cabelitos: Indeed.

cabelitos:
@edersondisouza @bdilly Sent a v2. I got rid of the null checks

edersondisouza:
r+

cabelitos:
merged

PR #1775, opened 2016-03-30
Nodejs bindings follow config settings

### LINKED ISSUE #1567: -O present in CFLAGS should have precedence over _FORTIFY_SOURCE
#### Task Description
We need to reconcile ```CFLAGS``` and ```_FORTIFY_SOURCE```. Currently, it's impossible to build with ```-O0``` because, if ```_FORTIFY_SOURCE``` is detected it automatically adds ```-O2``` which ends up being the setting honoured by the compiler. We need to modify things such that, if CFLAGS is incompatible with _FORTIFY_SOURCE, then _FORTIFY_SOURCE will not be applied.

gabrielschulhof:
If you configure soletta to not have OIC functionality then the resulting node.js bindings can't have such functionality either.

Addtitionally, this PR fixes #1567 by surrounding the addition of the fortify-related CFLAGS with a check to make sure that -O0 is not already part of the CFLAGS.


tools/build/Makefile.vars
```diff
@@ -286,7 +286,9 @@ COMMON_CFLAGS += -g -fno-lto
 endif
 
 ifeq (y,$(BUILD_TYPE_RELEASE))
-COMMON_CFLAGS += $(RELEASE_COMMON_CFLAGS)
+ifeq (,$(shell echo $(COMMON_CFLAGS) | grep -e "-O0"))
+COMMON_CFLAGS += $(FORTIFY_CFLAGS)
+endif
```
bdilly: Does it make sense to have a release build with -O0 ?

gabrielschulhof: Well, I've tried building with
```bash
BUILD_TYPE_RELEASE=n make
```
and it seems to ignore that, whereas building with
```bash
COMMON_CFLAGS="-g -O0"
```
correctly prepends ```-g -O0``` to the COMMON_CFLAGS.

The bottom line is that COMMON_CFLAGS may be set from outside the build environment, so the build environment has to account for this. Now, most of the time the COMMON_CFLAGS arriving from the outside are orthogonal to what we're adding inside the build environment, but in this case the existing COMMON_CFLAGS need to take precedence.

gabrielschulhof: Additional info: The use case for this is the situation where the user starts the build process with
```npm install``` or ```npm install --debug```. In both cases the npm install process ends up calling ```make alldefconfig``` followed by ```make```, but prefixed with a bunch of configuration options. In case of ```npm install```, it results in

```bash
USE_NODEJS="y" RPATH="y" make
```

whereas in the ```npm install --debug``` case it also has to at least turn on ```-g -O0``` in Soletta. It does this by also adding ```COMMON_CFLAGS="-g -O0"``` to the settings preceding the ```make```.

I also don't necessarily want to link Soletta's debug build feature to node.js' debug build, considering ```-g -O0``` as sufficient debugging support by Soletta for the node.js bindings.

Here's how it is currently: https://github.com/solettaproject/soletta/blob/5daa00eadbe275dbb696a749f22fd319e39d877f/bindings/nodejs/build-for-npm.sh#L28-L34

There are two problems with how it is currently:
0. It modifies the .config file. This is fixed in this PR by conditionally setting COMMON_CFLAGS for ```make``` instead.
0. ```-O0``` from external COMMON_CFLAGS is being overridden by RELEASE_COMMON_CFLAGS which later adds -O2 so as to support -DFORTIFY_SOURCE=2. This currently cannot be avoided by either modifying .config or by prefixing ```make``` with a variable AFAICT.

gabrielschulhof: OK, so I understand why I can't use ```BUILD_TYPE_RELEASE=n make```. It's because Makefile include .config, and in .config BUILD_TYPE_RELEASE is set explicitly to ```y```, not via ```?=```, so the environment variable is being ignored. Thus, my only option, if I wish not to modify the .config file is to use COMMON_CFLAGS which, in turn, leaves me with the solution in this PR.

bdilly:
What I'm suggesting is to use BUILD_TYPE_DEBUG=y instead of BUILD_TYPE_RELEASE=y (on this case BUILD_TYPE_RELEASE isn't even declared)

PR #1781, opened 2016-03-30
Use OIC Data Models [v2]

bdilly:
With this PR a few data models are ignored due to JSON errors.

Later I'll try to create some samples using new models so
we may check if auto generated code behaves as expected.

### DIFF
--- a/.gitmodules
+++ b/.gitmodules
@@ -18,3 +18,8 @@
 	url = https://github.com/mavlink/c_library.git
 	ignore = dirty
 	branch = master
+[submodule "src/thirdparty/oic-data-models"]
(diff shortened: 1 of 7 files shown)


otaviobp:
lgtm, but I have some points, that needs to be addressed later:
 - we could use _ as word separator for port names (json spec is using camelcase and we are just converting them to uppercase)
 - we don't have the interfaces implemented by each resource. This information is in raml files. Current code is using a hard coded interface
 - remove core.light and use oic.r.light.brigthness in soletta oic samples

bdilly:
ok, I'll create tickets for each suggestion and merge this PR.
Thanks

PR #1782, opened 2016-03-30
WIP - flow: Add support to lsm9ds0 sensor

mbelluzzo:
LSM9DS0 it's a chipset containing a gyroscope, magnetometer,
accelerometer and temperature sensor. It's divided in two I2C devices,
one for the gyroscope and another for the remaining sensors.

In an ideal world, I would be able to implement the support for this chipset in each node type related to its features: gyro node, accel node, temp node etc. That raised the first issue: multiple nodes handling the same I2C address isn't good in the way we structured our API and if this wasn't an issue, some capabilities share some registers avoiding that one node override the settings of another wouldn't be possible because each one only knows their context (since impossible is nothing, it would require a very complex scenario).

Alternative number 2 is to have a node for the gyro and another for the remaining sensors, that is how this PR is implemented. For the sake of simplicity, I kept both in the same module to test if it would work. The problem is that the dispute for the bus makes one of the nodes to fail the initialization after reaching max retries (even for large thresholds), this isn't a good way to handle resources. I still think we should have a queue per bus instead of check busy/pending.

I think a bus queue is necessary and changing the 'pending' use to check if your particular requested was processed. Also a handle for a "i2c slave device" would be handy, because would allow for different nodes to deal with the same I2C device. Dispatch of operations would be easier as well.

Another option is to ignore all API limitations and implement a single node for this chipset, but I really dislike this approach.

Chipset specific issues:
I wasn't able to turn off a specific axis, I checked my code and it was sending the correct configuration but for some reason if any axis is enabled, all registers correctly update it's value, so I removed the options to enable/disable independent axis for now.


barbieri:
this is why these things shouldn't be done in userspace but in IIO. This PR can get in as a stop gap, but please sync with kernel team to implement this at the proper layer.

src/modules/flow/lsm9ds0/lsm9ds0.c
```diff
@@ -0,0 +1,583 @@
...
+    uint32_t t;
+    struct sol_direction_vector xmt;
+    struct lsm9ds0_xmt_data *mdata = cb_data;
+
+    if (status <= 0) {
+        SOL_ERR("Couldn't read LSM9DS0 XMT device.");
+        return;
+    }
+
+    _parse_raw_data(&xmt, mdata->read_q, (mdata->accel_scale / 32768.0));
```
ceolin: `32768.0` will be better as a constant

src/modules/flow/lsm9ds0/lsm9ds0.c
```diff
@@ -0,0 +1,583 @@
...
+    mdata->curr_queue.cb_data = mdata;
+    mdata->curr_queue.queue = mdata->init_q;
+    mdata->curr_queue.queue_len = SOL_UTIL_ARRAY_SIZE(mdata->init_q);
+}
+
+static int
+lsm9ds0_xmt_open(struct sol_flow_node *node, void *data, const struct sol_flow_node_options *options)
+{
+    uint8_t accel_reg = 0x00;
+    uint8_t mag_reg = 0x20;
```
edersondisouza: Can't these values be configured by user? Shouldn't you allow that as well?

edersondisouza: Err, well, that's exactly what they do... disregard =D

edersondisouza:
looks ok
And that pattern to read/configure asynchronously is really common - luck this one didn't need specific timing between i2c transfers =D
@barbieri What about smallOSes that lack IIO?

mbelluzzo:
Good point about smallOSes and I'm not sure how IIO can solve the problem
if the device uses i2c to transfer data. The same issues may be a problem
in SPI.

barbieri:
small OS should get their native drivers, these kind of UPM or userspace drivers will suck badly.

for Zephyr they are working on IIO-like system for that exact reason.

PR #1787, opened 2016-03-31
Node.js bindings: Build bindings heeding configuration

gabrielschulhof:
This sacrifices our ability to build bindings against an existing soletta
installation identified by its pkgconfig file, because the pkgconfig file does
not contain information about the configuration options that were used when
building Soletta.

Replaces #1775.

Changes since v1:
* Liberally modify ```.config``` from ```bindings/nodejs/build-for-npm.sh```, turning on node.js bindings, and, when building for debug, then also ```-g -O0```, and turning off BUILD_TYPE_RELEASE and turning on BUILD_TYPE_DEBUG.


ceolin:
Soletta starts to build from the node bindings build ? That's the reason you change the .config manually to set DEBUG or RELEASE ? Also why are you setting these flags (`-g -O0`) ? It seems to be in the wrong place. Soletta's build should set this in debug mode.

gabrielschulhof:
@ceolin
> Soletta starts to build from the node bindings build ? That's the reason you change the .config manually to set DEBUG or RELEASE ?

Yes. There are two use cases:

0. You start with ```make <something>config``` and then you do ```make```
0. You start with ```npm install``` or ```npm install --debug```

In the latter case it is assumed that you downloaded the package from the npm repository so there's no need to worry about make install, so ```RPATH``` is set and ```.config``` is modified to ensure the node.js bindings are built. Additionally, if ```--debug``` is specified, the release build is turned off and the debug build is turned on.

> Also why are you setting these flags (```-g -O0```) ? It seems to be in the wrong place. Soletta's build should set this in debug mode.

BUILD_TYPE_DEBUG does not currently result in ```-O0``` being added to the COMMON_CFLAGS. I will modify [Makefile.vars](https://github.com/solettaproject/soletta/blob/5daa00eadbe275dbb696a749f22fd319e39d877f/tools/build/Makefile.vars#L285) to add ```-O0```.

PR #1793, opened 2016-03-31
Add http server/client json and blob nodes - v2

cabelitos:
Changes since v1:

* using sol_util_memdup()
* Removed some SOL_WRN()s in favor or error packets


cabelitos:
merged

PR #1794, opened 2016-03-31
Http sse v5

ceolin:
from v4:
- s/response_progressive/progressive_response/g
- Fix documentation
- sample's buffer is local
- the connections are handled properly on CTRL-D


barbieri:
+1, I'd just test the `SOL_STR_SLICE_LITERAL()` and if that works you can send it directly (you can force-update this PR to trigger CI to test).

Issue #1585, opened 2016-03-10
sol-fbp-runner: web inspector


barbieri:
the tool `sol-fbp-runner` provides a console inspector that will print all nodes, connections, packet send and delivery to `stdout`.

We should have `--webinspector[=0]` that changes the behavior to start an HTTP server on given port or 0 for random port (which should be printed to stdout). It would define `/events` with server-sent-events (SEE) about the current flow and export `/` as a redirect to `/static/index.html` that is served as static files to `${PREFIX}/share/soletta/webinspector/`. We can then provide some javascript and css to provide great experience, like opening an `new EventSource("/events")` in JavaScript and showing it in a table.

Currently `sol_http_server` doesn't allow us to keep connections alive after a response was sent. We need to provide that. Ideally we just get the socket ownership.

The  `--webinspector` should wait until there is a connection to `/events` before creating the flow, this way we can get all events dispatched to the inspector. We can then reject further connections on `/events` and exit the application once that connection is closed.

barbieri:
done in #1815

PR #1795, opened 2016-04-01
Fbp webinspector

barbieri:
first try, missing html/js/css.

Could you guys take a look and see with some of the designers if that could be useful for them? I'll do some static HTML/JS/CSS that will render a table with nodes and connected ports, with that I plan to put the last value and maybe plot a graph of values-over-time.

This is to address #1585


src/bin/sol-fbp-runner/main.c
```diff
@@ -120,7 +146,31 @@ parse_args(int argc, char *argv[])
...
+            break;
+#endif
+#ifdef WEB_INSPECTOR
+        case 'W':
+            if (optarg) {
+                unsigned long v;
+                char *endptr;
+
+                errno = 0;
+                v = strtoul(optarg, &endptr, 10);
```
ceolin: we have `sol_util_strtoul`  that is exported now.

src/bin/sol-fbp-runner/Makefile
```diff
@@ -3,4 +3,11 @@ bin-sol-fbp-runner-$(FBP_RUNNER) := main.c runner.c
 
 ifeq (y,$(INSPECTOR))
 bin-sol-fbp-runner-$(FBP_RUNNER) += inspector.c
+
+ifeq (y,$(HTTP_SERVER))
```
glima: I wonder if we don't want an *and ``SOL_FLOW_INSPECTOR_ENABLED``* here

src/bin/sol-fbp-runner/main.c
barbieri: that's not useful here, the regular version will do all right

src/bin/sol-fbp-runner/Makefile
barbieri: it's nested under `ifeq (y,$(INSPECTOR))` that is baaed on the same config value that results in `SOL_FLOW_INSPECTOR_ENABLED`

src/bin/sol-fbp-runner/web-inspector.c
```diff
@@ -0,0 +1,1396 @@
...
+        if (r < 0)
+            return r;
+
+        r = sol_json_serialize_string(buf, kv->value);
+        if (r < 0)
+            return r;
+
+        r = sol_buffer_append_char(buf, ']');
+        if (r < 0)
+            return r;
```
ceolin: Don't you think it's worth log this error ? Don't need to be here, can be in the top leve function, e. g. `web_inspector_will_deliver_packet`. As far as i could see the error is not spotted.

barbieri: will do the logging for these errors.

barbieri:
New version at #1802

PR #1796, opened 2016-04-01

sragavan:
changes.

### DIFF
--- a/bindings/nodejs/lib/gpio.js
+++ b/bindings/nodejs/lib/gpio.js
@@ -1,0 +1,116 @@
+/*
+ * This file is part of the Soletta Project
+ *
+ * Copyright (C) 2015 Intel Corporation. All rights reserved.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+ var soletta = require( 'bindings' )( 'soletta' ),
+    _ = require( 'lodash' );
+
+exports.open = function( init ) {
+    return new Promise( function( fulfill, reject ) {
+        var pin = init.pin;
+        var dir = 0;
+        var drive_mode = 0;
+        var config = null;
+        var gpiopin;
+        var callback_data = [];
+        var edge = "any";
+
+        if ( init.pullup )
+            drive_mode = soletta.sol_gpio_drive.SOL_GPIO_DRIVE_PULL_UP;
+        else if ( init.pullup === false )
+            drive_mode = soletta.sol_gpio_drive.SOL_GPIO_DRIVE_PULL_DOWN;
+        else
+            drive_mode = soletta.sol_gpio_drive.SOL_GPIO_DRIVE_NONE;
+
+        if ( init.edge )
+            edge = init.edge;
+
+        if ( init.direction == "in" ) {
+            config = {
+                dir: soletta.sol_gpio_direction.SOL_GPIO_DIR_IN,
+                active_low: init.activeLow,
+                poll_timeout: init.poll,
+                drive_mode: drive_mode,
+                trigger_mode: edge,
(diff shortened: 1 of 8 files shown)


sragavan:
@gabrielschulhof Can you merge this, it addresses the review comments from the previous one. Thanks.

bindings/nodejs/src/functions/gpio.cc
```diff
@@ -0,0 +1,144 @@
...
+    sol_gpio *gpio;
+
+    if (!gpio_data)
+        return;
+    gpio = gpio_data->gpio;
+
+    Nan::Callback *callback = gpio_data->callback;
+    sol_gpio_close(gpio);
+    if (callback) {
+        delete callback;
```
gabrielschulhof: You need to
```C++
Nan::SetInternalFieldPointer(jsGpio, 0, 0);
```
to make sure the pointer is removed from the JS object, so that if somebody accidentally calls ```bind_sol_gpio_close()``` twice with the same object, the second time around it throws an exception.

bindings/nodejs/src/functions/gpio.cc
```diff
@@ -0,0 +1,144 @@
...
+        Nan::ThrowError("Unable to extract sol_gpio_config\n");
+        return;
+    }
+
+    Nan::Callback *callback = gpio_data->callback;
+    if (callback) {
+        if (!hijack_ref()) {
+            delete callback;
+            delete gpio_data;
+            Nan::ThrowError("Failed to ref for mainloop");
```
gabrielschulhof: No need to throw an error here. ```hijack_ref()``` already throws an error for you. It's enough to simply return.

bindings/nodejs/src/functions/gpio.cc
```diff
@@ -0,0 +1,144 @@
...
+    }
+
+    gpio = sol_gpio_open(pin, &config);
+    if (gpio) {
+        gpio_data->gpio = gpio;
+        info.GetReturnValue().Set(SolGpio::New(gpio_data));
+        return;
+    } else {
+        if (callback)
+            delete callback;
```
gabrielschulhof: Don't you need to also ```hijack_unref()``` here?

src/lib/io/sol-gpio-common.c
```diff
@@ -74,3 +77,39 @@ sol_gpio_open(uint32_t pin, const struct sol_gpio_config *config)
 
     return gpio;
 }
+
+SOL_API enum sol_gpio_edge
+sol_gpio_edge_from_str(const char *edge)
+{
```
gabrielschulhof: I don't have a problem with doing the conversion from string to enum and vice versa here, but wouldn't it be easier to do this in the high-level JS API? After all, the soletta C API has no concept of strings that refer to edges. It only has the enum. Thus, the conversion from string to enum and vice versa should happen in the high-level JS API.

sragavan: It was there but @barbieri  wanted it this way.

sragavan:
ill make new PR

bindings/nodejs/lib/gpio.js
```diff
@@ -0,0 +1,116 @@
...
+        if ( init.edge )
+            edge = init.edge;
+
+        if ( init.direction == "in" ) {
+            config = {
+                dir: soletta.sol_gpio_direction.SOL_GPIO_DIR_IN,
+                active_low: init.activeLow,
+                poll_timeout: init.poll,
+                drive_mode: drive_mode,
+                trigger_mode: edge,
```
gabrielschulhof: Do the lookup here and set the enum value as the value of the "trigger_mode" property.

src/lib/io/sol-gpio-common.c
gabrielschulhof: If you're referring to https://github.com/solettaproject/soletta/pull/1767#discussion_r57982171, then that's not the same as what I'm saying. What I'm saying is that you don't need the functions ```sol_gpio_edge_from_str()``` and ```sol_gpio_edge_to_str()``` at all. You can write the lookup in both directions JS with a lot less code. See my comment [here](https://github.com/solettaproject/soletta/pull/1796/files#r58203275).

barbieri: It is based on that comment and I've asked him to provide these functions. I agree they are not strictly needed, but providing them is good for multiple reasons:
 - other bindings can use the same mapping, helping our users with consistency;
 - even in C, sometimes you want to debug or print something to the user, then having a enum<->string is helpful; Like you expose this enum via a OIC or HTTP/JSON.

These functions should be provided by soletta's C API, we missed them but I count on you guys to provide these as you notice they are missing. If you don't have time you can fill a ticket for others to tackle them.

PR #1798, opened 2016-04-01
Node.js bindings: Build bindings heeding configuration

gabrielschulhof:
This sacrifices our ability to build bindings against an existing soletta
installation identified by its pkgconfig file, because the pkgconfig file does
not contain information about the configuration options that were used when
building Soletta.

This replaces #1787.

Changes:

* Modify ```COMMON_CFLAGS``` in case of ```BUILD_TYPE_DEBUG``` adding ```-O0```


tools/build/Makefile.vars
```diff
@@ -282,7 +282,7 @@ COMMON_CFLAGS += -fPIC
 endif
 
 ifeq (y,$(BUILD_TYPE_DEBUG))
-COMMON_CFLAGS += -g -fno-lto
+COMMON_CFLAGS += -g -fno-lto -O0
 endif
```
ceolin: o yeah, here is much better

ibriano: On Fri, 01 Apr 2016 07:07:28 -0700, Flávio Vinicius Alvares Ceolin wrote:
> > @@ -282,7 +282,7 @@ COMMON_CFLAGS += -fPIC
> >  endif
> >  
> >  ifeq (y,$(BUILD_TYPE_DEBUG))
> > -COMMON_CFLAGS += -g -fno-lto
> > +COMMON_CFLAGS += -g -fno-lto -O0
> >  endif
> 
> o yeah, here is much better

I'd rather re-order the way we add stuff to COMMON_CFLAGS to make sure
that all the things that can break debugging are not there in the first
place.

PR #1801, opened 2016-04-01
dependency-resolver: check if --compiler is valid

dorileo:
If the --compiler argument is not valid the ccode and flags tests will
fail and generate invalid Kconfig.gen and Makefile.gen - due resulting
for example on wrong BASE_OS.


vcgomes:
Looks good. :+1:

bdilly:
+1

dorileo:
applied, thanks.

PR #1802, opened 2016-04-03
Fbp webinspector v2

barbieri:
Changes since v1 (#1795):
 - changed mime type detection so ".css" is recognized as "text/css", libmagic won't get that as it will infer the content as "text/plain", rendering the stylesheet useless to browsers. @ceolin please check.
 - SSE messages reshuffled based on web-inspector.js needs
 - HTML, JS and CSS to showcase the structure of data, it works but it ugly as hell.
 - Makefile rules now can install data per-directory, no need to change Makefile.rules/Makefile.vars/Makefile.targets manually like done for JSONs and the likes. @dorileo please check.

I'm not planning to work on the HTML/JS/CSS anytime soon, so let's commit like that and let one of the designers implement their nice views. @gpaes can you check this part?


tools/build/Makefile.rules
```diff
@@ -132,6 +132,23 @@ parse-bin = \
 	$(eval $(1)-deps     := $(subst .mod,,$(bin-$(1)-y-deps))) \
 	$(eval bins-out      += $(bin-$(1)-out)) \
 
+define install-resource
+$(2): $(1)
+	$(Q)echo "    "INST"   "$(1)
+	$(Q)$(MKDIR) -p $(dir $(2))
+	$(Q)$(CP) $(1) $(2)
+endef
```
ceolin: this is cool, first time i see.

barbieri: it was used manually at Makefile.target, each resource was installed with a rule there, see GDB_ variables in Makefile.vars and Makefile.targets

ceolin:
pr looks awesome. A minor issue:
```
./build/soletta_sysroot/usr/bin/sol-fbp-runner -W 8080 ./src/samples/flow/http-composed/http-composed.fbp

ERR: ./src/bin/sol-fbp-runner/runner.c:91 search_fbp_file() Couldn't find file 8080
```

barbieri:
ok, fixed in the help so `-W[PORT]` is together.

barbieri:
see v3 #1815

PR #1803, opened 2016-04-04
Node.js bindings: Add JavaScript bindings for UART

nagineni:
This patch provides native and highlevel JavaScript bindings
for UART. Also, it adds enum-string translation functions to
soletta.

Signed-off-by: Sudarsana Nagineni <sudarsana.nagineni@intel.com>
Signed-off-by: Srinivasa Ragavan <srinivasa.ragavan.venkateswaran@intel.com>

PR #1804, opened 2016-04-04
Node.js bindings: Add JavaScript bindings for UART

nagineni:
This patch provides native and highlevel JavaScript bindings
for UART. Also, it adds enum-string translation functions to
soletta.

This replaces #1803

Signed-off-by: Srinivasa Ragavan <srinivasa.ragavan.venkateswaran@intel.com>

### DIFF
--- a/bindings/nodejs/src/functions/uart.cc
+++ b/bindings/nodejs/src/functions/uart.cc
@@ -1,0 +1,176 @@
+/*
+ * This file is part of the Soletta Project
+ *
+ * Copyright (C) 2015 Intel Corporation. All rights reserved.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+#include <sol-uart.h>
+#include <node.h>
+#include <nan.h>
+
+#include "../common.h"
+#include "../hijack.h"
+#include "../structures/sol-js-uart.h"
+#include "../structures/js-handle.h"
+
+using namespace v8;
+
+class SolUART : public JSHandle<SolUART> {
+public:
+    static const char *jsClassName() { return "SolUART"; }
+};
+
+static void sol_uart_read_callback(void *user_data, struct sol_uart *uart,
+    unsigned char byte_read) {
+    Nan::HandleScope scope;
+    sol_uart_data *uart_data = (sol_uart_data *)user_data;
+    Nan::Callback *callback = uart_data->rx_cb;
+    if (!callback)
+        return;
+
+    Local<Value> arguments[2] = {
+        SolUART::New(uart_data),
+        Nan::New(byte_read)
+    };
+
+    callback->Call(2, arguments);
+}
+
+NAN_METHOD(bind_sol_uart_open) {
+    VALIDATE_ARGUMENT_COUNT(info, 2);
+    VALIDATE_ARGUMENT_TYPE_OR_NULL(info, 0, IsString);
+    VALIDATE_ARGUMENT_TYPE(info, 1, IsObject);
+    sol_uart_config config;
+    sol_uart *uart = NULL;
+
+    if (!hijack_ref())
+        return;
+
+    sol_uart_data *uart_data = new sol_uart_data;
+    uart_data->rx_cb = NULL;
+    uart_data->tx_cb = NULL;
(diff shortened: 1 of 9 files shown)


bindings/nodejs/src/functions/uart.cc
```diff
@@ -0,0 +1,176 @@
...
+static void sol_uart_read_callback(void *user_data, struct sol_uart *uart,
+    unsigned char byte_read) {
+    Nan::HandleScope scope;
+    sol_uart_data *uart_data = (sol_uart_data *)user_data;
+    Nan::Callback *callback = uart_data->rx_cb;
+    if (!callback)
+        return;
+
+    Local<Value> arguments[2] = {
+        SolUART::New(uart_data),
```
gabrielschulhof: I'm not sure it's such a good idea to create a new ```SolUART``` here, because you're storing the pointer ```uart_data``` in yet another object. Now you have the same pointer (given to you by soletta in ```sol_uart_open()```) in two JS objects. You invalidate one of them in ```bind_sol_uart_close()``` by doing ```Nan::SetInternalFieldPointer()```, but you never invalidate this object the same way. It might be better to store a persistent reference to the SolUART object you create in ```bind_sol_uart_open()``` in the structure ```uart_data``` and pass it to the JS callback here. That way, there's only one JS object corresponding to the ```sol_uart *``` in play.

gabrielschulhof: https://github.com/solettaproject/soletta/pull/1805/files#r58405204

bindings/nodejs/src/functions/uart.cc
```diff
@@ -0,0 +1,176 @@
...
+    if (status >= 0) {
+        Local <Object> bufObj;
+        bufObj = Nan::NewBuffer((char *)tx, status).ToLocalChecked();
+        buffer = bufObj;
+    } else {
+        buffer = Nan::Null();
+    }
+
+    Local<Value> arguments[3] = {
+        SolUART::New(uart_data),
```
gabrielschulhof: Same as https://github.com/solettaproject/soletta/pull/1804/files#r58404435, please use a persistent reference to the ```SolUART``` object you create in ```bind_sol_uart_open()``` and pass it to the callback here instead of creating a new ```SolUART``` object.

src/lib/io/sol-uart-common.c
```diff
@@ -0,0 +1,145 @@
...
+        SOL_STR_TABLE_ITEM("databits-8", SOL_UART_DATA_BITS_8),
+        SOL_STR_TABLE_ITEM("databits-7", SOL_UART_DATA_BITS_7),
+        SOL_STR_TABLE_ITEM("databits-6", SOL_UART_DATA_BITS_6),
+        SOL_STR_TABLE_ITEM("databits-5", SOL_UART_DATA_BITS_5),
+        { }
+    };
+
+    return sol_str_table_lookup_fallback(table,
+        sol_str_slice_from_str(data_bits), SOL_UART_DATA_BITS_8);
+}
```
gabrielschulhof: @barbieri I understand that it makes sense for this to return ```SOL_UART_DATA_BITS_8``` as a fallback, but that can cause unexpected behaviour the cause for which is hard to trace. Like, in this case, what if somebody looks up ```"dataBits:8"``` or even ```"dataBits-8"```, or even accidentally ```""```? In Javascript, you could throw an exception, making it easy to debug the application.

If we keep this, the implementation of ```sol_str_table_lookup_fallback()``` should at least issue a warning when it falls back.

nagineni:
Replaced by #1818

src/lib/io/sol-uart-common.c
barbieri: that is a good point, if you need to check, then use the `sol_str_table_lookup()`, which returns true/false. Then we change the signature to pass `const char **` and return a boolean. In this case I'd say we need to grep all cases and convert them all to the same signature.

@bdilly @ibriano @glima ?

ibriano: On Tue, 05 Apr 2016 11:16:00 -0700, Gustavo Sverzut Barbieri wrote:
> > +
> > +
> > +SOL_API enum sol_uart_data_bits
> > +sol_uart_data_bits_from_str(const char *data_bits)
> > +{
> > +    static const struct sol_str_table table[] = {
> > +        SOL_STR_TABLE_ITEM("databits-8", SOL_UART_DATA_BITS_8),
> > +        SOL_STR_TABLE_ITEM("databits-7", SOL_UART_DATA_BITS_7),
> > +        SOL_STR_TABLE_ITEM("databits-6", SOL_UART_DATA_BITS_6),
> > +        SOL_STR_TABLE_ITEM("databits-5", SOL_UART_DATA_BITS_5),
> > +        { }
> > +    };
> > +
> > +    return sol_str_table_lookup_fallback(table,
> > +        sol_str_slice_from_str(data_bits), SOL_UART_DATA_BITS_8);
> > +}
> 
> that is a good point, if you need to check, then use the `sol_str_table_lookup()`, which returns true/false. Then we change the signature to pass `const char **` and return a boolean. In this case I'd say we need to grep all cases and convert them all to the same signature.
> 
> @bdilly @ibriano @glima ?

I agree, we are being quite pedantic about error checking everywhere we
notice it's missing, and so we should provide the same facilities to
bindings.

PR #1805, opened 2016-04-04

sragavan:
changes.

### DIFF
--- a/bindings/nodejs/src/functions/gpio.cc
+++ b/bindings/nodejs/src/functions/gpio.cc
@@ -1,0 +1,145 @@
+/*
+ * This file is part of the Soletta Project
+ *
+ * Copyright (C) 2015 Intel Corporation. All rights reserved.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+#include <sol-gpio.h>
+
+#include <v8.h>
+#include <node.h>
+#include <nan.h>
+#include <map>
+
+#include "../common.h"
+#include "../hijack.h"
+#include "../structures/sol-js-gpio.h"
+#include "../structures/js-handle.h"
+
+using namespace v8;
+
+class SolGpio : public JSHandle<SolGpio> {
+public:
+    static const char *jsClassName() { return "SolGpio"; }
+};
+
+static void sol_gpio_read_callback(void *data, struct sol_gpio *gpio, bool value) {
+    Nan::HandleScope scope;
+    sol_gpio_data *gpio_data = (sol_gpio_data *)data;
+    Nan::Callback *callback = gpio_data->callback;
+    if (!callback)
+        return;
+
+    Local<Value> arguments[2] = {
+        SolGpio::New(gpio_data),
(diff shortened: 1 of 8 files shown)


sragavan:
@gabrielschulhof  fixed it.

gabrielschulhof:
:+1:

bindings/nodejs/src/functions/gpio.cc
```diff
@@ -0,0 +1,145 @@
...
+
+static void sol_gpio_read_callback(void *data, struct sol_gpio *gpio, bool value) {
+    Nan::HandleScope scope;
+    sol_gpio_data *gpio_data = (sol_gpio_data *)data;
+    Nan::Callback *callback = gpio_data->callback;
+    if (!callback)
+        return;
+
+    Local<Value> arguments[2] = {
+        SolGpio::New(gpio_data),
```
gabrielschulhof: Similarly to https://github.com/solettaproject/soletta/pull/1804/files#r58404435, I think it's better to store a persistent reference to the ```SolGpio``` object you create in ```bind_sol_gpio_open()``` in the structure ```gpio_data``` and pass that object to the callback here, instead of creating a second JS object holding the same pointer as the object you created in ```bind_sol_gpio_open()```.

gabrielschulhof: Conceptually, the parameter ```sol_gpio *gpio``` is given as a convenience to this callback. It's not a ```sol_gpio *``` newly created soletta. It's the same pointer that you have received from soletta when you earlier called ```sol_gpio_open()```. Therefore, you should not create a second JS object to hold it, because you've already created a JS object in ```bind_sol_gpio_open()```.

src/lib/io/sol-gpio-common.c
```diff
@@ -74,3 +77,34 @@ sol_gpio_open(uint32_t pin, const struct sol_gpio_config *config)
 
     return gpio;
 }
+
+SOL_API enum sol_gpio_edge
+sol_gpio_edge_from_str(const char *edge)
+{
+    static const struct sol_str_table table[] = {
```
cabelitos: You should check if edge is not NULL. Otherwise `sol_str_slice_from_str(edge)` will crash!

cabelitos: Do the same for the UART PR.

gabrielschulhof: @cabelitos will it crash even though```sol_str_slice_from_str()``` is declared ```SOL_ATTR_NONNULL(1)```?

bindings/nodejs/src/functions/gpio.cc
nagineni: Good catch. We don't really need to pass the gpio_data as an arg to the JS in the callbacks, so we will fix it by removing that.

src/lib/io/sol-gpio-common.c
nagineni: It does crash. NULL check is needed definitely.

sragavan:
Pushing another one.

src/lib/io/sol-gpio-common.c
ibriano: On Tue, 05 Apr 2016 00:08:44 -0700, gabrielschulhof wrote:
> > @@ -74,3 +77,34 @@ sol_gpio_open(uint32_t pin, const struct sol_gpio_config *config)
> >  
> >      return gpio;
> >  }
> > +
> > +SOL_API enum sol_gpio_edge
> > +sol_gpio_edge_from_str(const char *edge)
> > +{
> > +    static const struct sol_str_table table[] = {
> 
> @cabelitos will it crash even though```sol_str_slice_from_str()``` is declared ```SOL_ATTR_NONNULL(1)```?
> 

The nonnull attribute tells the compiler that the argument will never be
NULL, so if you pass an explicit NULL there, it will complain at build
time, and for other cases, it's free to optimize away any checks inside
the function.
That passes the responsibility to the user of the function to ensure
that the parameter passed will never be NULL.

PR #1806, opened 2016-04-04
Enable building the JS meta-type (and flows using it) on small OSes

ibriano:
Building a flow with JS will still require that the duktape.h and duk_config.h
files are copied to the application's source.

### DIFF
--- a/src/modules/flow-metatype/js/Kconfig
+++ b/src/modules/flow-metatype/js/Kconfig
@@ -1,6 +1,6 @@
 config FLOW_METATYPE_JAVASCRIPT
 	tristate "JavaScript flow metatype"
-	depends on FLOW_SUPPORT && PLATFORM_LINUX && HAVE_DUKTAPE_SRC
+	depends on FLOW_SUPPORT && HAVE_DUKTAPE_SRC
 	default m if MODULES=y
 	help
 	    Allow creating node types using JavaScript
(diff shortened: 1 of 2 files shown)


barbieri:
+1

I wonder if we should install the duktape.h/c/duk_config.h in some `${PREFIX}/share/soletta/flow/metatype/js/include` and in generated .c we include that one specifically. Maybe that is not that helpful, particularly for small OS where you do not install soletta at all.

@zolkis @gabrielschulhof this simple change enables us to build and run JS with Zephyr on small OS, problem is that you need a board big enough to hold the ROM/RAM, as well as increasing the stack to 16kb. We're using a Freescale board to test this.

ibriano:
On Mon, 04 Apr 2016 10:55:46 -0700, Gustavo Sverzut Barbieri wrote:
> +1
> 
> I wonder if we should install the duktape.h/c/duk_config.h in some `${PREFIX}/share/soletta/flow/metatype/js/include` and in generated .c we include that one specifically. Maybe that is not that helpful, particularly for small OS where you do not install soletta at all.
> 

I didn't need the C anymore after changing how Soletta is built for
Zephyr. But the .h would be nice to have available somewhow.

> @zolkis @gabrielschulhof this simple change enables us to build and run JS with Zephyr on small OS, problem is that you need a board big enough to hold the ROM/RAM, as well as increasing the stack to 16kb. We're using a Freescale board to test this.

16k for the mycounter.{fbp,js} under src/samples/flow/js
I tried the telegram sample with a bigger stack and I still couldn't get
it to work.

glima:
Looks ok

cabelitos:
+1

PR #1808, opened 2016-04-04
generator: use <> instead of "" for Soletta headers inclusion

glima:

### DIFF
--- a/src/bin/sol-fbp-generator/main.c
+++ b/src/bin/sol-fbp-generator/main.c
@@ -1590,11 +1590,11 @@ generate(struct sol_vector *fbp_data_vector)
         "#include <math.h>\n"
         "#include <stdint.h>\n"
         "#include <float.h>\n"
-        "#include \"sol-flow.h\"\n"
-        "#include \"sol-flow-static.h\"\n");
+        "#include <sol-flow.h>\n"
+        "#include <sol-flow-static.h>\n");
 
     if (!args.export_symbol) {
-        out("#include \"soletta.h\"\n");
+        out("#include <soletta.h>\n");
     }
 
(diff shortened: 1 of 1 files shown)


src/bin/sol-fbp-generator/main.c
```diff
@@ -1611,12 +1611,12 @@ generate(struct sol_vector *fbp_data_vector)
 
     /* Header name is currently inferred from the module name. */
     SOL_VECTOR_FOREACH_IDX (&ctx->headers, header, i) {
-        out("#include \"%.*s\"\n", SOL_STR_SLICE_PRINT(*header));
+        out("#include <%.*s>\n", SOL_STR_SLICE_PRINT(*header));
```
ibriano: This one may be problematic, as some of the node types used could have been generated locally and not be installed.

cabelitos:
Why are you changing that?

glima:
Because the "" syntax tells the compiler to start searching for that at PWD, while the other tell its to start the search on system include dirs and -I entries, directly  (and that's the most common case when using the generator, anyway).

glima:
https://gcc.gnu.org/onlinedocs/cpp/Include-Syntax.html

cabelitos:
@glima I know the difference between them. I just asked: "why the change that now?"

glima:
Just being anal and maybe better at header lookups for the most common use case.

cabelitos:
:new_moon_with_face:

glima:
It seems I did not test it, meh. Dropping.

bdilly:
It seems @ibriano is right about his assumptions.
So I suggest to drop this patch

PR #1811, opened 2016-04-04
Add SSE support for Soletta's HTTP nodes

cabelitos:


src/modules/flow/http-client/http-client.c
```diff
@@ -377,20 +355,11 @@ common_request_finished(void *data,
...
-
-        sol_json_scanner_init(&scanner, response->content.data, response->content.used);
-        SOL_JSON_SCANNER_OBJECT_LOOP (&scanner, &token, &key, &value, reason) {
-            if (!sol_json_token_str_eq(&token, mdata->key.data, mdata->key.len))
-                continue;
-            ret = type->process_token(node, &key, &value);
-            if (ret < 0)
-                goto err;
-        }
+        ret = type->process_token(node, &response->content);
```
barbieri: seems you only need the slice (ptr/size), so you can change the signature to take that, then you use `sol_buffer_get_slice(&response->content)`

src/modules/flow/http-client/http-client.c
```diff
@@ -402,17 +371,102 @@ common_request_finished(void *data,
...
+
+    type = (const struct http_client_node_type *)sol_flow_node_get_type(node);
+
+    while (slice.len) {
+        char *start, *end, *content;
+        struct sol_buffer content_buf;
+        size_t content_len, total_len;
+        int r;
+
+        start = memmem(slice.data, slice.len, "data: ", prefix_len);
```
barbieri: it's better to define prefix as a slice before the loop, then `sol_str_slice_starts_with(slice, prefix)` and `sol_str_slice_contains(slice, end)`.

barbieri: hum... `sol_str_slice_contains()` returns a boolean, you'll need another function that returns the ptr inside given slice, a nice addition to our api.

src/modules/flow/http-client/http-client.c
```diff
@@ -422,13 +476,20 @@ common_get_process(struct sol_flow_node *node, void *data, uint16_t port,
...
-    if (mdata->accept && !sol_http_param_add(&params,
-        SOL_HTTP_REQUEST_PARAM_HEADER("Accept", mdata->accept))) {
-        SOL_WRN("Failed to set query params");
-        r = -ENOMEM;
-        goto err;
+
+    if (mdata->accept) {
+        if (!sol_http_param_add(&params,
+            SOL_HTTP_REQUEST_PARAM_HEADER("Accept", mdata->accept))) {
+            SOL_WRN("Failed to set query params");
```
barbieri: which header, which value... improve error messages

src/modules/flow/http-client/http-client.c
```diff
@@ -422,13 +476,20 @@ common_get_process(struct sol_flow_node *node, void *data, uint16_t port,
...
+            r = -ENOMEM;
+            goto err;
+        }
+
+        if (strstr(mdata->accept, "text/stream"))
```
barbieri: hum... this is tricky, we should check based on order. The accept "proper handling" would also account for the "q=" parameter, basically you split it to an array of struct { struct sol_str_slice mime; double q; } and then order by `q`.

@ceolin did you ever implement that? maybe it's a function to add to `sol-http.h`, it would take a slice and returns a `struct sol_vector` (not pointer!) with the sorted vector. Our function would check if text/stream is the highest (first) priority.

src/modules/flow/http-client/http-client.c
```diff
@@ -550,33 +619,43 @@ common_post_process(struct sol_flow_node *node, void *data,
...
 {
     bool result;
 
-    SOL_HTTP_RESPONSE_CHECK_API(response, -EINVAL);
-
-    if (!strncasecmp("true", response->content.data, response->content.used))
+    if (!strncasecmp("true", buf->data, buf->used))
         result = true;
-    else if (!strncasecmp("false", response->content.data, response->content.used))
+    else if (!strncasecmp("false", buf->data, buf->used))
```
barbieri: `sol_str_slice_caseeq()`, if you move from buffer to slice it will be even cleaner.

src/modules/flow/http-client/http-client.c
```diff
@@ -73,10 +73,8 @@ struct create_url_data {
 
 struct http_client_node_type {
     struct sol_flow_node_type base;
-    int (*process_token)(struct sol_flow_node *node, struct sol_json_token *key,
-        struct sol_json_token *value);
-    int (*process_data)(struct sol_flow_node *node,
-        struct sol_http_response *response);
+    int (*process_token)(struct sol_flow_node *node, struct sol_buffer *buf);
```
barbieri: this method name must change from token to json.

src/modules/flow/http-client/http-client.c
```diff
@@ -377,20 +355,11 @@ common_request_finished(void *data,
...
-            ret = type->process_token(node, &key, &value);
-            if (ret < 0)
-                goto err;
-        }
+        ret = type->process_token(node, &response->content);
+        if (ret < 0)
+            goto err;
     } else {
-        ret = type->process_data(node, response);
+        ret = type->process_data(node, &response->content);
```
barbieri: same here, using slice will be better.

src/modules/flow/http-client/http-client.c
```diff
@@ -550,33 +619,43 @@ common_post_process(struct sol_flow_node *node, void *data,
...
-        result = true;
-    else if (sol_json_token_get_type(value) == SOL_JSON_TYPE_FALSE)
-        result = false;
-    else
-        return -EINVAL;
+    sol_json_scanner_init(&sub_scanner, buf->data, buf->used);
+    found = false;
+
+    SOL_JSON_SCANNER_OBJECT_LOOP (&sub_scanner, &token, &sub_key, &sub_value, reason) {
+        if (SOL_JSON_TOKEN_STR_LITERAL_EQ(&sub_key, "value")) {
```
barbieri: do we need `{"value": true}`? Why not just `true`? I'm not sure if that would be valid json for other clients doing json http requests, like browsers.

src/modules/flow/http-client/http-client.c
```diff
@@ -602,30 +681,44 @@ boolean_post_process(struct sol_flow_node *node, void *data, uint16_t port, uint
...
-string_process_data(struct sol_flow_node *node,
-    struct sol_http_response *response)
+string_process_data(struct sol_flow_node *node, struct sol_buffer *buf)
 {
     char *result;
 
-    SOL_HTTP_RESPONSE_CHECK_API(response, -EINVAL);
-
-    result = strndup(response->content.data, response->content.used);
+    result = strndup(buf->data, buf->used);
```
barbieri: `sol_str_slice_to_string()`

src/modules/flow/http-client/http-client.json
```diff
@@ -47,7 +47,7 @@
           },
           {
             "data_type": "string",
-            "default": "application/json",
+            "default": "text/stream",
```
barbieri: I'd make this a list of both, with text/stream in the highest priority. That way if the server doesn't accept text/stream, like some HTTP, then we get the json or text.

src/modules/flow/http-server/http-server.c
```diff
@@ -278,6 +268,70 @@ common_handle_response_cb(struct sol_flow_node *node,
...
+
+    if (method != SOL_HTTP_METHOD_GET)
+        return false;
+
+    SOL_HTTP_PARAMS_FOREACH_IDX (params, param, i) {
+        if (param->type != SOL_HTTP_PARAM_HEADER)
+            continue;
+
+        if (sol_str_slice_str_eq(param->value.key_value.key, "Accept") &&
+            sol_str_slice_str_contains(param->value.key_value.value, "text/stream"))
```
barbieri: as before we need to check the highest priority accept... more complex than "contains" :-(

src/modules/flow/http-server/http-server.c
```diff
@@ -311,22 +365,53 @@ common_response_cb(void *data, struct sol_http_request *request)
...
-            type->send_packet_cb(mdata, node);
-    }
+        r = sol_ptr_vector_append(&mdata->sse_clients, sse);
+        if (r < 0) {
+            sol_http_progressive_response_del(sse, false);
+            sol_flow_send_error_packet_str(node, -r,
+                "Could append the SSE response to the client's array");
+            goto exit;
+        }
+        r = send_sse_data(type, mdata);
```
barbieri: should we send just to the newly connected sse?

src/modules/flow/http-server/http-server.c
```diff
@@ -393,7 +479,13 @@ static void
 common_close(struct sol_flow_node *node, void *data)
 {
     struct http_data *mdata = data;
+    uint16_t i;
+    struct sol_http_progressive_response *sse_client;
+
+    SOL_PTR_VECTOR_FOREACH_IDX (&mdata->sse_clients, sse_client, i)
```
barbieri: not expected in the current implementation, but it is better to steal the vector and zero the existing before you dispatch these, so if the "on_close" callback fires it won't affect the vector you're iterating. Then you iterate and clear the copy you stored locally in this function

src/modules/flow/http-server/http-server.c
```diff
@@ -563,12 +660,14 @@ boolean_post_cb(struct http_data *mdata, struct sol_flow_node *node,
...
 
-    r = sol_buffer_append_printf(content, "%s", mdata->value.b == true ? "true" : "false");
-    SOL_INT_CHECK(r, < 0, r);
+    str = mdata->value.b ? "true" : "false";
 
-    return 0;
+    if (json)
+        return sol_buffer_append_printf(content, "{\"value\":%s}", str);
+    else
+        return sol_buffer_append_printf(content, "%s", str);
```
barbieri: ugh, append slice is more efficient :-)

src/modules/flow/http-server/http-server.c
```diff
@@ -600,10 +699,15 @@ boolean_process_cb(struct http_data *mdata, const struct sol_flow_packet *packet
...
-    int r = 0;
+    int r;
 
     if (json) {
+        r = sol_buffer_append_slice(content,
+            sol_str_slice_from_str("{\"value\":"));
+        SOL_INT_CHECK(r, < 0, r);
         r = sol_json_serialize_string(content, mdata->value.s);
+        SOL_INT_CHECK(r, < 0, r);
+        r = sol_buffer_append_slice(content, sol_str_slice_from_str("}"));
```
barbieri: append_char()

src/modules/flow/http-server/http-server.c
```diff
@@ -880,7 +984,17 @@ rgb_post_cb(struct http_data *mdata, struct sol_flow_node *node,
...
 rgb_process_cb(struct http_data *mdata, const struct sol_flow_packet *packet)
 {
-    return sol_flow_packet_get_rgb(packet, &mdata->value.rgb);
+    struct sol_rgb rgb;
+    int r;
+
+    r = sol_flow_packet_get_rgb(packet, &rgb);
+    SOL_INT_CHECK(r, < 0, r);
+
+    if (!memcmp(&mdata->value.rgb, &rgb, sizeof(struct sol_rgb)))
```
barbieri: we should provide a sol_rgb_is_equal, like we do for irange/drange.

src/modules/flow/http-server/http-server.c
```diff
@@ -1008,8 +1122,18 @@ static int
...
 {
-    return sol_flow_packet_get_direction_vector(packet,
-        &mdata->value.dir_vector);
+    struct sol_direction_vector dir;
+    int r;
+
+    r = sol_flow_packet_get_direction_vector(packet, &dir);
+    SOL_INT_CHECK(r, < 0, r);
+
+    if (!memcmp(&mdata->value.dir_vector, &dir,
```
barbieri: same as above, we should provide a sol_direction_vector_is_equal

src/modules/flow/http-server/http-server.c
```diff
@@ -1163,6 +1303,19 @@ blob_handle_response_cb(struct sol_flow_node *node,
...
+/*
+   Only implemented to be used by send_sse_data()
+ */
+static int
+json_response_cb(struct http_data *mdata, struct sol_buffer *buf, bool json)
+{
+    if (!mdata->value.blob)
+        return 0;
+
+    return sol_buffer_append_bytes(buf,
```
barbieri: i guess we have sol_str_slice_from_blob()

barbieri: ah, but in this case this won't work! We need to decide what yo do, maybe you need to convert to base64? there is a buffer append that does that. This will break if the lines contains "\n\n" or other chars in there

src/modules/flow/http-client/http-client.c
```diff
@@ -402,17 +371,102 @@ common_request_finished(void *data,
...
+            r = type->process_data(node, &content_buf);
+        sol_buffer_fini(&content_buf);
+        SOL_INT_CHECK(r, < 0, r);
+        slice.len -= total_len;
+        slice.data += total_len;
+    }
+
+exit:
+    SOL_DBG("Buf len: %zu - Consumed: %zu", buf->used, consumed);
+    return consumed;
```
barbieri: if we don't consume the data (ie; partial "data: ..." without \n\n), then the infrastructure will keep that remaining bytes for the next call?

src/modules/flow/http-server/http-server.c
cabelitos: ok

src/modules/flow/http-server/http-server.c
cabelitos: Indeed. My bad.

src/modules/flow/http-server/http-server.c
cabelitos: This function is used by the http-server/json and it's the json content itself. There's no need to convert to base64. However, this may be useful for the http-server/blob node.

barbieri: ok, I've overlooked it, it's the json-response... I got the wrong impression it was the blob returning json, my bad (which is the parameter, not the function name).

src/modules/flow/http-client/http-client.c
cabelitos: Ok.

src/modules/flow/http-client/http-client.c
cabelitos: Well, this is the case when accept is application/json

src/modules/flow/http-server/http-server.c
cabelitos: Indeed. Sorry

src/modules/flow/http-client/http-client.c
barbieri: i know, what I mean is: do we need an object, or can we have just the true/false, like buf == `true`. Then we'd not need to loop and check for key.

src/modules/flow/http-client/http-client.c
cabelitos: Yes it will!

src/modules/flow/http-client/http-client.c
cabelitos: ok

src/modules/flow/http-client/http-client.c
cabelitos: We do not have such function. I will implement it.

src/modules/flow/http-client/http-client.c
cabelitos: I don't think I will change this function to use a sol_str_slice, because `get_blob_process()` and `get_json_process()`
uses `sol_buffer_steal()` in order to create blob. If I change to sol_str_slice I will possible duplicate memory.

cabelitos: However, I can change the line above to `sol_str_slice_str_caseeq(sol_buffer_get_slice(buf), "false")`

src/modules/flow/http-client/http-client.c
cabelitos: ok

src/modules/flow/http-server/http-server.c
cabelitos: Ok.

src/modules/flow/http-client/http-client.c
cabelitos: Well according to http://jsonlint.com/
`true` is valid. I will change it, so we do not need to send an object

barbieri: test with a browser as well... do a json request and send it a single value to see if that is okay

src/modules/flow/http-client/http-client.c
barbieri: ok, that is a good justification :-)

cabelitos:
:new_moon_with_face:

src/modules/flow/http-client/http-client.json
cabelitos: ok

cabelitos:
#1827

PR #1812, opened 2016-04-04
grove: compile if i2c *or* aio present.

glima:
This way, when one of those are not available we still get some of this
family's nodes.

### DIFF
--- a/src/modules/flow/grove/grove.c
+++ b/src/modules/flow/grove/grove.c
@@ -20,14 +20,20 @@
 #include <math.h>
 
 #include "sol-flow/grove.h"
-#include "sol-flow/aio.h"
 
 #include "sol-flow-internal.h"
 #include "sol-flow-static.h"
-#include "sol-i2c.h"
 #include "sol-mainloop.h"
 #include "sol-util-internal.h"
 #include "sol-vector.h"
+
+#ifdef USE_I2C
+#include "sol-i2c.h"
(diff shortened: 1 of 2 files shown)


src/modules/flow/grove/grove.c
```diff
@@ -20,15 +20,21 @@
...
 
 #include "sol-flow-internal.h"
 #include "sol-flow-static.h"
-#include "sol-i2c.h"
 #include "sol-mainloop.h"
 #include "sol-util-internal.h"
 #include "sol-vector.h"
 
+#ifdef USE_I2C
+#include "sol-i2c.h"
```
barbieri: move this and other i2c related stuff near the other ifdef? pretty minor, you can continue with that

glima: You mean include it in the middle of the file? I almost went with that, but aborted due to ugliness.

barbieri: i know, but at least it's a "section". Alternatively you can split in their own file and conditionally include them.

glima: Ok, going with mid-inclusion, good enough.

glima:
After talking with people here, we decided to split the lcd nodes into their own module/family. Some sort of node aliasing will come later as well.

glima:
Patch being made now, to be landed on another PR.

PR #1822, opened 2016-04-05
http-client: set error value on function return

bdilly:
Actually there could be two reasons to return at
this point, no memory to add to a vector or
invalid version.

It should be just a matter of returning the value
returned by add_param() function, but it's returning
a bool instead of an error value. It'll be fixed
soon as these API review changes are done anyway.


glima:
OK, as we'll fix ``sol_http_param_add()``'s return value soon.

bdilly:
ok, merged

PR #1823, opened 2016-04-05
form/converter: make Python-like formatting code optional.

glima:
The user is warned of the consequences on both affected nodes. We could
accept printf-like only formatting strings on the form too, but not
now (maybe later).

### DIFF
--- a/src/modules/flow/converter/Kconfig
+++ b/src/modules/flow/converter/Kconfig
@@ -1,3 +1,31 @@
 config FLOW_NODE_TYPE_CONVERTER
 	tristate "Node type: converter"
 	default y
+	help
+		The converter family of nodes will convert Soletta packets
+		from one type to another (e.g. boolean to byte). Some of them
+		will do more fancy things, like splitting a byte into (boolean)
+		individual bits. Others will compose a final packet from
+		others, like assembling four bytes into a final integer. You'll
+		also find a node that splits an error packet in other two—its
+		integer code and its string message. There are also the nodes
+		dealing with JSON data (e.g. one that gets a string packet
+		and outputs a JSON array packet, if the string is valid).
+		Finally, some of the nodes converting packets to string,
+		namely float-to-string and int-to-string, have Python-like
+		string formatting capabilities, as in
+		https://docs.python.org/3/library/stdtypes.html#str.format.
+		This code can be disabled, if not meant to be used, and
+		feeding data to those node instances, in this case, will
+		always make them output error packets.
+
+config PYTHON_FORMATTING
+	bool "Build Python-like string formatting (float-to-string/int-to-string) nodes"
+	depends on FLOW_NODE_TYPE_CONVERTER
+	default y
+	help
+		Consider changing this if this functionality won't be used
+		and the final Soletta binary is meant for constrained
+		systems. If disabled, the format of those nodes
+		will be the pristine printf ones and only the value (not the
+		range) of the inputs will be formatted.
--- a/src/modules/flow/form/Kconfig
+++ b/src/modules/flow/form/Kconfig
@@ -11,3 +11,14 @@ config FLOW_NODE_TYPE_FORM
 		used by other node types doing visual output, such as
 		LCD display (grove/lcd-string), in conjuntion with others
 		providing input.
+
+config PYTHON_FORMATTING
+	bool "Build Python-like string formatting (string-formatted) nodes"
+	depends on FLOW_NODE_TYPE_CONVERTER
+	default y
+	help
+		Consider changing this if this functionality won't be used
+		and the final Soletta binary is meant for constrained
+		systems. Note that the string-formatted node is made useless
+		if this is not set (any interaction with these nodes will
+		lead to error packets).
--- a/src/modules/flow/form/form.c
+++ b/src/modules/flow/form/form.c
@@ -2601,6 +2601,7 @@ struct string_formatted_chunk {
     } state;
 };
 
+#ifdef PYTHON_FORMATTING
 static int
 string_formatted_format_do(struct sol_flow_node *node)
 {
@@ -2760,7 +2761,9 @@ err:
 
     return r;
 }
+#endif
 
+#ifdef PYTHON_FORMATTING
 static bool
 string_formatted_timeout(void *data)
 {
@@ -2801,6 +2804,7 @@ string_formatted_format(struct sol_flow_node *node)
 
     return 0;
 }
+#endif
 
 static void
 string_formatted_close(struct sol_flow_node *node, void *data)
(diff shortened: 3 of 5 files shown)


src/modules/flow/converter/Kconfig
```diff
@@ -1,3 +1,31 @@
...
+		feeding data to those node instances, in this case, will
+		always make them output error packets.
+
+config PYTHON_FORMATTING
+	bool "Build Python-like string formatting (float-to-string/int-to-string) nodes"
+	depends on FLOW_NODE_TYPE_CONVERTER
+	default y
+	help
+		Consider changing the Python-like formatting functionality
+		won't be used or the final Soletta binary is meant for
```
edersondisouza: I'm rather confused... consider changing to what?

src/modules/flow/form/form.c
```diff
@@ -3471,6 +3510,12 @@ string_formatted_select_set(struct sol_flow_node *node,
     return sol_flow_send_string_slice_packet(node,
         SOL_FLOW_NODE_TYPE_FORM_INT__OUT__SELECTED,
         sol_buffer_get_slice(&mdata->formatted_value));
+#else
+    sol_flow_send_error_packet(node, ENOTSUP, "The string-formatted node"
+        " won't work on this Soletta build -- Python-like formatting "
+        "capability was disabled");
+    return -EINVAL;
+#endif
```
bdilly: If this node won't work, why is it even built?

src/modules/flow/form/Kconfig
```diff
@@ -11,3 +11,14 @@ config FLOW_NODE_TYPE_FORM
...
 		providing input.
+
+config PYTHON_FORMATTING
+	bool "Build Python-like string formatting (string-formatted) nodes"
+	depends on FLOW_NODE_TYPE_CONVERTER
+	default y
+	help
+		Consider changing this if this functionality won't be used
+		and the final Soletta binary is meant for constrained
+		systems. Note that the string-formatted node is made useless
```
edersondisouza: Ahhhh... here the help is understandable... maybe you want to use the same words on previous Kconfig.

src/modules/flow/form/form.c
glima: We can't change the .jsons declaring them based on config options (same case for removing PCRE support).

src/modules/flow/converter/Kconfig
glima: My bad, some line got killed here it seems.

src/modules/flow/form/Kconfig
glima: :) Pushing on top of that now.

src/modules/flow/converter/Kconfig
ibriano: So, depending on how Soletta is configured, the same node may have different outputs? That doesn't feel right.

src/modules/flow/form/form.c
bdilly: I don't like much having dependencies / configuration for only part of the node types of a module.

So you have the module, correct version, but it will give a runtime error for lack of support in some configuration. How should a flow deal with such error packet? Is there any solution?

Wouldn't make more sense to provide it in another module?

barbieri:
-0

I don't like this approach that much, why do we need this configuration? Wouldn't it be better to separate this in different nodes? Then if the node is not used, their functions should go away.

glima:
@barbieri: note that one the nodes in question is *still* usable with the new config. I can separate the form one in another node (I'd have to do that with the pcre one as well). Everyone OK with that approach?

src/modules/flow/form/form.c
ibriano: On Tue, 05 Apr 2016 10:56:44 -0700, Bruno Dilly wrote:
> > @@ -3471,6 +3510,12 @@ string_formatted_select_set(struct sol_flow_node *node,
> >      return sol_flow_send_string_slice_packet(node,
> >          SOL_FLOW_NODE_TYPE_FORM_INT__OUT__SELECTED,
> >          sol_buffer_get_slice(&mdata->formatted_value));
> > +#else
> > +    sol_flow_send_error_packet(node, ENOTSUP, "The string-formatted node"
> > +        " won't work on this Soletta build -- Python-like formatting "
> > +        "capability was disabled");
> > +    return -EINVAL;
> > +#endif
> 
> I don't like much having dependencies / configuration for only part of the node types of a module.
> 
> So you have the module, correct version, but it will give a runtime error for lack of support in some configuration. How should a flow deal with such error packet? Is there any solution?
> 
> Wouldn't make more sense to provide it in another module?

I concur. The PCRE/ICU thing was done as a quick fix, and I have to
admit that seeing that big fat warning every time I build for !Linux it
annoys me. We need to work on providing a standard set of features that
will be available for all systems we support, and those things we can't
provide for all can be optional and on the side, not the standard to be
disabled when an obstacle is met.

ibriano:
On Tue, 05 Apr 2016 11:03:31 -0700, Gustavo Lima Chaves wrote:
> @barbieri: note that one the nodes in question is *still* usable with the new config. I can separate the form one in another node (I'd have to do that with the pcre one as well). Everyone OK with that approach?

They are usable, but the result is not the same, so anyone using these
nodes needs to keep in mind that the result might change under different
configurations.

bdilly:
@glima couldn't you explain in the commit msg what exactly is going on? Looks like these functions are only used by the node type "form/string-formatted", that I suppose you're not using in your demo for small os.

Unused nodes aren't removed from the binary?

ibriano:
On Tue, 05 Apr 2016 11:07:27 -0700, Bruno Dilly wrote:
> @glima couldn't you explain in the commit msg what exactly is going on? Looks like these functions are only used by the node type "form/string-formatted", that I suppose you're not using in your demo for small os.
> 
> Unused nodes aren't removed from the binary?

If nothing references them, yes. But if the functions they use to do
their work end up under some path referencing other functions, they will
all be brought in.
Same reason why anything using the console module will drag every packet
type we have, even if all you do is print an empty from a timer to check
that the program is alive.

glima:
@bdilly: two converter nodes also refer to those functions. @ibriano's point is valid, but I try to state the consequences on the Kconfig help text (not in the original node's documentation, though). I can tweak the commit message, sure.

glima:
I'll wait for the work on alias to proceed here. Maybe we can have a python .so with the nodes in question and alias them back the original names.

ibriano:
On Tue, 19 Apr 2016 13:03:09 -0700, Gustavo Lima Chaves wrote:
> I'll wait for the work on alias to proceed here. Maybe we can have a python .so with the nodes in question and alias them back the original names.

Will they provide the exact same behavior?

glima:
Yes, and PR made.

PR #1825, opened 2016-04-05
sol-bluetooth: Basic functionality

vcgomes:
This implements basic functionality for the Linux implementation of sol-bluetooth, this allows to power the adapter up (only the first adapter found) and scan for devices.

Also, there are a couple of sol-bus fixes, that were found while implementing this. I suggest reviewing this first. (Github messed with my ordering).


src/lib/comms/Kconfig
```diff
@@ -16,6 +16,16 @@ config DTLS
 
 	    If unsure, say N.
 
+config BLUETOOTH
+	bool "Bluetooth"
+	default n
```
barbieri: nah, default y. if unsure say y.

src/lib/comms/Makefile
```diff
@@ -121,12 +124,21 @@ obj-networking-$(MAVLINK)-extra-cflags += \
 obj-networking-$(LWM2M) += \
 	sol-lwm2m.o
 
+obj-bluetooth-$(BLUETOOTH) += \
+	sol-bluetooth.o
+
+obj-bluetooth-$(PLATFORM_LINUX) += \
```
barbieri: should we do "bluez" instead of linux here? I wonder if some crazy dudes may come with their own "small bluetooth"  one day. Calling it bluez is more specific and clear on what we do. Same for `impl` filename

src/lib/comms/include/sol-bluetooth.h
```diff
@@ -0,0 +1,90 @@
...
+	enum sol_bt_uuid_type type;
+	union {
+		uint16_t val16;
+		uint32_t val32;
+		uint8_t val128[16];
+		uint8_t val[0];
+	};
+};
+
+int sol_bt_uuid_from_str(struct sol_bt_uuid *uuid, const char *str);
```
barbieri: are these always null-terminated? if not then we should use a str_slice, that have an easy conversion from char* using `sol_str_slice_from_str()`

src/lib/comms/include/sol-bluetooth.h
```diff
@@ -0,0 +1,90 @@
...
+	int16_t rssi;
+};
+
+int sol_bt_set_connection_callbacks(void *(connected)(void *user_data,
+						      struct sol_bt_conn *conn),
+				    void *(disconnected)(void *user_data,
+							 struct sol_bt_conn *conn),
+				    const void *user_data);
+
+void sol_bt_unset_connection_callbacks(void);
```
barbieri: no new symbol, document people should call `set` with `NULL`

src/lib/comms/include/sol-bluetooth.h
```diff
@@ -0,0 +1,90 @@
...
+
+int sol_bt_enable(void (*enabled)(void *data, bool powered), const void *user_data);
+int sol_bt_disable(void);
+bool sol_bt_is_enabled(void);
+
+enum sol_bt_transport {
+	SOL_BT_TRANSPORT_ALL,
+	SOL_BT_TRANSPORT_LE,
+	SOL_BT_TRANSPORT_BREDR,
+};
```
barbieri: from bindings/nodejs we noticed we need to add enum<->string conversion, so when they expose these in their bindings they get the standard conversion to be applied. See their PR for gpio/uart. use sol_str_table_ to do the lookups from string and an array to convert to string

src/lib/comms/include/sol-gatt.h
```diff
@@ -0,0 +1,256 @@
...
+	SOL_GATT_CHR_FLAGS_ENCRYPT_WRITE = (1 << 2),
+	SOL_GATT_CHR_FLAGS_ENCRYPT_AUTHENTICATED_READ = (1 << 3),
+	SOL_GATT_CHR_FLAGS_ENCRYPT_AUTHENTICATED_WRITE = (1 << 4),
+};
+
+/*
+ * FIXME: need to add errors that make sense for the BlueZ D-Bus API
+ * and "raw" GATT.
+ */
+enum sol_gatt_error {
```
barbieri: usually we're not using specific errors like this, rather translating them to errno

src/lib/comms/include/sol-gatt.h
```diff
@@ -0,0 +1,256 @@
...
+struct sol_gatt_attr {
+	struct sol_bt_uuid uuid;
+	enum sol_gatt_attr_type type;
+	uint16_t flags;
+	/**
+	 * The response for this operation will be returned by calling
+	 * #sol_gatt_pending_reply() passing a valid buffer, or an error.
+	 */
+	int (*read)(struct sol_gatt_pending *op,
+		    uint16_t offset);
```
barbieri: what is offset?

src/lib/comms/include/sol-gatt.h
```diff
@@ -0,0 +1,256 @@
...
+	/**
+	 * The response for this operation will be returned by calling
+	 * #sol_gatt_pending_reply(), the @a buf parameter will be ignored.
+	 */
+	int (*write)(struct sol_gatt_pending *op,
+		     struct sol_buffer *buf,
```
barbieri: is this buffer const? (source data, data to be written?) maybe rename the parameter, right now I can't know if it's input or output buffer.

src/lib/comms/include/sol-gatt.h
```diff
@@ -0,0 +1,256 @@
...
+	/**
+	 * This operation doesn't require an reply.
+	 */
+	/*
+	 * FIXME: is there any case that it would make sense to
+	 * wait for a reply?
+	 */
+	int (*flush)(struct sol_gatt_pending *op,
+		     bool sync);
+	void *user_data;
```
barbieri: these should be const.

src/lib/comms/include/sol-gatt.h
```diff
@@ -0,0 +1,256 @@
...
+
+/*
+ * Helpers to the construction of the local attribute database
+ */
+#define SOL_GATT_SERVICE(service_uuid) {			\
+		.type = SOL_GATT_ATTR_TYPE_SERVICE,		\
+		.uuid = service_uuid,				\
+	}
+
+#define SOL_GATT_CHARACTERISTIC(uuid, flags, read, write,		\
```
barbieri: this and GATT_DESCRIPTOR macros will get out of hand to the reader. It's better to give then __VA_ARGS__ and then let the user call with: `SOL_GATT_CHARACTERISTIC(uuid, .flags = myflags, .read = myread...)`

src/lib/comms/include/sol-gatt.h
```diff
@@ -0,0 +1,256 @@
...
+		.read = read,						\
+		.write = write,						\
+		.flush = flush,						\
+		.user_data = user_data,					\
+	}
+
+/**
+ * @brief Registers attributes into the GATT database
+ *
+ * @param attrs Array of the attributes to be registered
```
barbieri: why not null-terminated? (define a `SOL_GATT_ATTR_TYPE_UNKNOWN = 0`)

should it be const? do you keep a reference? Say so here.

src/lib/comms/include/sol-gatt.h
```diff
@@ -0,0 +1,256 @@
...
+ *
+ * This only really makes sense for Characteristics and Descriptors,
+ * the Read Long attribute procedure will be performed if it's supported.
+ *
+ * @param conn Connection to a remote device
+ * @param attr Attribute to be read
+ * @param cb Callback to be called when the operation is finished
+ * @param user_data User pointer to be passed to the callback
+ */
+int sol_gatt_read_attr(struct sol_bt_conn *conn, struct sol_gatt_attr *attr,
```
barbieri: do we need a pending handle to cancel it? Or it is automatically gone if the connection dies? We need to write that here.

Case is:
 - ask read to show in some context;
 - context is gone (ie: window closed, widgets and local context is free()ed);
 - you must have a way to stop the read, otherwise it will call you back and you'll access data that is not there anymore.

In this case I believe we must return a handle to cancel, otherwise we'd have to kill the connection to support this case, but maybe the other parts of your app still needs that connection

src/lib/comms/include/sol-gatt.h
```diff
@@ -0,0 +1,256 @@
...
+int sol_gatt_unregister_attributes(struct sol_gatt_attr *attrs, uint16_t len);
+
+/**
+ * @brief Reads the value from a attribute
+ *
+ * This only really makes sense for Characteristics and Descriptors,
+ * the Read Long attribute procedure will be performed if it's supported.
+ *
+ * @param conn Connection to a remote device
+ * @param attr Attribute to be read
```
barbieri: should gatt be const?

src/lib/comms/include/sol-gatt.h
```diff
@@ -0,0 +1,256 @@
...
+ * @brief Discover attributes by type, restricted by UUID and parent attribute.
+ *
+ * Discover attributes belonging to a parent attribute, for example,
+ * discover all the characteristics under a service or all the
+ * descriptors under a characteristic, matching uuid (if provided).
+ */
+int sol_gatt_discover(struct sol_bt_conn *conn, enum sol_gatt_attr_type type,
+		      const struct sol_gatt_attr *parent,
+		      const struct sol_bt_uuid *uuid,
+		      bool (*func)(void *user_data, struct sol_bt_conn *conn,
```
barbieri: s/func/cb/g to be consistent.

also improve the documentation, for all parameters, write which one can be NULL and what it means to return true/false from callback.

src/lib/comms/include/sol-gatt.h
```diff
@@ -0,0 +1,256 @@
...
+ */
+int sol_gatt_unsubscribe(struct sol_bt_conn *conn, struct sol_gatt_attr *attr,
+		       bool (*cb)(void *user_data, struct sol_gatt_attr *attr,
+				  const uint8_t *value, uint16_t len));
+
+/**
+ * Sends an indication to the device represented by @a conn, if @a conn is NULL, send to all
+ * devices that have registered themselves via the CCC attribute.
+ *
+ * The value to be indicated is the value retrieved by calling @a read callback of the attribute.
```
barbieri: since this relates to `sol_gatt_subscribe()`, cross link them with `@see funcname()`

src/lib/comms/sol-bluetooth-impl-linux.c
```diff
@@ -0,0 +1,683 @@
...
+    void *enabled_user_data;
+    bool adapter_powered;
+} context;
+
+enum {
+    ADAPTER_PROPERTY_POWERED = 0,
+};
+
+static void
+strv_free(char **strv)
```
barbieri: this should be in sol-util.h

src/lib/comms/sol-bluetooth-impl-linux.c
```diff
@@ -0,0 +1,683 @@
...
+    { NULL, NULL }
+};
+
+static void
+adapter_property_changed(void *data, const char *path, uint64_t mask)
+{
+    struct context *ctx = data;
+
+    if (mask & (1 << ADAPTER_PROPERTY_POWERED)) {
+        ctx->adapter_powered = !ctx->adapter_powered;
```
barbieri: why invert it?

src/lib/comms/sol-bluetooth-impl-linux.c
```diff
@@ -0,0 +1,683 @@
...
+
+    info = &device->info;
+    free(info->name);
+    sol_vector_clear(&info->uuids);
+
+    free(device->path);
+    free(device);
+}
+
+static struct device_info *find_device_by_path(struct context *ctx, const char *path)
```
barbieri: const ctx and then device_info/d.

src/lib/comms/sol-bluetooth-impl-linux.c
```diff
@@ -0,0 +1,683 @@
...
+
+    info = &d->info;
+
+    if (info->name)
+        goto skip;
+
+    r = sd_bus_message_read_basic(m, 's', &name);
+    SOL_INT_CHECK_GOTO(r, < 0, skip);
+
+    /* When we receive the property again we will try to allocate again */
```
barbieri: shouldn't you update it and notify user?

src/lib/comms/sol-bluetooth-impl-linux.c
```diff
@@ -0,0 +1,683 @@
...
+    struct sol_bt_uuid *u = NULL;
+    char **uuids = NULL, **uuid;
+    int r;
+
+    d = find_device_by_path(ctx, path);
+    SOL_NULL_CHECK_GOTO(d, skip);
+
+    info = &d->info;
+
+    r = sd_bus_message_read_strv(m, &uuids);
```
barbieri: can't you open the array, iterate using on-wire data, then avoid the strv memory allocation and free? you just need it to decode the uuid to our sol_bt_uuid

src/lib/comms/sol-bluetooth-impl-linux.c
```diff
@@ -0,0 +1,683 @@
...
+    SOL_NULL_CHECK(d);
+
+    d->mask |= mask;
+
+    if (!scan->callback)
+        return;
+
+    info = &d->info;
+
+    if ((d->mask & all_present) == all_present)
```
barbieri: why only on all are present? Wouldn't be better to use a single `sol_timeout_add(0,...)` to dispatch it? so you delay it to the next mainloop iteration, but do it even if not all properties arrived?

src/lib/comms/sol-bluetooth-impl-linux.c
```diff
@@ -0,0 +1,683 @@
...
+
+    return 0;
+}
+
+SOL_API int
+sol_bt_set_connection_callbacks(void *(connected)(void *user_data,
+        struct sol_bt_conn *conn),
+    void *(disconnected)(void *user_data,
+        struct sol_bt_conn *conn),
+    const void *user_data)
```
barbieri: so why provide this? from the .h I was clueless why this was required and what was the use, expected to at least get a hint  here :-)

src/lib/comms/sol-bluetooth.c
```diff
@@ -0,0 +1,123 @@
...
+        break;
+    case SOL_BT_UUID_TYPE_128:
+        memcpy(&tmp0, &uuid->val128[0], sizeof(tmp0));
+        memcpy(&tmp1, &uuid->val128[2], sizeof(tmp1));
+        memcpy(&tmp2, &uuid->val128[6], sizeof(tmp2));
+        memcpy(&tmp3, &uuid->val128[8], sizeof(tmp3));
+        memcpy(&tmp4, &uuid->val128[10], sizeof(tmp4));
+        memcpy(&tmp5, &uuid->val128[12], sizeof(tmp5));
+
+        r = sol_buffer_append_printf(buffer , "%.8x-%.4x-%.4x-%.4x-%.8x%.4x",
```
barbieri: should it be 0-padded? or is it common to have "1-2-3"?

src/lib/comms/sol-bluetooth.c
```diff
@@ -0,0 +1,123 @@
...
+        return -EINVAL;
+    }
+
+    return r ? -errno : 0;
+}
+
+SOL_API bool
+sol_bt_uuid_equal(const struct sol_bt_uuid *u1, const struct sol_bt_uuid *u2)
+{
+    return false;
```
barbieri: at least a TODO... but i guess you could do easily with memcmp based on the given size.

src/lib/comms/sol-network-impl-linux.c
```diff
@@ -75,6 +75,68 @@ struct sol_network {
...
+bluetooth_addr_from_str(struct sol_network_link_addr *addr, const char *str)
+{
+    int i;
+    uint8_t *ba = addr->addr.bt_addr;
+
+    /* FIXME */
+    addr->family = SOL_NETWORK_FAMILY_BLUETOOTH;
+    addr->addr.bt_type = SOL_NETWORK_BT_ADDR_BASIC_RATE;
+
+    for (i = 5; i >= 0; i--, str += 3)
```
barbieri: you should iterate str based on endptr from strtol... and it should be the strtoul (unsigned)

src/lib/comms/sol-network-impl-linux.c
```diff
@@ -75,6 +75,68 @@ struct sol_network {
...
+    return (str - p) == 17;
+}
+
+static const char *
+bluetooth_addr_to_str(const struct sol_network_link_addr *addr,
+    struct sol_buffer *buffer)
+{
+    const uint8_t *ba = addr->addr.bt_addr;
+
+    sol_buffer_append_printf(buffer, "%2.2X:%2.2X:%2.2X:%2.2X:%2.2X:%2.2X",
```
barbieri: you're missing "0" to do zero padding, otherwise it will contain spaces to pad.

src/lib/comms/Makefile
barbieri: see what I did for crypto, that way you select the one for your platform and your makefile will be as simple as now, just a different symbol.

src/lib/comms/include/sol-bluetooth.h
```diff
@@ -0,0 +1,90 @@
...
+	char *name;
+	struct sol_vector uuids;
+	int16_t rssi;
+};
+
+int sol_bt_set_connection_callbacks(void *(connected)(void *user_data,
+						      struct sol_bt_conn *conn),
+				    void *(disconnected)(void *user_data,
+							 struct sol_bt_conn *conn),
+				    const void *user_data);
```
barbieri: thinking about usage, I believe that a global set of callbacks won't do. If you have multiple parts of your application, each doing its own connections, you'll have problem with a single callback.

Maybe add a monitor-like pattern or go back to give the "on_connect" and "on_disconnect" to `sol_bt_connect()` (I guess you did that in your first proposal and i've asked you to change?)

barbieri: ah, and if that is an async pattern, you need to return a handle so it can be cancelled.

src/lib/comms/include/sol-bluetooth.h
```diff
@@ -0,0 +1,90 @@
...
+
+void sol_bt_unset_connection_callbacks(void);
+
+int sol_bt_connect(const struct sol_network_link_addr *addr);
+int sol_bt_disconnect(const struct sol_network_link_addr *addr);
+
+int sol_bt_enable(void (*enabled)(void *data, bool powered), const void *user_data);
```
barbieri: as per above, if it's async you need a handle to cancel it.

barbieri:
overall I miss the documentation in header to know what you really want to do with those functions. This will help us to understand the API even without the actual bits in it.

In order to proceed, please remove the unsupported parts (ie: sol-gatt.h) and focus on a single part of the API, document it extensively and provide samples. Then we get it merged and you proceed to other parts of the API.

src/lib/comms/Kconfig
vcgomes: Ok.

src/lib/comms/Makefile
vcgomes: Will take a look. I was wondering about making "bluez" when writing this, but as BlueZ is the default Bluetooth stack for Linux, I thought this made sense. But no hard feelings about this.

src/lib/comms/include/sol-bluetooth.h
vcgomes: Yeah, I am assuming the `str` to be null-terminated. Will do.

src/lib/comms/include/sol-bluetooth.h
vcgomes: Ok.

src/lib/comms/include/sol-bluetooth.h
vcgomes: Will do.

src/lib/comms/include/sol-gatt.h
vcgomes: I was worried about awkward translations. But will leave it for later, when this API is added for real (using your suggestion later in this series).

src/lib/comms/include/sol-gatt.h
vcgomes: In GATT (and this information is exported in Zephyr) it's is valid to receive a read request asking to read starting in byte `offset`. 

I could read the entire attribute (it can be at most 512 bytes long) and do this inside `sol-gatt`, sure. Should be slightly worse for big attributes, but for the majority of cases (small attributes) it will be simpler.

src/lib/comms/include/sol-gatt.h
vcgomes: Will find a better name.

src/lib/comms/include/sol-gatt.h
vcgomes: Will fix.

src/lib/comms/include/sol-gatt.h
vcgomes: Cool. Didn't think of this. Will fix.

src/lib/comms/include/sol-gatt.h
vcgomes: Just a matter of taste. Being null-terminated sounds fine. Will fix.

src/lib/comms/include/sol-gatt.h
vcgomes: Will add an way to cancel this, I guess `sol_gatt_pending` can be reused for this.

src/lib/comms/include/sol-gatt.h
vcgomes: Most probably.

src/lib/comms/include/sol-gatt.h
vcgomes: yeah, will fix.

src/lib/comms/include/sol-gatt.h
vcgomes: Will fix.

src/lib/comms/sol-bluetooth-impl-linux.c
vcgomes: Yeah, this ended being more complicated than it needs to be.

This is how it works: `adapter_property_powered_set()` only returns `true` when the adapter powered state is different from the stored one, `_set()` returning `true` causes this function to be called with the appropriate mask. So I know that the powered is state is different than the one before.

src/lib/comms/sol-bluetooth-impl-linux.c
vcgomes: I couldn't understand the 'device_info' part of your comment.

src/lib/comms/sol-bluetooth-impl-linux.c
vcgomes: Yeah, will add a check for when 'Name' changed.

src/lib/comms/sol-bluetooth-impl-linux.c
vcgomes: I could, `_read_strv()` was just more convenient. Will fix.

src/lib/comms/sol-bluetooth-impl-linux.c
vcgomes: I forgot to change this name, it should say `basic_info` or something like it, it includes address, name and rssi. In "recent" controllers (made after 2003, I believe) this information will be always available.

src/lib/comms/sol-bluetooth-impl-linux.c
vcgomes: This function will only be needed when implementing GATT.

src/lib/comms/sol-bluetooth.c
vcgomes: ".8"/".4" will do the padding.

src/lib/comms/sol-bluetooth.c
vcgomes: It is a little more complicated, there are cases than a 16-bit UUID will for comparison purposes be equal to an 128-bit UUID, for comparing UUIDs of different sizes I will need to convert them to their 128-bit representation.

src/lib/comms/sol-network-impl-linux.c
vcgomes: Cool.

src/lib/comms/sol-network-impl-linux.c
vcgomes: specifying the precision does that, no?

src/lib/comms/include/sol-bluetooth.h
vcgomes: I tried this API so we could have a consistent API for both "initiator" and "acceptor" use cases. But perhaps this is indeed too simplistic.

What I am thinking right now is something like this: 
```C
struct sol_bt_pending_conn *
sol_bt_connect(addr, void (*on_connect)(), void (*on_disconnect)(), user_data)
``` 

if `addr` is a empty address (with just a family, or with family and port), I use it to signify that I am should listen for connections. And I call all matching callbacks in case of ambiguity.

src/lib/comms/include/sol-bluetooth.h
vcgomes: Yeah.

src/lib/comms/sol-bluetooth-impl-linux.c
barbieri: the iterator element you use below: `struct device_info *d` should be const.

src/lib/comms/sol-bluetooth.c
barbieri: ah, I've missed that. I guessed ".8" would limit to 8 digits, but would pad with spaces. Just tried and confirmed it does 0-padding.

src/lib/comms/sol-bluetooth.c
barbieri: hum...  in this case should we always bump uuid to 128bit? Or do we need to remember if it was 16-bit or 128?

src/lib/comms/include/sol-bluetooth.h
barbieri: now I understand it. Maybe we can have two API, one to trigger connect to a given device, the other one to accept incoming requests?

src/lib/comms/sol-bluetooth.c
vcgomes: I need to remember. If I want to save power I need to use the shorter representation always.

src/lib/comms/include/sol-bluetooth.h
vcgomes: I don't know. But I am having a slight preference for the suggestion I made above. I will try to produce a sample that exercises this and see how it would look.

vcgomes:
There's an updated version in https://github.com/solettaproject/soletta/pull/1948, closing this.

PR #1827, opened 2016-04-06
Add SSE support for Soletta's HTTP nodes v2

cabelitos:
Changes since v1:

* Added commits c3a27a1 and 4fcd1c1
* Added functions `sol_direction_vector_equal()` and `sol_rgb_equal()`
* Upon get request, the function `send_sse_data()` at http-server.c will not broadcast any more.
* In htt-server.c now copies the sse_clients array in order to close the sse connection.
* In http-client node - added more types to the accept option.
* Other minor fixes.

### DIFF
--- a/src/lib/common/include/sol-types.h
+++ b/src/lib/common/include/sol-types.h
@@ -116,6 +116,16 @@ struct sol_direction_vector {
 };
 
 /**
+ * @brief Checks the ranges of @c var0 and @c var1 for equality.
+ *
+ * @param var0 The first direction vector
+ * @param var1 The second direction vector
+ *
+ * @return @c true if both are equal, @c false otherwise.
+ */
+bool sol_direction_vector_equal(struct sol_direction_vector *var0, struct sol_direction_vector *var1);
+
+/**
  * @brief Data type to describe a location.
  */
 struct sol_location {
@@ -135,6 +145,16 @@ struct sol_rgb {
     uint32_t green_max; /**< @brief Green component maximum value */
     uint32_t blue_max; /**< @brief Blue component maximum value */
 };
+
+/**
+ * @brief Checks the ranges of @c var0 and @c var1 for equality.
+ *
+ * @param var0 The first RGB
+ * @param var1 The second RGB
+ *
+ * @return @c true if both are equal, @c false otherwise.
+ */
+bool sol_rgb_equal(struct sol_rgb *var0, struct sol_rgb *var1);
 
 /**
  * @brief Set a maximum value for all components of a RGB color
--- a/src/lib/comms/sol-http-common.c
+++ b/src/lib/comms/sol-http-common.c
@@ -20,6 +20,7 @@
 #include <stdio.h>
 #include <string.h>
 #include <inttypes.h>
+#include <stdlib.h>
 
 #include "sol-http.h"
 #include "sol-log.h"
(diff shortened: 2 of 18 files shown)


src/lib/common/include/sol-types.h
```diff
@@ -116,6 +116,16 @@ struct sol_direction_vector {
...
 
 /**
+ * @brief Checks the ranges of @c var0 and @c var1 for equality.
+ *
+ * @param var0 The first direction vector
+ * @param var1 The second direction vector
+ *
+ * @return @c true if both are equal, @c false otherwise.
+ */
+bool sol_direction_vector_equal(struct sol_direction_vector *var0, struct sol_direction_vector *var1);
```
barbieri: const....

src/lib/common/include/sol-types.h
```diff
@@ -137,6 +147,16 @@ struct sol_rgb {
...
 
 /**
+ * @brief Checks the ranges of @c var0 and @c var1 for equality.
+ *
+ * @param var0 The first RGB
+ * @param var1 The second RGB
+ *
+ * @return @c true if both are equal, @c false otherwise.
+ */
+bool sol_rgb_equal(struct sol_rgb *var0, struct sol_rgb *var1);
```
barbieri: const...

src/lib/common/include/sol-types.h
cabelitos: sorry

src/lib/common/include/sol-types.h
cabelitos: sorry

src/lib/comms/sol-http-common.c
```diff
@@ -882,3 +883,238 @@ sol_http_split_post_field(const char *query, struct sol_http_params *params)
...
+static char *
+is_qvalue_param(const struct sol_str_slice param)
+{
+#define STATE_NEEDS_Q_CHAR (0)
+#define STATE_NEEDS_EQUAL_CHAR (1)
+
+    size_t i;
+    int state = STATE_NEEDS_Q_CHAR;
+
+    for (i = 0; i < param.len; i++) {
```
barbieri: here you can iterate using pointers:
```c
const char *itr, *itr_end;

itr = param.data;
itr_end = itr + param.len;
for (; itr < itr_end; itr++) {
    if (isspace(*itr)) ...
```

src/lib/comms/sol-http-common.c
```diff
@@ -882,3 +883,238 @@ sol_http_split_post_field(const char *query, struct sol_http_params *params)
...
+        switch (state) {
+        case STATE_NEEDS_Q_CHAR:
+            if (param.data[i] != 'q')
+                return NULL;
+            state = STATE_NEEDS_EQUAL_CHAR;
+            break;
+        default:
+            if (param.data[i] != '=')
+                return NULL;
+            return (char *)param.data + i;
```
barbieri: it's better to return the qvalue_slice, you'll need it anyway. Return an empty slice if not found.

src/lib/comms/sol-http-common.c
```diff
@@ -882,3 +883,238 @@ sol_http_split_post_field(const char *query, struct sol_http_params *params)
...
+        return -1;
+
+    return 0;
+}
+
+static char *
+is_qvalue_param(const struct sol_str_slice param)
+{
+#define STATE_NEEDS_Q_CHAR (0)
+#define STATE_NEEDS_EQUAL_CHAR (1)
```
barbieri: instead of define you can declare a local enum:

```c
enum { state_needs_q_char, state_needs_equal_char } state;
```

cabelitos: I always forget about this... :/

src/lib/comms/sol-http-common.c
```diff
@@ -882,3 +883,238 @@ sol_http_split_post_field(const char *query, struct sol_http_params *params)
...
+    struct sol_vector tokens;
+    struct sol_str_slice *token;
+    struct sol_http_content_type_priority *pri;
+    uint16_t i;
+
+    SOL_NULL_CHECK(priorities, -EINVAL);
+
+    SOL_DBG("Parsing content priorities for: %.*s",
+        SOL_STR_SLICE_PRINT(content_type));
+    tokens = sol_str_slice_split(content_type, ",", 0);
```
barbieri: hum... it's bad that we need to allocate memory to just iterate. Also, below you do a manual iteration pattern as well (over `';'`), then my recommendation is to follow json-scanner pattern and offer a way to iterate without allocating memory, something like:

```c
struct sol_str_slice type = SOL_STR_SLICE_EMPTY; // first iteration takes an empty string, it is also used to keep the state, so if trim() is to be used, it must be done in a copy.
uint16_t i = 0;
while (sol_str_slice_split_iterate(content_type, &type, ',')) {
    struct sol_str_slice token = SOL_STR_SLICE_EMPTY;
    struct sol_http_content_type_priority *pri = NULL;

    while (sol_str_slice_split_iterate(type, &token, ';')) {
        struct sol_str_slice clean_token = sol_str_slice_trim(token);

        if (!pri)  {
            pri = sol_vector_append(priorities);
            pri->index = i;
            pri->qvalue = 1.0;
           pri->content_type = clean_token;
           sol_vector_init(&pri->tokens, sizeof(struct sol_str_slice));
           set_type_and_sub_type(pri->content_type, &pri->type, &pi->sub_type);
        } else if (!is_qvalue_start(clean_token)) {
           struct sol_str_slice *el = sol_vector_append(&pri->tokens);
           *el = clean_token;
        } else {
           pri->qvalue = ...;
        }
    }
    i++;
}
```

src/lib/comms/sol-http-common.c
cabelitos: fair. I wanted to mimic strstr(). But an empty slice is better anyway

src/lib/comms/sol-http-common.c
```diff
@@ -882,3 +883,238 @@ sol_http_split_post_field(const char *query, struct sol_http_params *params)
...
+                } else
+                    pri->qvalue = qvalue;
+                SOL_DBG("Type:%.*s with qvalue: %g",
+                    SOL_STR_SLICE_PRINT(pri->content_type), pri->qvalue);
+            }
+        } while (token->len);
+    }
+
+    qsort(priorities->data, priorities->len,
+        sizeof(struct sol_http_content_type_priority), sort_priority);
```
barbieri: you can use priorities->elem_size instead of sizeof.

src/modules/flow/http-client/http-client.c
```diff
@@ -402,17 +371,102 @@ common_request_finished(void *data,
...
+        total_len = content_len + prefix.len + suffix.len;
+        consumed += total_len;
+
+        SOL_DBG("Parsed SSE data:*%.*s*",
+            SOL_STR_SLICE_PRINT(sol_buffer_get_slice(&content_buf)));
+
+        if (type->process_json)
+            r = type->process_json(node, sol_buffer_get_slice(&content_buf));
+        else
+            r = type->process_data(node, &content_buf);
```
barbieri: is this right?

src/modules/flow/http-client/http-client.c
```diff
@@ -422,13 +476,29 @@ common_get_process(struct sol_flow_node *node, void *data, uint16_t port,
...
+
+        if (!sol_http_param_add(&params,
+            SOL_HTTP_REQUEST_PARAM_HEADER("Accept", mdata->accept))) {
+            SOL_WRN("Failed to set the 'Accept' header with value: %s",
+                mdata->accept);
+            r = -ENOMEM;
+            goto err;
+        }
+
+        r = sol_http_parse_content_type_priorities(sol_str_slice_from_str(mdata->accept), &priorities);
```
barbieri: you're parsing your own accept. You should get the accept from the HTTP request. In this case, the `GET` port, you should do the request with `mdata->accept` and wait for the server to reply with its `Content-Type: text/stream`, then, at http-client-request callback you can decide if you'll follow with SSE or not.

src/modules/flow/http-client/http-client.c
```diff
@@ -457,8 +527,13 @@ common_get_process(struct sol_flow_node *node, void *data, uint16_t port,
         SOL_INT_CHECK_GOTO(r, < 0, err);
     }
 
-    connection = sol_http_client_request(SOL_HTTP_METHOD_GET, mdata->url,
-        &params, type->http_response ? : common_request_finished, node);
+    if (!is_sse_request) {
+        connection = sol_http_client_request(SOL_HTTP_METHOD_GET, mdata->url,
+            &params, type->http_response ? : common_request_finished, node);
+    } else {
+        connection = sol_http_client_request_with_interface(SOL_HTTP_METHOD_GET,
```
barbieri: you should always do this one. If server replied with `Content-Type: text/steam` then you keep the progressive approach. Otherwise you wait all the content to arrive and process as usual.

src/lib/comms/sol-http-common.c
cabelitos: Ok

src/modules/flow/http-client/http-client.c
cabelitos: What's the problem?

src/modules/flow/http-client/http-client.c
```diff
@@ -602,30 +690,44 @@ boolean_post_process(struct sol_flow_node *node, void *data, uint16_t port, uint
...
+    sol_json_scanner_init(&sub_scanner, slice.data, slice.len);
+    found = false;
+
+    SOL_JSON_SCANNER_OBJECT_LOOP (&sub_scanner, &token, &sub_key, &sub_value, reason) {
+        if (SOL_JSON_TOKEN_STR_LITERAL_EQ(&sub_key, "value")) {
+            if (sol_json_token_get_type(&sub_value) == SOL_JSON_TYPE_STRING)
+                result = sol_json_token_get_unescaped_string_copy(&sub_value);
+            else
+                result = strndup(sub_value.start,
+                    sub_value.end - sub_value.start);
```
barbieri: `sol_json_token_get_size()`

src/modules/flow/http-client/http-client.c
```diff
@@ -602,30 +690,44 @@ boolean_post_process(struct sol_flow_node *node, void *data, uint16_t port, uint
...
-string_process_data(struct sol_flow_node *node,
-    struct sol_http_response *response)
+string_process_data(struct sol_flow_node *node, struct sol_buffer *buf)
 {
     char *result;
 
-    SOL_HTTP_RESPONSE_CHECK_API(response, -EINVAL);
-
-    result = strndup(response->content.data, response->content.used);
+    result = strndup(buf->data, buf->used);
```
barbieri: maybe follow the optimization done for json and use `sol_buffer_steal_or_copy()`?

src/modules/flow/http-client/http-client.c
```diff
@@ -652,15 +754,14 @@ string_post_process(struct sol_flow_node *node, void *data, uint16_t port, uint1
...
-    struct sol_json_token *key, struct sol_json_token *value)
+int_process_json(struct sol_flow_node *node, const struct sol_str_slice slice)
 {
     struct sol_irange irange = SOL_IRANGE_INIT();
     enum sol_json_loop_reason reason;
     struct sol_json_scanner sub_scanner;
     struct sol_json_token sub_key, sub_value, token;
 
-    sol_json_scanner_init(&sub_scanner, value->start, value->end - value->start);
+    sol_json_scanner_init(&sub_scanner, slice.data, slice.len);
```
barbieri: maybe add a helper `sol_json_scanner_init_from_slice()`?  this looks useful (static inline in the header).

src/modules/flow/http-client/http-client.c
```diff
@@ -682,15 +783,12 @@ int_process_token(struct sol_flow_node *node,
...
-    struct sol_http_response *response)
+int_process_data(struct sol_flow_node *node, struct sol_buffer *buf)
 {
     int value;
 
-    SOL_HTTP_RESPONSE_CHECK_API(response, -EINVAL);
-
     errno = 0;
-    value = strtol(response->content.data, NULL, 0);
+    value = sol_util_strtol(buf->data, NULL, buf->used, 0);
```
barbieri: maybe `sol_util_slice_strtol()` also looks useful.

src/modules/flow/http-client/http-client.c
```diff
@@ -888,15 +979,14 @@ hex_str_to_decimal(const char *start, ssize_t len, uint32_t *value)
...
+rgb_process_data(struct sol_flow_node *node, struct sol_buffer *buf)
 {
     struct sol_rgb rgb = { 0 };
     struct sol_str_slice rgb_str;
     int r;
 
-    rgb_str.data = response->content.data;
-    rgb_str.len = response->content.used;
+    rgb_str.data = buf->data;
+    rgb_str.len = buf->used;
```
barbieri: `sol_buffer_get_slice()`

src/modules/flow/http-client/http-client.c
```diff
@@ -1128,22 +1216,18 @@ generic_url_process(struct sol_flow_node *node, void *data, uint16_t port,
...
     size_t size;
     void *data;
     int r;
 
-    SOL_HTTP_RESPONSE_CHECK_API(response, -EINVAL);
-
-    SOL_DBG("Blob process - response from: %s", response->url);
-
-    data = sol_buffer_steal(&response->content, &size);
+    data = sol_buffer_steal(buf, &size);
```
barbieri: safer to call `steal_or_copy()`?

src/test/test-http.c
```diff
@@ -20,6 +20,93 @@
...
+            { SOL_STR_SLICE_LITERAL("image/*"), 0.5, 0, { 0 } },
+            { SOL_STR_SLICE_LITERAL("*/*"), 0.1, 0, { 0 } },
+        },
+    };
+
+    static const struct {
+        int r;
+        size_t priorities_len;
+        struct sol_str_slice accept;
+    } test[] = {
```
barbieri: ugh, can't you add the expected results here as well? This lookup will likely introduce errors in the future.

src/modules/flow/http-client/http-client.c
barbieri: do we handle SSE data for nodes that do not implement json?

cabelitos: In this case process_data() will be used by http-json only

barbieri: ah, ok. maybe a comment?

src/lib/comms/sol-http-common.c
cabelitos: I will create this new function

src/modules/flow/http-client/http-client.c
cabelitos: I will add it.

src/modules/flow/http-client/http-client.c
cabelitos: Indeed.

src/modules/flow/http-client/http-client.c
cabelitos: Indeed

src/modules/flow/http-client/http-client.c
cabelitos: Ok

src/modules/flow/http-client/http-client.c
cabelitos: Thank you

src/modules/flow/http-client/http-client.c
cabelitos: we have sol_str_slice_to_int(). I will use it

src/modules/flow/http-client/http-client.c
cabelitos: Indeed

src/test/test-http.c
cabelitos: ok

cabelitos:
#1841

PR #1836, opened 2016-04-06
oic-gen: better port names and support missing titles

bdilly:


barbieri:
+1

bdilly:
ok
I've merged it with a slight modification, I'm capitalizing titles and using spaces instead of dots so they will look all the same

PR #1837, opened 2016-04-06
mqtt: Gives user data in config

ceolin:
It is the same patter used by message digest.

### DIFF
--- a/src/lib/comms/include/sol-mqtt.h
+++ b/src/lib/comms/include/sol-mqtt.h
@@ -330,6 +330,11 @@ struct sol_mqtt_config {
     const struct sol_cert *private_key;
 
     /**
+     * User data provided to the callbacks
+     */
+    const void *data;
+
+    /**
      * Handlers to be used with this connection
      */
     const struct sol_mqtt_handlers handlers;
@@ -341,11 +346,10 @@ struct sol_mqtt_config {
  * @param host The host address of the MQTT broker
  * @param port The host port to connect to
  * @param config Configuration and callbacks
- * @param data User data provided to the callbacks
  *
  * @return New mqtt object on sucess, NULL otherwise
  */
-struct sol_mqtt *sol_mqtt_connect(const char *host, uint16_t port, const struct sol_mqtt_config *config, const void *data);
+struct sol_mqtt *sol_mqtt_connect(const char *host, uint16_t port, const struct sol_mqtt_config *config);
 
 /**
  * @brief Reestablish the connection to the MQTT broker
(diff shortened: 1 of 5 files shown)


src/lib/comms/include/sol-mqtt.h
```diff
@@ -341,11 +346,10 @@ struct sol_mqtt_config {
  * @param host The host address of the MQTT broker
  * @param port The host port to connect to
  * @param config Configuration and callbacks
- * @param data User data provided to the callbacks
  *
  * @return New mqtt object on sucess, NULL otherwise
  */
-struct sol_mqtt *sol_mqtt_connect(const char *host, uint16_t port, const struct sol_mqtt_config *config, const void *data);
+struct sol_mqtt *sol_mqtt_connect(const char *host, uint16_t port, const struct sol_mqtt_config *config);
```
barbieri: I'm thinking if we should move host/port to the config as well... what do you think?

src/lib/comms/sol-mqtt-impl-mosquitto.c
```diff
@@ -371,7 +371,7 @@ sol_mqtt_connect(const char *host, uint16_t port, const struct sol_mqtt_config *
 
     /* It comes as const, but goes back to user on the callbacks as
      * not const, for convenience */
-    mqtt->data = (void *)data;
+    mqtt->data = (void *)config->data;
```
barbieri: take the time to store this as `const` and only cast on callbacks?

src/lib/comms/include/sol-mqtt.h
vcgomes: Hm, now that there's support for naming resolving, we could consider using `sol_network_link_addr` here. Just an idea.

vcgomes:
typo on the commit message: s/patter/pattern/ .

cabelitos:
@ceolin Maybe we should do the same thing for LWM2M.

src/lib/comms/include/sol-mqtt.h
ceolin: i do not think is better, libmosquitto expects a string (host) and an int (port), so it'll be necessary extra convertions if we use `sol_network_link_addr`

src/lib/comms/sol-mqtt-impl-mosquitto.c
ceolin: yep

src/lib/comms/include/sol-mqtt.h
ceolin: @barbieri this is the connect function, i guess we can keep the address out of the config

ceolin:
https://github.com/solettaproject/soletta/pull/1860

PR #1838, opened 2016-04-06
Zephyr small fixes

glima:


src/lib/common/sol-platform-impl-zephyr.c
```diff
@@ -100,8 +101,13 @@ sol_platform_impl_get_serial_number(char **number)
 int
 sol_platform_impl_get_os_version(char **version)
 {
-    SOL_WRN("Not implemented");
-    return -ENOTSUP;
+    SOL_NULL_CHECK(version, -EINVAL);
+
+    *version = strdup(KERNEL_VERSION_STRING);
+    if (!*version)
+        return -ENOMEM;
```
cabelitos: why not `SOL_NULL_CHECK()` ?

glima: Fair (copy and pasta from RIOT's).

glima:
@ibriano, could you take a look, please?

src/shared/sol-random.c
```diff
@@ -91,7 +91,11 @@ get_platform_seed(uint64_t seed)
 
     /* Fall back to using a bad source of entropy if platform-specific,
      * higher quality random sources, are unavailable. */
+#ifdef SOL_PLATFORM_ZEPHYR
+    return ((uint64_t)sys_rand32_get() << 32) | sys_rand32_get();
```
ibriano: Is this always available or it depends on some driver configuration?

src/lib/common/sol-platform-impl-zephyr.c
```diff
@@ -85,10 +85,22 @@ sol_platform_impl_set_target(const char *target)
...
 {
-    SOL_WRN("Not implemented");
+#ifdef SOL_MACHINE_ID
+    static const char ID[] = SOL_MACHINE_ID;
+    const char *ptr;
+    int i = 0;
+
+    for (ptr = ID; *ptr && i < 33; ptr++) {
+        id[i++] = *ptr;
+    }
```
ibriano: SOL_MACHINE_ID is a binary string? How do you pass it as a define?

src/shared/sol-random.c
glima: It *used* to be on always, recently it seems we need a config (and the config was changed wrongly, I had to patch the Kconfig) :/ But it's the recommended way of getting an almost-random number in the OS, it seems.

src/lib/common/sol-platform-impl-zephyr.c
glima: It's meant to get a string, yes, as in "4d77755382fb459abdd2181de3316aaa".

src/shared/sol-random.c
ibriano: On Thu, 07 Apr 2016 13:49:44 -0700, Gustavo Lima Chaves wrote:
> > @@ -91,7 +91,11 @@ get_platform_seed(uint64_t seed)
> >  
> >      /* Fall back to using a bad source of entropy if platform-specific,
> >       * higher quality random sources, are unavailable. */
> > +#ifdef SOL_PLATFORM_ZEPHYR
> > +    return ((uint64_t)sys_rand32_get() << 32) | sys_rand32_get();
> 
> It *used* to be on always, recently it seems we need a config (and the config was changed wrongly, I had to patch the Kconfig) :/ But it's the recommended way of getting an almost-random number in the OS, it seems.

Ok then

src/lib/common/sol-platform-impl-zephyr.c
ibriano: On Thu, 07 Apr 2016 13:50:24 -0700, Gustavo Lima Chaves wrote:
> >  {
> > -    SOL_WRN("Not implemented");
> > +#ifdef SOL_MACHINE_ID
> > +    static const char ID[] = SOL_MACHINE_ID;
> > +    const char *ptr;
> > +    int i = 0;
> > +
> > +    for (ptr = ID; *ptr && i < 33; ptr++) {
> > +        id[i++] = *ptr;
> > +    }
> 
> It's meant to get a string, yes, as in "4d77755382fb459abdd2181de3316aaa".

Then why not just strcpy?

glima: My original attempt was with a memcpy(), but it does not work.

glima:
So, merging with the due fixes.

PR #1840, opened 2016-04-07
Export socket V2

ceolin:
from v1:

- Removed ifdefs in headers
- Remove unecessary ifdefs commit
- Socket options is given in the new struct `sol_socket_default_options`
- `set_on_can_read/write` replace by `set_read/write_monitor`
- api check macros moved to C source


src/lib/comms/sol-socket-impl-linux.c
```diff
@@ -33,42 +33,36 @@
...
+    struct sol_fd *watch;
+    const void *data;
     int fd;
-    struct {
-        bool (*cb)(void *data, struct sol_socket *s);
-        const void *data;
-        struct sol_fd *watch;
-    } read, write;
+    bool read_monitor;
+    bool write_monitor;
```
vcgomes: I guess the code would be simpler if instead of separate booleans for each kind of monitor you had a `flags` field.

cabelitos: Indeed

src/lib/comms/include/sol-socket.h
```diff
@@ -0,0 +1,315 @@
...
+     * @brief Allows multiple sockets to be bound to the same
+     * socket address. It is used by sol_socket_bind()
+     */
+    bool reuse_port;
+
+    /**
+     * @brief Allow reuse of local addresses. It is used by
+     * sol_socket_bind()
+     */
+    bool reuse_addr;
```
barbieri: do we really need both? I didn't follow all the discussion about the socket options on linux :-(

src/lib/comms/include/sol-socket.h
```diff
@@ -0,0 +1,315 @@
...
+
+    /**
+     * @brief Function to be called to read data from socket.
+     *
+     * @li @c s the socket pointer (this)
+     * @li @c buf the data buffer that will be used to receive the data
+     * @li @c len the size of @a buf
+     * @li @c cliaddr the source address of the message
+     */
+    ssize_t (*recvmsg)(struct sol_socket *s, void *buf, size_t len,
```
barbieri: at least in this internal api already move to sol_buffer?

src/lib/comms/include/sol-socket.h
```diff
@@ -0,0 +1,315 @@
...
+        struct sol_network_link_addr *cliaddr);
```
barbieri: const?

src/lib/comms/include/sol-socket.h
```diff
@@ -0,0 +1,315 @@
...
+
+    /**
+     * @brief Function to be called to write data in the socket.
+     *
+     * @li @c s the socket pointer (this)
+     * @param buf the data to be transmitted
+     * @param len the size of @a buf
+     * @param cliaddr the address which the data will be sent
+     */
+    int (*sendmsg)(struct sol_socket *s, const void *buf, size_t len,
```
barbieri: at least on unix it's `ssize_t` return.

src/lib/comms/include/sol-socket.h
```diff
@@ -0,0 +1,315 @@
...
+        const struct sol_network_link_addr *group);
+
+    /**
+     * @brief Function to be called to bind the socket to a network address.
+     *
+     * @li @c s the socket pointer (this)
+     * @li @c addr the address to bind.
+     */
+    int (*bind)(struct sol_socket *s, const struct sol_network_link_addr *addr);
+};
```
barbieri: in all of those say what needs to be returned and what happens on errors.

src/lib/comms/include/sol-socket.h
```diff
@@ -0,0 +1,315 @@
...
+ * @see sol_socket_type
+ */
+struct sol_socket {
+#ifndef SOL_NO_API_VERSION
+#define SOL_SOCKET_API_VERSION (1)  /**< compile time API version to be checked during runtime */
+    /**
+     * must match #SOL_SOCKET_API_VERSION at runtime.
+     */
+    uint16_t api_version;
+#endif
```
barbieri: hum... maybe we can skip this api_version and only check `type && type->api_version`?

src/lib/comms/include/sol-socket.h
```diff
@@ -0,0 +1,315 @@
...
+ * Destroy and release all the socket resources
+ *
+ * @param s The value got with @c sol_socket_new()
+ *
+ * @see sol_socket_new()
+ */
+void sol_socket_del(struct sol_socket *s);
+
+/**
+ * @brief Adds a function to be called when the socket had data
```
barbieri: oops, needs to fix the docs. Please refer to function specified on `new()` with config

src/lib/comms/include/sol-socket.h
```diff
@@ -0,0 +1,315 @@
...
+ * to be read.
+ *
+ * @param s The value got with @c sol_socket_new.
+ * @param on @c true to start to monitor the socket. When the socket is
+ * available to read the callback set on @c sol_socket_default_options
```
barbieri: default options?

src/lib/comms/include/sol-socket.h
```diff
@@ -0,0 +1,315 @@
...
+ * to send data.
+ *
+ * @param s The value got with @c sol_socket_new.
+ * @param on @c true to start to monitor the socket. When the socket is
+ * available to write the callback set on @c sol_socket_default_options
+ * will be called, if @c false it stops monitoring the socket to write.
+ *
+ * @return @c 0 on success, error code (always negative) otherwise.
+ *
+ * @see sol_socket_new()
```
barbieri: same here, fix docs

src/lib/comms/include/sol-socket.h
```diff
@@ -0,0 +1,315 @@
...
+    const struct sol_network_link_addr *cliaddr);
+
+/**
+ * @brief Joins a multicast group.
+ *
+ * @param s The value got with @c sol_socket_new()
+ * @param ifindex The index of the interface to be used.
+ * @param group The address of the group to join.
+ *
+ * @return @c 0 on success, error code (always negative) otherwise.
```
barbieri: which families/types can join groups?

src/lib/comms/include/sol-socket.h
```diff
@@ -0,0 +1,315 @@
...
+ * @see sol_socket_recvmsg()
+ */
+int sol_socket_sendmsg(struct sol_socket *s, const void *buf, size_t len,
+    const struct sol_network_link_addr *cliaddr);
+
+/**
+ * @brief Joins a multicast group.
+ *
+ * @param s The value got with @c sol_socket_new()
+ * @param ifindex The index of the interface to be used.
```
barbieri: where to get this index from?

src/lib/comms/include/sol-socket.h
```diff
@@ -0,0 +1,315 @@
...
+ * @ingroup Comms
+ *
+ * @brief API to create an endpoint communication using sockets.
+ *
+ * @{
+ */
+
+struct sol_socket;
+
+struct sol_socket_default_options {
```
barbieri: ah, this `_default_` here makes no sense, call it "base" or just remove the specializer.

barbieri: in sol_flow_options, it's just that. no specializers.

src/lib/comms/sol-coap.c
```diff
@@ -1746,40 +1746,30 @@ network_event(void *data, const struct sol_network_link *link, enum sol_network_
 }
 
 static struct sol_coap_server *
-sol_coap_server_new_full(enum sol_socket_type type, const struct sol_network_link_addr *servaddr)
+sol_coap_server_new_full(struct sol_socket_default_options *options, const struct sol_network_link_addr *servaddr)
```
barbieri: Move this to take the pre-built socket, this will avoid much of the hassle you have here. Add another function that creates your socket like for iotivity/oic needs (at sol-oic)

src/lib/comms/sol-socket-dtls-impl-tinydtls.c
```diff
@@ -371,7 +352,7 @@ call_user_read_cb(struct dtls_context_t *ctx, session_t *session, uint8_t *buf,
         return -EINVAL;
     }
 
-    if (socket->read.cb(socket->read.data, socket))
+    if (socket->read.cb(socket->data, socket))
```
barbieri: always store data as const and only cast before calling user.

src/lib/comms/sol-socket-impl-linux.c
barbieri: can't you get the flags from watch? no watch is nothing set.

src/lib/comms/sol-socket-impl-linux.c
```diff
@@ -437,85 +361,98 @@ sol_socket_linux_bind(struct sol_socket *socket, const struct sol_network_link_a
...
 
-const struct sol_socket_impl *
-sol_socket_get_impl(void)
+struct sol_socket *
+sol_socket_default_new(const struct sol_socket_default_options *options)
 {
-    static const struct sol_socket_impl impl = {
+    int ret;
+    struct sol_socket_linux *s;
+    int fd, socktype = SOCK_CLOEXEC | SOCK_NONBLOCK | SOCK_DGRAM;
```
barbieri: `SOCK_DGRAM` shouldn't be here. It will be a pain to remember to change this when other types are enabled.

src/lib/comms/sol-socket-impl-linux.c
```diff
@@ -437,85 +361,98 @@ sol_socket_linux_bind(struct sol_socket *socket, const struct sol_network_link_a
...
+    SOL_NULL_CHECK_GOTO(s, calloc_error);
+
+    s->base.type = &type;
+    SOL_SET_API_VERSION(s->base.api_version = SOL_SOCKET_API_VERSION; )
+    s->fd = fd;
+    s->data = options->data;
+    s->on_can_write = options->on_can_write;
+    s->on_can_read = options->on_can_read;
+
+    s->watch = sol_fd_add(fd, SOL_FD_FLAGS_NONE, on_socket_event, s);
```
barbieri: not good, do not keep a fd watch if there is no use for it.

src/lib/comms/sol-socket-dtls-impl-tinydtls.c
```diff
@@ -699,54 +675,58 @@ get_psk_info(struct dtls_context_t *ctx, const session_t *session,
...
         .event = handle_dtls_event,
         .get_psk_info = get_psk_info
     };
-    struct sol_socket_dtls *socket;
+    struct sol_socket_default_options opts = *options;
+    struct sol_socket_dtls *s;
 
     init_dtls_if_needed();
+    opts.on_can_read = read_encrypted;
+    opts.on_can_write = call_user_write_cb;
```
barbieri: this whole way of doing the dtls wrapping seems strange, you can leave this in but review it to be more natural later.

ceolin: yep, i agree with you. But i tried not change the internal things as much as possible in this first commit. I intend to review it later.

src/lib/comms/sol-socket-dtls-impl-tinydtls.c
ceolin: i've tried not changed things that were not necessary. But I'll change it before commit

src/lib/comms/sol-coap.c
ceolin: It comes in the next pr with the oic over ble.

src/lib/comms/include/sol-socket.h
ceolin: I am not sure if this options make sense for others sockets.

src/lib/comms/sol-socket-dtls-impl-tinydtls.c
barbieri: makes sense.

src/lib/comms/include/sol-socket.h
barbieri: then call it udp as you said?

src/lib/comms/include/sol-socket.h
ibriano: No, we store the address the message came from there.

ceolin:
fixes on https://github.com/solettaproject/soletta/pull/1859

PR #1841, opened 2016-04-07
Add SSE support for Soletta's HTTP nodes v3

cabelitos:
Chances since v2:
* Added commit de7362a
* Added function `sol_json_scanner_init_from_slice()` - check  7aa1e5f
* Some other minor fixes pointed by @barbieri

### DIFF
--- a/src/lib/datatypes/include/sol-str-slice.h
+++ b/src/lib/datatypes/include/sol-str-slice.h
@@ -162,9 +162,9 @@ sol_str_slice_caseeq(const struct sol_str_slice a, const struct sol_str_slice b)
  * @param haystack Slice that will be searched
  * @param needle Slice to search for in @c haystack
  *
- * @return @c true if @c needle is contained in @c haystack
+ * @return A pointer to the beginning of substring or @c NULL if not
  */
-bool sol_str_slice_contains(const struct sol_str_slice haystack, const struct sol_str_slice needle);
+char *sol_str_slice_contains(const struct sol_str_slice haystack, const struct sol_str_slice needle);
 
 /**
  * @brief Checks if @c haystack contains @c needle.
@@ -172,9 +172,13 @@ bool sol_str_slice_contains(const struct sol_str_slice haystack, const struct so
  * @param haystack Slice that will be searched
  * @param needle String to search for in @c haystack
  *
- * @return @c true if @c needle is contained in @c haystack
+ * @return A pointer to the beginning of substring or @c NULL if not
  */
-bool sol_str_slice_str_contains(const struct sol_str_slice haystack, const char *needle);
+static inline char *
+sol_str_slice_str_contains(const struct sol_str_slice haystack, const char *needle)
+{
+    return sol_str_slice_contains(haystack, SOL_STR_SLICE_STR(needle, strlen(needle)));
+}
 
 /**
  * @brief Copies the content of slice @c src into string @c dst.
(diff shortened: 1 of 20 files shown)


src/lib/datatypes/include/sol-str-slice.h
```diff
@@ -352,6 +352,28 @@ sol_str_slice_trim(struct sol_str_slice slice)
...
+ * }
+ * @endcode
+ *
+ * @param slice The slice the be splitted
+ * @param token A splitted token
+ * @param itr An iterator - It should be @c NULL on the first call.
+ * @param delim The delimiter string
+ * @return @c true if iteration should continue or @c false if iteration should be stopped.
+ */
+bool sol_str_slice_split_iterate(const struct sol_str_slice slice, struct sol_str_slice *token, const char **itr, const char *delim);
```
barbieri: I like delim being a string, should we change the other split to use it?

Also cross-link them with `@see`.

src/lib/datatypes/sol-str-slice.c
```diff
@@ -113,3 +115,40 @@ sol_str_slice_split(const struct sol_str_slice slice,
     sol_vector_clear(&v);
     return v;
 }
+
+SOL_API bool
+sol_str_slice_split_iterate(const struct sol_str_slice slice,
+    struct sol_str_slice *token, const char **itr, const char *delim)
```
barbieri: it would be better to make this a slice, and in the header you do one helper that converts str to slice. That way if delim is constant, the compiler will optimize the strlen and it will be slightly faster.

src/lib/datatypes/sol-str-slice.c
```diff
@@ -48,16 +48,18 @@ sol_str_slice_to_int(const struct sol_str_slice s, int *value)
     return 0;
 }
 
-SOL_API bool
+SOL_API char *
 sol_str_slice_contains(const struct sol_str_slice haystack, const struct sol_str_slice needle)
 {
-    return memmem(haystack.data, haystack.len, needle.data, needle.len) != NULL;
+    return memmem(haystack.data, haystack.len, needle.data, needle.len);
```
barbieri: you can do a very nice simple optimization here:

if (needle.len == 1) return memchr() else return memmem();

src/lib/datatypes/sol-str-slice.c
```diff
@@ -48,16 +48,18 @@ sol_str_slice_to_int(const struct sol_str_slice s, int *value)
...
 }
 
-SOL_API bool
+SOL_API char *
 sol_str_slice_str_contains(const struct sol_str_slice haystack, const char *needle)
 {
-    return memmem(haystack.data, haystack.len, needle, strlen(needle)) != NULL;
+    SOL_NULL_CHECK(needle, NULL);
+
+    return memmem(haystack.data, haystack.len, needle, strlen(needle));
```
barbieri: why not add these to header and convert to slice? Then all the slice conversions are done in header, where compiler will see `strlen()` and can optimize constants.

src/lib/datatypes/include/sol-str-slice.h
cabelitos: @barbieri You mean `sol_str_slice_split()` ? It already uses a string as delim

barbieri:
very good, just minor issues you can either ignore or fix later. no need to send a new PR.

on that line, I feel that if you used more slices and less strings, it would be simpler.

src/lib/datatypes/sol-str-slice.c
cabelitos: Indeed.

cabelitos:
Merged with @barbieri 's suggestions

PR #1843, opened 2016-04-07
python soletta asyncio mainloop implementation

tripzero:
Here's the initial PR for python integration.  This commit implements the soletta mainloop interface using asyncio.


cabelitos:
I didn't look at your PR carefully. However, I saw that at 4b47ef8 you add bunch of makefiles and at 30398c3 you delete them. Why committing them if you're deleting?

bindings/python/soletta/mainloop.pyx
```diff
@@ -0,0 +1,90 @@
...
+
+cdef void wrap_sol_run():
+	asyncio.get_event_loop().run_forever()
+
+cdef void *wrap_sol_timeout_add(uint32_t timeout_ms, bint (*cb)(void *), const void *data):
+	
+	@asyncio.coroutine
+	def task():
+		while cb(data):
+			yield from asyncio.sleep(timeout_ms / 1000)
```
barbieri: I'm not sure this changed in python3, but before this would be an integer division, add "1000.0" to be sure it's a float.

bindings/python/soletta/mainloop.pyx
```diff
@@ -0,0 +1,90 @@
...
+	return True
+
+cdef void *wrap_idle_add(bint (*cb)(void*), const void *data ):
+
+	@asyncio.coroutine
+	def task():
+		while True:
+			cb(data)
+			"""FIXME: this is probably not what we want"""
+			yield from asyncio.sleep(0.1)
```
barbieri: definitely no, first you need to check if cb returns False in order to stop. Then you should not sleep an amount, just run this as a low priority task (give timers, idlers and fd-handlers a chance to run).

bindings/python/soletta/mainloop.pyx
barbieri: btw in C it is fine to call "sol_timeout_del()" (or sol_idle_del()) and return False in cb(data). Do you need to do something to account for that?

Test in C using your asyncio:

```c
struct sol_timeout *t;
bool cb(void *data) {
    sol_timeout_del(t);
    return false;    
}

    struct sol_timeout *t = sol_timeout_add(100, cb, NULL);
```

barbieri:
looking good the end result. As said you don't need to add/remove the makefiles (`git rebase -i origin/master` provide a nice way to edit/drop commits)

bindings/python/soletta/mainloop.pyx
tripzero: AFAIK, asyncio doesn't have the sort of idle handler concept that exists in glib and others.  The typical pattern that is somewhat similar is to write a coroutine that yields to the event loop for some period of time.  The lower the priority, the longer the coroutine can suspend while the event loop continues with other tasks.

This allows for somewhat more deterministic behavior then what you'd get with glib.

tripzero:
Yes.  I need to squish these.

tripzero:
@cabelitos I have squished.  I'd still like to add a test.

Kconfig
```diff
@@ -246,4 +246,12 @@ config USE_NODEJS
 	default n
 	help
 		Enable Node.js bindings
+
+config USE_PYTHON
+	bool "Python bindings"
+	depends on HAVE_LIBPYTHON3 && HAVE_CYTHON3 && HAVE_PYTHON3
+	default n
```
cabelitos: I guess we should enable this be default, no?

cabelitos:
@tripzero Btw, add `bindings/python/mainloop.c` to gitignore. You should not commit that as well !

tripzero:
@cabelitos @barbieri new push.  I rearranged things a bit.  I think it's more organized now.  So far there is:
soletta.asyncio - soletta mainloop implementation with asyncio
soletta.tests - tests for the modules.

You can make install, however, on Ubuntu, it's installing to /usr/lib/python3.4/site-packages, which is outside ubuntu's python3 default path.  This is because I'm installing with --prefix=build/sysroot/... which overrides the default path of dist-packages.  This will be the same for any debian-based distro.

I'm not sure how to fix that yet, but a quick fix is to just sudo cp -r /usr/lib/python3.4/site-packages/soletta /usr/lib/python3.4/dist-packages

After that, you can python3 -c "import soletta.asyncio" and watch the pretty fireworks (I'll fix that soon).

I also fixed idle_add to check the callback for true/false.

Kconfig
tripzero: I'm okay with that.  @barbieri ?

barbieri: sure. People can always disable if they do not want to.

it would be nice to get the semaphore-ci to build it, then it needs the deps installed.

bindings/python/run_tests.py
```diff
@@ -0,0 +1,4 @@
+from soletta.tests import run_tests
+
+if __name__ == "__main__":
+	run_tests()
```
barbieri: missing newline at the end of file

bindings/python/setup.py
```diff
@@ -0,0 +1,67 @@
+"setup.py"
```
barbieri: add a `#!/usr/bin/python3` and flag this as executable, it may help if people want to manually execute this

bindings/python/soletta/asyncio/__init__.py
```diff
@@ -0,0 +1 @@
+from soletta.asyncio._soletta_mainloop import *
```
barbieri: missing newline.

bindings/python/soletta/tests/__init__.py
```diff
@@ -0,0 +1 @@
+from _soletta_tests import *
```
barbieri: missing newline

bindings/python/soletta/tests/_soletta_tests.pyx
```diff
@@ -0,0 +1,77 @@
...
+def add_test(func):
+	tests[func] = False
+
+cdef bint cb(void* data):
+	cdef bint success = dref(<bint*>data)
+	success = True
+	return False
+
+cdef bint cb2(void* data):
+		cdef int* d = <int*>data
```
barbieri: messed up indentation

bindings/python/soletta/tests/_soletta_tests.pyx
```diff
@@ -0,0 +1,77 @@
...
+	def post():
+		return success
+	
+
+def test_timeout_del():
+	cdef sol_timeout* t
+
+	cdef int callbackcounts = 0
+
+	t = sol_timeout_add(100, cb2, <void*>&callbackcounts)
```
barbieri: if you return true, why is the expected count == 1? It should be more.

bindings/python/soletta/tests/_soletta_tests.pyx
```diff
@@ -0,0 +1,77 @@
...
+
+	def test_timeout_del_post():
+		return callbackcounts == 1
+	
+	return test_timeout_del_post
+
+def run_tests():
+
+	import asyncio
+
```
barbieri: I miss a test that deletes the `sol_timeout_del()` from the callback and returns `False`, that should not double-free.

And of course review that returning true is running the CB, it seems not given `callbackcounts == 1` above. My hint there is from `cb2` you count and if ` == expected: return False`, this way you let it run for a while and see what happens.

It is also missing multiple tasks running at the same time.

barbieri:
@tripzero as for the installation bug, from what you describe it is the `site-packages` x `dist-packages`, right? I do not see you specifying that anywhere, where are you getting that?

Also, consider checking `import site; site.getsitepackages()`

tripzero:
>@tripzero as for the installation bug, from what you describe it is the site-packages x dist-packages, right? I do not see you specifying that anywhere, where are you getting that?

Debian patched python and distutils.  If you override the install root, it assumes you want to use site-packages.  The patched version of python debian ships with does not look in site-packages.  site.getsitepackages() returns dist-packages.

This is an annoying issue on debian/debian-based distros, but it does not affect other distros that I know of.

tripzero:
closing this for new one

PR #1846, opened 2016-04-08
Node.js bindings: Simplify code for locating the repository root

gabrielschulhof:
Use ```getRoot()``` from ```bindings``` instead of custom code.


bindings/nodejs/tests/closestSoletta.js
```diff
@@ -1,18 +0,0 @@
...
-var path = require( "path" );
-var dirToCheck;
-
-// Find the closest instance of soletta via its package.json
-function closestSoletta( startDir ) {
-	for ( dirToCheck = startDir; dirToCheck !== "/";
-			dirToCheck = path.resolve( dirToCheck, ".." ) ) {
-		try {
-			if ( JSON.parse( fs.readFileSync(
-					path.join( dirToCheck, "package.json" ), "utf8" ) ).name == "soletta" ) {
```
edersondisouza: Here you checked if `package.json` was named `soletta`, however, new method you are using doesn't check anything like that. Is that ok? If so, r+.

gabrielschulhof: You're right. The check with which I'm replacing the existing check is less strict. However, since this is for testing anyway, I think it's safe to assume that we're not gonna move the test suite info a new npm package anytime soon. If that should happen, more changes to the test suite are need. So yes, this is OK.

PR #1847, opened 2016-04-08

nagineni:

### DIFF
--- a/bindings/nodejs/src/functions/spi.cc
+++ b/bindings/nodejs/src/functions/spi.cc
@@ -1,0 +1,158 @@
+/*
+ * This file is part of the Soletta Project
+ *
+ * Copyright (C) 2015 Intel Corporation. All rights reserved.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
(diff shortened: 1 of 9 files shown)


nagineni:
@gabrielschulhof @barbieri could you please review this?

bindings/nodejs/src/functions/spi.cc
```diff
@@ -0,0 +1,158 @@
...
+        return;
+    }
+    memset(rxBuffer, 0, length * sizeof(uint8_t));
+    Nan::Callback *callback =
+        new Nan::Callback(Local<Function>::Cast(info[2]));
+    bool returnValue =
+        sol_spi_transfer(spi, txBuffer, rxBuffer, length,
+            sol_spi_transfer_cb, callback);
+
+    if (!returnValue || !hijack_ref()) {
```
gabrielschulhof: You should make sure that hijacking has not failed before attempting to transfer the data.

gabrielschulhof: ... and you should also not set the return value if it does fail.

nagineni: Thanks for reviewing @gabrielschulhof . I will create a new PR with the fix.

PR #1848, opened 2016-04-08
sol-connman: Add sol-connman basic API

wzhen12:
so-connman is for implementing connection manager features.
The APIs should be implemented in Zephyr, connman and so on.

The patches are for implementing sol-connman basic APIs to connman.
the connman D-Bus APIs are wrapped in the so-connman APIs.
The user can use the so-connman APIs to control connman.


src/lib/comms/include/sol-connman.h
```diff
@@ -0,0 +1,285 @@
...
+    SOL_CONNMAN_SERVICE_STATE_ONLINE        = 5,
+    SOL_CONNMAN_SERVICE_STATE_DISCONNECT    = 6,
+    SOL_CONNMAN_SERVICE_STATE_FAILURE       = 7,
+    SOL_CONNMAN_SERVICE_STATE_REMOVE        = 8,
+};
+
+#define SOL_CONNMAN_SERVICE_TYPE_ETHERNET   "ethernet"  /**< @brief ethernet service type string */
+#define SOL_CONNMAN_SERVICE_TYPE_WIFI       "wifi"      /**< @brief wifi service type string */
+#define SOL_CONNMAN_SERVICE_TYPE_BLUETOOTH  "bluetooth" /**< @brief bluetooth service type string */
+#define SOL_CONNMAN_SERVICE_TYPE_CELLULAR   "celluar"   /**< @brief bluetooth service type string */
```
ibriano: Typo here, and copy paste docs.

src/lib/comms/sol-connman-impl-connman.c.orig
```diff
@@ -0,0 +1,804 @@
+/*
```
ibriano: I don't think you wanted to add this file here.

src/lib/comms/sol-connman-impl-connman.c
```diff
@@ -0,0 +1,206 @@
...
+
+static void
+_free_connman_service(struct sol_connman_service *service)
+{
+    if (!service)
+        return;
+
+    service->pending.slot = sd_bus_slot_unref(service->pending.slot);
+
+    if (service->path)
```
edersondisouza: You don't need to check `if (x)` before `free(x);`. `free` handles `NULL` input.

src/lib/comms/sol-connman-impl-connman.c
```diff
@@ -0,0 +1,906 @@
...
+};
+
+struct dbus_monitor_callback {
+    void (*cb)(void *data, const struct sol_connman_service *service);
+    void *data;
+};
+
+struct ctx {
+    struct sol_vector service_vector;
+    struct dbus_monitor_callback monitor_callback;
```
ibriano: Should be a vector as well, so more than one callback can be added.

src/lib/comms/include/sol-connman.h
```diff
@@ -0,0 +1,285 @@
...
+ * @return 0 on success, -errno on failure.
+ */
+int sol_connman_service_disconnect(struct sol_connman_service *service);
+
+/**
+ * @brief Adds a monitor for the updated connman services
+ *
+ * To monitor the property of services, it gives information
+ * about the services. The callback is used to provide the information.
+ * sol_connman_add_service_monitor will call sol_connman_lazy_init
```
ibriano: No need to document implementation details. The API documentation should tell the user all the information he/she needs to use these functions properly, not how they work under the hood.

src/lib/comms/sol-connman-impl-connman.c
```diff
@@ -0,0 +1,906 @@
...
+
+int sol_connman_init(void);
+void sol_connman_shutdown(void);
+
+struct sol_connman_pending {
+    sd_bus_slot *slot;
+};
+
+struct sol_connman_service {
+    struct sol_connman_pending pending;
```
ibriano: This doesn't allow to have more than one request ongoing at the same time.

src/lib/comms/sol-connman-impl-connman.c
```diff
@@ -0,0 +1,906 @@
...
+sol_connman_init(void)
+{
+    _ctx.connman = NULL;
+    _ctx.state_slot.slot = NULL;
+    _ctx.manager_slot.slot = NULL;
+    _ctx.service_slot.slot = NULL;
+    _ctx.connman_state = SOL_CONNMAN_STATE_UNKNOWN;
+    _ctx.properties_changed = NULL;
+    _ctx.monitor_callback.cb = NULL;
+    _ctx.monitor_callback.data = NULL;
```
ibriano: These are largely unnecessary here, and considering init_lazy and init_shutdown may end up called more than once, if you do need to ensure that these variables are set to NULL on init, you need to set them after shutdown.

src/lib/comms/sol-connman-impl-connman.c
```diff
@@ -148,10 +150,101 @@ sol_connman_service_get_strength(const struct sol_connman_service *service)
...
+
+    if (state) {
+        if (strcmp(state, "online") == 0)
+            _ctx.connman_state = SOL_CONNMAN_STATE_ONLINE;
+        else if (strcmp(state, "ready") == 0)
+            _ctx.connman_state = SOL_CONNMAN_STATE_READY;
+        else if (!strcmp(state, "idle"))
+            _ctx.connman_state = SOL_CONNMAN_STATE_IDLE;
+        else if (!strcmp(state, "offline"))
+            _ctx.connman_state = SOL_CONNMAN_STATE_OFFLINE;
```
edersondisouza: Minor nitpick: be consistent and use ` == 0` (or `!strcmp`) in all comparisons.

src/lib/comms/Makefile
```diff
@@ -5,6 +5,11 @@ obj-networking-$(NETWORK) := \
     sol-comms.o \
     sol-socket.o
 
+obj-networking-$(CONNMAN) += \
+    sol-connman-impl-connman.o
+obj-networking-$(CONNMAN)-extra-cflags += $(SYSTEMD_CFLAGS)
+obj-networking-$(CONNMAN)-extra-ldflags += $(SYSTEMD_LDFLAGS)
+
```
vcgomes: Seems the header `sol-connman.h` is not being installed.

src/lib/comms/include/sol-connman.h
```diff
@@ -0,0 +1,285 @@
...
+ *
+ * This function gets the network address for the connman service,
+ * after the service is to give by service monitor.
+ *
+ * @param service The service structure which the network address is desired
+ * @param family which type of the network address is desired (ipv4 or ipv6).
+ *
+ * @return A point to sol_network_link_addr created on success, NULL on failure.
+ */
+struct sol_network_link_addr *sol_connman_service_get_network_address(
```
vcgomes: I am considering if this should return a list (or a vector) of addresses.

src/lib/comms/Makefile
vcgomes: I did try to build, got this error:
```
make: *** No rule to make target 'sol-connman.h', needed by 'build/stage/lib/comms/sol-connman-impl-connman.o'.  Stop.
```

src/lib/comms/sol-connman-impl-connman.c
```diff
@@ -0,0 +1,906 @@
...
+sol_connman_add_service_monitor(
+    void (*cb)(void *data, const struct sol_connman_service *service),
+    const void *data)
+{
+    int r;
+
+    r = sol_connman_init_lazy();
+    SOL_INT_CHECK_GOTO(r, < 0, fail_init);
+
+    r = dbus_connection_get_manager_properties();
```
vcgomes: What about if the user calls `sol_connman_add_service_monitor()` twice, it will send all these messages and try to parse them again. And if the user doesn't add any monitor, he/she will never get any meaningful information.

What I propose, is that you add watches for `Manager.ServicesChanged`, `Manager.PropertiesChanged` and each service `Service.PropertiesChanged`, and `call 'GetServices()' (did I miss anything?) during the module initialization. So if the user later calls any method, even if there's no monitor added, the user will get a valid and meaningful information. What do you think?

src/lib/comms/sol-connman-impl-connman.c
```diff
@@ -0,0 +1,906 @@
...
+    _ctx.connman_state = SOL_CONNMAN_STATE_UNKNOWN;
+    _ctx.properties_changed = NULL;
+    _ctx.monitor_callback.cb = NULL;
+    _ctx.monitor_callback.data = NULL;
+
+    return 0;
+}
+
+void
+sol_connman_shutdown(void)
```
vcgomes: If the user calls shutdown without removing the monitor, a lot of resources will be leaked.

src/lib/comms/sol-connman-impl-connman.c
```diff
@@ -0,0 +1,906 @@
...
+                     else if (!strcmp(str, "association"))
+                         service->state = SOL_CONNMAN_SERVICE_STATE_ASSOCIATION;
+                     else if (!strcmp(str, "configuration"))
+                         service->state = SOL_CONNMAN_SERVICE_STATE_CONFIGURATION;
+                     else if (!strcmp(str, "disconnect"))
+                         service->state = SOL_CONNMAN_SERVICE_STATE_DISCONNECT;
+                     else if (!strcmp(str, "idle"))
+                         service->state = SOL_CONNMAN_SERVICE_STATE_IDLE;
+                     else if (!strcmp(str, "failure"))
+                         service->state = SOL_CONNMAN_SERVICE_STATE_FAILURE;
```
vcgomes: Take a look at `sol-str-table`, I guess it can be used here.

src/lib/comms/sol-connman-impl-connman.c
```diff
@@ -0,0 +1,906 @@
...
+
+                 r = sd_bus_message_enter_container(m, SD_BUS_TYPE_VARIANT, "s");
+                 SOL_INT_CHECK(r, < 0, r);
+
+                 r = sd_bus_message_read_basic(m, SD_BUS_TYPE_STRING, &type);
+                 SOL_INT_CHECK(r, < 0, r);
+
+                 if (service->type)
+                     free(service->type);
+                 service->type = strdup(type);
```
vcgomes: Please check if `strdup()` returned a valid pointer.

src/lib/comms/sol-connman-impl-connman.c
```diff
@@ -0,0 +1,906 @@
...
+#include "sol-connman.h"
+
+SOL_LOG_INTERNAL_DECLARE(_sol_connman_log_domain, "connman");
+
+int sol_connman_init(void);
+void sol_connman_shutdown(void);
+
+struct sol_connman_pending {
+    sd_bus_slot *slot;
+};
```
vcgomes: Why not using a `sd_bus_slot` directly?

src/lib/comms/sol-connman-impl-connman.c
```diff
@@ -0,0 +1,906 @@
...
+            is_exist = true;
+            break;
+        }
+    }
+
+    if (is_exist == false) {
+        service = sol_vector_append(&_ctx.service_vector);
+        SOL_NULL_CHECK(service, -EINVAL);
+
+        _init_connman_service(service);
```
vcgomes: No need to call this, `sol_vector_append()` already `memset`s the memory to zero.

src/lib/comms/sol-connman-impl-connman.c
```diff
@@ -0,0 +1,906 @@
...
+
+                 r = sd_bus_message_enter_container(m, SD_BUS_TYPE_VARIANT, "s");
+                 SOL_INT_CHECK(r, < 0, r);
+
+                 r = sd_bus_message_read_basic(m, SD_BUS_TYPE_STRING, &name);
+                 SOL_INT_CHECK(r, < 0, r);
+
+                 if (service->name)
+                     free(service->name);
+                 service->name = strdup(name);
```
vcgomes: `strdup` can fail.

src/lib/comms/sol-connman-impl-connman.c
```diff
@@ -153,9 +154,34 @@ sol_connman_get_state(void)
...
+    int r;
+    struct sol_connman_pending *pending = &_ctx.state_slot;
+    sd_bus *bus = sol_bus_client_get_bus(_ctx.connman);
+
+    SOL_NULL_CHECK(bus, -EINVAL);
+
+    r = sd_bus_call_method_async(bus, &pending->slot, "net.connman", "/",
+        "net.connman.Manager", "SetProperty", _set_state_property_changed,
+        pending, "sv", "OfflineMode", "b", enabled);
+    SOL_INT_CHECK(r, < 0, -EINVAL);
```
cabelitos: return the R

src/lib/comms/sol-connman-impl-connman.c
```diff
@@ -153,9 +154,34 @@ sol_connman_get_state(void)
...
+    sd_bus_error *ret_error)
+{
+    struct sol_connman_pending *pending = userdata;
+
+    pending->slot = sd_bus_slot_unref(pending->slot);
+
+    if (sol_bus_log_callback(reply, userdata, ret_error) < 0)
+        return -EINVAL;
+
+    return 0;
```
cabelitos: `return sol_bus_log_callback()` ?

cabelitos:
Review all your `SOL_INT_CHECK()`, try to do like `SOL_INT_CHECK(r, <0 , r)` and not `SOL_INT_CHECK(r, < 0, -EINVAL)`

src/lib/comms/Kconfig
```diff
@@ -16,6 +16,16 @@ config DTLS
 
 	    If unsure, say N.
 
+config CONNMAN
+        bool "connection manager interfaces"
```
cabelitos: I would keep it simple and just call connman here

src/lib/comms/sol-connman-impl-connman.c
```diff
@@ -899,5 +899,8 @@ sol_connman_del_service_monitor(
 SOL_API int
 sol_connman_get_service_vector(struct sol_vector *vector)
 {
+    vector = &_ctx.service_vector;
+    SOL_NULL_CHECK(vector, -EINVAL);
```
cabelitos: This is wrong.
for two reasons:
First it should be `struct sol_vector **vector`
So you would do something like: `*vector = &_ctx.service_vector;`

You also need to invert the order here, NULL_CHECK must be placed before the attribution

src/lib/comms/include/sol-connman.h
wzhen12: Thanks. I will modify it.

src/lib/comms/sol-connman-impl-connman.c.orig
wzhen12: Sorry, delete it right now.

src/lib/comms/sol-connman-impl-connman.c
wzhen12: If it not needed int soletta, I will remove it.

src/lib/comms/sol-connman-impl-connman.c
wzhen12: To our design, add monitor can be added once successfully.
Therefore, only one callback exists.

src/lib/comms/include/sol-connman.h
wzhen12: Okay, modify it.

src/lib/comms/sol-connman-impl-connman.c
wzhen12: Yes.

src/lib/comms/sol-connman-impl-connman.c
wzhen12: Remove it.

src/lib/comms/sol-connman-impl-connman.c
wzhen12: Okay, Remove set to NULL on init

src/lib/comms/sol-connman-impl-connman.c
wzhen12: I will modify it.

src/lib/comms/Makefile
wzhen12: Sorry, add it right now

src/lib/comms/sol-connman-impl-connman.c
wzhen12: It is necessary that  the user calls sol_connman_add_service_monitor() before any operation.
If not calls sol_connman_add_service_monitor(), the other operation will be failure.

If the user calls sol_connman_add_service_monitor() twice, the first call is successful.
`call 'GetServices()' has been done in sol_connman_add_service_monitor() and not in  the module initialization.

src/lib/comms/sol-connman-impl-connman.c
wzhen12: I will try to modify it.

src/lib/comms/sol-connman-impl-connman.c
wzhen12: Okay.

src/lib/comms/sol-connman-impl-connman.c
wzhen12: Okay.

src/lib/comms/sol-connman-impl-connman.c
wzhen12: I will add the result check.

src/lib/comms/sol-connman-impl-connman.c
wzhen12: Okay.

src/lib/comms/sol-connman-impl-connman.c
wzhen12: Okay

src/lib/comms/sol-connman-impl-connman.c
wzhen12: Yes, sorry, it is lower error.

wzhen12:
I will submit the new version soon.

wzhen12:
Hi all,

I submit the new patches.

It is in https://github.com/wzhen12/soletta/commits/master

Can you help reopen the PR? 
So that the patches can continue to be reviewed.

Thanks.

wzhen12:
1. In the new patches, almost  all questions have been answered.

2. A few questions from you still need continue to be investigated.
After then, I will give you some response.
Thanks.

wzhen12:
Please check the patches again.

I will continue to try answer the related questions.
Then the new version patches will be submitted again.

vcgomes:
This suggestion may be a little late, but how about renaming this to `sol-netctl` or something like it? Even `connman` being a generic name, it may cause confusion.

wzhen12:
>This suggestion may be a little late, but how about renaming this to sol-netctl or something like it? Even connman being a generic name, it may cause confusion.

If it is not necessary, sol-connman should be ok for me.
It is a generic name, however, the first connman means that connection manager.

wzhen12:
1. sol-stro-table is used to replace "if .. else if .. else if .." in the new patches.
2. sd_bus_slot replace "struct sol_connman_pending" in the new patches.
3. add the test cases.
4. optimize  the Kconfig of connman. 

Please review it again. Thanks.

barbieri:
I'll take a look at the new series. But please use the contribution
pattern: open new PR v2,v3... And close the old, mentioning the new PR
number in the comments. This way we get the notifications about new code
coming in

Also consider the name Vinícus suggested: netctl, at least we can remove
the "connman-impl-connman" confusion. Everything else can stay the same.

Em terça-feira, 12 de abril de 2016, wzhen12 <notifications@github.com
<javascript:_e(%7B%7D,'cvml','notifications@github.com');>> escreveu:

>
>    1. sol-stro-table is used to replace "if .. else if .. else if .." in
>    the new patches.
>    2. sd_bus_slot replace "struct sol_connman_pending" in the new patches.
>    3. add the test cases.
>    4. optimize the Kconfig of connman.
>
> Please review it again. Thanks.
>
> —
> You are receiving this because you are subscribed to this thread.
> <https://github.com/solettaproject/soletta/pull/1848#issuecomment-208799596>
>


-- 
Gustavo Sverzut Barbieri
--------------------------------------
Mobile: +55 (19) 99225-2202
Contact: http://www.gustavobarbieri.com.br/contact

wzhen12:
@barbieri 
Okay, I will submit the new PR after you review.
I will modify the file name in the new PR too.

At the same time, I want to provide a new API of sol-connman too.
When the services are updated, the user callback will be invoked.
However, the user doesn't know which items of the services have been updated.

I want to add a new API of sol-connman.(such as sol_connman_service_get_flag)
So that the user can know which items of services have been changed, doesn't need to check all items of services.

some suggestion?

src/lib/comms/include/sol-connman.h
```diff
@@ -0,0 +1,285 @@
...
+ * One of these must be chosen to set the state of service,
+ * with sol_connman_service_get_state()
+ *
+ * @note SOL_CONNMAN_SERVICE_STATE_REMOVE is used to show
+ * the service has been removed. When it is notified,
+ * the service has been removed in the system.
+ */
+enum sol_connman_service_state {
+    SOL_CONNMAN_SERVICE_STATE_UNKNOWN       = 0,
+    SOL_CONNMAN_SERVICE_STATE_IDLE          = 1,
```
barbieri: no need to define all the values, only define the base (=0) and let the compiler do the rest.

src/lib/comms/include/sol-connman.h
barbieri: In sol-network.h we have `struct sol_network_link`, that contains a list of addresses and a flag, both look meaningful here, so maybe get that? Also the pattern could be to fill a given pointer instead of allocating and returning it, then return -errno.

src/lib/comms/include/sol-connman.h
```diff
@@ -0,0 +1,285 @@
...
+ * Since the connman function is asynchronous, the call return is not connection error,
+ * but just some dispatching/immediate error. The actual state change will be notified
+ * via sol_connman_add_service_monitor() callbacks.
+ * After the actual state change is notified via sol_connman_add_service_monitor()
+ * callbacks, sol_connman_service_get_call_result can be used to get the actual state.
+ *
+ * @return TRUE on success, FALSE on failure
+ */
+bool sol_connman_service_get_call_result(
+    const struct sol_connman_service *service);
```
barbieri: no need to expose this, you may have a monitor for error.

src/lib/comms/include/sol-connman.h
```diff
@@ -0,0 +1,285 @@
...
+
+/**
+ * @brief Gets the global connection state of system
+ *
+ * This function gets the global connection state for system.
+ *
+ * @return The state of the global connection state on success,
+ * SOL_CONNMAN_STATE_UNKNOWN on failure
+ */
+enum sol_connman_state sol_connman_get_state(void);
```
barbieri: we need a monitor for the base properties (State, Offline...). We can do a single monitor that says "something changed", and then you can get the new value if you're interested in it. Then you should also refer `@see sol_connman_...` and say that the pattern is to add the monitor BEFORE call get_state()/get_offline()...

src/lib/comms/include/sol-connman.h
```diff
@@ -0,0 +1,285 @@
...
+/**
+ * @brief Gets the connman services vector.
+ *
+ * The connman services vector can be given via sol_connman_get_vector.
+ *
+ * @param vector The retrieved content
+ *
+ * @return 0 if the content was successfully retrieved, error code (always negative) otherwise.
+ */
+int sol_connman_get_service_vector(struct sol_vector **vector);
```
barbieri: make this signature similar to sol-network.h `sol_network_get_available_links()`, see https://github.com/solettaproject/soletta/blob/master/src/lib/comms/include/sol-network.h#L273

Also say the vector is the last-known and more may be added/removed dynamically, then the pattern is to add a service monitor BEFORE calling this function.

src/lib/comms/sol-connman-impl-connman.c
```diff
@@ -0,0 +1,910 @@
...
+#define SOL_LOG_DOMAIN &_sol_connman_log_domain
+#include <sol-log-internal.h>
+
+#include "sol-str-table.h"
+#include "sol-str-slice.h"
+#include "sol-util-internal.h"
+#include "sol-bus.h"
+#include "sol-connman.h"
+
+SOL_LOG_INTERNAL_DECLARE(_sol_connman_log_domain, "connman");
```
barbieri: you can use the static version since this is restricted to one file.

src/lib/comms/sol-connman-impl-connman.c
```diff
@@ -0,0 +1,910 @@
...
+
+static struct ctx _ctx;
+
+static void
+call_monitor_callback(struct sol_connman_service *service)
+{
+    struct dbus_monitor_callback *monitor;
+    uint16_t i;
+
+    SOL_VECTOR_FOREACH_IDX (&_ctx.monitor_vector, monitor, i) {
```
barbieri: use our `sol-monitor.h`, it will do the right thing. In this case things can go wrong if from `monitor->cb` you remove the next element in `monitor_vector`. That's why we added the monitor helpers :-)

src/lib/comms/sol-connman-impl-connman.c
```diff
@@ -0,0 +1,910 @@
...
+{
+    if (!service)
+        return;
+
+    service->slot = sd_bus_slot_unref(service->slot);
+
+    free(service->path);
+    free(service->type);
+    free(service->link_addr);
+    free(service->link_addr6);
```
barbieri: you will need a `free(service)` here and change the vector to a "ptr_vector". The reason is that the service handles will be modified once you append/remove services and users have no way to know of that, this will result in them accessing invalid pointers.

src/lib/comms/sol-connman-impl-connman.c
```diff
@@ -0,0 +1,910 @@
...
+}
+
+SOL_API bool
+sol_connman_service_get_call_result(const struct sol_connman_service *service)
+{
+    SOL_NULL_CHECK(service, false);
+
+    return service->is_call_success;
+}
```
barbieri: as said, remove, there is no need to do this exposure. You may opt to add a `sol_connman_add_error_monitor()` and that would get called when an error happens, it may be connect to a service, set offline mode (global), etc.

src/lib/comms/sol-connman-impl-connman.c
```diff
@@ -0,0 +1,910 @@
...
+SOL_API struct sol_network_link_addr *
+sol_connman_service_get_network_address(const struct sol_connman_service *service,
+    enum sol_network_family family)
+{
+    SOL_NULL_CHECK(service, NULL);
+
+    if (family == SOL_NETWORK_FAMILY_INET)
+        return service->link_addr;
+    else if (family == SOL_NETWORK_FAMILY_INET6)
+        return service->link_addr6;
```
barbieri: maybe return the link, it can contain all the addresses in all families.

src/lib/comms/sol-connman-impl-connman.c
```diff
@@ -0,0 +1,910 @@
...
+    r = sd_bus_message_enter_container(m, SD_BUS_TYPE_ARRAY, "{sv}");
+    SOL_INT_CHECK(r, < 0, r);
+
+    do {
+        r = sd_bus_message_enter_container(m, SD_BUS_TYPE_DICT_ENTRY, "sv");
+        SOL_INT_CHECK_GOTO(r, < 1, end);
+
+        r = sd_bus_message_read_basic(m, SD_BUS_TYPE_STRING, &str);
+        SOL_INT_CHECK(r, < 0, r);
+        if (strcmp(str, "Address") == 0) {
```
barbieri: we have `streq()` to make it clear and avoid typing

src/lib/comms/sol-connman-impl-connman.c
```diff
@@ -0,0 +1,910 @@
...
+}
+
+static void
+get_service_ipv4(sd_bus_message *m, struct sol_connman_service *service)
+{
+    if (!service)
+        return;
+
+    if (!service->link_addr) {
+        service->link_addr = malloc(sizeof(*service->link_addr));
```
barbieri: why malloc it? Why not keep it inside the struct service?

src/lib/comms/sol-connman-impl-connman.c
```diff
@@ -0,0 +1,910 @@
...
+}
+
+static void
+get_service_ipv6(sd_bus_message *m, struct sol_connman_service *service)
+{
+    if (!service)
+        return;
+
+    if (!service->link_addr6) {
+        service->link_addr6 = malloc(sizeof(*service->link_addr6));
```
barbieri: same here, no need to malloc. And if you convert to keep a sol_network_link, it's basically append to the vector (that will malloc for you :-D)

src/lib/comms/sol-connman-impl-connman.c
```diff
@@ -0,0 +1,910 @@
...
+remove_services(const char *path)
+{
+    struct sol_connman_service *service;
+    uint16_t i;
+
+    if (!path)
+        return;
+
+    SOL_VECTOR_FOREACH_IDX (&_ctx.service_vector, service, i) {
+        if (strcmp(service->path, path) == 0) {
```
barbieri: `streq()`

src/lib/comms/sol-connman-impl-connman.c
```diff
@@ -0,0 +1,910 @@
...
+static int
+get_services_properties(sd_bus_message *m, const char *path)
+{
+    int r;
+    struct sol_connman_service *service;
+    uint16_t i;
+    bool is_exist = false;
+
+    SOL_VECTOR_FOREACH_IDX (&_ctx.service_vector, service, i) {
+        if (strcmp(service->path, path) == 0) {
```
barbieri: `streq()`, and this is used more than once, so what about a helper function:

```c
static struct sol_connman_service *
find_service_by_path(const char *path) { ... }
```

src/lib/comms/sol-connman-impl-connman.c
```diff
@@ -0,0 +1,910 @@
...
+            break;
+        }
+    }
+
+    if (is_exist == false) {
+        service = sol_vector_append(&_ctx.service_vector);
+        SOL_NULL_CHECK(service, -EINVAL);
+
+        service->path = strdup(path);
+        SOL_NULL_CHECK(service->path, -ENOMEM);
```
barbieri: here you leak the element in the array, you need to `SOL_NULL_CHECK_GOTO()` and `sol_vector_del_last()`

src/lib/comms/sol-connman-impl-connman.c
```diff
@@ -0,0 +1,910 @@
...
+
+    do {
+        char *str;
+
+        r = sd_bus_message_enter_container(m, SD_BUS_TYPE_DICT_ENTRY, "sv");
+        SOL_INT_CHECK_GOTO(r, < 1, end);
+
+        r = sd_bus_message_read_basic(m, SD_BUS_TYPE_STRING, &str);
+        SOL_INT_CHECK(r, < 0, r);
+        if (strcmp(str, "Name") == 0) {
```
barbieri: `streq()`

src/lib/comms/sol-connman-impl-connman.c
```diff
@@ -0,0 +1,910 @@
...
+
+            r = sd_bus_message_enter_container(m, SD_BUS_TYPE_VARIANT, "s");
+            SOL_INT_CHECK(r, < 0, r);
+
+            r = sd_bus_message_read_basic(m, SD_BUS_TYPE_STRING, &name);
+            SOL_INT_CHECK(r, < 0, r);
+
+            if (service->name)
+                free(service->name);
+            service->name = strdup(name);
```
barbieri: https://github.com/solettaproject/soletta/blob/master/src/shared/include/sol-util.h#L425 `sol_util_replace_str_if_changed()`

src/lib/comms/sol-connman-impl-connman.c
```diff
@@ -0,0 +1,910 @@
...
+            service->type = strdup(type);
+
+            r = sd_bus_message_exit_container(m);
+            SOL_INT_CHECK(r, < 0, r);
+            SOL_NULL_CHECK(service->type, -ENOMEM);
+        } else if (strcmp(str, "IPv4") == 0) {
+            r = sd_bus_message_enter_container(m, SD_BUS_TYPE_VARIANT, "a{sv}");
+            SOL_INT_CHECK(r, < 0, r);
+
+            get_service_ipv4(m, service);
```
barbieri: I'm wondering if we should store the gateway address as well. What about netmask?

Do you know if these would be used for an application?

src/lib/comms/sol-connman-impl-connman.c
```diff
@@ -0,0 +1,910 @@
...
+            SOL_INT_CHECK(r, < 0, r);
+        } else if (strcmp(str, "IPv6") == 0) {
+            r = sd_bus_message_enter_container(m, SD_BUS_TYPE_VARIANT, "a{sv}");
+            SOL_INT_CHECK(r, < 0, r);
+
+            get_service_ipv6(m, service);
+
+            r = sd_bus_message_exit_container(m);
+            SOL_INT_CHECK(r, < 0, r);
+        }  else {
```
barbieri: could you log with `SOL_DBG("ignored service property: %s", str);`?

src/lib/comms/sol-connman-impl-connman.c
barbieri: why no `sol_str_table` here?

src/lib/comms/sol-connman-impl-connman.c
```diff
@@ -0,0 +1,910 @@
...
+            r = sd_bus_message_skip(m, NULL);
+            SOL_INT_CHECK(r, < 0, r);
+        }
+
```
barbieri: shouldn't you call the service monitor here to say the service changed?

src/lib/comms/sol-connman-impl-connman.c
```diff
@@ -0,0 +1,910 @@
...
+        SOL_INT_CHECK(r, < 0, r);
+
+        if (strcmp(str, "State") == 0) {
+            r = get_manager_properties(m);
+            SOL_INT_CHECK(r, < 0, r);
+        } else {
+            r = sd_bus_message_skip(m, "v");
+            SOL_INT_CHECK(r, < 0, r);
+        }
+
```
barbieri: as said in the .h, here you would inform the generic monitor to say something changed (it could be state or offline.

also log what you are skipping `SOL_DBG()` and AFAIU you should get the OfflineMode.

src/lib/comms/sol-connman-impl-connman.c
```diff
@@ -0,0 +1,910 @@
...
+        &_ctx, "sv", "OfflineMode", "b", enabled);
+    SOL_INT_CHECK(r, < 0, r);
+
+    return 0;
+}
+
+SOL_API bool
+sol_connman_get_offline(void)
+{
+    if (_ctx.connman_state != SOL_CONNMAN_STATE_OFFLINE)
```
barbieri: no... this should be the `OfflineMode` property. I understand this is confusing, so should we rename the `get_offline()` and `set_offline()` to `set_radios_offline()`? Or `set_airplane_mode()`?

src/lib/comms/sol-connman-impl-connman.c
```diff
@@ -0,0 +1,910 @@
...
+    struct sol_connman_service *service = userdata;
+
+    SOL_NULL_CHECK(service, -EINVAL);
+
+    service->slot = sd_bus_slot_unref(service->slot);
+
+    if (sol_bus_log_callback(reply, userdata, ret_error) < 0)
+        service->is_call_success = false;
+    else
+        service->is_call_success = true;
```
barbieri: hint: in these cases it's okay to write: `service->is_call_success = sol_bus_log_callback(reply, userdata, ret_error) >= 0`

But likely you do not need this, just inform the errors using an error monitor.

src/lib/comms/sol-connman-impl-connman.c
```diff
@@ -0,0 +1,910 @@
...
+
+    SOL_NULL_CHECK(bus, -EINVAL);
+    SOL_NULL_CHECK(service, -EINVAL);
+    SOL_NULL_CHECK(service->path, -EINVAL);
+
+    service->is_call_success = false;
+
+    r = sd_bus_call_method_async(bus, &service->slot, "net.connman", service->path,
+        "net.connman.Service", "Connect", _service_connect, service, NULL);
+    SOL_INT_CHECK(r, < 0, r);
```
barbieri: it's okay to just `return sd_bus_call_method...` no need to `SOL_INT_CHECK()`

src/lib/comms/sol-connman-impl-connman.c
```diff
@@ -0,0 +1,910 @@
...
+
+    SOL_VECTOR_FOREACH_IDX (&_ctx.service_vector, service, i) {
+        if (strcmp(service->path, path) == 0) {
+            is_exist = true;
+            break;
+        }
+    }
+
+    if (is_exist == false) {
+        service = sol_vector_append(&_ctx.service_vector);
```
barbieri: you can't use the vector, must be a sol_ptr_vector here. The reason is that user will retain the pointer and it must be stable so he can call `sol_connman_service_connect()` and the likes later. He can keep the pointer until `sol_connman_service_get_state() != SOL_CONNMAN_SERVICE_STATE_REMOVE`.

But `sol_vector_append()` may call `realloc()` that will change the element pointer.

src/lib/comms/sol-connman-impl-connman.c
```diff
@@ -0,0 +1,910 @@
...
+
+    bus = sol_bus_get(NULL);
+    if (!bus) {
+        SOL_WRN("Unable to get sd bus\n");
+        return -EINVAL;
+    }
+
+    _ctx.connman = sol_bus_client_new(bus, "org.connman");
+    if (!_ctx.connman) {
+        SOL_WRN("Unable to new a bus client\n");
```
barbieri: leaks `bus`?

src/lib/comms/sol-connman-impl-connman.c
```diff
@@ -0,0 +1,910 @@
...
+{
+    const char *interface;
+    int r;
+
+    SOL_NULL_CHECK(_ctx.properties_changed, -EINVAL);
+
+    interface = sd_bus_message_get_interface(m);
+    SOL_NULL_CHECK(interface, -EINVAL);
+
+    if (strncmp(interface, "net.connman.", 12) != 0)
```
barbieri: we have `strstartswith()`

src/lib/comms/sol-connman-impl-connman.c
```diff
@@ -0,0 +1,910 @@
...
+    } else if (sd_bus_message_is_signal(m, "net.connman.Manager",
+        "PropertyChanged")) {
+        char *str = NULL;
+
+        if (_ctx.manager_slot)
+            return -EINVAL;
+
+        r = sd_bus_message_read_basic(m, SD_BUS_TYPE_STRING, &str);
+        SOL_INT_CHECK(r, < 0, r);
+        if (strcmp(str, "State") == 0) {
```
barbieri: as for `services_list_changed()`, couldn't you abstract this to a common function shared with the GetProperties?

src/lib/comms/sol-connman-impl-connman.c
```diff
@@ -0,0 +1,910 @@
...
+    int r;
+    uint16_t i;
+    struct dbus_monitor_callback *monitor;
+    bool is_exist = false;
+    char matchstr[256];
+    sd_bus *bus = sol_bus_client_get_bus(_ctx.connman);
+
+    SOL_NULL_CHECK(bus, -EINVAL);
+
+    SOL_VECTOR_FOREACH_IDX (&_ctx.monitor_vector, monitor, i) {
```
barbieri: we have `sol-monitors.h`, take a look as these are all handled.

src/lib/comms/sol-connman-impl-connman.c
```diff
@@ -0,0 +1,910 @@
...
+
+    return 0;
+}
+
+static void
+dbus_connection_del_monitor(
+    void (*cb)(void *data, const struct sol_connman_service *service),
+    const void *data)
+{
+
```
barbieri: empty?

src/lib/comms/sol-connman-impl-connman.c
```diff
@@ -0,0 +1,910 @@
...
+{
+    int r;
+
+    r = sol_connman_init_lazy();
+    SOL_INT_CHECK_GOTO(r, < 0, fail_init);
+
+    r = dbus_connection_add_monitor(cb, data);
+    SOL_INT_CHECK_GOTO(r, < 0, fail);
+
+    r = dbus_connection_get_manager_properties();
```
barbieri: why get manager properties? You should have a monitor for the general state, not related to services, then you can do this there.

src/lib/comms/sol-connman-impl-connman.c
```diff
@@ -0,0 +1,910 @@
...
+
+SOL_API int
+sol_connman_del_service_monitor(
+    void (*cb)(void *data, const struct sol_connman_service *service),
+    const void *data)
+{
+    SOL_NULL_CHECK(_ctx.connman, -EINVAL);
+
+    dbus_connection_del_monitor(cb, data);
+    sol_connman_shutdown_lazy();
```
barbieri: uh oh! You need to make `init_lazy()` and `shutdown_lazy()` reference counted, otherwise on the first one you're going to shutdown the whole thing.

You can't  rely on service_monitors.len since other users of the dbus may be in place, like GetServices() call and so on.

src/lib/comms/sol-connman-impl-connman.c
```diff
@@ -0,0 +1,910 @@
...
+    r = sol_connman_init_lazy();
+    SOL_INT_CHECK_GOTO(r, < 0, fail_init);
+
+    r = dbus_connection_add_monitor(cb, data);
+    SOL_INT_CHECK_GOTO(r, < 0, fail);
+
+    r = dbus_connection_get_manager_properties();
+    SOL_INT_CHECK_GOTO(r, < 0, fail);
+
+    r = dbus_connection_get_service_properties();
```
barbieri: only do the get_service_properties() on the first monitor.

src/test/test-connman.c
```diff
@@ -0,0 +1,79 @@
...
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+#include <stdbool.h>
+#include <inttypes.h>
+#include <stdio.h>
+#include <errno.h>
+
+#include "sol-mainloop.h"
```
barbieri: use `soletta.h`, it defines the `SOL_MAIN_DEFAULT()` macro that you should use.

src/test/test-connman.c
```diff
@@ -0,0 +1,79 @@
...
+    enum sol_connman_service_state state;
+    enum sol_connman_state g_state;
+    int r;
+    bool offline;
+
+    str = sol_connman_service_get_name(service);
+    if (!str)
+        SOL_INF("service name = %s\n", str);
+    else
+        SOL_INF("service name = NULL\n");
```
barbieri: for these samples we use regular printf() instead of logs.

barbieri:
@wzhen12 this PR is very good! I believe after you rename to `netctl` and fixes the minor issues I've pointed it can be merged. Basically you need 2 more monitors: one to report manager (global) changes and another to report errors.

Then use https://github.com/solettaproject/soletta/blob/master/src/shared/sol-monitors.h to have safe-walk and dispatch, that is when you walk the monitors vector and call back the user, which may lead to changes to the array you're walking. https://github.com/solettaproject/soletta/blob/master/src/lib/common/sol-platform.c can be used as example.

src/lib/comms/sol-connman-impl-connman.c
wzhen12: Yes. gateway address and netmask are useful for an application.

src/lib/comms/sol-connman-impl-connman.c
wzhen12: I will add it. Thanks.

src/lib/comms/sol-connman-impl-connman.c
wzhen12: Yes. It look like that sd_bus doesn't provide "RemoveMatch".
The match will be removed in sd-bus internal code.

src/lib/comms/sol-connman-impl-connman.c
wzhen12: I will add it.

src/lib/comms/sol-connman-impl-connman.c
wzhen12: Okay, modify it.

src/lib/comms/sol-connman-impl-connman.c
wzhen12: okay.

src/lib/comms/sol-connman-impl-connman.c
wzhen12: Remove it.

src/lib/comms/sol-connman-impl-connman.c
wzhen12: Yes.

src/lib/comms/sol-connman-impl-connman.c
wzhen12: services_list_changed has been used in GetProperties.

wzhen12: Please check _services_properties_changed

src/lib/comms/sol-connman-impl-connman.c
wzhen12: I will modify it

src/lib/comms/sol-connman-impl-connman.c
wzhen12: The new dbus_connection_del_monitor will be not empty.

src/lib/comms/sol-connman-impl-connman.c
wzhen12: Will add a monitor

src/lib/comms/sol-connman-impl-connman.c
wzhen12: Okay.

src/lib/comms/sol-connman-impl-connman.c
wzhen12: I will do it.

src/test/test-connman.c
wzhen12: Okay

src/test/test-connman.c
wzhen12: I will modify it.

src/lib/comms/sol-connman-impl-connman.c
wzhen12: I will modify it.

src/lib/comms/sol-connman-impl-connman.c
wzhen12: I will add it.

src/lib/comms/sol-connman-impl-connman.c
wzhen12: I will modify it.

src/lib/comms/sol-connman-impl-connman.c
wzhen12: Okay

src/lib/comms/sol-connman-impl-connman.c
wzhen12: Yes. Thanks.

src/lib/comms/sol-connman-impl-connman.c
wzhen12: I will modify it.

src/lib/comms/sol-connman-impl-connman.c
wzhen12: I will add the help function

src/lib/comms/sol-connman-impl-connman.c
wzhen12: Okay

src/lib/comms/sol-connman-impl-connman.c
wzhen12: I will remove malloc.

src/lib/comms/sol-connman-impl-connman.c
wzhen12: I will remove malloc

src/lib/comms/sol-connman-impl-connman.c
wzhen12: okay.

src/lib/comms/sol-connman-impl-connman.c
wzhen12: okay

src/lib/comms/sol-connman-impl-connman.c
wzhen12: I will modify it.

src/lib/comms/sol-connman-impl-connman.c
wzhen12: I will add sol_connman_add_error_monitor.

src/lib/comms/sol-connman-impl-connman.c
wzhen12: Okay.

src/lib/comms/sol-connman-impl-connman.c
wzhen12: Okay.

src/lib/comms/sol-connman-impl-connman.c
wzhen12: I will modify it.

src/lib/comms/include/sol-connman.h
wzhen12: I will modify it.

src/lib/comms/include/sol-connman.h
wzhen12: Okay.

src/lib/comms/include/sol-connman.h
wzhen12: I will modify it.

wzhen12:
Thanks. after I fix the issues, I will submit a new version for it in a few days.

src/lib/comms/sol-connman-impl-connman.c
wzhen12: After finish "the while", service monitor is invoked.
Please check "end: ... call_monitor_callback(service).

barbieri: ah! I've missed it!

src/lib/comms/include/sol-connman.h
vcgomes: I didn't think about `struct sol_network_link` that looks like a good match.

wzhen12: @vcgomes , I submit the new version of sol-connman. some modifications are for sol_network_link.
Please check it. Thanks.

wzhen12:
The patches have been closed.
The new version sol-connman patches have been submitted.
The link is https://github.com/solettaproject/soletta/pull/1905

PR #1849, opened 2016-04-08

nagineni:

This replaces #1847

### DIFF
--- a/bindings/nodejs/src/structures/sol-js-spi.cc
+++ b/bindings/nodejs/src/structures/sol-js-spi.cc
@@ -1,0 +1,42 @@
+/*
+ * This file is part of the Soletta Project
+ *
+ * Copyright (C) 2015 Intel Corporation. All rights reserved.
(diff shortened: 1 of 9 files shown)


nagineni:
@gabrielschulhof PTAL. Thanks!

bindings/nodejs/src/structures/sol-js-spi.cc
```diff
@@ -0,0 +1,42 @@
...
+ * limitations under the License.
+ */
+
+#include <nan.h>
+#include "../common.h"
+#include "sol-js-spi.h"
+
+using namespace v8;
+
+bool c_sol_spi_config(v8::Local<v8::Object> jsUARTConfig,
```
gabrielschulhof: Please rename the variable to something SPI-related :)

nagineni: Sure :)

PR #1850, opened 2016-04-08
Flow uniform string to packet

barbieri:
I'm starting a huge rework of the sol-flow internals in order to help with memory consumption. Since the patchset is likely to be big and I can't work on all of it in one go, I'll submit intermediate stages as I do it.

This is the first step and makes the string to packet type conversion unified in `sol-flow-packet.h`, this one will cover the simple types and also composed, creating if required, so we don't spread that parsing all over the place.

There is a drawback: if this new function is used, then all packet types will be held. I've noticed that most of the users of the packet type don't actually need them if we offer introspection to serialize/parse based on memory description, like JS or HTTP nodes. Then with the future cleanups this may be minimized.

### DIFF
--- a/src/lib/datatypes/include/sol-str-slice.h
+++ b/src/lib/datatypes/include/sol-str-slice.h
@@ -206,7 +206,7 @@ sol_str_slice_copy(char *dst, const struct sol_str_slice src)
 static inline bool
 sol_str_slice_starts_with(const struct sol_str_slice slice, const struct sol_str_slice prefix)
 {
-    return slice.len >= prefix.len && strncmp(slice.data, prefix.data, prefix.len);
+    return slice.len >= prefix.len && strncmp(slice.data, prefix.data, prefix.len) == 0;
 }
 
 /**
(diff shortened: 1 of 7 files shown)


ibriano:
Seems fine

src/lib/datatypes/include/sol-str-slice.h
```diff
@@ -206,7 +206,7 @@ sol_str_slice_copy(char *dst, const struct sol_str_slice src)
 static inline bool
 sol_str_slice_starts_with(const struct sol_str_slice slice, const struct sol_str_slice prefix)
 {
-    return slice.len >= prefix.len && strncmp(slice.data, prefix.data, prefix.len);
+    return slice.len >= prefix.len && strncmp(slice.data, prefix.data, prefix.len) == 0;
```
cabelitos: God, how did we miss this one?

cabelitos:
+1

src/lib/datatypes/include/sol-str-slice.h
edersondisouza: :new_moon_with_face:

edersondisouza:
lgtm

barbieri:
merged

PR #1851, opened 2016-04-08

nagineni:

This replaces #1849

### DIFF
--- a/bindings/nodejs/src/functions/spi.cc
+++ b/bindings/nodejs/src/functions/spi.cc
@@ -1,0 +1,166 @@
+/*
+ * This file is part of the Soletta Project
+ *
+ * Copyright (C) 2015 Intel Corporation. All rights reserved.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+#include <node.h>
+#include <nan.h>
+#include <sol-spi.h>
+
+#include "../common.h"
+#include "../data.h"
+#include "../hijack.h"
+#include "../structures/js-handle.h"
+#include "../structures/sol-js-spi.h"
+
+using namespace v8;
+
+class SolSpi : public JSHandle<SolSpi> {
+public:
+    static const char *jsClassName() { return "SolSpi"; }
+};
+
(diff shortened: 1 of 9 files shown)


nagineni:
@gabrielschulhof

bindings/nodejs/src/functions/spi.cc
```diff
@@ -0,0 +1,166 @@
...
+
+    size_t length = node::Buffer::Length(info[1]);
+    uint8_t *txBuffer = (uint8_t *) malloc(length * sizeof(uint8_t));
+    if (!txBuffer) {
+        Nan::ThrowError("Failed to allocate memory for output buffer");
+        return;
+    }
+    memcpy(txBuffer, node::Buffer::Data(info[1]), length);
+
+    uint8_t *rxBuffer = (uint8_t *) malloc(length * sizeof(uint8_t));
```
edersondisouza: Why not `calloc` instead of `malloc + memset`?

cabelitos: `calloc` also is safer than `malloc`, because it will check of overflow. 
(`length *s izeof(uint8_t)` may overflow

bindings/nodejs/src/functions/spi.cc
```diff
@@ -0,0 +1,166 @@
...
+    if (!spi)
+        return;
+
+    if (!node::Buffer::HasInstance(info[1])) {
+        Nan::ThrowTypeError("Argument 1 must be a node Buffer");
+        return;
+    }
+
+    size_t length = node::Buffer::Length(info[1]);
+    uint8_t *txBuffer = (uint8_t *) malloc(length * sizeof(uint8_t));
```
edersondisouza: As @cabelitos said, here it's better use `calloc` as well, or check for overflows using `sol_util_size_mul`.

src/lib/io/include/sol-spi.h
```diff
@@ -82,6 +84,32 @@ struct sol_spi_config {
...
+ *
+ * @see sol_spi_mode_from_str().
+ *
+ * @param spi_mode sol_spi_mode
+ *
+ * @return String representation of the sol_spi_mode
+ */
+const char *sol_spi_mode_to_str(enum sol_spi_mode spi_mode) SOL_ATTR_WARN_UNUSED_RESULT;
+
+/**
```
edersondisouza: I would separate this changes to sol_spi itself on a different commit. What do you think?

bindings/nodejs/src/functions/spi.cc
nagineni: Make sense. I will fix  this.

bindings/nodejs/src/functions/spi.cc
nagineni: I will fix this.

src/lib/io/include/sol-spi.h
nagineni: I'm okay to split this patch, since these are common enum-string conversion functions in sol_spi and not dependent on bindings.

nagineni:
@cabelitos @edersondisouza thanks a lot for reviewing. I will split the patch and submit new PRs with the fixes.

PR #1852, opened 2016-04-08
log: make it possible to skip getting file/function names on output.

glima:
Besides logging those entries or not (which was already a run-time
paramemeter), we now have the option of never including those strings to
the final binay (so even if the user asks to log those, it gets empty
strings for them). This helps save space on restricted system, while
still having log enabled, *a lot*. Without this, lots of applications
would have to be debugged with custom printfs (and Soletta general
logging turned off), what is a total PITA.

### DIFF
--- a/src/lib/common/include/sol-log.h
+++ b/src/lib/common/include/sol-log.h
@@ -458,6 +458,24 @@ sol_log_domain_init_level(struct sol_log_domain *domain)
 #define SOL_LOG_LEVEL_POSSIBLE(level) (0)
 #endif
 
+#ifdef SOL_LOG_FILES
+#ifdef SOL_LOG_FUNCTIONS
+#define SOL_LOG_FILE __FILE__
+#define SOL_LOG_FUNCTION __PRETTY_FUNCTION__
+#else
+#define SOL_LOG_FILE __FILE__
+#define SOL_LOG_FUNCTION NULL
+#endif
+#else
+#ifdef SOL_LOG_FUNCTIONS
+#define SOL_LOG_FILE NULL
+#define SOL_LOG_FUNCTION __PRETTY_FUNCTION__
+#else
+#define SOL_LOG_FILE NULL
+#define SOL_LOG_FUNCTION NULL
+#endif
+#endif
+
 /**
  * @def SOL_LOG(level, fmt, ...)
  *
@@ -472,7 +490,7 @@ sol_log_domain_init_level(struct sol_log_domain *domain)
     do { \
         if (SOL_LOG_LEVEL_POSSIBLE(level)) { \
             sol_log_print(SOL_LOG_DOMAIN, level, \
-                __FILE__, __PRETTY_FUNCTION__, __LINE__, \
+                SOL_LOG_FILE, SOL_LOG_FUNCTION,  __LINE__, \
                 fmt, ## __VA_ARGS__); \
         } \
     } while (0)
--- a/src/lib/common/Kconfig
+++ b/src/lib/common/Kconfig
@@ -367,6 +367,24 @@ choice
           information will be hidden and sol-fbp-runner users may also
           have a bad experience.
 
+          On Linux, one is able to control some logging parameters by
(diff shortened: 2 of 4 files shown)


src/lib/common/include/sol-log.h
```diff
@@ -468,14 +468,55 @@ sol_log_domain_init_level(struct sol_log_domain *domain)
...
+#ifdef SOL_LOG_FILES
+#ifdef SOL_LOG_FUNCTIONS
 #define SOL_LOG(level, fmt, ...) \
     do { \
         if (SOL_LOG_LEVEL_POSSIBLE(level)) { \
             sol_log_print(SOL_LOG_DOMAIN, level, \
-                __FILE__, __PRETTY_FUNCTION__, __LINE__, \
+                __FILE__, \
+                __PRETTY_FUNCTION__, \
+                __LINE__, \
```
ibriano: Would changing __FILE__ and friends to something like LOG_FILE that is defined to __FILE__ or NULL accordingly work?

glima: I guess so. Thanks, trying that.

src/lib/common/include/sol-log.h
```diff
@@ -458,6 +458,24 @@ sol_log_domain_init_level(struct sol_log_domain *domain)
...
+#endif
+#else
+#ifdef SOL_LOG_FUNCTIONS
+#define SOL_LOG_FILE NULL
+#define SOL_LOG_FUNCTION __PRETTY_FUNCTION__
+#else
+#define SOL_LOG_FILE NULL
+#define SOL_LOG_FUNCTION NULL
+#endif
+#endif
```
ibriano: Much shorter in general, but why not
```C
#ifdef SOL_LOG_FILES
#define SOL_LOG_FILE __FILE__
#else
#define SOL_LOG_FILE NULL
#endif

#ifdef SOL_LOG_FUNCTIONS
#define SOL_LOG_FUNCTION __PRETTY_FUNCTION__
#else
#define SOL_LOG_FUNCTION NULL
#endif
```

glima: Because go :horse:. I'll fix that too.

ibriano: On Fri, 08 Apr 2016 12:06:27 -0700, Gustavo Lima Chaves wrote:
> > +#ifdef SOL_LOG_FUNCTIONS
> > +#define SOL_LOG_FILE __FILE__
> > +#define SOL_LOG_FUNCTION __PRETTY_FUNCTION__
> > +#else
> > +#define SOL_LOG_FILE __FILE__
> > +#define SOL_LOG_FUNCTION NULL
> > +#endif
> > +#else
> > +#ifdef SOL_LOG_FUNCTIONS
> > +#define SOL_LOG_FILE NULL
> > +#define SOL_LOG_FUNCTION __PRETTY_FUNCTION__
> > +#else
> > +#define SOL_LOG_FILE NULL
> > +#define SOL_LOG_FUNCTION NULL
> > +#endif
> > +#endif
> 
> Because go :horse:. I'll fix that too.

And merge it

barbieri:
This is a great commit, but in that sense I'm thinking we could go one step
further:

What if we run an ELF editor that replaces the message with "{id}
{in_order_replacements}", then we'd see messages such as "DBG: 2af 2", with
that you get a decode script on desktop that converts it to "DBG: some
var=2"

This could be used even for the file names, then you use an id instead of
long path.

Since this may be hard to implement, I'd  just add a ticket and someday we
get back to it.

Em sexta-feira, 8 de abril de 2016, Gustavo Lima Chaves <
notifications@github.com> escreveu:

> Closed #1852 <https://github.com/solettaproject/soletta/pull/1852>.
>
> —
> You are receiving this because you are subscribed to this thread.
> <https://github.com/solettaproject/soletta/pull/1852#event-620576212>
>


-- 
Gustavo Sverzut Barbieri
--------------------------------------
Mobile: +55 (19) 99225-2202
Contact: http://www.gustavobarbieri.com.br/contact

PR #1853, opened 2016-04-10
SPI: Add enum-string mapping functions for sol_spi_mode

nagineni:
This patch adds enum-string mapping functions for sol_spi_mode, so that these common functions can be used in the bindings for the conversion from string to enum and vice versa.

Part of #1851 

### DIFF
--- a/src/lib/io/include/sol-spi.h
+++ b/src/lib/io/include/sol-spi.h
@@ -20,8 +20,10 @@
 
 #include <stdint.h>
 #include <stdbool.h>
+#include <sys/types.h>
 
 #include <sol-common-buildopts.h>
+#include <sol-macros.h>
 
 #ifdef __cplusplus
 extern "C" {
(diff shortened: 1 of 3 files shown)


nagineni:
@cabelitos @edersondisouza could you please have a look and merge this? I will create a PR for SPI bindings once this is in.

Thanks!

edersondisouza:
+1

nagineni:
@edersondisouza @cabelitos Thanks again for looking into this. Could you please merge?

src/lib/io/include/sol-spi.h
```diff
@@ -20,8 +20,10 @@
 
 #include <stdint.h>
 #include <stdbool.h>
+#include <sys/types.h>
```
cabelitos: Not sure about this one. Why did you include sys/types?

cabelitos: I guess this can lead to problems depending what's your target.

nagineni: @cabelitos I remember seeing compilation error _‘ssize_t’ has not been declared_ when sol-spi.h was included in the bindings, so I added this. Now, I tried with the latest sources and I don't see the error even after removing this include, so we probably don't need it.

cabelitos: Awesome!
Removed it and send a new version plz!

nagineni: @cabelitos  thanks!

I will submit a new PR.

PR #1856, opened 2016-04-11
Parse the test cases results log by json

wanghongjuan:
- Generate a json file to collect the test results.
- Locate easily the failed tests in result file when using the results file
- Quickly generate results to match the specified format that QA required
- JSON dose not need to take up extra workspace.

Signed-off-by: Wang Hongjuan hongjuan.wang@intel.com

### DIFF
--- a/bindings/nodejs/tests/getresult.js
+++ b/bindings/nodejs/tests/getresult.js
@@ -1,0 +1,51 @@
+/*
+ * This file is part of the Soletta Project
+ *
+ * Copyright (C) 2016 Intel Corporation. All rights reserved.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+var fs = require( "fs" );
+
+var jsonFilename = require( "path" ).resolve( __dirname, "results.json" );
+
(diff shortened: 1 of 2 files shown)


wanghongjuan:
@gabrielschulhof Could you please review it? Thanks.

bindings/nodejs/tests/getresult.js
```diff
@@ -0,0 +1,57 @@
...
+  caseList = [],
+  setList = [],
+  allList = [],
+  test = {},
+  testInfo = {};
+
+exports.getTestResult = function( status, success, failure  ) {
+
+  if ( status.result ) {
+    success = "PASS";
```
edersondisouza: Is this right? It appears that you get `success` and `failure` as parameters and change their values. But the caller already set a value for them, why do you do that?

bindings/nodejs/tests/setup.js
```diff
@@ -56,6 +57,8 @@ QUnit.config.callbacks.log.push( function( status ) {
 
 	// Parameters: status: { module, result(t/f), message, actual, expected, testId, runtime }
 
+	results.getTestResult( status, success, failure );
```
edersondisouza: Here it appears that `success` and `failure` will have their value changed by this call. Is this right?

bindings/nodejs/tests/getresult.js
```diff
@@ -0,0 +1,57 @@
...
+  } else if ( !status.result ) {
+    failure = "FAIL";
+  }
+
+  testInfo = {
+    "message": (  status.runtime  + ": " + status.message ),
+    "result": ( status.result ? success : failure ),
```
edersondisouza: Maybe all that you wanted here was something like:
```js
"result": ( status.result ? "PASS" : "FAIL" ),
```
With no need for `success` and `failure` parameters?

wanghongjuan: Many thanks for propose a good way, I will update my code.

wanghongjuan:
@edersondisouza I create a new PR https://github.com/solettaproject/soletta/pull/1921, please review it. Thanks.

PR #1859, opened 2016-04-11
Socket v3

ceolin:
from v2:

- Change options's name
- Use sub api in options
- Do not create the fd watch if this is not used
- Fixes on doc
- sendmsg() returns ssize_t
- added client/server echo sample


ceolin:
@barbieri I intend to change the socket api to use `sol_buffer` in the next pr, please take it in consideration when reviewing. So the server sample will not need to allocate the memory as it is doing now.

src/lib/comms/sol-socket-impl-linux.c
```diff
@@ -437,85 +351,104 @@ sol_socket_linux_bind(struct sol_socket *socket, const struct sol_network_link_a
...
+    bool ret;
+    struct sol_socket_linux *sock = (struct sol_socket_linux *)s;
 
-    option = sol_socket_option_to_linux(optname);
-    l = sol_socket_level_to_linux(level);
+    SOL_NULL_CHECK(sock->on_can_write, -EINVAL);
 
-    ret = getsockopt(s->fd, l, option, optval, &len);
-    SOL_INT_CHECK(ret, < 0, ret);
+    if (!sock) {
```
vcgomes: This should be before `SOL_NULL_CHECK()`.

src/lib/comms/sol-socket-impl-linux.c
```diff
@@ -437,85 +351,104 @@ sol_socket_linux_bind(struct sol_socket *socket, const struct sol_network_link_a
...
-    case SOL_SOCKET_LEVEL_SOCKET:
-        return SOL_SOCKET;
-    case SOL_SOCKET_LEVEL_IP:
-        return IPPROTO_IP;
-    case SOL_SOCKET_LEVEL_IPV6:
-        return IPPROTO_IPV6;
-    default:
-        SOL_WRN("Invalid level %d", level);
-        break;
+    if (!sock) {
```
vcgomes: Ah, I guess you meant `sock->watch`, no?

(my other comment missed that)

src/samples/network/echo-client.c
```diff
@@ -0,0 +1,179 @@
...
+    r = sol_socket_sendmsg(sock, value, strlen(value), &address);
+    if (r < 0) {
+        fprintf(stderr, "ERROR: Could not send data\n");
+        goto end;
+    }
+
+    r = 0;
+
+end:
+    sol_socket_set_write_monitor(sock, false);
```
vcgomes: What's the difference of returning `false` and disabling the write monitor? From the documentation I got the impression that they are equivalent?

src/samples/network/echo-client.c
```diff
@@ -0,0 +1,179 @@
...
+    len = (size_t)(itr - addr);
+    memcpy(addr_str, addr, len);
+
+    if (!sol_network_link_addr_from_str(&address, addr_str)) {
+        fprintf(stderr, "Could not convert the address: %s\n", addr_str);
+        goto err;
+    }
+
+    itr++;
+    address.port = atoi(itr);
```
vcgomes: Isn't a `sol_socket_bind()` missing?

ceolin: for client ?

src/lib/comms/sol-socket-impl-linux.c
ceolin: o yeah,  i messed it during a rebase. Fixing

src/samples/network/echo-client.c
vcgomes: Ah! That answers it :-)

src/samples/network/echo-client.c
ceolin: They are the same, it was my fault

src/samples/network/echo-server.c
```diff
@@ -0,0 +1,190 @@
...
+ * limitations under the License.
+ */
+
+/**
+ * @file
+ * @brief Network link status
+ *
+ * Network sample that monitors specified links. It checks if the link is
+ * up/down and its addresses.
+ * To see the usage help, -h or --help.
```
glima: At least fix the doc string (copy and pasta)

src/samples/network/echo-client.c
```diff
@@ -0,0 +1,178 @@
...
+ * limitations under the License.
+ */
+
+/**
+ * @file
+ * @brief Network link status
+ *
+ * Network sample that monitors specified links. It checks if the link is
+ * up/down and its addresses.
+ * To see the usage help, -h or --help.
```
glima: Ditto

src/lib/comms/sol-coap.c
```diff
@@ -1746,40 +1747,30 @@ network_event(void *data, const struct sol_network_link *link, enum sol_network_
 }
 
 static struct sol_coap_server *
-sol_coap_server_new_full(enum sol_socket_type type, const struct sol_network_link_addr *servaddr)
+sol_coap_server_new_full(struct sol_socket_ip_options *options, const struct sol_network_link_addr *servaddr)
```
glima: I was expecting a further patch changing this to receive an external socket struct (or adding variants of that kind to those), but I did not find it. Are we going that direction later, enabling CoAP for every socket, or not?

src/samples/network/echo-client.c
ceolin: thanks !

src/lib/comms/sol-coap.c
ceolin: yes, it will come in the next pr. I'm doing it together with the oic over ble in order to have on more client of this api and see the best way to change it.

src/lib/comms/sol-socket-impl-linux.c
```diff
@@ -33,42 +33,30 @@
...
-    ret = s->read.cb((void *)s->read.data, &s->base);
-    if (!ret) {
-        s->read.cb = NULL;
-        s->read.data = NULL;
-        s->read.watch = NULL;
+    if (flags & SOL_FD_FLAGS_IN) {
+        ret = s->on_can_read((void *)s->data, &s->base);
+        if (!ret)
+            sol_fd_unset_flags(s->watch, SOL_FD_FLAGS_IN);
+    } else if (flags & SOL_FD_FLAGS_OUT) {
```
vcgomes: I guess it is possible the socket to be available for reading and writing at the same time, no?

src/lib/comms/sol-socket-impl-linux.c
```diff
@@ -33,42 +33,30 @@
...
-    bool ret;
 
-    ret = s->write.cb((void *)s->write.data, &s->base);
-    if (!ret) {
-        s->write.cb = NULL;
-        s->write.data = NULL;
-        s->write.watch = NULL;
-    }
-    return ret;
+    return true;
```
vcgomes: Just a question (I don't know the answer), do we want to keep the watch even if there are no flags set? Something to think about.

src/lib/comms/sol-socket-impl-linux.c
ceolin: argg, yeap. Fixing it.
Thanks a lot

ceolin:
fixes on https://github.com/solettaproject/soletta/pull/1872

PR #1860, opened 2016-04-11
mqtt: Gives user data in config

ceolin:
It is the same pattern used by message digest.


ceolin:
from v1:
 - Host and port in the config
 - store the user data as const pointer

vcgomes:
Looks good.

glima:
:+1:

cabelitos:
looks goodd

PR #1861, opened 2016-04-12
Node.js bindings: Copy high-level APIs to repo root

gabrielschulhof:
This only happens when building an npm package

### DIFF
--- a/binding.gyp
+++ b/binding.gyp
@@ -38,6 +38,21 @@
 					} ]
 				},
 				{
+					"target_name": "copyapis",
+					"type": "none",
+					"actions": [ {
+						"action_name": "copyapis",
+						"message": "Copying JS APIs",
+						"inputs": [ "./bindings/nodejs/lib" ],
+						"outputs": [ "" ],
+						"action": [
+							"sh",
+							"-c",
+							"cp -a ./bindings/nodejs/lib/* ."
+						]
+					} ]
+				},
+				{
 					"target_name": "soletta",
 					"includes": [
 						"bindings/nodejs/generated/nodejs-bindings-sources.gyp"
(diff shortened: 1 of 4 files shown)


binding.gyp
```diff
@@ -38,6 +38,23 @@
 					} ]
 				},
 				{
+					"target_name": "copyapis",
+					"type": "none",
+					"conditions": [ [ "'<!(echo $BUILD_NPM_PACKAGE)'=='y'", {
+						"actions": [ {
+							"action_name": "copyapis",
+							"message": "Copying JS APIs",
+							"outputs": [ "./uart.js" ],
```
edersondisouza: Any reason this should apply only to uart? Why not aio, pwm or gpio?

gabrielschulhof: Hmmm ... maybe this needs a rethink. What if soletta is built without uart support? The fact is that this is just an artifact to make sure the target has some inputs and outputs. Maybe this target must also be generated by configure-bindings.js. Good catch, @edersondisouza!

gabrielschulhof: OTOH, if the support is not there, it doesn't matter if we copy the corresponding high-level APIs or not, because applications not taking these things into account will fail anyway. Whether they fail because the low-level soletta object lacks the appropriate bindings or whether they fail because
```JS
require( "soletta/uart" );
```
fails, it doesn't matter. They fail.

gabrielschulhof: In fact, the hardware bindings will fail exactly as intended, in that the promises they return will be rejected, because the promise function will throw an exception when it attempts to call ```soletta.sol_<hardware_api>_*()```. It's not as clean as rejecting ```if ( typeof soletta.sol_<hardware_api>_*() === "undefined" )```, but it works just as well.

PR #1863, opened 2016-04-12
HTTP: Json nodes should support simple types like int, float, etc...

cabelitos:


src/modules/flow/http-client/http-client.json
```diff
@@ -667,6 +667,46 @@
...
+          "data_type": "string",
+          "description": "POST a json string",
+          "name": "POST_STRING",
+          "methods": {
+            "process": "json_string_post_process"
+          }
+        },
+        {
+          "data_type": "int",
+          "description": "POST a json int",
```
barbieri: say it will use "val" from irange. similar for float.

src/modules/flow/http-client/http-client.json
```diff
@@ -679,6 +719,31 @@
...
+          "name": "INT"
+        },
+        {
+          "data_type": "float",
+          "description": "The response as json float",
+          "name": "FLOAT"
+        },
+        {
+          "data_type": "empty",
+          "description": "The response as json NULL",
```
barbieri: "if json object is `null`, then an empty packet is emitted"

src/modules/flow/http-client/http-client.json
```diff
@@ -679,6 +719,31 @@
...
+          "name": "STRING"
+        },
+        {
+          "data_type": "int",
+          "description": "The response as json int",
+          "name": "INT"
+        },
+        {
+          "data_type": "float",
+          "description": "The response as json float",
```
barbieri: "if the json object is a float, the float object is created with that as value and default range (min as -DBL_MAX, max as DBL_MAX and step as DBL_MIN)"

src/modules/flow/http-client/http-client.json
```diff
@@ -679,6 +719,31 @@
...
+          "name": "BOOLEAN"
+        },
+        {
+          "data_type": "string",
+          "description": "The response as json string",
+          "name": "STRING"
+        },
+        {
+          "data_type": "int",
+          "description": "The response as json int",
```
barbieri: if the json object is an integer, the int packet is created with that as value and default range (min as INT32_MIN, max as INT32_MAX and step as 1)

src/modules/flow/http-server/http-server.c
```diff
@@ -1317,50 +1313,263 @@ blob_handle_response_cb(struct sol_flow_node *node,
...
+    if (type == SOL_FLOW_PACKET_TYPE_JSON_ARRAY)
+        return SOL_JSON_TYPE_ARRAY_START;
+    if (type == SOL_FLOW_PACKET_TYPE_BOOLEAN)
+        return SOL_JSON_TYPE_TRUE;
+    if (type == SOL_FLOW_PACKET_TYPE_STRING)
+        return SOL_JSON_TYPE_STRING;
+    if (type == SOL_FLOW_PACKET_TYPE_DRANGE)
+        return SOL_JSON_TYPE_NUMBER;
+    if (type == SOL_FLOW_PACKET_TYPE_IRANGE) {
+        *is_irange = true;
```
barbieri: what's up with this is_irange? why no is_drange?

barbieri: forget about that, saw the reason below

src/modules/flow/http-server/http-server.c
```diff
@@ -1317,50 +1313,263 @@ blob_handle_response_cb(struct sol_flow_node *node,
...
+            SOL_FLOW_NODE_TYPE_HTTP_SERVER_JSON__OUT__BOOLEAN, mdata->value.b);
+        break;
+    case SOL_JSON_TYPE_STRING:
+        sol_flow_send_string_packet(node,
+            SOL_FLOW_NODE_TYPE_HTTP_SERVER_JSON__OUT__STRING, mdata->value.s);
+        break;
+    case SOL_JSON_TYPE_NUMBER:
+        sol_flow_send_drange_packet(node,
+            SOL_FLOW_NODE_TYPE_HTTP_SERVER_JSON__OUT__FLOAT, &mdata->value.d);
+        if (mdata->value.d.val >= INT32_MIN && mdata->value.d.val <= INT32_MAX)
```
barbieri: shouldn't you check there are no decimal places? something like:
```c
sol_drange_value_equal(mdata->value.d.val - (int32_t)mdata->value.d.val , 0.0)
```

barbieri:
just minor things, you can push without new PR

src/modules/flow/http-server/http-server.c
cabelitos: I thought about that too, however when I look at the json nodes I saw that they don't do that.
Then I did the same thing, to keep it consistent

src/modules/flow/http-client/http-client.json
cabelitos: I think we should say "if the json value is null, then an..."

cabelitos:
merged with the proposed fixes

PR #1866, opened 2016-04-12
Iio create sync v2

edersondisouza:
v2:
v1 was flawed: it was busy waiting on a directory where a file was supposed to appear - and not on file itself. As we have no way to know what will be file name, we now keep scanning directory till file appears - up to one second. I was fooled into thinking it was working thanks to slowness caused by running Soletta with DBG log level =/
Thanks to this change, first patch - sol-util-file one - may be unnecessary. I'll leave it if no complains appears, as it's a cool patch.

### DIFF
--- a/src/shared/sol-util-file.c
+++ b/src/shared/sol-util-file.c
@@ -565,3 +565,21 @@ err_rename:
 
     return r;
 }
+
+SOL_API bool
+sol_util_busy_wait_file(const char *path, uint64_t nanoseconds)
+{
+    struct stat st;
+    struct timespec start = sol_util_timespec_get_current();
+
+    while (stat(path, &st)) {
+        struct timespec elapsed, now = sol_util_timespec_get_current();
+
+        sol_util_timespec_sub(&now, &start, &elapsed);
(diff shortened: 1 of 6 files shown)


src/shared/sol-util-file.c
```diff
@@ -565,3 +565,21 @@ sol_util_move_file(const char *old_path, const char *new_path, mode_t mode)
...
+{
+    struct stat st;
+    struct timespec start = sol_util_timespec_get_current();
+
+    while (stat(path, &st)) {
+        struct timespec elapsed, now = sol_util_timespec_get_current();
+
+        sol_util_timespec_sub(&now, &start, &elapsed);
+        if ((uint64_t)elapsed.tv_sec >= (nanoseconds / SOL_NSEC_PER_SEC) &&
+            (uint64_t)elapsed.tv_nsec > (nanoseconds % SOL_NSEC_PER_SEC))
```
glima: Why not >= on the second clause as well?

glima:
for the rest

src/shared/sol-util-file.c
edersondisouza: Fair.

edersondisouza:
Integrated with fix. Tks!

PR #1868, opened 2016-04-12
Remove unneeded header include from sol-spi.h

nagineni:
This removes an unneeded header include from sol-spi.h and
also fixes a minor error in the API documentation.


edersondisouza:
@nagineni Thank you very much, integrated!

nagineni:
@edersondisouza thanks a lot for your time. :)

PR #1869, opened 2016-04-12

nagineni:

This replaces #1851

### DIFF
--- a/bindings/nodejs/src/functions/spi.cc
+++ b/bindings/nodejs/src/functions/spi.cc
@@ -1,0 +1,165 @@
+/*
+ * This file is part of the Soletta Project
+ *
+ * Copyright (C) 2015 Intel Corporation. All rights reserved.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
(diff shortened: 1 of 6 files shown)


nagineni:
@cabelitos @edersondisouza Addressed review comments from #1851. Could you please take another look? Thanks!

bindings/nodejs/src/functions/spi.cc
```diff
@@ -0,0 +1,165 @@
...
+    Nan::Callback *callback = (Nan::Callback *)cb_data;
+    Local<Value> txBuffer;
+    Local<Value> rxBuffer;
+
+    if (status >= 0) {
+        txBuffer = Nan::NewBuffer((char *)tx, status).ToLocalChecked();
+        rxBuffer = Nan::NewBuffer((char *)rx, status).ToLocalChecked();
+    } else {
+        txBuffer = Nan::Null();
+        rxBuffer = Nan::Null();
```
cabelitos: Don't you need to free the buffers?
`tx` and `rx`

cabelitos:
Overall looks good.

bindings/nodejs/src/functions/spi.cc
nagineni: @cabelitos Good point. Instead of checking the _status_, it make sense to check the _tx_ and _rx_ and pass the ownership to NewBuffer if not null. I will make this change in a new PR.

PR #1870, opened 2016-04-12
lib/sol-iio: release buffer on sol_iio_close()

bdilly:
Otherwise after using a Soletta application the following
issue should happen when trying to access it:

root@edison:/sys/bus/iio/devices/iio:device1# cat *raw
cat: in_voltage0_raw: Device or resource busy
cat: in_voltage1_raw: Device or resource busy
cat: in_voltage2_raw: Device or resource busy

@edersondisouza could you please confirm that?
I didn't try to reproduce this issue (or fix), but
doing some code browsing the reason to stay busy
is thanks to buffer keep set as enabled.

Thanks

### DIFF
--- a/src/lib/io/sol-iio.c
+++ b/src/lib/io/sol-iio.c
@@ -873,6 +873,9 @@ sol_iio_close(struct sol_iio_device *device)
     }
     sol_ptr_vector_clear(&device->channels);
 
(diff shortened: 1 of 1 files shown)


edersondisouza:
Good catch! Tested and it works!
:+1:

bdilly:
cool, merged.
thanks

PR #1872, opened 2016-04-13
Socket v4

ceolin:
from v3:

- Fix samples docs
- Do not keep the watch alive if it's no being monitored


src/lib/comms/include/sol-socket.h
```diff
@@ -0,0 +1,343 @@
...
+     * @brief User data data will be given in on_can_read() and
+     * on_can_write()
+     */
+    const void *data;
+
+    /**
+     * @brief The family that should be used when creating the socket.
+     * @see sol_network_family
+     */
+    enum sol_network_family family;
```
barbieri: didn't you say the family would be in the ip options? not that I disagree with it being here, just reminded of that.

src/lib/comms/include/sol-socket.h
```diff
@@ -0,0 +1,343 @@
...
+};
+
+struct sol_socket_ip_options {
+    struct sol_socket_options base;
+#ifndef SOL_NO_API_VERSION
+#define SOL_SOCKET_IP_OPTIONS_API_VERSION (1)  /**< compile time API version to be checked during runtime */
```
barbieri: sub_api, right? you need to make that clear. And as done with sol_flow_options, it would be nice to offer a macro that sets that.

src/lib/comms/include/sol-socket.h
```diff
@@ -0,0 +1,343 @@
...
+ * This function creates a socket using the system default implementation.
+ *
+ * @param options The socket's options.
+ *
+ * @return a handle to the socket on success, otherwise @c NULL is returned.
+ *
+ * @see sol_socket_del()
+ * @see sol_socket_options
+ */
+struct sol_socket *sol_socket_new(const struct sol_socket_options *options);
```
barbieri: AFAIU you'd call this the `sol_socket_ip_new()`, no?

src/lib/comms/sol-coap.c
```diff
@@ -1746,40 +1747,30 @@ network_event(void *data, const struct sol_network_link *link, enum sol_network_
...
+    int ret;
 
     SOL_LOG_INTERNAL_INIT_ONCE;
 
-    s = sol_socket_new(servaddr->family, type, 0);
+    server = calloc(1, sizeof(*server));
+    SOL_NULL_CHECK(server, NULL);
+
+    options->base.data = server;
+    s = sol_socket_new((struct sol_socket_options *)options);
```
barbieri: avoid  casts, use `&options->base`

src/lib/comms/sol-socket-impl-linux.c
```diff
@@ -437,85 +369,105 @@ sol_socket_linux_bind(struct sol_socket *socket, const struct sol_network_link_a
...
+        sock->watch = sol_fd_add(sock->fd, SOL_FD_FLAGS_OUT, on_socket_event, sock);
+        SOL_NULL_CHECK(sock->watch, -EBADF);
+        return 0;
+    }
 
-    *optlen = len;
-    return 0;
+    if (on)
+        ret = sol_fd_set_flags(sock->watch,
+            sol_fd_get_flags(sock->watch) | SOL_FD_FLAGS_OUT);
```
barbieri: this is strange, because set_flags and unset_flag are similar names and different behaviors. Would you do another PR that changes `unset_flags` to `add/remove` flags?

src/lib/comms/sol-socket.c
```diff
@@ -31,121 +31,103 @@
...
+     * This SUB_API check should be done by constructors. This is being done
+     * here only because the strange behaviour/implementation of socket_dtls.
+     * TODO: Fix sol_socket_dtls
+     */
+    SOL_SOCKET_OPTIONS_CHECK_SUB_API_VERSION(options, SOL_SOCKET_IP_OPTIONS_API_VERSION, NULL);
 
-    socket = impl->new(domain, type, protocol);
-    if (socket)
-        socket->impl = impl;
+    opts = (struct sol_socket_ip_options *)options;
```
barbieri: no casts

src/samples/network/echo-client.c
```diff
@@ -0,0 +1,178 @@
...
+        goto err;
+    }
+
+    options.base.data = argv[optind];
+    if (!options.base.data) {
+        fprintf(stderr, "ERROR: No data was given, use -h for help\n");
+        goto err;
+    }
+
+    sock = sol_socket_new((struct sol_socket_options *)&options);
```
barbieri: no casts, get the base.

src/samples/network/echo-client.c
```diff
@@ -0,0 +1,178 @@
...
+    if (!sock) {
+        fprintf(stderr, "ERROR: Could not create the socket\n");
+        goto err;
+    }
+
+    address.family = SOL_NETWORK_FAMILY_INET6;
+
+    itr = strstr(addr, "-");
```
barbieri: minor, but `strchr()`

barbieri: also `-` is not standard to split port and address. Usually it is `':'`.

Then `echo` service has a specific port, 7 as per `/etc/services`. It would be nice to default to that and also implement the actual echo service protocol (if there is anything special)

ceolin:
merged, thanks

PR #1873, opened 2016-04-13
composed:http:client: Fix wrong check

ceolin:
The generated code was checking if the response content was empty.
This check is not valid because the case of a POST the response has no content.


cabelitos:
+1

ceolin:
merged, thanks

Issue #1826, opened 2016-04-05
bindings/nodejs doesn't install


barbieri:
Although `npm` usage is not to install its files to the system, soletta expects all is placed at `build/soletta_sysroot/`, using given `PREFIX` such as `build/soletta_sysroot/$(PREFIX)/lib/node_modules` so our `make install` target will place the files at the proper location (which will account for `DESTDIR` in package environments).

Then after `Makefile.targets` generates the build, we must copy the required files to `build/soletta_sysroot`, actually the tests and samples should run based on that copy.

Failing to do so will not allow us to create packages for our bindings.

I'm assigning this to Gabriel as he seems to know best the nodejs build system, but feel free to reassign.

PR #1875, opened 2016-04-13
Node.js bindings: Add the JS APIs to the root and the package to the sysroot

gabrielschulhof:
Fixes gh-1826

This PR replaces https://github.com/solettaproject/soletta/pull/1861 and adds scripting that copies the npm package to the sysroot.

### DIFF
--- a/binding.gyp
+++ b/binding.gyp
@@ -38,6 +38,21 @@
 					} ]
 				},
 				{
+					"target_name": "copyapis",
+					"type": "none",
+					"actions": [ {
+						"action_name": "copyapis",
+						"message": "Copying JS APIs",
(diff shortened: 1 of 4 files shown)


barbieri:
looks okay, but I'm not aware of the nodejs behavior... so there I just trust you :-)

gabrielschulhof:
@barbieri So, is it OK if I land it?

barbieri:
yes, if build/sysroot_soletta is usable as "/" by nodeapps, like iotivity-node on Ostro, then it should be merged.

gabrielschulhof:
Yeah, I tested that scenario. I'll land it in the morning.

gabrielschulhof:
Thanks for reviewing!

PR #1876, opened 2016-04-13

nagineni:

### DIFF
--- a/bindings/nodejs/src/functions/spi.cc
+++ b/bindings/nodejs/src/functions/spi.cc
@@ -1,0 +1,166 @@
+/*
+ * This file is part of the Soletta Project
+ *
+ * Copyright (C) 2015 Intel Corporation. All rights reserved.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+#include <node.h>
+#include <nan.h>
+#include <sol-spi.h>
+
+#include "../common.h"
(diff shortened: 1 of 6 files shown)


bindings/nodejs/src/functions/spi.cc
```diff
@@ -0,0 +1,166 @@
...
+    if (tx)
+        txBuffer = Nan::NewBuffer((char *)tx, status).ToLocalChecked();
+    else
+        txBuffer = Nan::Null();
+
+    if (rx)
+        rxBuffer = Nan::NewBuffer((char *)rx, status).ToLocalChecked();
+    else
+        rxBuffer = Nan::Null();
+
```
cabelitos: Hey, I think this still quite wrong. The last version was almost correct.
You should do something like.
```C
if (status >= 0) {
  txBuffer = Nan::NewBuffer((char *)tx, status).ToLocalChecked();
  rxBuffer = Nan::NewBuffer((char *)rx, status).ToLocalChecked();
} else {
  free(tx);
  free(tx);
  txBuffer = Nan::Null();
  rxBuffer = Nan::Null();
}
`

File: bindings/nodejs/src/functions/spi.cc [hunk 2]
```diff
@@ -0,0 +1,166 @@
...
+    Nan::SetInternalFieldPointer(jsSpi, 0, 0);
+}
+
+static void sol_spi_transfer_cb(void *cb_data, struct sol_spi *spi,
+    const uint8_t *tx, uint8_t *rx, ssize_t status) {
+    Nan::HandleScope scope;
+    Nan::Callback *callback = (Nan::Callback *)cb_data;
+    Local<Value> txBuffer, rxBuffer;
+
+    if (tx)
```
cabelitos: This will never be null

bindings/nodejs/src/functions/spi.cc
```diff
@@ -0,0 +1,166 @@
...
+        txBuffer = Nan::NewBuffer((char *)tx, status).ToLocalChecked();
+    else
+        txBuffer = Nan::Null();
+
+    if (rx)
```
cabelitos: this will never be null

bindings/nodejs/src/functions/spi.cc
nagineni: Ah! Got the point. We have null checks for tx & rx in bind_sol_spi_transfer.

PR #1880, opened 2016-04-14
Socket fixes

ceolin:
- set api version in dtls
- Fix echo client (avoid writing the same data several times)
- Fix dtls's sendmsg() signature


otaviobp:
lgtm

ceolin:
merged, thanks

PR #1882, opened 2016-04-14
coap: fix coverity-pointed error

glima:
CID 143930.

Since sol_socket_sendmsg()'s return type changed to ssize_t and it's
we're now storing its return value on ret var, fix the error check below
it.


otaviobp:
+1

glima:
Thanks, merging.

PR #1884, opened 2016-04-14
socket: Set errno error stating that method is not implemented

otaviobp:
In Oic server and client we need to know if secure server creation has
failed because it is not enabled or if an error occurred. If we have a
real error we need to abort server creation. If it is not supported, we
need to start an insecure server.

### DIFF
--- a/src/lib/comms/sol-socket.c
+++ b/src/lib/comms/sol-socket.c
@@ -51,6 +51,7 @@ sol_socket_ip_new(const struct sol_socket_options *options)
 #ifdef DTLS
         s = sol_socket_default_dtls_new(options);
 #else
+        errno = ENOSYS;
         SOL_WRN("DTLS is not enabled, secure socket is not possible");
         return NULL;
 #endif


ceolin:
:+1:

src/lib/comms/sol-socket.c
```diff
@@ -51,6 +51,7 @@ sol_socket_ip_new(const struct sol_socket_options *options)
 #ifdef DTLS
         s = sol_socket_default_dtls_new(options);
 #else
+        errno = ENOSYS;
         SOL_WRN("DTLS is not enabled, secure socket is not possible");
         return NULL;
```
barbieri: ok, but you should also check `#ifdef DTLS` before trying to set the secure flag in coap/oic.

otaviobp: I am not checking for #ifdef DTLS in oic. I count on coap to check that.
Only using ifdefs in oic in security module. Is it better to check for DTLS in oic creation and don't even trying to create a secure coap connection? Mixed feelings about that.

otaviobp: If we do that, we don't need to apply this patch

ceolin:
The patch itself is good to me. But i agree that the DTLS  (or other socket security) check could be done before trying to use.

otaviobp:
merged

PR #1889, opened 2016-04-14
Uart improvements

cabelitos:


src/lib/io/include/sol-uart.h
```diff
@@ -81,22 +89,27 @@ enum sol_uart_parity {
...
+/**
+ * @brief A configuration struct used to set the UART paramenters.
+ *
+ * @see sol_uart_open()
+ */
 struct sol_uart_config {
 #ifndef SOL_NO_API_VERSION
 #define SOL_UART_CONFIG_API_VERSION (1)
-    uint16_t api_version;
+    uint16_t api_version; /**< The API version */
```
barbieri: use the usual text relating it to `SOL_UART_CONFIG_API_VERSION` in runtime...

src/lib/io/include/sol-uart.h
```diff
@@ -81,22 +89,27 @@ enum sol_uart_parity {
...
-    enum sol_uart_stop_bits stop_bits;
-    void (*rx_cb)(void *user_data, struct sol_uart *uart, uint8_t byte_read); /** Set a callback to be called every time a character is received on UART */
-    const void *rx_cb_user_data;
-    bool flow_control; /** Enables software flow control(XOFF and XON) */
+    enum sol_uart_baud_rate baud_rate; /**< The baud rate value */
+    enum sol_uart_data_bits data_bits; /**< The data bits value */
+    enum sol_uart_parity parity; /**< The parity value*/
+    enum sol_uart_stop_bits stop_bits; /**< The stop bits value */
+    void (*rx_cb)(void *user_data, struct sol_uart *uart); /** Set a callback to be called every time a character is received on UART */
+    const void *rx_cb_user_data; /**< User data to @c rx_cb */
```
barbieri: you can use simply `user_data`, if more callbacks are added they can use the same data.

barbieri: also move the cb up, leave the other enums/bool close.

src/lib/io/include/sol-uart.h
```diff
@@ -238,7 +251,18 @@ void sol_uart_close(struct sol_uart *uart);
  * @note Caller should guarantee that tx buffer will not be freed until
  * callback is called.
  */
-bool sol_uart_write(struct sol_uart *uart, const uint8_t *tx, unsigned int length, void (*tx_cb)(void *data, struct sol_uart *uart, uint8_t *tx, int status), const void *data);
+int sol_uart_write(struct sol_uart *uart, const uint8_t *tx, size_t length, void (*tx_cb)(void *data, struct sol_uart *uart, uint8_t *tx, int status), const void *data);
```
barbieri: as per today's talk in person, I'm wondering if we should convert these to the same pattern, taking a blob and a single CB given to config.

@ceolin is doing the socket, at least trying to unify with that. IMO the blob is a good trade-off for memory consumption.

src/lib/io/include/sol-uart.h
```diff
@@ -238,7 +251,18 @@ void sol_uart_close(struct sol_uart *uart);
...
+/**
+ * @brief Read from UART bus
+ *
+ * @param uart The UART bus handle
+ * @param rx Where to store the UART data
+ * @param length number of bytes to read
+ *
+ * @return the number of bytes read or negative errno on error.
+ */
+int sol_uart_read(struct sol_uart *uart, uint8_t *rx, size_t length);
```
barbieri: sol_buffer! and document the flags, you may take an already initialized buffer, if fixed capacity you read up to that, if it's not, then you can allocate to return all that is available.

src/lib/io/sol-uart-impl-linux.c
```diff
@@ -40,15 +41,15 @@ SOL_LOG_INTERNAL_DECLARE_STATIC(_log_domain, "uart");
...
     int fd;
     struct {
-        void *rx_fd_handler;
-        void (*rx_cb)(void *data, struct sol_uart *uart, uint8_t byte_read);
+        struct sol_fd *rx_fd_handler;
+        void (*rx_cb)(void *data, struct sol_uart *uart);
         const void *rx_user_data;
 
-        void *tx_fd_handler;
+        struct sol_fd *tx_fd_handler;
```
barbieri: as @ceolin did in socket, use a single `sol_fd` and set the flags.

src/lib/io/include/sol-uart.h
cabelitos: To be honest I prefer a blob as well.

src/lib/io/include/sol-uart.h
cabelitos: ok

src/lib/io/sol-uart-impl-linux.c
cabelitos: ok

ceolin:
@cabelitos i think is better you squash the last commits. All of them are changing the api i think it's better you put in the same commit

cabelitos:
#1892

src/lib/io/include/sol-uart.h
zolkis: I am late to this discussion, but if there is one callback given to config, and suppose call write 10 times after each other, how do you associate the callback to which write exactly failed? Does the write pass some user data to the callback from which the client can figure out which write went wrong?

zolkis: Actually the transmit callback API is more efficient IMHO: less code to write on client side, less things to maintain, and the API implementation doesn't care whether to store user data or a function pointer. 

On the receive side, IMHO the best would be to have a flag like ```new_data_available``` (containing the size of the available data), and a notification when it is **set**. Not a notification (or callback) every time there is data, which can be very frequent. The client decides when to do the read (bulk read even), for better efficiency. For instance it can start a timer and read the data once per second. If all data is read, the flag is reset.

barbieri: see PR #1910 with improved behavior. The goal is to find a common pattern for all functions that may have pending information and will asynchronously deliver progressive data to you, so HTTP, UART, socket...

In order to save memory on small systems, instead of keeping a buffer that is realloc()ed to fit pending memory, we take memory segments (blobs) that can be reference counted and have hierarchy (so one can have windows/slices, take references to that and it keeps references to parent). It's a bit more cumbersome to use, but at least making it a common pattern that is efficient on small OS helps. Usability we can help with wrappers.

This also address you request: which data was written/failed? You can use the blob to identify that.

As for one-callback, this is the standard use case where you have only one callback per instance. Thus makes it easier to use and consumes less memory.

barbieri: ah, it is worth to mention we're trying to have the pattern to fulfill the requirement: fast data producers must have a way to know the pending data is piling up so they can stop producing data.

Like "cat /dev/zero | slow-reader", once the pipe is filled, the `cat` process will be stuck into blocking `write()`. We can't just block in soletta due the main loop, so at least a way to notify data wasn't fully sent must be available.

PR #1891, opened 2016-04-15
coap: fix leaks on error cases [v2]

glima:
Coverity CID 144066

-----

Changes since v1:
- new patch fixing unrelated issue pointed out by @vcgomes.

### DIFF
--- a/src/lib/comms/sol-coap.c
+++ b/src/lib/comms/sol-coap.c
@@ -1214,12 +1214,14 @@ well_known_get(struct sol_coap_server *server,
     r = sol_coap_header_set_code(resp, SOL_COAP_RSPCODE_CONTENT);
     if (r < 0) {
         SOL_WRN("Failed to set header code on packet %p", resp);
+        sol_coap_packet_unref(resp);
         return -EINVAL;
     }
 
     r = coap_get_header_len(resp);
     if (r < 0) {
         SOL_WRN("Failed to get header len from packet %p", resp);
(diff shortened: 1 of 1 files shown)


src/lib/comms/sol-coap.c
```diff
@@ -1214,12 +1214,14 @@ well_known_get(struct sol_coap_server *server,
...
     if (r < 0) {
         SOL_WRN("Failed to set header code on packet %p", resp);
+        sol_coap_packet_unref(resp);
         return -EINVAL;
     }
 
     r = coap_get_header_len(resp);
     if (r < 0) {
         SOL_WRN("Failed to get header len from packet %p", resp);
+        sol_coap_packet_unref(resp);
```
vcgomes: This patch is correct. But the more correct solution for this would be to make this use `sol_coap_get_payload()`, but I will leave that as optional.

glima: you mean %s and the payload? or %p and that?

vcgomes: Sorry for the delay, I was thinking about `coap_get_header_len()`, that it can be replaced by `sol_coap_get_payload()`. But it is unrelated to the patch/

glima: Thanks, new patch added for that.

vcgomes:
Looking good. The CI errors look unrelated, please confirm.

glima:
Indeed they are. Thanks, merging.

PR #1892, opened 2016-04-15
Uart API changes v2

cabelitos:


src/lib/io/include/sol-uart.h
```diff
@@ -224,21 +239,35 @@ struct sol_uart *sol_uart_open(const char *port_name, const struct sol_uart_conf
 void sol_uart_close(struct sol_uart *uart);
 
 /**
- * @brief Perform a UART asynchronous transmission.
+ * @brief Perform an UART asynchronous transmission.
+ *
+ * @param uart The UART bus handle
+ * @param blob The blob to be written
+ * @return 0 on success negative errno on error
+ *
```
barbieri: beef up this doc. Say that this function will take a reference of the blob and right before the blob is automatically unreferrenced (remove the extra reference), `tx_cb` is called so users can know when a specific blob was fully written. Say that you can query pending amount of bytes with `sol_uart_get_pending_bytes()` in cases where the producer is faster than the UART.

Also link this to a general doxygen section on this pattern. Then link this with crypto `sol_message_crypto` and http progressive (when it is converted). This will help us to ensure the pattern is reused in the future (and keep an eye when people send PR with patterns matching this one, you'll be responsible to keep consistency!)

src/lib/io/include/sol-uart.h
```diff
@@ -224,21 +239,35 @@ struct sol_uart *sol_uart_open(const char *port_name, const struct sol_uart_conf
...
+ */
+int sol_uart_write(struct sol_uart *uart, struct sol_blob *blob);
+
+/**
+ * @brief Perform an UART synchronous read
+ *
+ * @param uart The UART bus handle
+ * @param rx Where to store the UART data
+ * @param length number of bytes to read
```
barbieri: docs makes no sense, and document what is expected from buf (it must be initialized and all). Also document this "read to buffer" in a general section since this is repeating in many places.

barbieri: and of course, link this to `rx_cb` to say how to know when it can be read or not.

I'm thinking if we need to offer monitor enable/disable like `socket` did. @ceolin?

src/lib/io/include/sol-uart.h
```diff
@@ -224,21 +239,35 @@ struct sol_uart *sol_uart_open(const char *port_name, const struct sol_uart_conf
...
- * success the status parameter on tx_cb should be equal to length otherwise
- * an error happen during the transmission
- * @param data the first parameter of tx_cb
- * @return true if transfer was started
- *
- * @note Caller should guarantee that tx buffer will not be freed until
- * callback is called.
- */
-bool sol_uart_write(struct sol_uart *uart, const uint8_t *tx, unsigned int length, void (*tx_cb)(void *data, struct sol_uart *uart, uint8_t *tx, int status), const void *data);
+ * @param pending_bytes Where to store the pending bytes amount
```
barbieri: should we document this is the sum of all pending blobs? or that's internal detail?

src/lib/io/sol-uart-impl-linux.c
```diff
@@ -32,49 +33,70 @@
...
+
     if (active_flags & SOL_FD_FLAGS_IN) {
-        uint8_t c; /* Backing storage for next sol_buffer */
-        struct sol_buffer buf = SOL_BUFFER_INIT_FLAGS(&c, sizeof(c),
-            SOL_BUFFER_FLAGS_MEMORY_NOT_OWNED | SOL_BUFFER_FLAGS_NO_NUL_BYTE);
-        int status = sol_util_fill_buffer(uart->fd, &buf, 1);
-        if (status > 0)
-            uart->async.rx_cb((void *)uart->async.rx_user_data, uart, c);
+        uart->rx_cb((void *)uart->user_data, uart);
+        return true;
```
barbieri: if we can read & write, why return here and force yet-another mainloop iteration?

src/lib/io/sol-uart-impl-linux.c
```diff
@@ -32,49 +33,70 @@
...
+    }
+
+    //SOL_FD_FLAGS_OUT
+
+    blob = sol_ptr_vector_get_nocheck(&uart->pending_blobs, 0);
+    r = write(uart->fd, (char *)blob->mem + uart->written,
+        blob->size - uart->written);
+
+    if (r < 0) {
+        if (errno != EAGAIN) {
```
barbieri: `EINTR` is also okay t retry.

src/lib/io/sol-uart-impl-linux.c
```diff
@@ -32,49 +33,70 @@
...
+            int aux_errno = errno;
+            SOL_WRN("Could not write at the UART fd: %d - Reason:%s", uart->fd,
+                sol_util_strerrora(aux_errno));
+            if (uart->tx_cb)
+                uart->tx_cb((void *)uart->user_data, uart, blob, -aux_errno);
+        }
```
barbieri: leaks blob reference! You must take a reference on `sol_uart_write()` and unref it here when it 's unused.

also, you must pop the pending blob from list. If you will retry, then I do not think we should callback with the actual blob, use `NULL` in the `tx_cb`,

src/lib/io/sol-uart-impl-linux.c
```diff
@@ -32,49 +33,70 @@
...
+        return false;
+    }
+    uart->written += r;
+    if (uart->written == blob->size) {
+        if (uart->tx_cb)
+            uart->tx_cb((void *)uart->user_data, uart, blob, 0);
```
barbieri: if `sol_uart_close()` is called from `tx_cb`, then it crashes. You should postpone this as the last thing to do (of course move `sol_blob_unref(blob)` as well).

src/lib/io/sol-uart-impl-linux.c
```diff
@@ -172,72 +200,90 @@ sol_uart_open(const char *port_name, const struct sol_uart_config *config)
...
-    return false;
+    r = ioctl(uart->fd, FIONREAD, &total);
+    SOL_INT_CHECK(r, < 0, r);
+    return (int)sol_util_fill_buffer(uart->fd, buf, total);
 }
 
-SOL_API bool
-sol_uart_write(struct sol_uart *uart, const uint8_t *tx, unsigned int length, void (*tx_cb)(void *data, struct sol_uart *uart, uint8_t *tx, int status), const void *data)
+SOL_API int
+sol_uart_get_pending_bytes(struct sol_uart *uart, size_t *pending_bytes)
```
barbieri: at a second look, this function should reflect it's write bytes, not read bytes. So rename to `sol_uart_get_pending_write_bytes()`

src/lib/io/sol-uart-impl-riot.c
```diff
@@ -29,57 +29,58 @@
...
     struct sol_uart *uart = arg;
+    int r;
 
-    if (!uart->async.rx_cb)
-        return;
-    uart->async.rx_cb((void *)uart->async.rx_user_data, uart,
-        (uint8_t)data);
+    r = sol_buffer_append_char(&uart->rx_buffer, (char)data);
+    SOL_INT_CHECK(r, < 0);
+    uart->rx_cb((void *)uart->user_data, uart);
```
barbieri: what if we use a 0-timeout to dispatch this? Then we can accumulate all that comes from `uart_rx_cb` (scheduled from uart thread) and dispatch to user in one go. It can be something like "if not 0-timeout scheduled, schedule one".

src/lib/io/sol-uart-impl-riot.c
```diff
@@ -29,57 +29,58 @@
...
+uart_tx_dispatch(struct sol_uart *uart, struct sol_blob *blob, int status)
 {
-    uint8_t *tx = (uint8_t *)uart->async.tx_buffer;
-
-    uart->async.tx_buffer = NULL;
-    if (!uart->async.tx_cb)
+    if (!uart->tx_cb)
         return;
-    uart->async.tx_cb((void *)uart->async.tx_user_data, uart, tx, status);
+    uart->tx_cb((void *)uart->user_data, uart, blob, status);
```
barbieri: unref blob?

barbieri: ah, forget about it, that is done at `uart_tx_cb`

barbieri: @ibriano is this a safe context to call the user? main thread?

src/lib/io/sol-uart-impl-riot.c
```diff
@@ -117,50 +118,124 @@ sol_uart_open(const char *port_name, const struct sol_uart_config *config)
...
+            &uart->handler);
+        SOL_INT_CHECK_GOTO(ret, != 0, fail);
+        uart->rx_cb = config->rx_cb;
+    }
+
+    uart->tx_cb = config->tx_cb;
+    uart->user_data = config->user_data;
+    sol_ptr_vector_init(&uart->pending_blobs);
+    sol_buffer_init(&uart->rx_buffer);
+    uart->rx_buffer.flags |= SOL_BUFFER_FLAGS_NO_NUL_BYTE;
```
barbieri: why not `init_flags()`?

src/lib/io/sol-uart-impl-riot.c
```diff
@@ -29,57 +29,58 @@
...
+    struct sol_blob *blob;
 
-    uart_write(uart->id, uart->async.tx_buffer, uart->async.tx_length);
-    uart->async.tx_writer = NULL;
-    uart_tx_dispatch(uart, uart->async.tx_length);
+    blob = sol_ptr_vector_take(&uart->pending_blobs, 0);
 
-    return false;
+    uart_write(uart->id, blob->mem, blob->size);
+    uart_tx_dispatch(uart, blob, 0);
```
barbieri: if tx_cb closes the uart, the rest of this function will break.

src/lib/io/sol-uart-impl-riot.c
```diff
@@ -117,50 +118,124 @@ sol_uart_open(const char *port_name, const struct sol_uart_config *config)
...
+
+    SOL_NULL_CHECK(uart, -EINVAL);
+    SOL_NULL_CHECK(buf, -EINVAL);
+
+    if (!uart->rx_buffer.used)
+        return 0;
+
+    if (buf->flags & SOL_BUFFER_FLAGS_FIXED_CAPACITY) {
+        min = sol_min(uart->rx_buffer.used, (buf->capacity - buf->used));
+        memcpy(sol_buffer_at_end(buf), uart->rx_buffer.data, min);
```
barbieri: nul_byte is not being ensured if it's wanted. Use `append_bytes()` as it should do it.

src/lib/io/sol-uart-impl-riot.c
```diff
@@ -117,50 +118,124 @@ sol_uart_open(const char *port_name, const struct sol_uart_config *config)
...
+    } else {
+        min = uart->rx_buffer.used;
+        r = sol_buffer_append_buffer(buf, &uart->rx_buffer);
```
barbieri: eh, I'm thinking if here we should do a nasty hack to save memory: you could swap `buf` and `uart->rx_bufer`, then no malloc is done at this point.

barbieri: just a note, we'd have to guarantee all flags match before doing that, such as nul-byte.

src/samples/common/uart.c
```diff
@@ -20,104 +20,213 @@
...
-    uart_data->rx_buffer[uart_data->rx_index] = (char)read_char;
-    if (read_char == 0) {
-        uart_data->rx_index = 0;
-        printf("Data received on UART%d: %s\n",
-            uart_data->id, uart_data->rx_buffer);
-        missing_strings--;
-        if (missing_strings == 0)
+    struct sol_str_slice slice;
+
+    slice.data = (const char *)blob->mem;
```
barbieri: `sol_str_slice_from_blob()`

src/lib/io/sol-uart-impl-linux.c
ibriano: On Sat, 16 Apr 2016 08:00:51 -0700, Gustavo Sverzut Barbieri wrote:
> > -            SOL_BUFFER_FLAGS_MEMORY_NOT_OWNED | SOL_BUFFER_FLAGS_NO_NUL_BYTE);
> > -        int status = sol_util_fill_buffer(uart->fd, &buf, 1);
> > -        if (status > 0)
> > -            uart->async.rx_cb((void *)uart->async.rx_user_data, uart, c);
> > +        uart->rx_cb((void *)uart->user_data, uart);
> > +        return true;
> > +    }
> > +
> > +    //SOL_FD_FLAGS_OUT
> > +
> > +    blob = sol_ptr_vector_get_nocheck(&uart->pending_blobs, 0);
> > +    r = write(uart->fd, (char *)blob->mem + uart->written,
> > +        blob->size - uart->written);
> > +
> > +    if (r < 0) {
> > +        if (errno != EAGAIN) {
> 
> `EINTR` is also okay t retry.

In fact, EINTR should be retried immediately.

src/lib/io/sol-uart-impl-riot.c
ibriano: Yes

src/lib/io/sol-uart-impl-linux.c
cabelitos: Indeed.

src/lib/io/sol-uart-impl-linux.c
cabelitos: Forgot about it.

src/lib/io/sol-uart-impl-linux.c
cabelitos: I did an `sol_blob_ref()` on `sol_uart_write()`, I just forgot to unref it here :[

src/lib/io/sol-uart-impl-linux.c
cabelitos: Ok

src/lib/io/sol-uart-impl-linux.c
cabelitos: Ok

src/lib/io/sol-uart-impl-riot.c
cabelitos: I will replace this call. and sol_blob_unref as well.

src/lib/io/sol-uart-impl-riot.c
cabelitos: I will take care of that. Forgot about it.

cabelitos: Ok

src/lib/io/sol-uart-impl-riot.c
cabelitos: Indeed ! :new_moon_with_face:

src/samples/common/uart.c
cabelitos: Ok

src/lib/io/sol-uart-impl-riot.c
cabelitos: No special reason. I will change it.

src/lib/io/include/sol-uart.h
cabelitos: Yeah, we should. I will add it

src/lib/io/sol-uart-impl-riot.c
cabelitos: Do we have a guarantee that this will be called before the timer?

src/lib/io/include/sol-uart.h
cabelitos: I forgot to change the doc.
I'm sorry

src/lib/io/sol-uart-impl-riot.c
ibriano: Timers are processed before OS events, if you have several UART IRQ queued in the main loop, they will be called one after the other until there are no more events, and only then the loop iterates again, so if you have several bytes to read, you can add a timer to call the user callback for all those bytes instead of calling them several times.

src/lib/io/include/sol-uart.h
cabelitos: I will add this info. I guess we should mention that.

src/lib/io/include/sol-uart.h
cabelitos: I don't think it's really necessary to offer the enable/disable functions.

src/lib/io/sol-uart-impl-riot.c
cabelitos: I will add that

cabelitos:
#1900

PR #1893, opened 2016-04-15
oic: Improve query handling in oic server and client

otaviobp:


src/lib/comms/sol-oic-server.c
```diff
@@ -183,12 +183,41 @@ static const struct sol_oic_resource_type oic_p_resource_type = {
...
+
+static bool
+oic_query_split(struct sol_str_slice query, struct sol_str_slice *key, struct sol_str_slice *value)
+{
+    char *sep;
+
+    sep = memchr(query.data, '=', query.len);
+    SOL_NULL_CHECK(sep, false);
+    *key = SOL_STR_SLICE_STR(query.data, sep - query.data);
+    *value = SOL_STR_SLICE_STR(sep + 1, query.len - (sep - query.data + 1));
```
vcgomes: The case that the received query is something like `if=` could cause problems, no?

otaviobp: No, it won't fail. value->len will be 0, so it will be ignored.

vcgomes: Cool. So please make `sep` const. (I really don't like these libc functions that drop the const)

otaviobp:
merged

PR #1894, opened 2016-04-15
Socket buffer v2

ceolin:
from v1:
- Add macro to check if is possible resize the buffer
- Fix to build using g++
- Export sol_min/max
- Handle null byte flag properly

### DIFF
--- a/src/lib/comms/sol-socket-impl-linux.c
+++ b/src/lib/comms/sol-socket-impl-linux.c
@@ -25,6 +25,7 @@
 #include "sol-mainloop.h"
 #include "sol-network.h"
 #include "sol-network-util.h"
+#include "sol-util.h"
 #include "sol-util-internal.h"
 
 #include "sol-socket.h"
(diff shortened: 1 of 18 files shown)


barbieri:
+1

src/lib/comms/sol-socket-impl-linux.c
```diff
@@ -289,13 +301,13 @@ is_multicast(enum sol_network_family family, const struct sockaddr *sockaddr)
...
-sol_socket_linux_sendmsg(struct sol_socket *socket, const void *buf, size_t len,
+sol_socket_linux_sendmsg(struct sol_socket *socket, const struct sol_buffer *buf,
     const struct sol_network_link_addr *cliaddr)
 {
     struct sol_socket_linux *s = (struct sol_socket_linux *)socket;
     uint8_t sockaddr[sizeof(struct sockaddr_in6)] = { };
-    struct iovec iov = { .iov_base = (void *)buf,
-                         .iov_len = len };
+    struct iovec iov = { .iov_base = (void *)buf->data,
+                         .iov_len = buf->used };
```
barbieri: should we check nul_byte here and use `+1`? likely not... but shouldn't hurt if people use the proper flags (or use `NO_NUL_BYTE` for binary protocols)

edersondisouza:
looks ok

ceolin:
merged, thanks

PR #1895, opened 2016-04-15
oic: Make oic compatible with IoTivity 1.1

otaviobp:


src/shared/include/sol-util.h
```diff
@@ -372,19 +372,25 @@ int sol_util_uint32_mul(const uint32_t op1, const uint32_t op2, uint32_t *out);
  *
  * @return 0 on success, negative error code otherwise.
  */
-int sol_util_uuid_gen(bool upcase, bool with_hyphens, char id[SOL_STATIC_ARRAY_SIZE(37)]);
+int sol_util_uuid_gen(bool uppercase, bool with_hyphens, struct sol_buffer *uuid_buf);
```
glima: doxy check will fail, please fix the docstring

src/shared/include/sol-util.h
```diff
@@ -372,19 +372,25 @@ int sol_util_uint32_mul(const uint32_t op1, const uint32_t op2, uint32_t *out);
...
  * @return @c true if it's valid, @c false otherwise.
  */
-bool sol_util_uuid_str_valid(const char *str);
+bool sol_util_uuid_str_is_valid(const struct sol_str_slice uuid);
+
+SOL_API int
+sol_util_uuid_string_from_bytes(bool uppercase, bool with_hyphens, const uint8_t uuid_bytes[SOL_STATIC_ARRAY_SIZE(16)], struct sol_buffer *uuid_str);
+
+int
+sol_util_uuid_bytes_from_string(struct sol_str_slice uuid_str, struct sol_buffer *uuid_bytes);
```
glima: No docs?

src/lib/comms/sol-oic-server.c
```diff
@@ -95,7 +95,13 @@ static sol_coap_responsecode_t
 _sol_oic_server_d(const struct sol_network_link_addr *cliaddr, const void *data,
     const struct sol_oic_map_reader *input, struct sol_oic_map_writer *output)
 {
+    SOL_BUFFER_DECLARE_STATIC(dev_id, 37);
     bool ret;
+    int r;
+
+    r = sol_util_uuid_string_from_bytes(true, true,
+        sol_platform_get_machine_id_as_bytes(), &dev_id);
+    SOL_INT_CHECK_GOTO(r, < 0, error);
```
glima: I don't get from this patch why this would not be filled elsewhere anymore... Can we do it at one place only?

src/lib/comms/sol-oic-server.c
```diff
@@ -189,6 +189,28 @@ static const struct sol_oic_resource_type oic_p_resource_type = {
...
+encode_array_from_bsv(CborEncoder *map, const char *val)
+{
+    const char *p, *last;
+    CborEncoder array;
+    CborError err;
+
+    err = cbor_encoder_create_array(map, &array, CborIndefiniteLength);
+
+    last = val + strlen(val);
+    while (val < last) {
```
barbieri: @cabelitos added a nice `sol_str_slice_split_iterate()` that helps simplifying this kind of loops, consider using it.

src/lib/comms/sol-oic-client.c
```diff
@@ -169,6 +169,21 @@ sol_oic_resource_ref(struct sol_oic_resource *r)
...
+    uint16_t i;
+    struct sol_str_slice *str;
+
+    if (data)
+        free(data);
+    else {
+        SOL_VECTOR_FOREACH_IDX(vector, str, i)
+            free((char *)str->data);
+    }
+    sol_vector_clear(vector);
```
barbieri: this looks like those `strv` (vector of strings) where the actual strings are allocated in a single blob, is this really the case? Usually one comes with that by getting a single string and splitting it, then if enough space (usually the delimiter), you `strdup()` the source string and replace delimiters with `\0`.

If that is the case, then you can avoid the `data` parameter and explicit storage in `r`, document the `[0]` is the memory and you can free in one go.

src/shared/include/sol-util.h
otaviobp: fixing documentation issues.

src/lib/comms/sol-oic-server.c
otaviobp: I'll use it. I didn't know we have that. thanks

src/lib/comms/sol-oic-client.c
otaviobp: Yes and no.
We may have all strings in a single blob or multiple strings. If data is not NULL it is the first case and I can do a single free. If the opposite, I need to free all strings.
Well, I could have spited the blob string to have always the same case in free, but I didn't want to do it.
Can you see another way of checking if I have to free all strings or just the first one? If so, I'll do it.

src/shared/sol-util.c
```diff
@@ -883,20 +910,19 @@ sol_util_uint32_mul(const uint32_t a, const uint32_t b, uint32_t *out)
...
-    } else if (len == 36) {
+    } else if (uuid.len == 36) {
         char c;
-        for (i = 0; i < len; i++) {
-            c = str[i];
+        for (i = 0; i < uuid.len; i++) {
+            c = uuid.data[i];
 
             if (i == 8 || i == 13 || i == 18 || i == 23) {
                 if (c != '-')
```
edersondisouza: Don't we have tests for this functions? Do you think they deserve it?

src/lib/comms/sol-oic-client.c
```diff
@@ -244,12 +244,44 @@ _parse_platform_info_payload(struct sol_oic_platform_information *info,
...
+        return false;
+
+    if (cbor_value_is_byte_string(&value))
+        return cbor_value_dup_byte_string(&value, (uint8_t **)&device_id->data,
+            &device_id->used, NULL) == CborNoError;
+
+    if (cbor_value_is_text_string(&value)) {
+        r = cbor_value_dup_text_string(&value, (char **)&slice.data,
+            &slice.len, NULL) == CborNoError;
+        SOL_INT_CHECK_GOTO(r, < 0, error);
```
edersondisouza: This appears wrong. As `r` receives `==` result, it's going to be `0` or `1`, so it'll never be `< 0`. Maybe you wanted to use something like `SOL_EXP_CHECK_GOTO(r, error)`?
And this brings me another question, if `cbor_value_dup_text_string` fails, may it still allocate `slice.data`? And if it doesn't, `slice` may be uninitialised, so `free` on label `error` may crash.

otaviobp: error handling was wrong. Fixing that. 
And if cbor_value_dup_text_string fails, slice data is freed by function, so we don't need to free that.

src/shared/sol-util.c
otaviobp: I'll add tests to that. Why not?

src/lib/comms/sol-oic-server.c
otaviobp: fixed

src/shared/include/sol-util.h
otaviobp: fixed too

src/shared/include/sol-util.h
```diff
@@ -362,29 +362,57 @@ int sol_util_uint32_mul(const uint32_t op1, const uint32_t op2, uint32_t *out);
...
+ *                     (e.g. "de305d54-75b4-431b-adb2-eb6b9e546014") or
+ *                     without them.
+ * @param uuid_bytes A 16 byte array containing the UUID in byte format.
+ * @param uuid_str n initialized buffer to be used to append the converted uuid.
+ *        It will have 36 bytes of lenght if with_hyphens is true or 32 bytes
+ *        of lenght if with_hyphens is false.
+ *
+ * @return 0 on success, negative error code otherwise.
+ */
+int sol_util_uuid_string_from_bytes(bool uppercase, bool with_hyphens, const uint8_t uuid_bytes[SOL_STATIC_ARRAY_SIZE(16)], struct sol_buffer *uuid_str);
```
cabelitos: Why not a slice or another buffer for `uuid_bytes` ?

src/lib/comms/sol-oic-client.c
```diff
@@ -1045,11 +1076,11 @@ _observe_with_polling(struct sol_oic_client *client, struct sol_oic_resource *re
     void *data)
 {
     struct resource_request_ctx *ctx = sol_util_memdup(&(struct resource_request_ctx) {
```
cabelitos: Do you need this memdup?

cabelitos: I guess you can do it without it

src/lib/comms/sol-oic-server.c
```diff
@@ -189,6 +189,24 @@ static const struct sol_oic_resource_type oic_p_resource_type = {
 };
 
 static CborError
+encode_array_from_bsv(CborEncoder *map, const char *val)
+{
+    const char *p = NULL;
+    struct sol_str_slice token, slice = sol_str_slice_from_str(val),
```
cabelitos: Are you sure `val` can never be nulll?

src/lib/comms/sol-oic-client.c
```diff
@@ -920,11 +951,11 @@ _resource_request(struct sol_oic_client *client, struct sol_oic_resource *res,
     struct sol_network_link_addr addr;
     struct sol_oic_map_writer map_encoder;
     struct resource_request_ctx *ctx = sol_util_memdup(&(struct resource_request_ctx) {
```
cabelitos: I guess you can remove this memdup

src/lib/comms/sol-oic-client.c
```diff
@@ -778,10 +809,10 @@ sol_oic_client_find_resource(struct sol_oic_client *client,
     SOL_NULL_CHECK(client, false);
 
     ctx = sol_util_memdup(&(struct find_resource_ctx) {
```
cabelitos: same thing

src/lib/comms/sol-oic-client.c
```diff
@@ -169,6 +169,21 @@ sol_oic_resource_ref(struct sol_oic_resource *r)
     return r;
 }
 
+static void
+clear_vector_list(struct sol_vector *vector, void *data)
```
cabelitos: I think this function is odd. I prefer the older approuch

src/shared/include/sol-util.h
otaviobp: As far as I remember, buffer is used for writing. in this case we are reading. And we have agreed that we would use sol_str_slice for strings only, right?

src/lib/comms/sol-oic-client.c
otaviobp: We are using this context in a sol_timeout, so we need to allocated memory for this.

src/lib/comms/sol-oic-server.c
otaviobp: yes. I check this before calling this function

src/lib/comms/sol-oic-client.c
otaviobp: same as above.

src/lib/comms/sol-oic-client.c
otaviobp: What other approach? Take a look at the comment I did to @barbieri. if you have a better idea, I can do it.

otaviobp:
merged

PR #1900, opened 2016-04-18
Uart API changes v3

cabelitos:
Changes since v2
* Added suggestions pointed by @barbieri 

Note: I will add the common docs later.

### DIFF
--- a/src/lib/io/sol-uart-impl-riot.c
+++ b/src/lib/io/sol-uart-impl-riot.c
@@ -29,57 +29,78 @@
 #include "sol-mainloop.h"
 #include "sol-util-internal.h"
 #include "sol-interrupt_scheduler_riot.h"
+#include "sol-util.h"
+#include "sol-vector.h"
 
 SOL_LOG_INTERNAL_DECLARE_STATIC(_log_domain, "uart");
 
 struct sol_uart {
-    uart_t id;
-    struct {
-        void *handler;
-
-        void (*rx_cb)(void *data, struct sol_uart *uart, uint8_t byte_read);
-        const void *rx_user_data;
-
-        void (*tx_cb)(void *data, struct sol_uart *uart, uint8_t *tx, int status);
-        const void *tx_user_data;
-        struct sol_timeout *tx_writer;
-        const uint8_t *tx_buffer;
-        unsigned int tx_length;
-    } async;
+    void (*rx_cb)(void *data, struct sol_uart *uart);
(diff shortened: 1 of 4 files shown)


src/lib/io/sol-uart-impl-riot.c
```diff
@@ -119,48 +140,119 @@ sol_uart_open(const char *port_name, const struct sol_uart_config *config)
...
+    memcpy(buf, &uart->rx_buffer, sizeof(struct sol_buffer));
+
+    sol_buffer_init_flags(&uart->rx_buffer, NULL, 0,
+        SOL_BUFFER_FLAGS_DEFAULT | SOL_BUFFER_FLAGS_NO_NUL_BYTE);
+    buf->flags = flags;
+
+    if (!(buf->flags & SOL_BUFFER_FLAGS_NO_NUL_BYTE)) {
+        r = sol_buffer_ensure_nul_byte(buf);
+        SOL_INT_CHECK(r, < 0, r);
+    }
```
cabelitos: @barbieri Thinking about this one again. I don't think this is the right solution. We should append the whole buffer here and not swap it. If one is using the `buf` to accumulate all the received data, the swap will override the previous read data.

I think we should change the code to
```c
if (buf->flags & SOL_BUFFER_FLAGS_FIXED_CAPACITY) { 
//Leave as is.
}
//The new code
sol_buffer_append_buffer(buf, &uart->rx_buffer);
sol_buffer_reset(&uart->rx_buffer);
return 0;

```

cabelitos: Or
we can do the folliwing:
```c
if (buf->flags & SOL_BUFFER_FLAGS_FIXED_CAPACITY) { 
//Leave as is.
}
//The new code
if (buf->used > 0) {
   sol_buffer_append_buffer(buf, &uart->rx_buffer);
   sol_buffer_reset(&uart->rx_buffer);
   return 0;
}
flags = buf->flags;
memcpy(buf, &uart->rx_buffer, sizeof(struct sol_buffer));
sol_buffer_init_flags(&uart->rx_buffer, NULL, 0,
SOL_BUFFER_FLAGS_DEFAULT | SOL_BUFFER_FLAGS_NO_NUL_BYTE);
buf->flags = flags;
if (!(buf->flags & SOL_BUFFER_FLAGS_NO_NUL_BYTE)) {
   r = sol_buffer_ensure_nul_byte(buf);
   SOL_INT_CHECK(r, < 0, r);
 }
 return 0;
}
```

cabelitos:
#1910

PR #1903, opened 2016-04-19
Memdesc

barbieri:
This PR introduces infrastructure to describe memory segments. The idea is to use this as a common description for packets, options and serialization to/from CBOR in OIC.

So far nothing was converted to use it, I plan to do that in other PRs. First I plan to convert the `struct sol_flow_packet` to use that, then remove custom packet serialization from inspector, console and possibly others. Then I plan to proceed with flow options.

Someone should check converting CBOR/OIC to use that instead of custom conversions. This way `sol-oic-gen.py` would only generate the `struct sol_memdesc` for each resource, the actual conversion would be done automatically. This may expose new needs, such as enumerations, that can be added, let me know.

There is a sample serialization and load from JSON. The serialization was made into a common function that can be reused. The load I didn't think of any way to make it more general.

This is a followup to my PR #271.

### DIFF
--- a/src/lib/datatypes/include/sol-memdesc.h
+++ b/src/lib/datatypes/include/sol-memdesc.h
@@ -25,6 +25,7 @@
 #include <sol-common-buildopts.h>
 #include <sol-str-slice.h>
 #include <sol-macros.h>
+#include <sol-buffer.h>
 
 #ifdef __cplusplus
 extern "C" {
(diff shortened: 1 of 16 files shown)


barbieri:
sol-vector.h indentation was wrong, I did not touch it.

src/lib/datatypes/include/sol-memdesc.h
```diff
@@ -0,0 +1,1010 @@
...
+#endif
+        /**
+         * @brief initialize the defaults of memory.
+         *
+         * If provided, will be used to initialize the memory instead of
+         * the traditional use of struct sol_memdesc::defcontent.
+         *
+         * Should return 0 on success, negative errno on errors.
+         */
+        int (*init_defaults)(const struct sol_memdesc *desc, void *container);
```
edersondisouza: Maybe explain `container` parameter? I could only get, for sure, what it was after reading some of these callbacks - maybe is too soon in the morning as well - hence my troubles. But adding more about it shouldn't hurt.

barbieri: thanks for your comment, I was so biased that I forgot to review this. When I created it I was basically looking at structures and wrote the code assuming that. However the container should only be used for arrays. And with that, some properties like "name" and "offset".

I'll redesign this, have a `struct sol_memdesc_structure_member`, it will be the one with the name, offset and few other flags that doesn't make sense for all memory descriptions. Should also help save memory

barbieri:
See v2 #1909

PR #1907, opened 2016-04-19
oic-gen: Fix invalid code generation

otaviobp:
Read Only property was incorrectly handled in code generation and some functions were being called when they shouldn't


ceolin:
worked as expected :)

bdilly:
@otaviobp afaik @ceolin was the only guy able to reproduce it. If it's fixed for him, I suppose we're good. Commits make sense for me

PR #1909, opened 2016-04-19
Memdesc v2

barbieri:
Changes after v1 (#1903):
 * removed container complexity, it's all about the target memory now;
 * offset, name, detail and required moved to new description for struct members, saving space in the common case and avoiding confusion;
 * array operations is now a pointer, this allows to save memory for non-arrays;
 * Stop abusing children pointer, now each type has its own pointer: pointed_item, array_item and structure_members;
 * exported SOL_MEMDESC_SERIALIZE_OPTIONS_DEFAULT
 * `sol_json_load_memdesc()` now will return `-ENODATA` if required members were not found, but tries to load as much as possible.

After merging this I plan to work on enumerations.

### DIFF
--- a/src/lib/datatypes/include/sol-memdesc.h
+++ b/src/lib/datatypes/include/sol-memdesc.h
@@ -1,0 +1,1089 @@
+/*
+ * This file is part of the Soletta Project
+ *
+ * Copyright (C) 2016 Intel Corporation. All rights reserved.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+#pragma once
+
+#include <errno.h>
+#include <stdbool.h>
+#include <stdint.h>
+#include <stdlib.h>
+#include <sol-common-buildopts.h>
+#include <sol-str-slice.h>
+#include <sol-macros.h>
+
+#ifdef __cplusplus
+extern "C" {
+#endif
+
+/**
+ * @file
+ * @brief These are routines that Soletta provides for its memory description (memdesc) implementation.
+ */
+
+/**
+ * @defgroup MemDesc Memory Description
+ * @ingroup Datatypes
+ *
+ * @brief A memory description (memdesc) allows code to know how to
+ * handle it in runtime, such as decode/parse from some other
+ * representation (text/json), or serialize/encode. It will, as well,
+ * offer special handling such as memory being duplicated and freed
+ * for strings, or defined per-description with
+ * struct sol_memdesc::ops.
+ *
+ * @{
+ */
+
+struct sol_memdesc;
+
+/**
+ * @brief Designates the type of the memory description
+ */
+enum sol_memdesc_type {
+    SOL_MEMDESC_TYPE_UNKNOWN = 0, /**< @brief not to be used. */
+    /**
+     * @brief uint8_t equivalent (one unsigned byte).
+     *
+     * Initial content is specified in
+     * struct sol_memdesc::defcontent::u8.
+     */
+    SOL_MEMDESC_TYPE_UINT8,
+    /**
+     * @brief uint16_t equivalent (two unsigned bytes).
+     *
+     * Initial content is specified in
+     * struct sol_memdesc::defcontent::u16.
+     */
+    SOL_MEMDESC_TYPE_UINT16,
+    /**
+     * @brief uint32_t equivalent (four unsigned bytes).
+     *
+     * Initial content is specified in
+     * struct sol_memdesc::defcontent::u32.
+     */
+    SOL_MEMDESC_TYPE_UINT32,
+    /**
+     * @brief uint64_t equivalent (four unsigned bytes).
+     *
+     * Initial content is specified in
+     * struct sol_memdesc::defcontent::u64.
+     */
+    SOL_MEMDESC_TYPE_UINT64,
+    /**
+     * @brief unsigned long equivalent.
+     *
+     * Initial content is specified in
+     * struct sol_memdesc::defcontent::ul.
+     */
+    SOL_MEMDESC_TYPE_ULONG,
+    /**
+     * @brief size_t equivalent (four or eight unsigned bytes, depends on platform).
+     *
+     * Initial content is specified in
+     * struct sol_memdesc::defcontent::sz.
+     */
+    SOL_MEMDESC_TYPE_SIZE,
+    /**
+     * @brief int8_t equivalent (one signed byte).
+     *
+     * Initial content is specified in
+     * struct sol_memdesc::defcontent::i8.
+     */
+    SOL_MEMDESC_TYPE_INT8,
+    /**
+     * @brief int16_t equivalent (two signed bytes).
+     *
+     * Initial content is specified in
+     * struct sol_memdesc::defcontent::i16.
+     */
+    SOL_MEMDESC_TYPE_INT16,
+    /**
+     * @brief int32_t equivalent (four signed bytes).
+     *
+     * Initial content is specified in
+     * struct sol_memdesc::defcontent::i32.
+     */
+    SOL_MEMDESC_TYPE_INT32,
+    /**
+     * @brief int64_t equivalent (eight signed bytes).
+     *
+     * Initial content is specified in
+     * struct sol_memdesc::defcontent::i64.
+     */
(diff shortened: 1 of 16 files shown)


barbieri:
sol-vector.h indentations errors, did not touch

src/lib/datatypes/sol-memdesc.c
```diff
@@ -0,0 +1,804 @@
...
+            return 0;
+        else
+            return strcmp(*a, *b);
+    }
+    case SOL_MEMDESC_TYPE_PTR: {
+        const void *const *a = a_mem;
+        const void *const *b = b_mem;
+
+        if (!desc->pointed_item || !*a || !*b) {
+            if (*a < *b)
```
vcgomes: https://www.securecoding.cert.org/confluence/display/c/ARR36-C.+Do+not+subtract+or+compare+two+pointers+that+do+not+refer+to+the+same+array

src/lib/datatypes/sol-vector.c
```diff
@@ -139,6 +139,34 @@ sol_vector_del(struct sol_vector *v, uint16_t i)
     return 0;
 }
 
+SOL_API int
+sol_vector_del_range(struct sol_vector *v, uint16_t start, uint16_t len)
+{
+    size_t tail_len;
+
```
cabelitos: Should we add `SOL_NULL_CHECK()` for `v` here?

src/lib/datatypes/include/sol-vector.h
```diff
@@ -198,6 +198,20 @@ sol_vector_del_last(struct sol_vector *v)
...
+ * Removes the range starting at index @a start from the vector and
+ * goes until @a start + @a len.
+ *
+ * @param v Vector pointer
+ * @param start Index of the first element to remove
+ * @param len the number of elements to remover
+ *
+ * @return @c 0 on success, error code (always negative) otherwise
+ */
+int sol_vector_del_range(struct sol_vector *v, uint16_t start, uint16_t len);
```
cabelitos: I guess we should add the same function for sol_ptr_vector. Maybe a static inline function that calsl `sol_vector_del_range()`

src/lib/datatypes/include/sol-memdesc.h
```diff
@@ -0,0 +1,1089 @@
...
+     *      @c sol_memdesc_free_content(desc->array_item, it) for
+     *      every item that will be removed, then call
+     *      sol_vector_del_range(). If growing, call
+     *      sol_vector_append_n() and initialze items with
+     *      @c sol_memdesc_init_defaults(desc->array_item, it).
+     *
+     * @see SOL_MEMDESC_OPS_VECTOR and SOL_MEMDESC_OPS_PTR_VECTOR.
+     */
+    SOL_MEMDESC_TYPE_ARRAY
+};
```
cabelitos: I saw that you added `unsigned long` and `long`. Does it worth adding `int` and `unsigned int `?

src/lib/datatypes/include/sol-memdesc.h
```diff
@@ -0,0 +1,1089 @@
...
+    /**
+     * @brief size in bytes of the member memory.
+     *
+     * Usually this is @c sizeof(type), if a structure it will account
+     * for all members plus paddings.
+     *
+     * This is only used for SOL_MEMDESC_TYPE_STRUCTURE and
+     * SOL_MEMDESC_TYPE_ARRAY.
+     */
+    uint16_t size;
```
cabelitos: Why not a `size_t` ?

src/lib/datatypes/include/sol-memdesc.h
```diff
@@ -0,0 +1,1089 @@
...
+     *
+     * @see sol_memdesc_set_content()
+     */
+    int (*set_content)(const struct sol_memdesc *desc, void *memory, const void *ptr_content);
+    /**
+     * @brief copy the content from another memory.
+     *
+     * If provided, will be used to set the memory instead of the
+     * traditional code that will, for example, strdup() and free()
+     * strings.
```
edersondisouza: This is the very same paragraph from `set_content`. Is this right?

src/lib/datatypes/include/sol-memdesc.h
```diff
@@ -0,0 +1,1089 @@
...
+     * If provided, will be used to free the contents of a memory
+     * instead of the traditional code that will, for example, free()
+     * strings.
+     *
+     * Should return 0 on success, negative errno on errors.
+     *
+     * @see sol_memdesc_free_content()
+     */
+    int (*free_content)(const struct sol_memdesc *desc, void *memory);
+    union {
```
edersondisouza: Why this union of one?

src/lib/datatypes/include/sol-memdesc.h
edersondisouza: Why this should be used only for `SOL_MEMDESC_TYPE_STRUCTURE` and `SOL_MEMDESC_TYPE_STRUCTURE`? Wouldn't be easier ask for size when defining each `sol_memdesc`? Or are you trying to avoid mistakes on this use? I'm afraid that having this size to be used sometimes may lead to more confusion than explicitly asking for it all the time....

src/lib/datatypes/include/sol-memdesc.h
```diff
@@ -0,0 +1,1089 @@
...
+ */
+void *sol_memdesc_get_array_element(const struct sol_memdesc *array_desc, const void *memory, size_t idx);
+
+/**
+ * @brief Resize the length of an array.
+ *
+ * This function must be applied to SOL_MEMDESC_TYPE_ARRAY and will
+ * call struct sol_memdesc::ops::array::resize.
+ *
+ * The returned value is about the number of items according to
```
edersondisouza: What 'returned value'? Return of this function? If so, this description doesn't match `@return` one.

src/lib/datatypes/sol-memdesc.c
barbieri: makes sense. I'll keep NULL x some-pointer in consideration, but no less/greater then as it's not stable.

src/lib/datatypes/sol-vector.c
barbieri: no sol-vector.c does that, so for consistency no.  However we should flag them as `SOL_ATTR_NON_NULL()` and add the checks so we're safe and we hint address sanitizer and optimizations can remove the checks. Can be done in a different commit.

src/lib/datatypes/include/sol-vector.h
barbieri: makes sense.

src/lib/datatypes/include/sol-memdesc.h
barbieri: no as usually they do not change size and you should stick with a fixed version. long is a special case since it can be used to store pointers and other platform dependent features.

src/lib/datatypes/include/sol-memdesc.h
barbieri: it leads to mistake, actually I did that mistake and decided to force types to have known size because of that.

@cabelitos the reason is to keep structure small, size_t may be 4/8 bytes. The enum is usually short and keeping size small will also help fill the hole

src/lib/datatypes/include/sol-memdesc.h
barbieri: yes, it's the way to override that call back implementation

src/lib/datatypes/include/sol-memdesc.h
barbieri: will add another one for enumerations in near future.

src/lib/datatypes/include/sol-memdesc.h
barbieri: oops, copy & pasta. :-)

src/lib/datatypes/sol-memdesc.c
```diff
@@ -0,0 +1,804 @@
...
+SOL_API const struct sol_memdesc_ops SOL_MEMDESC_OPS_PTR_VECTOR = {
+    SOL_SET_API_VERSION(.api_version = SOL_MEMDESC_OPS_API_VERSION, )
+    .init_defaults = ptr_vector_ops_init_defaults,
+    .array = &(const struct sol_memdesc_ops_array){
+        SOL_SET_API_VERSION(.api_version = SOL_MEMDESC_OPS_ARRAY_API_VERSION, )
+        .get_length = vector_ops_get_array_length,
+        .get_element = vector_ops_get_array_element,
+        .resize = vector_ops_resize_array,
+    },
+};
```
edersondisouza: Why implement `SOL_MEMDESC_OPS_VECTOR` and `SOL_MEMDESC_OPS_PTR_VECTOR` here? Imho, they belong more to `sol-vector.[hc]` than here.

barbieri: I thought about this as well, likewise why serialize is here and not in sol-buffer.h

The reasoning I've come with is that these headers are much more used than sol-memdesc.h and forcing the dependency/inclusion will offer no good. Then these constants are provided here.

Also, if the API ever changes, these will be in the same source file and it's easier to fix.

edersondisouza:
Other than that was said, looks ok.

barbieri:
See #1920.

PR #1910, opened 2016-04-19
Uart API changes v4

cabelitos:
Changes since v3:

* remove sol_uart_read()

@barbieri @ceolin, Please could you guys take a look?

### DIFF
--- a/src/lib/io/include/sol-uart.h
+++ b/src/lib/io/include/sol-uart.h
@@ -42,6 +42,13 @@ extern "C" {
  * @{
  */
 
+/**
+ * @struct sol_uart
+ * @brief A handle to a UART device.
+ * @see sol_uart_open()
+ * @see sol_uart_close()
+ * @see sol_uart_write()
+ */
 struct sol_uart;
 
 /**
@@ -58,13 +65,13 @@ enum sol_uart_baud_rate {
 };
 
 /**
- * @brief Amount of stop bits.
+ * @brief Amount of data bits.
  */
 enum sol_uart_data_bits {
-    SOL_UART_DATA_BITS_8 = 0,
-    SOL_UART_DATA_BITS_7,
-    SOL_UART_DATA_BITS_6,
-    SOL_UART_DATA_BITS_5
+    SOL_UART_DATA_BITS_8 = 0, /**< Use 8 data bits */
+    SOL_UART_DATA_BITS_7, /**< Use 7 data bits */
+    SOL_UART_DATA_BITS_6, /**< Use 6 data bits */
+    SOL_UART_DATA_BITS_5 /**< Use 5 data bits */
 };
 
 /**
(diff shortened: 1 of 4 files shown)


src/lib/io/sol-uart-impl-riot.c
```diff
@@ -227,3 +227,22 @@ sol_uart_write(struct sol_uart *uart, struct sol_blob *blob)
...
+    SOL_NULL_CHECK(pending_bytes, -EINVAL);
+
+    *pending_bytes = 0;
+    SOL_PTR_VECTOR_FOREACH_IDX (&uart->pending_blobs, blob, i) {
+        r = sol_util_size_add(*pending_bytes, blob->size, pending_bytes);
+        SOL_INT_CHECK(r, < 0, r);
+    }
+
+    return 0;
+}
```
ceolin: I prefer do not have this function. I guess it's much better fail on `write` when the buffer if full. We can return `-EAGAIN` in this case.

The size of the internal buffer can be se through config.

glima: +1 for failing on the write side only

src/lib/io/sol-uart-impl-linux.c
```diff
@@ -32,49 +33,133 @@
...
+{
+    ssize_t r;
+    struct sol_buffer buf = SOL_BUFFER_INIT_CONST(uart->rx, uart->rx_used);
+
+    r = uart->rx_cb((void *)uart->user_data, uart, &buf);
+    SOL_INT_CHECK(r, < 0);
+    SOL_INT_CHECK((size_t)r, > uart->rx_used);
+
+    memmove(uart->rx, uart->rx + uart->rx_used, r);
+    uart->rx_used -= r;
```
barbieri: why can't you use `sol_buffer_remove()`? if the memory is fixed capacity, it shouldn't do the realloc, but the memmove/used could be updated. Then you keep the uart->rx as a buffer.

barbieri: the positive side of this is that if we ever want to allow changing the internal buffer to a dynamic one (malloc), then it's a simple change to how the buffer is initialized.

barbieri: and if you change that, make the buffer->data malloc(). Since it's 4096 (better to use `sysconf(_SC_PAGESIZE)`), it's likely to be allocated as a page, then if that memory is inline in a structure, it won't fit in a single page... which is worse.

src/lib/io/sol-uart-impl-linux.c
```diff
@@ -32,49 +33,133 @@
...
+            if (r < 0) {
+                if (errno == EAGAIN || errno == EINTR) {
+                    return true;
+                } else {
+                    status = errno;
+                    SOL_WRN("Could not read from the UART fd: %d - Reason:%s", uart->fd,
+                        sol_util_strerrora(status));
+                    return true;
+                }
+            }
```
barbieri: why not the sol-util-file buffer reading?

src/lib/io/sol-uart-impl-riot.c
```diff
@@ -119,48 +150,99 @@ sol_uart_open(const char *port_name, const struct sol_uart_config *config)
...
 
-    uart->async.rx_cb = config->rx_cb;
-    uart->async.rx_user_data = config->rx_cb_user_data;
-    uart->async.tx_buffer = NULL;
+    uart->rx_cb = config->rx_cb;
+    uart->tx_cb = config->tx_cb;
+    uart->user_data = config->user_data;
+    sol_ptr_vector_init(&uart->pending_blobs);
+    sol_buffer_init_flags(&uart->rx_buffer, NULL, 0,
+        SOL_BUFFER_FLAGS_DEFAULT | SOL_BUFFER_FLAGS_NO_NUL_BYTE);
```
barbieri: since you do fixed-size buffer in a much more capable system (linux), why not do the same here? As I said before, we keep using sol_buffer and eventually, if there is need, we can change the buffer to fixed capacity or not based on kconfig.

src/lib/io/sol-uart-impl-riot.c
barbieri: maybe `-ENOSPC`... but then you need a way in config to limit the pending size.

barbieri: also, what to do regarding partial? Say you have 10 bytes left, arrives 20bytes blob, what do you do?
 - queue the blob anyway, will overflow the allowed memory;
 - reject the blob anyway. Since the data was already produced, then it can be queued at user side or dropped, then it must be produced again.

src/samples/common/uart.c
```diff
@@ -20,104 +20,180 @@
...
+    };
+
+    char **argv;
+    int argc;
+
+    argc = sol_argc();
+    argv = sol_argv();
+
+    if (argc < 3) {
+        fprintf(stderr, "Usage: ./uart-sample <producerUART> <consumerUART>\n");
```
edersondisouza: Is any setup needed to actually run this sample, like the cross-link of ftdi cables? If so, please, explain that.

edersondisouza: Also consider explaining what and how this sample does what it does. You commit message is a good start.

src/lib/io/include/sol-uart.h
```diff
@@ -102,12 +105,32 @@ struct sol_uart_config {
 #define SOL_UART_CONFIG_API_VERSION (1) /**< compile time API version to be checked during runtime */
     uint16_t api_version; /**< must match #SOL_UART_CONFIG_API_VERSION at runtime */
 #endif
+    /**
+     * @brief Callback used containing data was read from UART
```
edersondisouza: s/used//

cabelitos:
#2007

PR #1911, opened 2016-04-19
http-server: Add support for HTTPS

ceolin:


src/lib/comms/include/sol-http-server.h
```diff
@@ -94,10 +95,35 @@ struct sol_http_progressive_response;
...
+ *
+ * @param port The port where the server will bind.
+ * @param cert The certificate that should be used to crypto the data.
+ * @param key The key of @a cert.
+ *
+ * @return a handle to the server on success, otherwise @c NULL is returned.
+ *
+ * @see sol_http_server_new()
+ */
+struct sol_http_server *sol_http_server_secure_new(uint16_t port,
```
cabelitos: I prefer that we have only one API for creating a server. 
`sol_http_server_new()`
Just like we do in CoAP

src/samples/http/server-https.c
```diff
@@ -0,0 +1,139 @@
...
+ * @brief HTTPS sample
+ *
+ * Basic HTTPS server. It uses a given certificate to crypto its contents.
+ *
+ * To test it:
+ *
+ * run ./server-https -p 8080 -d "Hello HTTPS" -c certificate_name.pem -k key_name.pem
+ *
+ * To see the usage help, -h or --help.
+ */
```
edersondisouza: Could we provide instructions/links to someone generate certificates, just for testing purposes?

src/lib/comms/include/sol-http-server.h
edersondisouza: ? Coap has `sol_coap_server_new` and `sol_coap_secure_server_new`...

cabelitos: @edersondisouza The idea was to change that

ceolin: It will change but not as you are thinking. The secuity will come from the socket. I really prefer two functions instead of receive unused parameters

src/samples/http/server-https.c
ceolin: yes, i can. I'll put together with how to test.

src/lib/comms/include/sol-http-server.h
barbieri: struct config... as usual..

ceolin: i was trying not introduce another type, but ... if you want guys I can proceed

edersondisouza: On 21-04-2016 11:00, Flávio Vinicius Alvares Ceolin wrote:
> In src/lib/comms/include/sol-http-server.h
> <https://github.com/solettaproject/soletta/pull/1911#discussion_r60584726>:
>
>> + * @c sol_http_server_register_handler()
>> + * and dirs with @c sol_http_server_add_dir().
>> + *
>> + * @note Only one instance of sol_http_server is possible per port.
>> + * Try to run a second instance in the
>> + * same port will result in fail.
>> + *
>> + * @param port The port where the server will bind.
>> + * @param cert The certificate that should be used to crypto the data.
>> + * @param key The key of @a cert.
>> + *
>> + * @return a handle to the server on success, otherwise @c NULL is returned.
>> + *
>> + * @see sol_http_server_new()
>> + */
>> +struct sol_http_server *sol_http_server_secure_new(uint16_t port,
>
> i was trying not introduce another type, but ... if you want guys I can
> proceed

@barbieri wants guys?

barbieri: I like that approach for multiple reasons:
 - you don't have to set the parameters you don't need (as opposed to pass multiple NULL, 0...)
 - we can keep reduced number of symbols;
 - no need to change the symbol version on changes, only the struct API version
 - API version can be handled in a customized version, it's just us changing the `==` check to `<` and code to handle the changes. This can offer a better API stability.

ceolin:
fixes in  https://github.com/solettaproject/soletta/pull/1932

PR #1912, opened 2016-04-19
zephyr: fix usage of net_context_get()

glima:
It receives references to {remote,local} addresses and expects them to
be reachable memory at all times (net_receive() and others will later
de-reference that memory). For some weird reason, Soletta did not crash
passing NULL to signal an all zeroed-out remote address and a stack
placed variable reference to the local one, beats us why.


ibriano:
Go for it

glima:
Ack

PR #1914, opened 2016-04-20
Iio/i2c: config pinmuxing before load I2C devices

yongli3:
On some platforms(Galileo/Edison), these GPIO pins must be configured before using I2C bus
This patch perform the pin mux configuration before loading I2C device drivers

### DIFF
--- a/bindings/nodejs/configure-bindings.js
+++ b/bindings/nodejs/configure-bindings.js
@@ -115,6 +115,14 @@ for ( oneVariable in process.env ) {
 				"sol-spi.h"
 			] );
 			break;
+		case "USE_I2C":
+			sources = sources.concat( [
+				"../src/functions/i2c.cc"
+			] );
+			headers = headers.concat( [
+				"sol-i2c.h"
+			] );
+			break;
 		default:
 			break;
 	}
--- a/bindings/nodejs/lib/i2c.js
+++ b/bindings/nodejs/lib/i2c.js
@@ -1,0 +1,144 @@
+/*
+ * This file is part of the Soletta Project
+ *
+ * Copyright (C) 2015 Intel Corporation. All rights reserved.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+var soletta = require( 'bindings' )( 'soletta' ),
+    _ = require( 'lodash' );
+
+exports.open = function( init ) {
+    return new Promise( function( fulfill, reject ) {
+        var bus = init.bus;
+        var spd = soletta.sol_i2c_speed_from_str(init.speed);
+        var i2cbus = null;
+
+        if (init.raw) {
+            i2cbus = I2CBus( soletta.sol_i2c_open_raw( bus, spd ) );
+        } else {
+            i2cbus = I2CBus( soletta.sol_i2c_open( bus, spd ) );
+        }
+
+        //copy the properties
+        _.extend(i2cbus, init);
+
+        fulfill( i2cbus );
+    });
+
+}
+
+var I2CBus = function( i2cbus ) {
+    if ( !this._isI2CBus )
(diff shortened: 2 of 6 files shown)


yongli3:
@bdilly please help to review

barbieri:
Indentation is wrong, run tools/check-style.py before sending patches.

As for the pinmux, check with @mbelluzzo

mbelluzzo:
Just by the commit, it doesn't look like the right place to set the mux on Soletta. I would like to check the full context before confirm this but I'm not in my desk right now.

src/lib/io/sol-i2c-impl-linux.c
```diff
@@ -968,8 +968,22 @@ create_device_iter_cb(void *data, const char *dir_path, struct dirent *ent)
     char path[PATH_MAX];
     int r;
     struct stat st;
+#ifdef USE_I2C
```
edersondisouza: I'd say that this file is only compiled when using I2C, so this ifdef is useless.

edersondisouza: I guess you wanted 'SOL_PIN_MUX' here.

src/lib/io/sol-i2c-impl-linux.c
```diff
@@ -968,8 +968,22 @@ create_device_iter_cb(void *data, const char *dir_path, struct dirent *ent)
...
 
     if (strstartswith(ent->d_name, "i2c-")) {
+	#ifdef USE_I2C
+		errno = 0;
+        i2c_bus = strtoul(ent->d_name + strlen("i2c-"), &end_ptr, 0);
+        if (errno || *end_ptr != '\0') {
+			SOL_WRN("Could not get I2C bus number");
+		} else {
+			sol_pin_mux_setup_i2c(i2c_bus);
+		}
```
edersondisouza: Please, fix indentation. Always run 'tools/check-style.py' to check indentation problems.

cabelitos:
You have a typo in your commit: "pinmuing" . Could you fix that?

src/lib/io/sol-i2c-impl-linux.c
mbelluzzo: +1 to USE_PIN_MUX

src/lib/io/sol-i2c-impl-linux.c
```diff
@@ -968,8 +968,22 @@ create_device_iter_cb(void *data, const char *dir_path, struct dirent *ent)
...
+#endif
 
     if (strstartswith(ent->d_name, "i2c-")) {
+	#ifdef USE_I2C
+		errno = 0;
+        i2c_bus = strtoul(ent->d_name + strlen("i2c-"), &end_ptr, 0);
+        if (errno || *end_ptr != '\0') {
+			SOL_WRN("Could not get I2C bus number");
+		} else {
+			sol_pin_mux_setup_i2c(i2c_bus);
```
mbelluzzo: You should check the return of this function, because in case a mux recipe was found but went on error during execution, this return will tell you that and you should consider everything in an error situation.

Check 'sol_i2c_open' in sol-i2c-common file, it even closes the bus that was trying to open in case of fault execution of 'sol_pin_mux_setup_i2c', maybe something similar should be done here.

src/lib/io/sol-i2c-impl-linux.c
```diff
@@ -968,8 +968,22 @@ create_device_iter_cb(void *data, const char *dir_path, struct dirent *ent)
...
+	unsigned int i2c_bus;
+	char *end_ptr;
+#endif
 
     if (strstartswith(ent->d_name, "i2c-")) {
+	#ifdef USE_I2C
+		errno = 0;
+        i2c_bus = strtoul(ent->d_name + strlen("i2c-"), &end_ptr, 0);
+        if (errno || *end_ptr != '\0') {
+			SOL_WRN("Could not get I2C bus number");
```
mbelluzzo: If this happens, you should considerer an error and abort everything, don't you?

mbelluzzo:
After a little discussing with @edersondisouza about the usage context of this code, we both agree that is indeed the right place to setup the mux. Just take inline comment into consideration.

src/lib/io/sol-i2c-impl-linux.c
yongli3: Done, thanks

src/lib/io/sol-i2c-impl-linux.c
yongli3: Done, thanks

src/lib/io/sol-i2c-impl-linux.c
yongli3: I add a logic to check if the API call fails, there is a warning, but continue to load i2c device drivers

PR #1915, opened 2016-04-20
node.js: Copy correct soletta.node file in installation

otaviobp:
Node.js build was failing everytime we rebuild the project. That was
happening because installing process used a find command to locate the
soletta.node file to be installed in sysroot. But after first build,
soletta.node file located in sysroot is found and we try to copy it to
itself.

Some people may not be affected by this bug because depending on locale
settings, find result order will be different

### DIFF
--- a/tools/build/Makefile.targets
+++ b/tools/build/Makefile.targets
@@ -220,7 +220,7 @@ bindings-nodejs: $(SOL_LIB_OUTPUT)
 	$(Q) mkdir -p $(build_nodejs_bindingsdir)
 	$(Q) cp -a *js package.json $(build_nodejs_bindingsdir)
 	$(Q) mkdir -p $(build_nodejs_bindingsdir)/build
-	$(Q) cp $$(find build -type f -name soletta.node | head -n 1) \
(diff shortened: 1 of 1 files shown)


tools/build/Makefile.targets
```diff
@@ -220,7 +220,7 @@ bindings-nodejs: $(SOL_LIB_OUTPUT)
 	$(Q) mkdir -p $(build_nodejs_bindingsdir)
 	$(Q) cp -a *js package.json $(build_nodejs_bindingsdir)
 	$(Q) mkdir -p $(build_nodejs_bindingsdir)/build
-	$(Q) cp $$(find build -type f -name soletta.node | head -n 1) \
+	$(Q) cp $$(find build -type f -name soletta.node | grep obj.target | head -n 1) \
```
barbieri: no, just use the correct directory in `find`, instead of `build` use a more specific folder that is not `soletta_sysroot`

otaviobp: This was already fixed in #1916 with a similar approach.

PR #1918, opened 2016-04-20
HTTP: sol_http_progressive_response_feed() should receive a blob.

cabelitos:

@ceolin

### DIFF
--- a/src/bin/sol-fbp-runner/web-inspector.c
+++ b/src/bin/sol-fbp-runner/web-inspector.c
@@ -36,6 +36,47 @@ static struct sol_http_progressive_response *events_response;
 
 static struct timespec start;
 
+
+static int
+web_inspector_send_sse_data(struct sol_http_progressive_response *client, struct sol_buffer *buf)
+{
+    int r;
+    struct sol_blob *blob;
+    static struct sol_blob *sse_prefix = NULL;
+    static struct sol_blob *sse_suffix = NULL;
+
+    if (!sse_prefix) {
+        sse_prefix = sol_blob_new(SOL_BLOB_TYPE_NO_FREE, NULL, "data: ", strlen("data: "));
+        if (!sse_prefix)
+            return -ENOMEM;
+    }
+
+    if (!sse_suffix) {
+        sse_suffix = sol_blob_new(SOL_BLOB_TYPE_NO_FREE, NULL, "\n\n", strlen("\n\n"));
+        if (!sse_suffix)
+            return -ENOMEM;
+    }
+
+    blob = sol_buffer_to_blob(buf);
+    if (!blob)
+        return -ENOMEM;
+
+    r = sol_http_progressive_response_feed(client, sse_prefix);
+    if (r < 0)
+        return r;
+
+    r = sol_http_progressive_response_feed(client, blob);
+    if (r < 0)
+        return r;
+
+    r = sol_http_progressive_response_feed(client, sse_suffix);
+    if (r < 0)
+        return r;
+
+    sol_blob_unref(blob);
+    return 0;
+}
+
 static int
 web_inspector_add_json_key_value(struct sol_buffer *buf, const char *k, const char *v)
 {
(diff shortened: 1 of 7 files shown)


src/bin/sol-fbp-runner/web-inspector.c
```diff
@@ -36,6 +36,47 @@ static struct sol_http_progressive_response *events_response;
...
+static int
+web_inspector_send_sse_data(struct sol_http_progressive_response *client, struct sol_buffer *buf)
+{
+    int r;
+    struct sol_blob *blob;
+    static struct sol_blob *sse_prefix = NULL;
+    static struct sol_blob *sse_suffix = NULL;
+
+    if (!sse_prefix) {
+        sse_prefix = sol_blob_new(SOL_BLOB_TYPE_NO_FREE, NULL, "data: ", strlen("data: "));
```
barbieri: this should be a global `static struct sol_blob sse_prefix`, no need to call `sol_blob_new()`. However I see why you did it and we should improve that: `SOL_BLOB_TYPE_NO_FREE` should be renamed to `SOL_BLOB_TYPE_NO_FREE_DATA`, and `SOL_BLOB_TYPE_NO_FREE` should be an empty free function in `struct sol_blob_type`, this is because if `type->free == NULL`, then it will free the blob structure, which is not wanted.

src/lib/comms/sol-http-server-impl-microhttpd.c
```diff
@@ -444,39 +445,48 @@ build_mhd_response(const struct sol_http_response *response, time_t last_modifie
...
-    len = sol_util_min(size, progressive->buffer.used);
-    memcpy(buf, progressive->buffer.data, len);
+    blob = sol_ptr_vector_get_nocheck(&progressive->pending_blobs, 0);
+    len = sol_util_min(size, blob->size);
+    memcpy(buf, (char *)blob->mem + progressive->written, len);
+    progressive->written += len;
 
-    r = sol_buffer_remove_data(&progressive->buffer, 0, len);
-    SOL_INT_CHECK(r, < 0, MHD_CONTENT_READER_END_WITH_ERROR);
+    if (progressive->written == blob->size) {
```
barbieri: the `on_feed_done` is something you plan to add later?

src/lib/comms/sol-http-server-impl-microhttpd.c
```diff
@@ -1260,22 +1269,28 @@ sol_http_progressive_response_del(struct sol_http_progressive_response *progress
...
     SOL_NULL_CHECK(progressive, -EINVAL);
     SOL_EXP_CHECK(progressive->delete_me == true, -EINVAL);
+    SOL_NULL_CHECK(blob, -EINVAL);
+
 
-    ret = sol_buffer_append_slice(&progressive->buffer, data);
-    SOL_INT_CHECK(ret, < 0, ret);
+    SOL_NULL_CHECK(sol_blob_ref(blob), -EOVERFLOW);
+    ret = sol_ptr_vector_append(&progressive->pending_blobs, blob);
+    SOL_INT_CHECK_GOTO(ret, < 0, err_exit);
```
barbieri: unref after append, will simplify the checks

src/lib/datatypes/sol-buffer.c
```diff
@@ -868,3 +868,19 @@ sol_buffer_fini(struct sol_buffer *buf)
...
 }
+
+SOL_API struct sol_blob *
+sol_buffer_to_blob(struct sol_buffer *buf)
+{
+    struct sol_blob *blob;
+
+    SOL_NULL_CHECK(buf, NULL);
+
+    blob = sol_blob_new(SOL_BLOB_TYPE_DEFAULT, NULL, buf->data, buf->used);
```
barbieri: if you go with `SOL_BLOB_TYPE_DEFAULT` then you must check flags OR steal_copy the buffer.

cabelitos: Indeed, I forgot about that;

src/lib/datatypes/include/sol-buffer.h
```diff
@@ -977,6 +978,17 @@ int sol_buffer_ensure_nul_byte(struct sol_buffer *buf);
  */
 int sol_buffer_remove_data(struct sol_buffer *buf, size_t offset, size_t size);
 
+
+/**
+ * @brief Convert a buffer to a @ref struct sol_blob
+ *
+ * The buffer will be stealed by the created blob.
```
edersondisouza: s/stealed/stolen/

cabelitos: ty

src/lib/comms/sol-http-server-impl-microhttpd.c
cabelitos: Yeah

src/lib/datatypes/sol-buffer.c
ceolin: hmm, here i think we have two approaches.  Convert the buffer to blob (then you can steal the buffer memory) or create a blob from buffer `sol_blob_from_buffer(const struct sol_buffer *buf)`. The later you will copy the memory.

barbieri: problem if you take the whole buffer into account for blob, this won't be useful for buffers allocated on the stack, that are pretty common. Taking only the buffer memory is more likely to be useful.

However I'd document this, that once you do the conversion, only the data is kept alive (and duplicated if needed), the buffer is empty (zeroed, data, used and capacity if not "memory_owned") and may be freed afterwards

cabelitos:
#2015

PR #1919, opened 2016-04-20
Add aliases support for node types

cabelitos:


src/shared/include/sol-util-file.h
```diff
@@ -52,6 +52,18 @@ extern "C" {
...
+ * @see sol_util_iterate_dir()
+ */
+#define SOL_UTIL_ITERATE_DIR_CONTINUE (1)
+
+/**
+ * @brief Used to flag the directory iteration should stop
+ * @see sol_util_iterate_dir()
+ */
+#define SOL_UTIL_ITERATE_DIR_STOP (0)
+
```
barbieri: enum would be better than defines.

src/shared/include/sol-util-file.h
```diff
@@ -256,12 +268,16 @@ sol_util_fill_buffer_exactly(const int fd, struct sol_buffer *buffer, const size
...
  * @param data The user data which will be passed in the callback.
  *
- * @return @c true on success, otherwise @c false.
+ * @return 0 on success or negative errno on error.
+ * @note @c "." and @c ".." will be excluded from the iteration.
+ * @see #SOL_UTIL_ITERATE_DIR_CONTINUE
+ * @see #SOL_UTIL_ITERATE_DIR_STOP
  */
-bool sol_util_iterate_dir(const char *path, bool (*iterate_dir_cb)(void *data, const char *dir_path,
+int sol_util_iterate_dir(const char *path, int (*iterate_dir_cb)(void *data, const char *dir_path,
```
barbieri: then here the return of `iterate_dir_cb` will be an enum, much cleaner.

src/shared/sol-conffile.c
```diff
@@ -917,9 +943,162 @@ _load_vector_defaults(void)
...
+    unsigned long precedence = 0;
+    size_t name_len;
+    int r;
+    uint16_t i;
+
+    SOL_BUFFER_DECLARE_STATIC(node_name, 128);
+
+    name_len = strlen(en->d_name);
+
+    if (extension > name_len || strcmp(en->d_name + name_len - extension, ".json"))
```
barbieri: we have an `strendswith()`? If not, add it.

src/shared/sol-conffile.c
```diff
@@ -917,9 +943,162 @@ _load_vector_defaults(void)
...
+}
+
+
+static int
+sort_aliases(const void *v1, const void *v2)
+{
+    const struct sol_alias_ctx *ctx1, *ctx2;
+
+    ctx1 = (const struct sol_alias_ctx *)v1;
+    ctx2 = (const struct sol_alias_ctx *)v2;
```
barbieri: no cast should be required.

src/shared/sol-conffile.c
```diff
@@ -37,9 +37,16 @@
 #include "sol-memmap-storage.h"
 #endif
 
+struct sol_alias_ctx {
+    char *node;
```
barbieri: store as `struct sol_str_slice` then you can get a more efficient `_eq()` that first checks the length and avoid strcmp()

barbieri: ah, and the data you can abuse and add to the arena in this file, so no need to free manually

src/shared/sol-conffile.c
```diff
@@ -948,26 +1135,50 @@ _bsearch_entry_cb(const void *data1, const void *data2)
...
+
 static int
 _resolve_config(const char *id, const char **type, const char ***opts)
 {
     struct sol_conffile_entry key;
     struct sol_conffile_entry *entry;
     void **vector_pointer;
+    const char *real_type;
+
+    real_type = get_real_type(id);
```
barbieri: I'd do this as a last-try, not first. Then you only pay the price of the lookup when everything else fails.

tools/build/Makefile.targets
```diff
@@ -234,3 +234,7 @@ check-bindings-nodejs: bindings-nodejs
 PHONY += check-bindings-nodejs
 
 endif #neq($(USE_NODEJS),)
+
+gen-aliases:
+	$(Q)$(MKDIR) -p $(build_aliasesdir)
+	$(FLOW_NODE_TYPE_ALIAS) --file-name=default --node-description-dir=$(flow-dir) --output-dir=$(build_aliasesdir)
```
barbieri: not a good interface, get the output filename (including number + `.json`) then all the files you want to use for that, not a directory. This way to do an alias file for 2 files you'd need to create a directory for it.

src/shared/sol-conffile.c
cabelitos: C++ feelings...

src/shared/sol-conffile.c
```diff
@@ -917,9 +943,162 @@ _load_vector_defaults(void)
...
 static int
+iterate_dir_cb(void *data, const char *dir_path, struct dirent *en)
+{
+    char path[PATH_MAX], *sep, *endptr = NULL, *str;
+    struct sol_ptr_vector aliases = SOL_PTR_VECTOR_INIT;
+    struct sol_buffer *file_contents = data;
+    struct sol_json_scanner scanner;
+    struct sol_json_token value;
+    enum sol_json_loop_reason end_reason;
+    const size_t extension = strlen(".json");
```
ceolin: Why don't you use `sizeof()` ? `sizeof` count the size in compile time while `strlen` is in run time. Maybe the compilers are smart enough to do this in runtime, but I think it's better do not presume it.

edersondisouza: On 22-04-2016 13:48, Flávio Vinicius Alvares Ceolin wrote:
> In src/shared/sol-conffile.c
> <https://github.com/solettaproject/soletta/pull/1919#discussion_r60768904>:
>
>> @@ -917,9 +943,162 @@ _load_vector_defaults(void)
>>  }
>>
>>  static int
>> +iterate_dir_cb(void *data, const char *dir_path, struct dirent *en)
>> +{
>> +    char path[PATH_MAX], *sep, *endptr = NULL, *str;
>> +    struct sol_ptr_vector aliases = SOL_PTR_VECTOR_INIT;
>> +    struct sol_buffer *file_contents = data;
>> +    struct sol_json_scanner scanner;
>> +    struct sol_json_token value;
>> +    enum sol_json_loop_reason end_reason;
>> +    const size_t extension = strlen(".json");
>
> Why don't you use |sizeof()| ? |sizeof| count the size in compile time
> while |strlen| is in run time. Maybe the compilers are smart enough to
> do this in runtime, but I think it's better do not presume it.

But remember `sizeof` will include `\0` on its counting.

barbieri: Strlen() is optimized on constants.

src/lib/common/sol-power-supply-impl-linux.c
```diff
@@ -181,37 +181,33 @@ _get_enum_prop(const char *name, enum sol_power_supply_prop prop,
...
     int r;
 
-    if ((!strcmp(ent->d_name, ".")) || (!strcmp(ent->d_name, ".."))) {
-        return false;
-    }
-
     if (ldata->check_type) {
         r = sol_power_supply_get_type(ent->d_name, &type);
-        SOL_INT_CHECK(r, < 0, false);
+        SOL_INT_CHECK(r, < 0, r);
```
edersondisouza: Here you are changing previous behaviour. What if on error the idea is really scan next file?

src/lib/common/sol-power-supply-impl-linux.c
```diff
@@ -181,37 +181,33 @@ _get_enum_prop(const char *name, enum sol_power_supply_prop prop,
...
 
         if (type != ldata->type)
-            return false;
+            return SOL_UTIL_ITERATE_DIR_CONTINUE;
     }
 
     name = strdup(ent->d_name);
-    SOL_NULL_CHECK(name, false);
+    SOL_NULL_CHECK(name, -ENOMEM);
```
edersondisouza: Ditto. Usage may be wrong, but it may as well be intentional.

src/lib/common/sol-power-supply-impl-linux.c
```diff
@@ -181,37 +181,33 @@ _get_enum_prop(const char *name, enum sol_power_supply_prop prop,
...
 
     r = sol_ptr_vector_append(ldata->list, name);
     if (r < 0) {
         SOL_WRN("Failed to append supply to list %s", name);
         free(name);
-        return false;
```
edersondisouza: Ditto.

src/lib/io/sol-i2c-impl-linux.c
```diff
@@ -1041,10 +1039,9 @@ sol_i2c_create_device(const char *address, const char *dev_name, unsigned int de
         }
     }
 
-    if (!sol_util_iterate_dir(real_path, create_device_iter_cb, &result)) {
+    r = sol_util_iterate_dir(real_path, create_device_iter_cb, &result);
+    if (r < 0)
```
edersondisouza: I'd add `&& r != -EEXIST` to avoid warning about a normal condition.

data/schemas/node-type-genspec.schema
```diff
@@ -417,6 +423,12 @@
           "minItems": 1,
           "uniqueItems": true,
           "items": { "$ref": "#/definitions/node" }
+        },
+        "aliases": {
+          "type": "array",
```
edersondisouza: Why not the `"$ref": "#/definitions/node"`way, instead of duplicating these options?

src/shared/sol-conffile.c
```diff
@@ -917,9 +943,162 @@ _load_vector_defaults(void)
...
+err_exit:
+    SOL_PTR_VECTOR_FOREACH_IDX (&aliases, str, i)
+        free(str);
+    sol_ptr_vector_clear(&aliases);
+    return r;
+}
+
+
+static int
+sort_aliases(const void *v1, const void *v2)
```
edersondisouza: Nitpick: this function doesn't sort aliases... it just compares them...

src/lib/common/sol-power-supply-impl-linux.c
glima: Fair, according to the next return.

src/lib/common/sol-power-supply-impl-linux.c
glima: Ack.

src/lib/common/sol-power-supply-impl-linux.c
glima: Ack.

src/lib/io/sol-i2c-impl-linux.c
glima: Fair.

src/lib/common/sol-power-supply-impl-linux.c
glima: Actually I'd keep this, as we're short on memory.

src/shared/sol-conffile.c
glima: Ok

data/schemas/node-type-genspec.schema
glima: Ok

glima:
All set here, just running some tests now, as I just changed stuff.

glima:
The patch did not handle aliases on conf file entries, now fixed

src/shared/sol-conffile.c
glima: Done.

glima:
v2 coming

PR #1920, opened 2016-04-20
Memdesc v3

barbieri:
This is the one I want to merge if nothing goes wrong:

Changes since v2 (#1909):
 - fixed comments from @vcgomes and @edersondisouza 
 - uncrustified vector so semaphore stops bitching
 - added `SOL_MEMDESC_TYPE_ENUMERATION`
 - added `sol_memdesc_set_as_int64()` and `sol_memdesc_set_as_uint64()` to simplify writing to various integers sizes.
 - added description serialization

### DIFF
--- a/src/lib/datatypes/include/sol-memdesc.h
+++ b/src/lib/datatypes/include/sol-memdesc.h
@@ -289,15 +289,17 @@ const char *sol_memdesc_type_to_str(enum sol_memdesc_type type) SOL_ATTR_WARN_UN
 #endif
 
 /**
- * @def SOL_MEMDESC_SET_DESCRIPTION(text)
+ * @def SOL_MEMDESC_SET_DESCRIPTION()
  *
  * Helper to set the description member of struct sol_memdesc if that
  * is available (conditional to #SOL_MEMDESC_SET_DESCRIPTION).
+ *
+ * It takes C statements and conditionally uses or eliminates them.
  */
 #ifdef SOL_MEMDESC_DESCRIPTION
-#define SOL_MEMDESC_SET_DESCRIPTION(text) .description = (text)
+#define SOL_MEMDESC_SET_DESCRIPTION(...) __VA_ARGS__
 #else
-#define SOL_MEMDESC_SET_DESCRIPTION(text)
+#define SOL_MEMDESC_SET_DESCRIPTION(...)
 #endif
 
 #ifndef SOL_NO_API_VERSION
@@ -1786,7 +1788,11 @@ struct sol_memdesc_serialize_options {
      * struct sol_memdesc_serialize_options::structure::show_key, then
      * struct sol_memdesc_serialize_options::structure::key::start and
      * struct sol_memdesc_serialize_options::structure::key::end are used around the
-     * struct sol_memdesc::name that is dumped as-is.
+     * struct sol_memdesc::name that is dumped as-is. If SOL_MEMDESC_DESCRIPTION
+     * is enabled, then struct sol_memdesc_serialize_options::structure::show_description
+     * is true, then the description is printed after value surrounded by
+     * struct sol_memdesc_serialize_options::structure::description::start and
+     * struct sol_memdesc_serialize_options::structure::description::end.
      *
      * If multiple members exist, they will be separated with
      * struct sol_memdesc_serialize_options::structure::separator.
@@ -1866,6 +1872,22 @@ struct sol_memdesc_serialize_options {
              */
             const struct sol_str_slice indent;
         } value;
+#ifdef SOL_MEMDESC_DESCRIPTION
+        struct {
+            /**
+             * Used when starting a new structure member description.
+             */
+            const struct sol_str_slice start;
+            /**
+             * Used when finishing structure member description.
+             */
+            const struct sol_str_slice end;
+            /**
+             * Used to indent a new member description.
+             */
+            const struct sol_str_slice indent;
+        } description;
+#endif
         /**
          * Used if multiple members exist.
          */
@@ -1882,6 +1904,12 @@ struct sol_memdesc_serialize_options {
          * non-detail members will.
          */
         bool detailed;
+#ifdef SOL_MEMDESC_DESCRIPTION
+        /**
+         * Controls whenever the description is to be serialized.
+         */
+        bool show_description;
+#endif
     } structure;
     /**
      * @brief options used by serialize_array_item
--- a/src/lib/datatypes/include/sol-memdesc.h
+++ b/src/lib/datatypes/include/sol-memdesc.h
@@ -1,0 +1,1093 @@
+/*
+ * This file is part of the Soletta Project
+ *
+ * Copyright (C) 2016 Intel Corporation. All rights reserved.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+#pragma once
+
+#include <errno.h>
+#include <stdbool.h>
+#include <stdint.h>
+#include <stdlib.h>
+#include <sol-common-buildopts.h>
+#include <sol-str-slice.h>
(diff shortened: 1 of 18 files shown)


src/lib/parsers/sol-json.c
```diff
@@ -789,6 +790,492 @@ sol_json_serialize_boolean(struct sol_buffer *buffer, bool val)
...
+    case SOL_MEMDESC_TYPE_SIZE: {
+        size_t *m = memory;
+        uint64_t v;
+
+        r = sol_json_token_get_uint64(token, &v);
+        if (r < 0)
+            return r;
+
+        if (v > SIZE_MAX)
+            return -ERANGE;
```
ceolin: You are comparing unsigned `v` with signed `SIZE_MAX`. We are appending `Wsign-compare` if available to `CFLAGS`. Compilers will complain with this comparation. There are many other places doing it.

src/lib/datatypes/sol-str-table.c
```diff
@@ -30,16 +30,21 @@ sol_str_table_lookup_fallback(const struct sol_str_table *table,
...
 
     for (iter = table; iter->key; iter++) {
         if (iter->len == len && memcmp(iter->key, key.data, len) == 0) {
+            errno = 0;
             return iter->val;
         }
     }
+
+    errno = ENOENT;
     return fallback;
```
ceolin: The doc does not say anything about set `errno` to `EINVAL` i guess would be good tells that in case the error `errno` will be set properly.

src/lib/datatypes/sol-memdesc.c
```diff
@@ -0,0 +1,808 @@
...
+        b_item = sol_memdesc_get_array_element(desc, b_memory, idx);
+        if (!b_item)
+            return 0;
+
+        r = compare_content(desc->array_item, a_item, b_item);
+        if (r != 0 || errno)
+            return r;
+    }
+
+    if (idx < len) /* loop failed */
```
glima: I guess you could do like ``copy_array()`` does here: ``r = errno ? -errno : -EINVAL;``.

src/test/test-memdesc.c
```diff
@@ -0,0 +1,1210 @@
...
+    const struct sol_memdesc otherdesc = {
+        SOL_SET_API_VERSION(.api_version = SOL_MEMDESC_API_VERSION, )
+        .size = sizeof(struct otherst),
+        .type = SOL_MEMDESC_TYPE_STRUCTURE,
+        .structure_members = (const struct sol_memdesc_structure_member[]){
+            {
+                .base = {
+                    SOL_SET_API_VERSION(.api_version = SOL_MEMDESC_API_VERSION, )
+                    .type = SOL_MEMDESC_TYPE_BOOLEAN,
+                    .defcontent.i64 = true,
```
glima: i64??

src/lib/parsers/sol-json.c
```diff
@@ -789,6 +790,440 @@ sol_json_serialize_boolean(struct sol_buffer *buffer, bool val)
...
+    case SOL_MEMDESC_TYPE_UINT64:
+    case SOL_MEMDESC_TYPE_ULONG:
+    case SOL_MEMDESC_TYPE_SIZE: {
+        uint64_t v;
+
+        r = sol_json_token_get_uint64(token, &v);
+        if (r < 0)
+            return r;
+
+        return sol_memdesc_set_as_uint64(desc, memory, v);
```
ceolin: AFAIK json does not have all those numeric types, so the token value can easily be different than the described.
For example, if one describes the memory as `UINT8` and the value in the token is greater shouldn't return an error ?

src/lib/parsers/sol-json.c
```diff
@@ -789,6 +790,440 @@ sol_json_serialize_boolean(struct sol_buffer *buffer, bool val)
...
+        tt = sol_json_token_get_type(token);
+        if (tt != SOL_JSON_TYPE_OBJECT_START)
+            return -EINVAL;
+
+        len = 0;
+        SOL_MEMDESC_FOREACH_STRUCTURE_MEMBER(desc, itr) {
+            len++;
+        }
+
+        done = alloca(len * sizeof(bool));
```
ceolin: Shouldn't we avoid `alloca` from user input ?

src/lib/parsers/sol-json.c
barbieri: There are no signs in the SIZE_MAX, at least did not trigger any warnings.

src/lib/datatypes/sol-memdesc.c
barbieri: Compare returns 0 on errors, errno is set. Or you mean check if the errno is set, if not then set, before returning?

src/lib/parsers/sol-json.c
barbieri: This is why I added set_as_uint64(), it does the overflow check

src/test/test-memdesc.c
barbieri: Copy and pasta detected

src/lib/parsers/sol-json.c
barbieri: Not bothering with that since the expected use is not to get this from end users, like network, text files or stdin.

src/lib/parsers/sol-json.c
```diff
@@ -789,6 +790,492 @@ sol_json_serialize_boolean(struct sol_buffer *buffer, bool val)
...
+    SOL_NULL_CHECK(memory, -EINVAL);
+
+    switch (desc->type) {
+    case SOL_MEMDESC_TYPE_UNKNOWN:
+        return -EINVAL;
+    case SOL_MEMDESC_TYPE_UINT8: {
+        uint8_t *m = memory;
+        uint64_t v;
+
+        r = sol_json_token_get_uint64(token, &v);
```
edersondisouza: Maybe it's a silly question, but if we have `sol_json_token_get_uint32`, why use `sol_json_token_get_uint64` when getting things < 32?

src/lib/datatypes/include/sol-memdesc.h
```diff
@@ -1085,6 +1086,525 @@ sol_memdesc_get_structure_member_memory(const struct sol_memdesc *structure_desc
...
+        i64 = *(const long *)memory;
+        goto check_signed;
+    case SOL_MEMDESC_TYPE_SSIZE:
+        i64 = *(const ssize_t *)memory;
+        goto check_signed;
+    case SOL_MEMDESC_TYPE_BOOLEAN:
+        return *(const bool *)memory;
+    case SOL_MEMDESC_TYPE_DOUBLE:
+        i64 = *(const double *)memory;
+        goto check_signed;
```
edersondisouza: Do you care about `-0.0`? It will fail range checking below, but maybe it should pass.

src/lib/datatypes/include/sol-memdesc.h
```diff
@@ -1085,6 +1086,525 @@ sol_memdesc_get_structure_member_memory(const struct sol_memdesc *structure_desc
...
+        return false;
+    case SOL_MEMDESC_TYPE_INT8:
+    case SOL_MEMDESC_TYPE_INT16:
+    case SOL_MEMDESC_TYPE_INT32:
+    case SOL_MEMDESC_TYPE_INT64:
+    case SOL_MEMDESC_TYPE_LONG:
+    case SOL_MEMDESC_TYPE_SSIZE:
+        return true;
+    case SOL_MEMDESC_TYPE_BOOLEAN:
+    case SOL_MEMDESC_TYPE_DOUBLE:
```
edersondisouza: Why double falls on not signed?

edersondisouza: Err... now I saw documentation... why not used `is_signed_integer` in the name of this function?

barbieri: good point :-)

src/lib/datatypes/include/sol-memdesc.h
barbieri: wouldn't bother. Initially I thought about only returning for UINT types, not even doing the checks, but it wasn't not so much effort and then I did the others. The code is all in the .h, so people can see what is going to happen (and compilers can optimize out based on previous type checks)

src/lib/parsers/sol-json.c
barbieri: no need to do 2 range comparisons, the uint64 doesn't do one, the 32 does but then we need to check for uint8_max here.

And as you can see in next commit, by introducing `set_as_uint64()` I move this to a helper.

src/lib/datatypes/sol-str-table.c
barbieri: it's always the case for invalid parameters, but I'll add it.

barbieri:
merged with proposed fixes.

PR #1924, opened 2016-04-21
Node.js bindings: Fix a crash and memory leak in GPIO JS bindings

nagineni:
This includes the following fixes:
- Fixes a crash where the argument count is wrong.
- We call hijack_ref() only  when callback is true, so unref it in
  the same condition.
- Fixes a possible memory leak by deleting gpio_data.


nagineni:
@gabrielschulhof @cabelitos Could you please have a look?

edersondisouza:
looks ok

PR #1925, opened 2016-04-21
Node.js bindings: Fix errors in Uart JS bindings

nagineni:
This includes the following changes:
- Fix possible memory leaks.
- Use correct argument type in enum_to_str translation functions.


gabrielschulhof:
:+1:

PR #1926, opened 2016-04-21
.gitignore: Add files moved to repo root as part of npm install

gabrielschulhof:

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -24,3 +24,10 @@ config.log*
 /node_modules/
 /bindings/nodejs/generated/main.cc
 /bindings/nodejs/generated/main.h
+aio.js
+bindings/nodejs/generated/header-files-list
+bindings/nodejs/generated/nodejs-bindings-sources.gyp
+gpio.js
+pwm.js
+spi.js
+uart.js


gabrielschulhof:
@otaviobp PTAL

ceolin:
this is ok, but why those files are not generated under the `build` directory ? whether `stage` or `soletta_sysroot`

gabrielschulhof:
@ceolin the build process is designed such that, if node.js bindings are enabled, then at the end of ```make``` the soletta repo is a perfectly well-formed and installed npm package in addition to having a copy of the perfectly well-formed npm package under ```build/soletta_sysroot/<prefix>/lib/node_modules/soletta``` for the purposes of a possible subsequent ```make install```. So, if we want to support writing ```require( "soletta/uart" );``` or ```require( "soletta/gpio" );``` then the files ```uart.js``` resp. ```gpio.js``` have to reside in the repo root at the end of the build process, because npm packages do not have a ```make install``` step. They are built in their final resting place.

otaviobp:
+1, if there isn't really a way of generating those files in build directory.

PR #1928, opened 2016-04-22
json: Always finalize the buffer in case of error

ceolin:
sol_json_token_get_unescaped_string was initializing the buffer but in
some errors case was not finalizing it.


barbieri:
Great, this was on my todo list after memdesc. However, from what I recall, there are some returns in this function, please check those.

ceolin:
I will check for others returns an commit after it. Thanks for reviewing it.

PR #1929, opened 2016-04-22
Js sample

edersondisouza:
Small JS API sample for Calamari Lure. While writing this sample, I've found some bugs, whose fixes are on this PR as well.
@gabrielschulhof @nagineni @sragavan Could you please review this PR? See if sample uses correctly JS bindings - please, share what can be improved, as I'm not expert on JavaScript, I may have chosen not so optimal solutions =D
Please notice my 'hack' comment on sample: I'm not sure if that is the best approach. I wanted to open all GPIO pins and if any of them failed, close all that were opened.
Another thing that I'd like to know is if there's something like conffile for js: what if I want to run it on a different setup than Calamari? How do I change pin numbers and so on?

### DIFF
--- a/bindings/nodejs/lib/aio.js
+++ b/bindings/nodejs/lib/aio.js
@@ -23,16 +23,23 @@ var soletta = require( 'bindings' )( 'soletta' ),
 exports.open = function( init ) {
 	return new Promise( function( fulfill, reject ) {
 		var precision = 12;
+		var aio;
 
 		if (init.precision)
 			precision = init.precision;
 
 		if (init.name && init.name != "") {
-			fulfill( AIOPin( soletta.sol_aio_open_by_label( init.name, precision ) ) );
+			aio = soletta.sol_aio_open_by_label( init.name, precision );
 		} else if (init.raw) {
-			fulfill( AIOPin( soletta.sol_aio_open_raw( init.device, init.pin, precision ) ) );
+			aio = soletta.sol_aio_open_raw( init.device, init.pin, precision );
 		} else {
-			fulfill( AIOPin( soletta.sol_aio_open( init.device, init.pin, precision ) ) );
+			aio = soletta.sol_aio_open( init.device, init.pin, precision );
+		}
+
+		if ( aio ) {
+			fulfill( AIOPin( aio ) );
+		} else {
+			reject( new Error( "Could not open AIO device" ) );
 		}
 	});
 
--- a/bindings/nodejs/lib/gpio.js
+++ b/bindings/nodejs/lib/gpio.js
@@ -32,6 +32,7 @@ exports.open = function( init ) {
         var pull = init.pull ? init.pull : "none";
         var active_low = init.activeLow ? init.activeLow : false;
         var poll = init.poll ? init.poll : 1000;
+        var gpio;
 
         if ( direction == "in" ) {
             config = {
@@ -56,9 +57,14 @@ exports.open = function( init ) {
             }
         }
 
-        gpiopin = GPIOPin( soletta.sol_gpio_open( pin, config ) );
-        callback_data.push( gpiopin );
-        fulfill( gpiopin );
+        gpio = soletta.sol_gpio_open( pin, config );
+        if ( gpio ) {
+            gpiopin = GPIOPin( gpio );
+            callback_data.push( gpiopin );
+            fulfill( gpiopin );
+        } else {
+            reject( new Error( "Could not open GPIO device" ) );
+        }
     });
 
 }
--- a/bindings/nodejs/lib/gpio.js
+++ b/bindings/nodejs/lib/gpio.js
@@ -30,12 +30,14 @@ exports.open = function( init ) {
         var direction = init.direction ? init.direction : "out";
         var edge = init.edge ? init.edge : "any";
         var pull = init.pull ? init.pull : "none";
+        var active_low = init.activeLow ? init.activeLow : false;
+        var poll = init.poll ? init.poll : 1000;
 
         if ( direction == "in" ) {
             config = {
                 dir: soletta.sol_gpio_direction_from_str( direction ),
-                active_low: init.activeLow,
-                poll_timeout: init.poll,
+                active_low: active_low,
+                poll_timeout: poll,
                 drive_mode: soletta.sol_gpio_drive_from_str( pull ),
                 trigger_mode: soletta.sol_gpio_edge_from_str( edge ),
                 callback: function( value ) {
@@ -49,7 +51,7 @@ exports.open = function( init ) {
         } else {
             config = {
                 dir: soletta.sol_gpio_direction_from_str( direction ),
-                active_low: init.activeLow,
+                active_low: active_low,
                 drive_mode: soletta.sol_gpio_drive_from_str( pull ),
             }
         }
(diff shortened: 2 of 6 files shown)


edersondisouza:
@gabrielschulhof @nagineni @sragavan Another thing that I noticed is that I needed to put some effort to assure that all opened IOs would be close. Is it possible to do so automatically for users?

bindings/nodejs/samples/calamari.js
```diff
@@ -0,0 +1,110 @@
+/*
+ * This file is part of the Soletta Project
```
barbieri: put all samples under `src/samples`, like `src/samples/nodejs`.

barbieri:
@edersondisouza about your question for conffile: I'd not go that route.

To start, the idea of using low-level I/O from userspace is a mistake done by others so we're only exposing that to be covering the same feature set. SPI, I2C and the likes are not something you should be using in a low-level API like our C or JS APIs. The "matching feature with other libraries" is also what force us to implement this API in higher priority.

People should use the FBP node types as they are higher level and matches closely what people will use in their applications. Of course without the flow, as I have demonstrated with "sol_flow_single_node" code. The nice side-effect is that it would give you the conffile for free AND consistent with other languages (Python, C all would use the same).

However, as explained above, we did not cover the FBP due the time pressure to match what was provided by MRAA and other JS projects. There is also a psychological component that people are afraid of what they do not know, but they will get to it and start to understand the reasons we push that path.

bindings/nodejs/lib/gpio.js
```diff
@@ -30,12 +30,14 @@ exports.open = function( init ) {
         var direction = init.direction ? init.direction : "out";
         var edge = init.edge ? init.edge : "any";
         var pull = init.pull ? init.pull : "none";
+        var active_low = init.activeLow ? init.activeLow : false;
+        var poll = init.poll ? init.poll : 1000;
```
nagineni: @edersondisouza First of all, I would like to thank you for fixing these and adding sample for JS APIs. 

The default poll value should be 0 as per the spec. 
https://github.com/zolkis/soletta/blob/master/doc/js-spec/gpio.md
`unsigned long long poll = 0;  // no polling`

bindings/nodejs/lib/pwm.js
```diff
@@ -43,6 +44,13 @@ exports.open = function( init ) {
             else
                 fulfill( PWMPin( soletta.sol_pwm_open( init.device, init.channel, config ) ) );
         }
+
+        if ( pwm ) {
```
nagineni: Please assign the instance of  sol_pwm to pwm before the `if`.
```
        if ( typeof init.name === 'string' && init.name !== "" ) {
            pwm = soletta.sol_pwm_open_by_label( init.name, config );
        } else {
            if ( raw )
                pwm = soletta.sol_pwm_open_raw( init.device, init.channel, config );
            else
                pwm = soletta.sol_pwm_open( init.device, init.channel, config );
        }
```

bindings/nodejs/samples/calamari.js
```diff
@@ -0,0 +1,110 @@
...
+Promise.all(gpio_map).then(_ => {
+    if (gpios.length == 6) {
+        btnLedControllers = [
+            {led: gpios[0], btn: gpios[3]},
+            {led: gpios[1], btn: gpios[4]},
+            {led: gpios[2], btn: gpios[5]},
+        ];
+        btnLedControllers.map(function(controller) {
+            var status = {status: true};
+            controller.btn.addEventListener("change", function(status, event) {
```
nagineni: We should use  `onchange` event instead of adding event listener. 

```
        btnLedControllers.map(function(controller) {                           
            var status = {status: true};                                       
            controller.btn.onchange = function(event) {                        
                status.status = !status.status;                                
                controller.led.write(status.status);                           
            };                                                                 
        });        
```

bindings/nodejs/samples/calamari.js
```diff
@@ -0,0 +1,110 @@
...
+    .then(function(led){
+        pwmLed = led;
+    })
+    .catch(function(fail){
+        console.log("Could not open PWM: ", fail);
+        process.exit();
+    });
+
+process.on("exit", (code) => {
+    gpios.map((gpio) => {
```
nagineni: Instead of `map` we can use traditional `for` loop to iterate over gpios array.

bindings/nodejs/lib/gpio.js
edersondisouza: Fair. I've just followed GPIO node type default.

bindings/nodejs/lib/pwm.js
edersondisouza: OMG! :tired_face: 
Thanks for catching!

nagineni:
@edersondisouza I tested gpio parts in the sample on my edison with led and button sensors. It works very well. Thanks again for adding it.

bindings/nodejs/samples/calamari.js
edersondisouza: Ok. Is there any special reason for that? Or just to save some typing?

nagineni: addEventListener() should work, but we are doing the same in high-level bindings and dispatching `onchange` event to clients as specified in the JS spec. Better to use the API specified in the spec so that we can test the EventHandler also.

bindings/nodejs/samples/calamari.js
```diff
@@ -0,0 +1,110 @@
...
+pwm.open({device: 0, channel: 0, period: 10000, dutyCycle: 0, enabled: true})
+    .then(function(led){
+        pwmLed = led;
+    })
+    .catch(function(fail){
+        console.log("Could not open PWM: ", fail);
+        process.exit();
+    });
+
+process.on("exit", (code) => {
```
nagineni: This is the correct way of closing opened IOs on exit. AFAIK, there is no automatic way of doing it properly, so we leave it up to developers to do the cleanup properly on exit.

edersondisouza:
v2 on #1930

bindings/nodejs/lib/aio.js
```diff
@@ -23,16 +23,23 @@ var soletta = require( 'bindings' )( 'soletta' ),
 exports.open = function( init ) {
 	return new Promise( function( fulfill, reject ) {
 		var precision = 12;
+		var aio;
 
 		if (init.precision)
 			precision = init.precision;
 
 		if (init.name && init.name != "") {
```
brunobottazzini: Use !== instead of !=

edersondisouza: On 25-04-2016 14:24, Bruno Bottazzini wrote:
> In bindings/nodejs/lib/aio.js
> <https://github.com/solettaproject/soletta/pull/1929#discussion_r60952408>:
>
>> @@ -23,16 +23,23 @@ var soletta = require( 'bindings' )( 'soletta' ),
>>  exports.open = function( init ) {
>>  	return new Promise( function( fulfill, reject ) {
>>  		var precision = 12;
>> +		var aio;
>>
>>  		if (init.precision)
>>  			precision = init.precision;
>>
>>  		if (init.name && init.name != "") {
>
> Use !== instead of !=

Hmmm... I didn't touch this line in my patch... But I may do that...

PR #1931, opened 2016-04-25
memdesc: Fix warning when comparing signed and unsigned expressions

otaviobp:
Cast value to unsigned before comparing with signed constants. As we are
checking if value is lower than 0 before comparing, we won't cast
negative numbers.


glima:
:+1:

PR #1932, opened 2016-04-25
HTTPS v2

ceolin:
from v1:
- add sol_http_server_config (avoid two functions to create a server)
- Added better explanation about how to test the sample

### DIFF
--- a/src/lib/common/sol-certificate-impl-linux.c
+++ b/src/lib/common/sol-certificate-impl-linux.c
@@ -21,6 +21,7 @@
 #include "sol-certificate.h"
 #define SOL_LOG_DOMAIN &_sol_certificate_log_domain
 #include "sol-log-internal.h"
+#include "sol-util-file.h"
 #include "sol-util-internal.h"
 #include "sol-vector.h"
 
@@ -146,4 +147,25 @@ sol_cert_get_filename(const struct sol_cert *cert)
     SOL_NULL_CHECK(cert, NULL);
 
     return cert->filename;
+}
+
+SOL_API struct sol_blob *
+sol_cert_get_contents(const struct sol_cert *cert)
+{
+    struct sol_blob *blob;
+    char *data;
+
+    SOL_NULL_CHECK(cert, NULL);
+
+    data = sol_util_load_file_string(cert->filename, NULL);
+    SOL_NULL_CHECK(data, NULL);
+
+    blob = sol_blob_new(SOL_BLOB_TYPE_DEFAULT, NULL, data, strlen(data));
+    if (!blob) {
+        SOL_WRN("Could not allocate memory for sol_blob");
+        free(data);
+        return NULL;
+    }
+
+    return blob;
 }
--- a/src/lib/comms/include/sol-http-server.h
+++ b/src/lib/comms/include/sol-http-server.h
@@ -18,6 +18,7 @@
 
 #pragma once
 
+#include <sol-certificate.h>
 #include <sol-http.h>
 #include <sol-network.h>
 #include <time.h>
@@ -52,6 +53,29 @@ extern "C" {
  * deleted with sol_http_server_del()
  */
 struct sol_http_server;
+
+/**
+ * @struct sol_http_server_config
+ * @brief
+ *
+ * It's created with sol_http_server_new() and should be later
+ * deleted with sol_http_server_del()
+ */
+struct sol_http_server_config {
+#ifndef SOL_NO_API_VERSION
+#define SOL_HTTP_SERVER_CONFIG_API_VERSION (1)  /**< compile time API version to be checked during runtime */
+    /**
+     * must match #SOL_HTTP_SERVER_CONFIG_API_VERSION at runtime.
+     */
+    uint16_t api_version;
+#endif
+
+    uint16_t port;
+    struct {
+        const struct sol_cert *cert;
+        const struct sol_cert *key;
+    } security;
+};
 
 /**
  * @struct sol_http_request
@@ -94,8 +118,10 @@ struct sol_http_progressive_response;
  * @param port The port where the server will bind.
  *
  * @return a handle to the server on success, otherwise @c NULL is returned.
+ *
+ * @see sol_http_server_config
  */
-struct sol_http_server *sol_http_server_new(uint16_t port);
+struct sol_http_server *sol_http_server_new(const struct sol_http_server_config *config);
 
 /**
  * @brief Destroy the @a server instance.
--- a/src/lib/comms/sol-http-server-impl-microhttpd.c
+++ b/src/lib/comms/sol-http-server-impl-microhttpd.c
@@ -52,6 +52,21 @@
 
 #define SOL_HTTP_REQUEST_BUFFER_SIZE 4096
 
+#ifndef SOL_NO_API_VERSION
+#define SOL_HTTP_SERVER_CONFIG_CHECK_API_VERSION(config, ...) \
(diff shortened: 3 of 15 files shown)


ceolin:
It's breaking the doxygen, i will fix it before commit.

src/lib/common/sol-certificate-impl-linux.c
```diff
@@ -147,3 +148,24 @@ sol_cert_get_filename(const struct sol_cert *cert)
...
+
+SOL_API struct sol_blob *
+sol_cert_get_contents(const struct sol_cert *cert)
+{
+    struct sol_blob *blob;
+    char *data;
+
+    SOL_NULL_CHECK(cert, NULL);
+
+    data = sol_util_load_file_string(cert->filename, NULL);
```
edersondisouza: Why don't you use second parameter to get `data` size instead of doing a `strlen`?

src/lib/comms/sol-http-server-impl-microhttpd.c
```diff
@@ -52,6 +52,21 @@
 
 #define SOL_HTTP_REQUEST_BUFFER_SIZE 4096
 
+#ifndef SOL_NO_API_VERSION
+#define SOL_HTTP_SERVER_CONFIG_CHECK_API_VERSION(config, ...) \
+    if (!(config)->api_version) { \
```
edersondisouza: Shouldn't this be  `if  ((config)->api_version != SOL_HTTP_SERVER_CONFIG_API_VERSION)`?

src/lib/comms/sol-http-server-impl-microhttpd.c
```diff
@@ -1035,12 +1052,28 @@ sol_http_server_new(uint16_t port)
...
-    server->daemon = MHD_start_daemon(MHD_USE_SUSPEND_RESUME,
-        port, NULL, NULL,
-        http_server_handler, server,
-        MHD_OPTION_NOTIFY_CONNECTION, notify_connection_cb, server,
-        MHD_OPTION_NOTIFY_COMPLETED, notify_connection_finished_cb, server,
-        MHD_OPTION_END);
+    if (config->security.cert && config->security.key) {
+        struct sol_blob *cert_contents, *key_contents;
+        cert_contents = sol_cert_get_contents(config->security.cert);
+        key_contents = sol_cert_get_contents(config->security.key);
```
edersondisouza: Shouldn't you NULL check `cert_contents` and `key_contents`?

src/lib/comms/include/sol-http-server.h
```diff
@@ -54,6 +55,29 @@ extern "C" {
...
+struct sol_http_server_config {
+#ifndef SOL_NO_API_VERSION
+#define SOL_HTTP_SERVER_CONFIG_API_VERSION (1)  /**< compile time API version to be checked during runtime */
+    /**
+     * must match #SOL_HTTP_SERVER_CONFIG_API_VERSION at runtime.
+     */
+    uint16_t api_version;
+#endif
+
+    uint16_t port;
```
edersondisouza: I know this is more of style, but why is `port` in config? It makes any new server creation more cumbersome when they couldn't be. SPI for instance, takes a config but separately bus number. Couldn't we live with something like that?

src/samples/http/server-https.c
```diff
@@ -0,0 +1,152 @@
...
+ *
+ * How to generate a certificate:
+ *
+ * openssl req -new > new.cert.csr
+ * openssl rsa -in privkey.pem -out new.cert.key
+ * openssl x509 -in new.cert.csr -out new.cert.cert -req -signkey new.cert.key -days 3652 -sha1
+ *
+ * To test it:
+ *
+ * run ./server-https -p 8080 -d "Hello HTTPS" -c new.cert.cert -k new,.cert.key
```
edersondisouza: I believe that also something one could do on browser/curl. I think that could be nice to point it to, as simple as it be, as you know, some people may not figure it out =D

edersondisouza: I'd guess that browser presents a warning message, it could be an opportunity to elaborate on that as well.

src/lib/common/sol-certificate-impl-linux.c
ceolin: damn it, i ve change to use blob and forgot to change it. Thanks

src/lib/comms/sol-http-server-impl-microhttpd.c
ceolin: copy&pasta fail

src/lib/comms/include/sol-http-server.h
ceolin: i do not know, i will check other places doing the same to identify the proper pattern.
I guess SPI is the different.

edersondisouza: SPI, UART, PWM and IIO are like that (but IIO doesn't count, I made it following the previous ones =P).

src/lib/comms/sol-http-server-impl-microhttpd.c
ceolin: not necessary, microhttpd will fail in this case, as the error treatment will be the same ...

src/lib/comms/include/sol-http-server.h
ceolin: ok, message-digest, socket, mqtt are doing the same of this patch

ceolin: I was thinking about, in this case it's possible consider the use of the default port in case any config is given. What do you think ?

src/lib/comms/sol-http-server-impl-microhttpd.c
edersondisouza: But you dereference it below...

src/lib/comms/sol-http-server-impl-microhttpd.c
```diff
@@ -1035,12 +1052,28 @@ sol_http_server_new(uint16_t port)
...
+        server->daemon = MHD_start_daemon(MHD_USE_SUSPEND_RESUME | MHD_USE_SSL,
+            config->port, NULL, NULL,
+            http_server_handler, server,
+            MHD_OPTION_NOTIFY_CONNECTION, notify_connection_cb, server,
+            MHD_OPTION_NOTIFY_COMPLETED, notify_connection_finished_cb, server,
+            MHD_OPTION_HTTPS_MEM_KEY, (char *)key_contents->mem,
+            MHD_OPTION_HTTPS_MEM_CERT, (char *)cert_contents->mem,
```
edersondisouza: ...here

src/lib/comms/include/sol-http-server.h
barbieri: I rather add all in config, we do not take a reference to it, so creating on the stack or as the sample is doing is okay.

src/lib/comms/sol-http-server-impl-microhttpd.c
barbieri: ederson is right, you'll get a segfault on `NULL->mem`

ceolin: there is a check for the certficates, what can happen is the function `sol_cer_get_contents` returns NULL, but in this case `MHD_start_daemon` will fail.

ceolin: ops, this was before the changes, sorry

ceolin:
fixes in https://github.com/solettaproject/soletta/pull/1934

Issue #1721, opened 2016-03-23
API Changes: OIC


cabelitos:
#### Task Description
- [ ] Remove `int:0` from `struct sol_oic_platform_information`
- [ ] `struct sol_oic_platform_information` -> `struct sol_oic_platform_info`
- [ ] Use `int64_t` for `system_time` in `struct sol_oic_platform_information`
- [ ] Remove `int:0` from `struct sol_oic_server_information`
- [ ] `struct sol_oic_server_information` -> `struct sol_oic_device_info`
- [ ] Documents the fields of `struct sol_oic_map_reader`
- [ ] `enum sol_oic_map_loop_reason` -> `enum sol_oic_map_loop_status`
- [ ] `href` -> `path` in `struct sol_oic_resource`
- [ ] `is_observing` -> `is_observed` in `struct sol_oic_resource`
- [ ] `sol_oic_notify_observers()` -> `sol_oic_server_notify_observers()`
- [ ] The macros `OIC_PLATFORM_ID`, `OIC_MANUFACTURER_NAME` and so on, should be set using Kconfig (make menuconfig)
- [ ] Functions that return `bool` - return type should be changed to `int` and they should return `0` (zero) on success and `-errno` on error.
- [ ] First callback parameter should be the user's data.

#### Dependencies
None.

PR #1933, opened 2016-04-25
oic: Make oic server calls async and improve API

### LINKED ISSUE #1821: OIC Make resource API async
#### Change OIC resource API to be async

``
    struct {
        sol_coap_responsecode_t (*handle)(const struct sol_network_link_addr *cliaddr,
            const void *data, const struct sol_oic_map_reader *input, struct sol_oic_map_writer *output);
    }
    get,
    put,
    post,
    del;
``

This OIC API should be async and follow the Soletta's pattern.

otaviobp:
- Oic server requests are now ASYNC
 - Improve server and client API so we won't need a callback to fill data in packets. The packets are first created and then sent.

JS bindings were updated and tests are not broken, but the API is still async and using same API. I'll count on people that are working on JS api to improve that.

Fixes #1821

### DIFF
--- a/src/lib/comms/include/sol-oic-client.h
+++ b/src/lib/comms/include/sol-oic-client.h
@@ -113,6 +113,18 @@ struct sol_oic_resource {
 };
 
 /**
+ * @struct sol_oic_client_request
+ *
+ * @brief Information about a client request.
+ *
+ * @see sol_oic_client_create_request
+ * @see sol_oic_client_create_non_confirmable_request
+ * @see sol_oic_client_request_free
+ * @see sol_oic_client_resource_request
+ */
+struct sol_oic_client_request;
+
+/**
  * @brief Creates a new OIC client intance.
  *
  * @return A new OIC client instance, or NULL in case of failure.
@@ -304,22 +316,14 @@ bool sol_oic_client_get_server_info_by_addr(struct sol_oic_client *client,
     const void *data);
 
 /**
- * @brief Send a request packet to server for specific @a resource.
+ * @brief Send a @a request packet to server.
  *
- * Send a CoAP confirmable request packet to server that contains the
- * @a resource and wait for a response. When the response arrives, @a callback
- * will be called.
+ * Send a CoAP @a request packet to server and wait for a response. When the
+ * response arrives, @a callback will be called.
  *
  * @param client An oic client instance.
- * @param res The resource that is going to receive the request.
- * @param method The coap request method as documented in @ref
- *        sol_coap_method_t.
- * @param fill_repr_map A callback to be called to fill the request data.
- *        Parameter @a data is a pointer to user's @a fill_repr_map_data and
- *        @a repr_map is a handler to write data to request packet. Use @ref
- *        sol_oic_map_append() to append data to @a repr_map. If @c NULL,
- *        no data will be added to request.
- * @param fill_repr_map_data User's data to be passed to @a fill_repr_map.
+ * @param request A request created using @ref sol_oic_client_create_request()
+ *        or sol_oic_client_create_non_confirmable_request().
  * @param callback Callback to be called when a response from this request
  *        arrives. Parameter @a response_code is the header response code of
  *        this request, @a cli is the @a client used to perform the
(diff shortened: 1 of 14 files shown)


src/lib/comms/include/sol-oic-client.h
```diff
@@ -328,59 +332,52 @@ bool sol_oic_client_get_server_info_by_addr(struct sol_oic_client *client,
...
+ * @return @c 0 on success or a negative number on errors.
  */
-bool sol_oic_client_resource_request(struct sol_oic_client *client, struct sol_oic_resource *res,
-    sol_coap_method_t method,
-    bool (*fill_repr_map)(void *data, struct sol_oic_map_writer *repr_map),
-    void *fill_repr_map_data,
-    void (*callback)(sol_coap_responsecode_t response_code, struct sol_oic_client *cli, const struct sol_network_link_addr *addr,
-    const struct sol_oic_map_reader *repr_vec, void *data),
-    const void *callback_data);
+int sol_oic_client_resource_request(struct sol_oic_client *client, struct sol_oic_client_request *request, void (*callback)(sol_coap_responsecode_t response_code, struct sol_oic_client *cli, const struct sol_network_link_addr *addr, const struct sol_oic_map_reader *repr_vec, void *data), const void *callback_data);
```
barbieri: always provide `void *data` as the first callback parameter, that's our convention. OIC does pretty bad regarding that consistency :-(

src/lib/comms/include/sol-oic-client.h
```diff
@@ -328,59 +332,52 @@ bool sol_oic_client_get_server_info_by_addr(struct sol_oic_client *client,
...
+void sol_oic_client_request_free(struct sol_oic_client_request *request);
+
+/**
+ * @brief Get the packet data from a client request.
+ *
+ * @param request The request to retrieve the data.
+ *
+ * @return The packet data from this request.
+ */
+struct sol_oic_map_writer *sol_oic_client_request_get_data(struct sol_oic_client_request *request);
```
barbieri: you return a writer, but neither the docs or the name make it clear why.

src/lib/comms/sol-oic-client.c
```diff
@@ -76,6 +76,14 @@ struct server_info_ctx {
     int64_t token;
 };
 
+struct sol_oic_client_request {
+    struct sol_coap_packet *pkt;
+    struct sol_oic_map_writer data;
+    bool (*cb)(struct sol_coap_server *server, struct sol_coap_packet *req, const struct sol_network_link_addr *addr, void *data);
```
barbieri: where do you store the callback's data?

src/lib/comms/sol-oic-client.c
```diff
@@ -927,153 +935,180 @@ _one_shot_resource_request_cb(struct sol_coap_server *server,
...
+    struct sol_oic_client_request *request;
+    char *href;
+
+    if (type != SOL_COAP_TYPE_CON && type != SOL_COAP_TYPE_NONCON) {
+        SOL_WRN("Only SOL_COAP_TYPE_CON and SOL_COAP_TYPE_NONCON requests are"
+            " supported");
+        return NULL;
     }
 
+    request = calloc(1, sizeof(struct sol_oic_client_request));
```
barbieri: review all memory allocation, here it's missing NULL_CHECK :-(

src/lib/comms/sol-oic-client.c
```diff
@@ -927,153 +935,180 @@ _one_shot_resource_request_cb(struct sol_coap_server *server,
...
+    return -ECOMM;
+}
 
-        res->observe.token = ctx->token;
-        sol_coap_add_option(req, SOL_COAP_OPTION_OBSERVE, &reg, sizeof(reg));
-        cb = _resource_request_cb;
-    } else {
-        cb = _one_shot_resource_request_cb;
+static struct sol_oic_client_request *
+create_request(sol_coap_method_t method, sol_coap_msgtype_t type, struct sol_oic_resource *res, bool isObserve)
```
barbieri: nowhere we're using `isObserve` in camelCase

src/lib/comms/sol-oic-client.c
```diff
@@ -927,153 +935,180 @@ _one_shot_resource_request_cb(struct sol_coap_server *server,
...
-SOL_API bool
-sol_oic_client_resource_non_confirmable_request(struct sol_oic_client *client, struct sol_oic_resource *res,
-    sol_coap_method_t method,
-    bool (*fill_repr_map)(void *data, struct sol_oic_map_writer *repr_map),
-    void *fill_repr_map_data,
-    void (*callback)(sol_coap_responsecode_t response_code, struct sol_oic_client *cli, const struct sol_network_link_addr *addr,
+SOL_API void
+sol_oic_client_request_free(struct sol_oic_client_request *request)
+{
+    if (request->pkt)
```
barbieri: NULL check all API entry point parameters

barbieri:
needs a more careful review regarding checks and API consistency. Since you're breaking the API, take the time to align it with http and others.

src/lib/comms/include/sol-oic-client.h
otaviobp: I didn't change that in this patch and I think this should be done in other commit. I updated the issue to fix OIC API (https://github.com/solettaproject/soletta/issues/1721) adding this.

src/lib/comms/include/sol-oic-client.h
otaviobp: It was clear to me, but I agree that it may be confusing because data can be a reader or a writer.

PR #1934, opened 2016-04-25
HTTPS v3

ceolin:
from v2:
 - Added commit fixing blob check api macro
 - Fix possible null pointer access
 - Avoid strelen on certificate
 - Added more description in the test

### DIFF
--- a/src/lib/common/sol-certificate-impl-linux.c
+++ b/src/lib/common/sol-certificate-impl-linux.c
@@ -21,6 +21,7 @@
 #include "sol-certificate.h"
 #define SOL_LOG_DOMAIN &_sol_certificate_log_domain
 #include "sol-log-internal.h"
+#include "sol-util-file.h"
 #include "sol-util-internal.h"
 #include "sol-vector.h"
 
@@ -146,4 +147,26 @@ sol_cert_get_filename(const struct sol_cert *cert)
     SOL_NULL_CHECK(cert, NULL);
 
     return cert->filename;
+}
+
+SOL_API struct sol_blob *
+sol_cert_get_contents(const struct sol_cert *cert)
+{
+    struct sol_blob *blob;
+    size_t size;
+    char *data;
+
+    SOL_NULL_CHECK(cert, NULL);
+
+    data = sol_util_load_file_string(cert->filename, &size);
+    SOL_NULL_CHECK(data, NULL);
+
+    blob = sol_blob_new(SOL_BLOB_TYPE_DEFAULT, NULL, data, size);
+    if (!blob) {
+        SOL_WRN("Could not allocate memory for sol_blob");
+        free(data);
+        return NULL;
+    }
+
+    return blob;
 }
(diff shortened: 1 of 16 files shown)


src/lib/common/sol-certificate-impl-linux.c
```diff
@@ -147,3 +148,25 @@ sol_cert_get_filename(const struct sol_cert *cert)
...
+
+    data = sol_util_load_file_string(cert->filename, &size);
+    SOL_NULL_CHECK(data, NULL);
+
+    blob = sol_blob_new(SOL_BLOB_TYPE_DEFAULT, NULL, data, size);
+    if (!blob) {
+        SOL_WRN("Could not allocate memory for sol_blob");
+        free(data);
+        return NULL;
+    }
```
barbieri: btw, just remembered that in `src/modules/flow/file/file.c`  we have `file_reader_blob_type`. We should make that a public one in sol-util-file.h and share it here. It would be more efficient than using `sol_util_load_file_string()` since if possible it wouldn't copy the contents to memory, rather use mmap()

barbieri:
+1, the blob/file-reader you can do in a different PR

src/lib/comms/sol-http-server-impl-microhttpd.c
```diff
@@ -1035,12 +1052,32 @@ sol_http_server_new(uint16_t port)
...
+        server->daemon = MHD_start_daemon(MHD_USE_SUSPEND_RESUME | MHD_USE_SSL,
+            config->port, NULL, NULL,
+            http_server_handler, server,
+            MHD_OPTION_NOTIFY_CONNECTION, notify_connection_cb, server,
+            MHD_OPTION_NOTIFY_COMPLETED, notify_connection_finished_cb, server,
+            MHD_OPTION_HTTPS_MEM_KEY, (char *)key_contents->mem,
+            MHD_OPTION_HTTPS_MEM_CERT, (char *)cert_contents->mem,
+            MHD_OPTION_END);
+        sol_blob_unref(cert_contents);
+        sol_blob_unref(key_contents);
```
vcgomes: Just thinking, this information is sensitive, so perhaps it is a good idea to clear this memory when are done using it, to reduce any risk of exposing it.

src/samples/http/server-https.c
```diff
@@ -0,0 +1,154 @@
...
+    }
+
+    server = sol_http_server_new(&(struct sol_http_server_config) {
+        SOL_SET_API_VERSION(.api_version = SOL_HTTP_SERVER_CONFIG_API_VERSION, )
+        .port = port,
+        .security = {
+            .cert = cert,
+            .key = key,
+        }
+    });
```
vcgomes: Move the unref of the certificate/key here? So we show the right way of doing stuff: keep sensitive stuff in memory for as little as possible.

vcgomes:
Just a few comments, see if they make sense. The code looks good.

PR #1936, opened 2016-04-26
gpio: Adapt the Zephyr changes in GPIO callbacks

ibriano:
Zephyr now supports adding multiple callbacks to a port, so we make use
of the new API.
As a bonus, change the implementation to make it ready to work with
more than one port, so when we have a way of knowing how many there are
(and how to identify them), most of the code is ready for it.

### DIFF
--- a/src/lib/io/sol-gpio-impl-zephyr.c
+++ b/src/lib/io/sol-gpio-impl-zephyr.c
@@ -22,6 +22,7 @@
 /* Zephyr includes */
 #include "atomic.h"
 #include "gpio.h"
+#include "misc/util.h"
 
 #define SOL_LOG_DOMAIN &_log_domain
 #include "sol-gpio.h"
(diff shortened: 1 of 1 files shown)


src/lib/io/sol-gpio-impl-zephyr.c
```diff
@@ -33,35 +34,37 @@
...
         return;
 
-    irq = irq_pins;
-    irq_pins = 0;
+    irq = port->irq_pins;
+    port->irq_pins = 0;
 
-    SOL_PTR_VECTOR_FOREACH_IDX (&opened_pins, g, i) {
+    SOL_PTR_VECTOR_FOREACH_IDX (&port->opened_pins, g, i) {
         if (irq & (1U << g->pin)) {
```
glima: You could use ``BIT()`` on those as well.

glima:
The rest of the patch is fine.

src/lib/io/sol-gpio-impl-zephyr.c
ibriano: On Tue, 26 Apr 2016 15:15:43 -0700, Gustavo Lima Chaves wrote:
> >  
> > -    SOL_PTR_VECTOR_FOREACH_IDX (&opened_pins, g, i) {
> > +    SOL_PTR_VECTOR_FOREACH_IDX (&port->opened_pins, g, i) {
> >          if (irq & (1U << g->pin)) {
> 
> You could use ``BIT()`` on those as well.

Ugh, that was already like that and I didn't notice it.

src/lib/io/sol-gpio-impl-zephyr.c
```diff
@@ -179,32 +198,24 @@ sol_gpio_open_raw(uint32_t pin, const struct sol_gpio_config *config)
...
-            if (gpio_pin_disable_callback(device, g->pin) < 0) {
-                SOL_WRN("Couldn't disable gpio callback");
-            }
-            sol_ptr_vector_del(&opened_pins, i);
-            break;
-        }
-    }
+    port = gpio->port;
+    port->cb.pin_mask &= ~BIT(gpio->pin);
+    gpio_pin_disable_callback(port->device, gpio->pin);
```
vcgomes: I guess this can cause problems for multiple `sol_gpio_open()`s, for example, two different pieces (A and B) of code call `_open()` for pin 10, when A calls `_close()` the callback is disabled for B also. Or am I missing something?

ibriano: On Wed, 27 Apr 2016 07:04:34 -0700, Vinicius Gomes wrote:
> >      SOL_NULL_CHECK(gpio);
> >  
> > -    SOL_PTR_VECTOR_FOREACH_IDX (&opened_pins, g, i) {
> > -        if (g->pin == gpio->pin) {
> > -            if (gpio_pin_disable_callback(device, g->pin) < 0) {
> > -                SOL_WRN("Couldn't disable gpio callback");
> > -            }
> > -            sol_ptr_vector_del(&opened_pins, i);
> > -            break;
> > -        }
> > -    }
> > +    port = gpio->port;
> > +    port->cb.pin_mask &= ~BIT(gpio->pin);
> > +    gpio_pin_disable_callback(port->device, gpio->pin);
> 
> I guess this can cause problems for multiple `sol_gpio_open()`s, for example, two different pieces (A and B) of code call `_open()` for pin 10, when A calls `_close()` the callback is disabled for B also. Or am I missing something?

Two different pieces opening the same pin of the same port is wrong.

vcgomes: Makes sense. Don't know how I missed the check above.

vcgomes:
Looks good.

ibriano:
Merged

PR #1937, opened 2016-04-26
Node.js bindings: Add basic examples for GPIO JS bindings

nagineni:
This adds the following basic examples for GPIO JS bindings.
 - gpio_write.js: Blinks an LED connected to Edison GPIO pin
 - gpio_read.js: Read the value of the GPIO pin
 - gpio_monitor.js: Monitor the button press state and turn on/off an LED connected to Edison GPIO pin.

### DIFF
--- a/src/samples/nodejs/gpio_read.js
+++ b/src/samples/nodejs/gpio_read.js
@@ -1,0 +1,49 @@
+/*
+ * This file is part of the Soletta Project
+ *
+ * Copyright (C) 2015 Intel Corporation. All rights reserved.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+/*
+ * This sample code returns the value read from the GPIO
+ * input pin 48 on Edison.
+ */
+var gpio = require( "soletta/gpio" ),
+    gpioPin = null;
+
+gpio.open( {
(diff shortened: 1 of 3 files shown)


nagineni:
@edersondisouza @gabrielschulhof PTAL

src/samples/nodejs/gpio_read.js
```diff
@@ -0,0 +1,49 @@
...
+ * This sample code returns the value read from the GPIO
+ * input pin 48 on Edison.
+ */
+var gpio = require( "soletta/gpio" ),
+    gpioPin = null;
+
+gpio.open( {
+    pin: 48, // Setup pin 48 for reading
+    direction: "in", // Set the gpio direction to input
+    edge: "rising"
```
edersondisouza: As you only read value when 'rising', it only returns `true`, right? Maybe you could state that on doc, or read on both edges, so user sees both changes...

src/samples/nodejs/gpio_write.js
```diff
@@ -0,0 +1,58 @@
...
+    pin: 48
+} ).then( function( pin ) {
+    ledPin = pin; // Save the handle
+    changeState(); // Change the LED state.
+} ).catch( function( error ) {
+    console.log( "Could not open GPIO: ", error );
+    process.exit();
+} );
+
+process.on( "SIGINT", function() {
```
edersondisouza: I'd guess this is for 'Ctrl + C' to work, right? State that. I shall update my sample soon, as well.

edersondisouza:
Minor comments, looks ok.

src/samples/nodejs/gpio_read.js
nagineni: I'm using sol_gpio_read() in the sample, so it returns the value of the pin(i.e true/false) at the moment the function is called. Edge property makes difference in case of listening for the pin change event using cb() callback. To avoid the confusion, I will remove the edge property so that it will use the default value (i.e both).

src/samples/nodejs/gpio_write.js
nagineni: You are right. I will add a comment.

nagineni:
@edersondisouza Thanks for reviewing. I will fix review comments and submit a new PR.

PR #1942, opened 2016-04-27
dependency-resolver: split args to parse the compiler name for Yocto

brunobottazzini:
Yocto's TARGETCC does not only contain the compiler information but yet it
comes with some flags to crosscompile.

Example:
TARGETCC=i586-ostro-linux-gcc -m32 -march=i586 -Wa,-momit-lock-prefix=yes

### DIFF
--- a/data/scripts/dependency-resolver.py
+++ b/data/scripts/dependency-resolver.py
@@ -540,7 +540,7 @@ if __name__ == "__main__":
 
     args = parser.parse_args()
 
-    if args.compiler != default_compiler and not which(args.compiler):
+    if args.compiler != default_compiler and not which(args.compiler.split()[0]):
         print("ERROR: Invalid --compiler argument, no such file: %s" % args.compiler)
         exit(1)
 


bdilly:
+1

dorileo:
+1

data/scripts/dependency-resolver.py
```diff
@@ -540,7 +540,7 @@ def log_environment(context):
 
     args = parser.parse_args()
 
-    if args.compiler != default_compiler and not which(args.compiler):
+    if args.compiler != default_compiler and not which(args.compiler.split(" ")[0]):
```
vcgomes: [minor] You could use `split()`, if it's possible that tabs are present in the variable.

brunobottazzini: Indeed, using split() should cover more possibilities. I will merge with this thanks

data/scripts/dependency-resolver.py
```diff
@@ -540,7 +540,7 @@ def log_environment(context):
 
     args = parser.parse_args()
 
-    if args.compiler != default_compiler and not which(args.compiler):
+    if args.compiler != default_compiler and not which(args.compiler.split()[0]):
```
barbieri: this is wrong and will break `-DX="a b"` kind of option.

Use `shlex.split()` https://docs.python.org/2/library/shlex.html

brunobottazzini: Good call I just opened a PR with shlex.split() 

https://github.com/solettaproject/soletta/pull/1944

PR #1943, opened 2016-04-27
Fix issues on PR 1244

bdilly:
I've fixed a few issues in a commit on top of
PR #1244 commits

### DIFF
--- a/doc/doxygen/Doxyfile
+++ b/doc/doxygen/Doxyfile
@@ -272,7 +272,7 @@ EXPAND_ONLY_PREDEF     = YES
 SEARCH_INCLUDES        = YES
 INCLUDE_PATH           =
 INCLUDE_FILE_PATTERNS  =
-PREDEFINED             = __attribute__((x))=
+PREDEFINED             = __attribute__((x))=  DOXYGEN_RUN
 EXPAND_AS_DEFINED      =
 SKIP_FUNCTION_MACROS   = YES
 #---------------------------------------------------------------------------
--- a/doc/doxygen/Doxyfile
+++ b/doc/doxygen/Doxyfile
@@ -272,7 +272,7 @@ EXPAND_ONLY_PREDEF     = YES
 SEARCH_INCLUDES        = YES
 INCLUDE_PATH           =
 INCLUDE_FILE_PATTERNS  =
-PREDEFINED             = __attribute__((x))=  DOXYGEN_RUN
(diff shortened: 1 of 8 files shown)


doc/doxygen/Doxyfile
```diff
@@ -272,7 +272,7 @@ EXPAND_ONLY_PREDEF     = YES
 SEARCH_INCLUDES        = YES
 INCLUDE_PATH           =
 INCLUDE_FILE_PATTERNS  =
-PREDEFINED             = __attribute__((x))=
+PREDEFINED             = __attribute__((x))=  DOXYGEN_RUN
```
glima: I ran doxygen a couple of times and never saw this being set. Beats me when it's supposed to be set on to what.

src/lib/common/include/sol-atomic.h
```diff
@@ -0,0 +1,229 @@
...
+/* Order levels */
+#define SOL_ATOMIC_RELAXED memory_order_relaxed
+#define SOL_ATOMIC_CONSUME memory_order_consume
+#define SOL_ATOMIC_ACQUIRE memory_order_acquire
+#define SOL_ATOMIC_RELEASE memory_order_release
+#define SOL_ATOMIC_ACQ_REL memory_order_acq_rel
+#define SOL_ATOMIC_SEQ_CST memory_order_seq_cst
+
+/* Types */
+typedef atomic_int sol_atomic_int;
```
glima: We got atomic_uchar but I guess it's no big deal to fallback to int on boolean atomics.

src/lib/common/sol-worker-thread-impl-riot.c
```diff
@@ -137,7 +137,7 @@ sol_worker_thread_do(void *data)
     /* From this point forward, we can't allow a context switch. IRQ will be
      * re-enabled by the scheduler once this function returns */
     irq_disable();
-    sol_atomic_store(&thread->finished, true, SOL_ATOMIC_SEQ_CST);
+    sol_atomic_store(&thread->finished, true, SOL_ATOMIC_RELEASE);
```
glima: I can agree we do not require ordering there, but I hope there are no pitfalls.

doc/doxygen/Doxyfile
bdilly: Indeed I don't think it's having any effect in our project

I'll drop it.

PR #1944, opened 2016-04-27
dependency-resolver: replace the use of split to shlex.split.

brunobottazzini:
The shlex split will be able to parse correctly in case of flags like the
following:

-DX="a b"

### DIFF
--- a/data/scripts/dependency-resolver.py
+++ b/data/scripts/dependency-resolver.py
@@ -27,6 +27,7 @@ import socket
 import subprocess
 import sys
 import tempfile
+import shlex
 from shutil import which, move
 
 default_compiler = "gcc"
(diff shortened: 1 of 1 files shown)


brunobottazzini:
@barbieri, could you review this patch please ?

ibriano:
This makes me wonder, will that take care of something like: CC=/opt/i\ use\ spaces\ in\ paths/gcc ?

brunobottazzini:
@ibriano,

I ran a small test here: 

```
>>> import shlex
>>> a="/opt/i\ use\ spaces\ in\ paths/gcc"
>>> shlex.split(a)
['/opt/i use spaces in paths/gcc']
```

barbieri:
that's the idea :-D It handles shell escaping and splitting rules in account.

barbieri:
+1

just check if we need more of that in the dep-resolver, maybe other naive str.split() were used.

brunobottazzini:
I checked here and this is the only one split used in dep-resolver :)

Thanks!

Issue #1566, opened 2016-03-07
Mainloop: Make signal handling contingent upon an API call


gabrielschulhof:
#### Task Description
Soletta's default-on signal handling interferes with node.js signal handling. Signal handlers should only be touched as part of an API call (```sol_mainloop_handle_signals()``` maybe).

barbieri:
I'll take this one, I'm just back from PRC and after post-travel settles I'll review your comments about mainloop integration.

If we get that mainloop integration right, the posix "init" is not to be called, so no signal handler is to be installed by soletta.

Right now I already envision we miss some callbacks like to spawn threads (worker threads) and processes (fork_run), that must be fixed. However I will carefully review your comments and do some testing, maybe I can send a PR that includes a proposal for the node.js part as well.

Thanks for reporting.

barbieri:
Today I did an initial investigation based on my assumptions and the my early comment above. To answer myself first:

> Right now I already envision we miss some callbacks like to spawn threads (worker threads) and processes (fork_run), that must be fixed. However I will carefully review your comments and do some testing, maybe I can send a PR that includes a proposal for the node.js part as well.

Worker threads shouldn't need any extra work. Although we do call `sigprocmask()` to unblock signals, the `libuv` doesn't do anything special in https://github.com/joyent/libuv/blob/master/src/unix/thread.c#L66 Since we'd **not** be calling our `mainloop_init()`, we'd have no signals added to the mask set and the operation would be no-op. So far so good.

Fork-run is more-complex:
 - Libuv doesn't offer `uv_fork()`, just `uv_spawn()` that will not just `fork()`, but also `execvp()`. Likely because of windows port. Looking at https://github.com/joyent/libuv/blob/master/src/unix/process.c we can see that no special `sigprocmask()` are set, but since it will `execvp()`, some of its local setup will go away without special care, like fds marked with close-on-exec.
 - soletta basically just use "spawn" behavior, as we do `fork()` and then `exec()`. However we do simplify our interface by allowing pre-exec code to be done, like `subprocess` module will setup pipes. Libuv solves this by exposing many controls, those would enable you to create the pipes on your behalf.
 - glib provides its own API that does that in a similar way https://developer.gnome.org/glib/stable/glib-Spawning-Processes.html#g-spawn-async-with-pipes

[...]

PR #1946, opened 2016-04-27
Node.js: Handle main loop interruption

gabrielschulhof:
Also consolidates test helper files into a single directory.

This does not yet include the signal handler chaining discussed in https://github.com/solettaproject/soletta/issues/1566#user-content-signal-handler-juggling, because this is a real issue causing busy loops sometimes.


gabrielschulhof:
@nagineni this should make soletta handles behave just like node.js handles. So, if you don't clean them up and you handle SIGINT, it will hang.

The only remaining shortcoming is that if you add a ```SIGINT``` handler before you ```require( "bindings" )( "soletta" );``` then it won't fire. That's #1566, which I'm also working on.

Does this fix your issue?

nagineni:
@gabrielschulhof Tested this locally and it works as you mentioned. Cleaning up soletta handles in SIGINT handler makes the process exit as expected.

PR #1948, opened 2016-04-27
sol-bluetooth: Basic functionality [v2]

vcgomes:
This implements basic functionality for the Linux implementation of sol-bluetooth, this allows to power the adapter up (only the first adapter found) and scan for devices.

Changes from last version:
 - sol-gatt is implemented from Linux (BlueZ);
 - added `impl-none` for sol-bluetooth and sol-gatt, this allows the headers to be installed even if the sol-gatt is not available;

This is an updated version of https://github.com/solettaproject/soletta/pull/1825


vcgomes:
The style check fail is in unrelated code.

src/lib/comms/sol-network-impl-linux.c
```diff
@@ -75,6 +75,78 @@ struct sol_network {
...
+    const char *ptr;
+    char *endptr = NULL;
+
+    /* FIXME */
+    addr->family = SOL_NETWORK_FAMILY_BLUETOOTH;
+    addr->addr.bt_type = SOL_NETWORK_BT_ADDR_BASIC_RATE;
+
+    ptr = str;
+
+    for (i = 0; i < 6; i++) {
```
glima: Nitpick: ``sizeof(addr->addr.bt_addr)``

src/lib/comms/sol-network-impl-linux.c
```diff
@@ -75,6 +75,78 @@ struct sol_network {
...
+        ba[i] = strtoul(ptr, &endptr, 16);
```
glima: Maybe ``sol_util_strtoul()``?

src/lib/comms/sol-bluetooth.c
```diff
@@ -0,0 +1,200 @@
...
+    SOL_NULL_CHECK(uuid, -EINVAL);
+
+    switch (str->len) {
+    case 4:
+        uuid->type = SOL_BT_UUID_TYPE_16;
+        uuid->val16 = strtoul(str->data, NULL, 16);
+        break;
+    case 8:
+        uuid->type = SOL_BT_UUID_TYPE_32;
+        uuid->val32 = strtoull(str->data, NULL, 16);
```
glima: ``sol_util_strtoul()`` again

edersondisouza: On 29-04-2016 11:19, Gustavo Lima Chaves wrote:
> In src/lib/comms/sol-bluetooth.c
> <https://github.com/solettaproject/soletta/pull/1948#discussion_r61584488>:
>
>> +SOL_API int
>> +sol_bt_uuid_from_str(struct sol_bt_uuid *uuid, const struct sol_str_slice *str)
>> +{
>> +    int r;
>> +
>> +    SOL_NULL_CHECK(str, -EINVAL);
>> +    SOL_NULL_CHECK(uuid, -EINVAL);
>> +
>> +    switch (str->len) {
>> +    case 4:
>> +        uuid->type = SOL_BT_UUID_TYPE_16;
>> +        uuid->val16 = strtoul(str->data, NULL, 16);
>> +        break;
>> +    case 8:
>> +        uuid->type = SOL_BT_UUID_TYPE_32;
>> +        uuid->val32 = strtoull(str->data, NULL, 16);
>
> |sol_util_strtoul()| again

`sol_util_strtoul` only advantage is being able to work with 
`sol_str_slice`, isn't it?

src/lib/comms/sol-bluetooth-impl-bluez.c
```diff
@@ -0,0 +1,981 @@
...
+
+    changed = ctx->adapter_powered != powered;
+
+    ctx->adapter_powered = powered;
+
+    return changed;
+
+skip:
+    r = sd_bus_message_skip(m, "b");
+    SOL_INT_CHECK(r, < 0, false);
```
glima: Nitpick: you return false anyway

src/lib/comms/sol-bluetooth-impl-bluez.c
```diff
@@ -0,0 +1,981 @@
...
+
+    SOL_PTR_VECTOR_FOREACH_IDX (&ctx->devices, d, idx)
+        destroy_device(d);
+    sol_ptr_vector_clear(&ctx->devices);
+
+    SOL_PTR_VECTOR_FOREACH_IDX (&ctx->sessions, s, idx) {
+        s->enabled((void *)s->user_data, false);
+        free(s);
+    }
+    sol_ptr_vector_clear(&ctx->sessions);
```
glima: Is adapter_path freed?

src/lib/comms/sol-bluetooth-impl-bluez.c
```diff
@@ -0,0 +1,981 @@
...
+    bool paired;
+    int r;
+
+    d = find_device_by_path(ctx, path);
+    SOL_NULL_CHECK(d, false);
+
+    info = &d->info;
+
+    r = sd_bus_message_read_basic(m, 'b', &paired);
+    SOL_INT_CHECK(r, < 0, false);
```
glima: You don't skip of failure here?

src/lib/comms/sol-bluetooth-impl-bluez.c
```diff
@@ -0,0 +1,981 @@
...
+    bool connected;
+    int r;
+
+    d = find_device_by_path(ctx, path);
+    SOL_NULL_CHECK(d, false);
+
+    info = &d->info;
+
+    r = sd_bus_message_read_basic(m, 'b', &connected);
+    SOL_INT_CHECK(r, < 0, false);
```
glima: And here

src/lib/comms/sol-bluetooth-impl-bluez.c
```diff
@@ -0,0 +1,981 @@
...
+    int16_t rssi;
+    int r;
+
+    d = find_device_by_path(ctx, path);
+    SOL_NULL_CHECK(d, false);
+
+    info = &d->info;
+
+    r = sd_bus_message_read_basic(m, 'n', &rssi);
+    SOL_INT_CHECK(r, < 0, false);
```
glima: Ditto.

src/lib/comms/sol-bluetooth-impl-bluez.c
```diff
@@ -0,0 +1,981 @@
...
+    int r;
+
+    d = find_device_by_path(ctx, path);
+    if (d)
+        return;
+
+    d = calloc(1, sizeof(*d));
+    SOL_NULL_CHECK(d);
+
+    info = &d->info;
```
glima: Not used?

src/lib/comms/include/sol-bluetooth.h
```diff
@@ -0,0 +1,312 @@
...
+ * different types, and compared returning if they are equal.
+ *
+ * @param u1 UUID to be compared.
+ * @param u2 UUID to be compared.
+ *
+ * @return True if UUIDs are equal, False otherwise.
+ */
+bool sol_bt_uuid_equal(const struct sol_bt_uuid *u1, const struct sol_bt_uuid *u2);
+
+struct sol_bt_conn;
```
glima: No docs?

src/lib/comms/include/sol-bluetooth.h
```diff
@@ -0,0 +1,312 @@
...
+ * In case the connection is not already established, the connection attempt
+ * will be cancelled after this is called.
+ *
+ * @param conn The reference to a connection.
+ *
+ * @return 0 on success, -errno otherwise.
+ */
+int sol_bt_disconnect(struct sol_bt_conn *conn);
+
+struct sol_bt_session;
```
glima: Ditto

src/lib/comms/include/sol-bluetooth.h
```diff
@@ -0,0 +1,312 @@
...
+    bool in_range;
+};
+
+/**
+ * @brief Over which transport should a scan be performed.
+ */
+enum sol_bt_transport {
+    SOL_BT_TRANSPORT_ALL,
+    SOL_BT_TRANSPORT_LE,
+    SOL_BT_TRANSPORT_BREDR,
```
glima: No entry docs?

src/lib/comms/include/sol-bluetooth.h
```diff
@@ -0,0 +1,312 @@
...
+ * @brief Converts string to a transport, @see sol_bt_transport.
+ *
+ * @param str A with the transport representation.
+ *
+ * @return a pointer to a string representing the transport on success,
+ *         NULL otherwise.
+ */
+enum sol_bt_transport sol_bt_transport_from_str(const char *str);
+
+struct sol_bt_scan_pending;
```
glima: Ditto

src/lib/comms/include/sol-gatt.h
```diff
@@ -0,0 +1,351 @@
...
+/**
+ * @brief Set of flags for Descriptor attributes
+ */
+enum sol_gatt_desc_flags {
+    SOL_GATT_DESC_FLAGS_READ = (1 << 0),
+    SOL_GATT_DESC_FLAGS_WRITE = (1 << 1),
+    SOL_GATT_DESC_FLAGS_ENCRYPT_READ = (1 << 2),
+    SOL_GATT_DESC_FLAGS_ENCRYPT_WRITE = (1 << 3),
+    SOL_GATT_DESC_FLAGS_ENCRYPT_AUTHENTICATED_READ = (1 << 4),
+    SOL_GATT_DESC_FLAGS_ENCRYPT_AUTHENTICATED_WRITE = (1 << 5),
```
glima: Hell of a job to document each entry... :/

src/lib/comms/include/sol-gatt.h
```diff
@@ -0,0 +1,351 @@
...
+ */
+#define SOL_GATT_CHARACTERISTIC_UUID_16(_uuid, _flags, ...) \
+    { .type = SOL_GATT_ATTR_TYPE_CHARACTERISTIC, \
+      .uuid = { .type = SOL_BT_UUID_TYPE_16, \
+                .val16 = (_uuid) }, \
+      .flags = (_flags), \
+      __VA_ARGS__ }
+
+/**
+ * @brief Helper to the construction of a GATT Characteristic with 16-bit UUID.
```
glima: descriptor

src/lib/comms/sol-bluetooth-impl-bluez.c
```diff
@@ -440,6 +494,27 @@ device_property_rssi_set(void *data, const char *path, sd_bus_message *m)
...
+    struct context *ctx = data;
+    struct device_info *d;
+    bool resolved;
+    int r;
+
+    d = find_device_by_path(ctx, path);
+    SOL_NULL_CHECK(d, false);
+
+    r = sd_bus_message_read_basic(m, 'b', &resolved);
+    SOL_INT_CHECK(r, < 0, false);
```
glima: Skip?

src/lib/comms/sol-bluetooth-impl-bluez.c
```diff
@@ -465,6 +540,93 @@ static const struct sol_bus_properties device_properties[] = {
...
+static bool
+attr_property_uuid_set(void *data, const char *path, sd_bus_message *m)
+{
+    struct sol_gatt_attr *attr = data;
+    const char *str;
+    struct sol_str_slice slice;
+    int r;
+
+    r = sd_bus_message_read_basic(m, 's', &str);
+    SOL_INT_CHECK(r, < 0, false);
```
glima: Ditto

src/lib/comms/sol-bluetooth-impl-bluez.c
```diff
@@ -465,6 +540,93 @@ static const struct sol_bus_properties device_properties[] = {
...
+    const struct sol_gatt_attr *attr = data;
+    struct subscription *sub;
+    struct sol_buffer buf;
+    const void *buf_data;
+    size_t len;
+    uint16_t idx;
+    int r;
+
+    r = sd_bus_message_read_array(m, 'y', &buf_data, &len);
+    SOL_INT_CHECK(r, < 0, false);
```
glima: Ditto

src/lib/comms/sol-bluetooth-impl-bluez.c
```diff
@@ -584,13 +813,288 @@ device_removed(void *data, const char *path)
...
 }
 
+static struct device_info *
+match_device_by_prefix(struct context *ctx, const char *path)
+{
+    struct device_info *d;
+    uint16_t i;
+
+    SOL_PTR_VECTOR_FOREACH_IDX (&ctx->devices, d, i) {
+        if (strncmp(d->path, path, strlen(d->path)) == 0)
```
glima: ``strstartswith()``

src/lib/comms/sol-bluetooth-impl-bluez.c
```diff
@@ -584,13 +813,288 @@ device_removed(void *data, const char *path)
...
+
+    op->slot = sd_bus_slot_unref(op->slot);
+
+    if (sol_bus_log_callback(m, userdata, ret_error)) {
+        r = -EINVAL;
+        goto done;
+    }
+
+    r = sd_bus_message_read_array(m, 'y', &data, &len);
+    SOL_INT_CHECK_GOTO(r, < 0, done);
```
glima: buf may not be init-ed

src/lib/comms/sol-bluetooth.c
glima: Fair

src/lib/comms/sol-bluetooth-impl-bluez.c
vcgomes: Yeah, but I want the warning, it could help debugging sessions.

src/lib/comms/sol-bluetooth-impl-bluez.c
vcgomes: Ah, I missed it. Thanks.

src/lib/comms/sol-bluetooth-impl-bluez.c
vcgomes: Yeah, things are pretty inconsistent right now, I need to think a little more about this, and see how should we handle this kind of failure. Thanks.

src/lib/comms/sol-bluetooth-impl-bluez.c
vcgomes: it is used when initializing `info->uuids`

src/lib/comms/sol-bluetooth-impl-bluez.c
vcgomes: Cool. Didn't know we had this. Will fix.

src/lib/comms/sol-bluetooth-impl-bluez.c
vcgomes: Ugh. Thanks. Fixed.

src/lib/comms/sol-network-impl-linux.c
vcgomes: Fixed. Thanks.

vcgomes:
The version I just pushed should have a more consistent way of handling with properties, and the rest of the structures and entries should be documented. Thanks.

src/lib/comms/include/sol-bluetooth.h
```diff
@@ -0,0 +1,327 @@
...
+
+/**
+ * @brief Convert an string to an UUID.
+ *
+ * @param uuid The uuid in which to store the value.
+ * @param str The string from which to convert.
+ *
+ * @return 0 on success, -errno on failure.
+ */
+int sol_bt_uuid_from_str(struct sol_bt_uuid *uuid, const struct sol_str_slice *str);
```
ceolin: If i am not mistaken, our API usually receive `const struct sol_str_slice` instead of `const struct sol_str_slice *`

src/lib/comms/include/sol-bluetooth.h
```diff
@@ -0,0 +1,327 @@
...
+ * should be called.
+ *
+ * @param enabled Function to be called when the controller changes state.
+ * @param user_data User data to be provided to the enabled() callback.
+ *
+ * @return a reference to a session, used for returning the system to its
+ * previous state, @see sol_bt_disable().
+ */
+struct sol_bt_session *sol_bt_enable(
+    void (*enabled)(void *data, bool powered),
```
ceolin: I am confuse with the usage of `powered`.

When this function is called it tryies to power on the adapter, in case of fail it will return NULL and the callback `enabled` will neve be called. So, this should be used to monitor when the adapter is powered off ?

src/lib/comms/include/sol-bluetooth.h
```diff
@@ -0,0 +1,327 @@
...
+
+/**
+ * @brief Converts string to a transport, @see sol_bt_transport.
+ *
+ * @param str A with the transport representation.
+ *
+ * @return a pointer to a string representing the transport on success,
+ *         NULL otherwise.
+ */
+enum sol_bt_transport sol_bt_transport_from_str(const char *str);
```
ceolin: This is inconsistent with `sol_bt_uuid_from_str`

src/lib/comms/include/sol-bluetooth.h
```diff
@@ -0,0 +1,327 @@
...
+ * @return a reference to a session, used for returning the system to its
+ * previous state, @see sol_bt_disable().
+ */
+struct sol_bt_session *sol_bt_enable(
+    void (*enabled)(void *data, bool powered),
+    const void *user_data);
+
+/**
+ * @brief Disables a session, returning the controller to its previous state.
+ *
```
ceolin: the function is not doing it. It is alwayes powering off the adapter.

` r = adapter_set_powered(&context, context.adapter_path, false); `

It should be fixed.

src/samples/bluetooth/connect-paired.c
```diff
@@ -0,0 +1,114 @@
...
+
+    if (session)
+        sol_bt_disable(session);
+}
+
+static void
+startup(void)
+{
+    session = sol_bt_enable(enabled, NULL);
+    SOL_NULL_CHECK(session);
```
ceolin: in case of fail shouldn't you call `sol_quit_with_code()` ? I mean, if it fails not even the function `enabled` will be called so it's better to exit.

src/lib/comms/include/sol-bluetooth.h
vcgomes: Yeah, this is basically a monitor, that also enables the controller. The entry point for the all the API.

The callback will only be called if it returns successfully, I will document this.

This behaviour was thought to make users handle the case where some other part of the system changes the state of the Bluetooth controller, for example, the user enables airplane mode.

src/lib/comms/include/sol-bluetooth.h
vcgomes: Not always, only when there are no more active sessions. But will give this a little more thought.

src/samples/bluetooth/connect-paired.c
vcgomes: Yeah, makes sense. Will fix.

ceolin:
I suggest you commit what was already reviewed and there is nothing to fix

vcgomes:
I already have most of comments already fixed. Just giving the series another round of testing.

I will re-send the series, and start applying from the `sol-network` changes tomorrow. Thanks.

src/lib/comms/include/sol-bluetooth.h
vcgomes: That seems to be the convention when writing code that will convert to/from enums.

As this kind of conversion will be mostly used by language bindings, this is easier it seems.

vcgomes:
Please note that this series exposed a bug in BlueZ, a proposed fix was already sent for review. The problem is the ordering of the `ServicesResolved` PropertyChanged signal, which is used to trigger the callback of `sol_gatt_discover()`, without the fix, it's possible that the callback is called without all the attributes being discovered.

src/lib/comms/include/sol-bluetooth.h
```diff
@@ -0,0 +1,333 @@
...
+ * @brief Over which transport should a scan be performed.
+ */
+enum sol_bt_transport {
+    /** @brief Discover devices over All transports */
+    SOL_BT_TRANSPORT_ALL,
+    /** @brief Discover devices over the Bluetooth Low Energy transport */
+    SOL_BT_TRANSPORT_LE,
+    /** @brief Discover devices over the Bluetooth Basic Rate transport */
+    SOL_BT_TRANSPORT_BREDR,
+};
```
ceolin: Would not be better:

```
enum sol_bt_transport {
    SOL_BT_TRANSPORT_LE,
    SOL_BT_TRANSPORT_BREDR,'
    SOL_BT_TRANSPORT_ALL = SOL_BT_TRANSPORT_LE | SOL_BT_TRANSPORT_BREDR,
};
```

src/lib/comms/include/sol-bluetooth.h
ceolin: Actually after the first call, all further calls will not have its callback called (unless the adapter changes) and they will not know if the adpater is powered on.
I think this can be a source of mistakes.

src/lib/comms/include/sol-bluetooth.h
vcgomes: At least from the BlueZ's side, it would be the same really, but if it makes your life any easier, I can happily change it.

src/lib/comms/include/sol-bluetooth.h
vcgomes: Ugh, nice catch. Fixed. Thanks.

src/lib/comms/include/sol-bluetooth.h
vcgomes: After discussing it offline, I did change it to be like you suggested.

vcgomes:
Already integrated.

PR #1954, opened 2016-04-28
zephyr: Don't use task_pipe from non-task contexts

ibriano:
As it turns out, we don't need two main loop implementations, and in
fact, one of them was entirely wrong. We can't use task_pipe from fiber
or ISR contexts.
If anyone knows what kind of drugs I was in when I initially wrote
that, I may need them again in the future... for science.


edersondisouza:
looks ok.

vcgomes:
:+1:

ibriano:
Merged

PR #1955, opened 2016-04-28
sol-atomic: remove DOXYGEN_RUN usage

bdilly:
Now we should be good on doxygen


glima:
OK

glima:
Not tested, though (relying on bot)

PR #1958, opened 2016-05-02
oic: Make oic server calls async and improve API - V2

### LINKED ISSUE #1722: Testing with Soletta flows: Improve page text
#### Task Description
Page: https://github.com/solettaproject/soletta/wiki/Testing-with-Soletta-flows
We can discuss why we should use the test nodes, give more examples and talk about how they are used in the Soletta development workflow itself.

#### Dependencies
None

otaviobp:
Fixes #1722 and #1821

Replaces #1933


src/lib/comms/include/sol-oic-server.h
```diff
@@ -215,59 +215,74 @@ struct sol_oic_resource_type {
...
      * @brief Callback handle to POST requests
      *
-     * @param cliaddr The client address.
      * @param data The user's data pointer.
-     * @param input Handler to read request packet data using, for example,
-     *        @ref SOL_OIC_MAP_LOOP() macro.
-     * @param output Handler to write data to response packet using, for
-     *        example, @ref sol_oic_map_append().
+     * @param request Information about this request. Use
+     *        sol_oic_server_request_get_reader() to get the request data
```
glima: All of them deal with the reader? What about the writer?

src/lib/comms/sol-oic-server.c
```diff
@@ -87,18 +99,20 @@ static void sol_oic_server_unref(void);
...
-    const struct sol_oic_map_reader *input, struct sol_oic_map_writer *output)
+static int
+_sol_oic_server_d(void *data, struct sol_oic_request *request)
 {
     SOL_BUFFER_DECLARE_STATIC(dev_id, 37);
+    struct sol_oic_response *response;
     bool ret;
     int r;
 
+    response = sol_oic_server_response_new(request);
```
glima: Error check?

src/lib/comms/include/sol-oic-server.h
glima: It looks right, after a second read.

src/lib/comms/sol-oic-server.c
```diff
@@ -111,22 +125,28 @@ _sol_oic_server_d(const struct sol_network_link_addr *cliaddr, const void *data,
...
-_sol_oic_server_p(const struct sol_network_link_addr *cliaddr, const void *data,
-    const struct sol_oic_map_reader *input, struct sol_oic_map_writer *output)
+static int
+_sol_oic_server_p(void *data, struct sol_oic_request *request)
 {
+    struct sol_oic_response *response;
     const char *os_version;
     bool ret;
 
+    response = sol_oic_server_response_new(request);
```
glima: Ditto.

src/lib/comms/sol-oic-server.c
otaviobp: my bad. I'll recheck that

src/lib/comms/sol-oic-server.c
```diff
@@ -687,6 +688,15 @@ sol_oic_server_response_new(struct sol_oic_request *request)
...
 
+    return response;
+}
+
+SOL_API struct sol_oic_response *
+sol_oic_server_response_new(struct sol_oic_request *request)
+{
+    struct sol_oic_response *response;
+
+    response = create_response();
```
glima: Null check

src/lib/comms/include/sol-oic-client.h
```diff
@@ -328,59 +320,52 @@ bool sol_oic_client_get_server_info_by_addr(struct sol_oic_client *client,
...
+ * @return @c 0 on success or a negative number on errors.
  */
-bool sol_oic_client_resource_request(struct sol_oic_client *client, struct sol_oic_resource *res,
-    sol_coap_method_t method,
-    bool (*fill_repr_map)(void *data, struct sol_oic_map_writer *repr_map),
-    void *fill_repr_map_data,
-    void (*callback)(sol_coap_responsecode_t response_code, struct sol_oic_client *cli, const struct sol_network_link_addr *addr,
-    const struct sol_oic_map_reader *repr_vec, void *data),
-    const void *callback_data);
+int sol_oic_client_request(struct sol_oic_client *client, struct sol_oic_request *request, void (*callback)(sol_coap_responsecode_t response_code, struct sol_oic_client *cli, const struct sol_network_link_addr *addr, const struct sol_oic_map_reader *repr_vec, void *data), const void *callback_data);
```
glima: The line breaks were good :(

glima: You don't seem to cite that the function will free the request as well

src/lib/comms/sol-oic-common.h
```diff
@@ -0,0 +1,34 @@
...
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+#pragma once
+
+#include "sol-coap.h"
+
+struct sol_oic_request {
+    bool isServerRequest;
```
glima: CamelCase?

src/lib/comms/sol-oic-client.c
```diff
@@ -927,154 +942,188 @@ _one_shot_resource_request_cb(struct sol_coap_server *server,
...
-        SOL_WRN("Could not create CoAP packet");
-        goto out_no_req;
+    err = sol_oic_packet_cbor_close(request->base.pkt, &request->writer);
+    SOL_INT_CHECK_GOTO(err, != CborNoError, cbor_error);
+
+    server = _best_server_for_resource(client, request->res, &addr);
+    req = request->base.pkt;
+    request->base.pkt = NULL;
+    if (!sol_coap_send_packet_with_reply(server, req, &addr, request->cb,
+        ctx) == 0) {
```
glima: Ouch, why not store the func's return and pass it to the error label?

src/lib/comms/sol-oic-client.c
```diff
@@ -927,154 +942,188 @@ _one_shot_resource_request_cb(struct sol_coap_server *server,
...
+    return -ECOMM;
+}
 
-        res->observe.token = ctx->token;
-        sol_coap_add_option(req, SOL_COAP_OPTION_OBSERVE, &reg, sizeof(reg));
-        cb = _resource_request_cb;
-    } else {
-        cb = _one_shot_resource_request_cb;
+static struct sol_oic_request *
+request_new(sol_coap_method_t method, sol_coap_msgtype_t type, struct sol_oic_resource *res, bool isObserve)
```
glima: moar CamelCase

src/samples/coap/iotivity-test-client.c
```diff
@@ -107,24 +107,24 @@ found_resource_print(struct sol_oic_client *cli, struct sol_oic_resource *res, v
     }
```
glima: Actually the sane rationale would be to use printf directly for *all* samples.

src/samples/coap/iotivity-test-client.c
```diff
@@ -588,29 +588,29 @@ found_resource(struct sol_oic_client *cli, struct sol_oic_resource *res, void *d
...
+    SOL_WRN("11 :  Confirmable DELETE Requests");
+    SOL_WRN("12 :  Confirmable OBSERVE Requests");
+    SOL_WRN("13 :  Unsupported");
+    SOL_WRN("14 :  Unsupported");
+    SOL_WRN("15 :  Unsupported");
+    SOL_WRN("16 :  Unsupported");
+    SOL_WRN("17 :  Unsupported");
+    SOL_WRN("18 :  Unsupported");
+    SOL_WRN("19 :  Discover Platforms");
+    SOL_WRN("20 :  Discover Devices");
```
glima: Ditto

src/samples/coap/oic-client.c
```diff
@@ -27,7 +27,7 @@ static void
 got_get_response(sol_coap_responsecode_t response_code, struct sol_oic_client *cli, const struct sol_network_link_addr *srv_addr, const struct sol_oic_map_reader *map_reader, void *data)
 {
     struct sol_oic_repr_field field;
-    enum sol_oic_map_loop_reason end_reason;
+    enum sol_oic_map_loop_status end_reason;
```
glima: Hey, we got other ``_reason``s! We got to stick to one

glima:
Jeez, so much to read here. Overall I like it, but for sure there might be more little mistakes

data/scripts/sol-oic-gen.py
```diff
@@ -1370,39 +1370,58 @@ def master_c_as_string(generated, oic_gen_c, oic_gen_h):
...
+    struct sol_oic_map_writer *output;
+    struct sol_oic_response *response;
 
     if (!resource->funcs->to_repr_vec)
-        return SOL_COAP_RSPCODE_NOT_IMPLEMENTED;
+        return sol_oic_server_send_response(request, NULL,
+            SOL_COAP_RSPCODE_NOT_IMPLEMENTED);
 
+    response = sol_oic_server_response_new(request);
+    SOL_NULL_CHECK_GOTO(response, error);
```
ceolin: Create a new label to avoid call `sol_oic_server_response_free` with NULL pointer in this case.

src/lib/comms/sol-oic-server.c
```diff
@@ -613,54 +638,183 @@ sol_oic_server_unref(void)
...
+}
+
+SOL_API int
+sol_oic_server_send_response(struct sol_oic_request *request, struct sol_oic_response *response, sol_coap_responsecode_t code)
+{
+    struct sol_coap_packet *pkt;
+    int r = -EINVAL;
 
-    return sol_coap_send_packet(server, response, cliaddr);
+    OIC_SERVER_CHECK(-ENOTCONN);
```
ceolin: This is not right. This function is responsible to release the response memory, in this case the response memory will leak

src/lib/comms/sol-oic-server.c
```diff
@@ -613,54 +638,183 @@ sol_oic_server_unref(void)
...
+}
+
+SOL_API void
+sol_oic_server_response_free(struct sol_oic_response *response)
+{
+    int32_t i;
+
+    if (!response)
+        return;
+
```
ceolin: Is it expecting recceive NULL  ? otherwise you should use `SOL_NULL_CHECK`

src/lib/comms/sol-oic-server.c
```diff
@@ -613,54 +638,183 @@ sol_oic_server_unref(void)
     sol_oic_server_shutdown_internal();
 }
 
+static void
+request_free(struct sol_oic_request *request)
+{
+    int32_t i;
+
+    if (!request)
+        return;
```
ceolin: Is it expecting recceive NULL ? otherwise you should use `SOL_NULL_CHECK`

src/lib/comms/sol-oic-server.c
```diff
@@ -613,54 +638,183 @@ sol_oic_server_unref(void)
...
+
+SOL_API struct sol_oic_response *
+sol_oic_server_response_new(struct sol_oic_request *request)
+{
+    struct sol_oic_response *response;
+
+    response = calloc(1, sizeof(struct sol_oic_response));
+    SOL_NULL_CHECK(response, NULL);
+
+    if (sol_ptr_vector_append(&oic_server.responses, response) < 0) {
```
ceolin: `SOL_INT_CHECK_GOTO` to `error` will be better to my mind

src/lib/comms/sol-oic-server.c
```diff
@@ -1014,57 +1024,71 @@ sol_oic_server_del_resource(struct sol_oic_server_resource *resource)
...
+
+error:
+    free(notification);
+    return NULL;
+}
 
-    SOL_NULL_CHECK(resource, false);
-    SOL_NULL_CHECK(fill_repr_map, false);
+SOL_API int
+sol_oic_server_send_notification_to_observers(struct sol_oic_response *notification)
```
ceolin: This function is probably wrong, please check the comments bellow

src/lib/comms/sol-oic-server.c
```diff
@@ -1014,57 +1024,71 @@ sol_oic_server_del_resource(struct sol_oic_server_resource *resource)
...
+{
+    int r = 0;
 
-    OIC_SERVER_CHECK(false);
+    SOL_NULL_CHECK(notification, -EINVAL);
+    OIC_SERVER_CHECK(-ENOTCONN);
```
ceolin: in this case it will return a negative value and will not release `notification`

src/lib/comms/sol-oic-server.c
```diff
@@ -1014,57 +1024,71 @@ sol_oic_server_del_resource(struct sol_oic_server_resource *resource)
...
 
-    sent_server = send_notification_to_server(resource, oic_server.server, fill_repr_map, data);
+    if (!notification->resource) {
+        SOL_WRN("Response is not a notification response.");
+        return -EINVAL;
```
ceolin: same here

src/lib/comms/sol-oic-server.c
```diff
@@ -1014,57 +1024,71 @@ sol_oic_server_del_resource(struct sol_oic_server_resource *resource)
...
+    }
+    r = send_notification_to_server(notification, oic_server.server);
+    SOL_INT_CHECK_GOTO(r, < 0, end);
```
ceolin: here will return a negative value but will release `notification` memory

src/lib/comms/sol-oic-server.c
```diff
@@ -1014,57 +1024,71 @@ sol_oic_server_del_resource(struct sol_oic_server_resource *resource)
...
     if (oic_server.dtls_server)
-        sent_dtls_server = send_notification_to_server(resource, oic_server.dtls_server, fill_repr_map, data);
+        r = send_notification_to_server(notification, oic_server.dtls_server);
```
ceolin: here it can return a non negative value and will release `notification`

src/lib/comms/sol-oic-client.c
```diff
@@ -56,6 +57,12 @@
 #define OIC_RESOURCE_CHECK_API(ptr, ...)
 #endif
 
+#define CHECK_REQUEST(request, ...) \
```
ceolin: several public functions are calling this macro directly without check if `request` is `NULL`. You can add this check in this macro to avoid call `SOL_NULL_CHECK` before.

If you not want to do this, review where this macro is being used and check whether the null check  is necessary.

src/samples/coap/iotivity-test-client.c
ceolin: +1 for using `printf`. Use `SOL_WRN` for output makes no sense, it should be used to log problems.

src/lib/comms/sol-oic-server.c
```diff
@@ -39,6 +39,12 @@
 #include "sol-oic-server.h"
 #include "sol-oic-security.h"
 
+#define STR(s) #s
+#define TO_STRING(s) STR(s)
+
+#define OIC_DATA_MODEL_VERSION "Unknown"
+#define OIC_SPEC_VERSION ""
```
ceolin: Why those options can not be configured in the build as the others ?

At least, check if it is alread defined before define.
```
#ifndef OIC_DATA_MODEL_VERSION
#define OIC_DATA_MODEL_VERSION
#endif
```

ceolin:
You can take the oportunity (commit that fix API name convention) to remove the `typedefs` and also remove the usage of camelcase in this module.

src/lib/comms/sol-oic-server.c
otaviobp: Yes. it is expecting to receive NULL, that's why I am not using SOL_NULL_CHECK.

src/lib/comms/sol-oic-server.c
otaviobp: same here

data/scripts/sol-oic-gen.py
otaviobp: I prefer to call it with NULL, sol_oic_server_response_free works fine with response == NULL.
I am using same approach of free function here. It doesn't fail or create any issue if free(NULL) is called.

src/lib/comms/sol-oic-server.c
otaviobp: notification memory should always be released here. Fixing the mistakes

src/lib/comms/sol-oic-server.c
otaviobp: Those are information related to the oic implementation we are using. There is no need to let user change them. I'll just fix OIC_SPEC_VERSION to have the current OIC spec implemented.

otaviobp:
merged with proposed changes

PR #1964, opened 2016-05-03
Node.js bindings: Throw a proper error object in single-file test

gabrielschulhof:


gabrielschulhof:
@zqzhang, @wanghongjuan this should fix the ```NaN``` response when a test is torn down. The invasive tests need ```sudo``` to run, so if you don't have that in your environment, the test will fail.

Can you please confirm that this produces an informative error message for you?

wanghongjuan:
Thanks a lot. The issue doesn't reproduced when using the modified code.The Ostro image initial role is root. Is it viable that run the invasive tests by using root role without using `sudo` command?

gabrielschulhof:
@wanghongjuan I was thinking about that. I need to learn how to detect whether a process is running as root, and prepend ```sudo``` contingent upon that.

PR #1965, opened 2016-05-03
Bug fixes

ipuustin:
I took a look at Soletta comms and parser code and used also static analysis to find potential problem spots. I fixed some minor issues as I went. The commits are not really tested, but mostly highlight issues and provide some potential fixes for them.

### DIFF
--- a/src/lib/comms/sol-lwm2m.c
+++ b/src/lib/comms/sol-lwm2m.c
@@ -2243,6 +2243,7 @@ sol_lwm2m_resource_clear(struct sol_lwm2m_resource *resource)
     LWM2M_RESOURCE_CHECK_API(resource);
 
     free(resource->data);
+    resource->data = NULL;
 }
 
 static int
--- a/src/lib/parsers/sol-json.c
+++ b/src/lib/parsers/sol-json.c
@@ -1092,6 +1092,10 @@ sol_json_load_memdesc(const struct sol_json_token *token, const struct sol_memde
             len++;
         }
 
+        /* never allocate zero bytes */
+        if (len == 0)
+            return -EINVAL;
+
         done = alloca(len * sizeof(bool));
         memset(done, 0, len * sizeof(bool));
 
--- a/src/modules/flow/regexp/regexp.c
+++ b/src/modules/flow/regexp/regexp.c
@@ -281,7 +281,7 @@ get_unescaped_regexp_replacement_str(struct string_regexp_replace_data *mdata,
     int *match_vector,
     size_t match_cnt)
 {
-    int r;
+    int r = -EINVAL;
     struct sol_str_slice slice;
     char *ptr = mdata->to_regexp;
 


src/lib/comms/sol-lwm2m.c
```diff
@@ -2822,19 +2822,25 @@ handle_read(struct sol_lwm2m_client *client,
...
         &format, sizeof(format));
     SOL_INT_CHECK_GOTO(r, < 0, err_exit);
 
     r = set_packet_payload(resp, (const uint8_t *)buf.data, buf.used);
     SOL_INT_CHECK_GOTO(r, < 0, err_exit);
 
+    SOL_VECTOR_FOREACH_IDX (&resources, res, i) {
+        r = setup_tlv(res, &buf);
+        SOL_INT_CHECK_GOTO(r, < 0, err_exit);
+    }
```
barbieri: I'm not sure we can change the order here, @cabelitos or someone needs to check the rest of the code, but usually setting coap payload depends on all headers to be set as it will follow from that position (encodes to the wire). Then if setup_tlv() does headers, it maybe of problems. Another possibility of problem is that `set_packet_payload()` uses buffer, that is set by `setup_tlv()`.

src/lib/comms/sol-lwm2m.c
```diff
@@ -2822,19 +2822,25 @@ handle_read(struct sol_lwm2m_client *client,
...
+
+    /* Go through the list another time to prevent double free in case of
+     * an error in setup_tlv (which triggers free for all resources in the
+     * error path processing). */
+    SOL_VECTOR_FOREACH_IDX (&resources, res, i) {
```
barbieri: but error in `setup_tlv()` is `goto err_exit`, it would not walk the list once more. I don't get what you mean here :-(

src/lib/parsers/sol-json.c
```diff
@@ -1092,6 +1092,10 @@ sol_json_load_memdesc(const struct sol_json_token *token, const struct sol_memde
             len++;
         }
 
+        /* never allocate zero bytes */
+        if (len == 0)
+            return -EINVAL;
```
barbieri: the check is good, not sure about len == 0 being `-EINVAL`. @cabelitos ?

src/lib/comms/sol-lwm2m.c
ipuustin: Well, the problem is that `goto err_exit` leads to another freeing of the data from all items in the vector. If it was already freed half way, it would lead to double free for the already freed items. Anyway, I think a simpler fix is just setting the freed pointer to `NULL`, because after that the error path can call `free()` as many times it likes. Will update the patch now...

src/lib/parsers/sol-json.c
ceolin: i think it's ok. `len == 0` will happen due a not correct `desc`, that is a user input. For those cases we are always returning `-EINVAL`

src/lib/comms/sol-lwm2m.c
```diff
@@ -2992,6 +2992,7 @@ handle_resource(struct sol_coap_server *server,
     r = extract_path(client, req, path, &path_size);
     header_code = SOL_COAP_RSPCODE_BAD_REQUEST;
     SOL_INT_CHECK_GOTO(r, < 0, exit);
+    SOL_INT_CHECK_GOTO(path_size, == 0, exit);
```
ceolin: if `path_size == 0` is not valid shouldn't the function `extract_path` return an error when this happen ?
Otherwise one will always need to do two checks, wheter return `< 0` or `path_size == 0`

src/lib/comms/sol-lwm2m.c
```diff
@@ -2243,6 +2243,7 @@ sol_lwm2m_resource_clear(struct sol_lwm2m_resource *resource)
     LWM2M_RESOURCE_CHECK_API(resource);
 
     free(resource->data);
+    resource->data = NULL;
 }
```
ceolin: :+1:

src/lib/comms/sol-lwm2m.c
ipuustin: I took another look and it seems that you are right, or at least I could not anymore reproduce the thought process I had when I was originally looking at the code. :-) This line appears to check that there is at least one item to process, which in turn meas that `path_size` can't be zero and `r` be non-negative at the same time:

https://github.com/ipuustin/soletta/blob/static-analysis-fixes/src/lib/comms/sol-lwm2m.c#L2261

I'll remove this patch from the PR...

bdilly:
:+1:

merged, thanks

PR #1966, opened 2016-05-03
oic: Update oic-data-models with new master updates

otaviobp:
And fix script to support them.
Some data models are still not working because they include some json files that don't exist in the repository. There is nothing we can do about it.


otaviobp:
Just a note. There is nothing we can do about that in soletta code. I reported the issue.

bdilly:
:+1: 

go for it

PR #1969, opened 2016-05-03
flow: Update 7-seg node to use board labels

mbelluzzo:


bdilly:
@mbelluzzo don't you think it's required to have a "raw" option here as well?

mbelluzzo:
Yep, I'm already adding that. Will push the fix soon.

mbelluzzo:
Updated!

bdilly:
:+1:

Btw, using a single raw options is a good idea. No reason to have granularity per pin.

I'll adopt the same approach in the ultrasonic sensor.

mbelluzzo:
Merged, and btw build was already b0rked :-P

PR #1970, opened 2016-05-03
Ultrasonic sensor

bdilly:
It doesn't work very well on Linux thanks to imprecision when receiving callbacks from GPIO, pulse widths around 200 microseconds.

It would be better to have it as a Linux driver.

Anyway, it's enough to play a bit and should work better on small oses.


src/modules/flow/hc-sr04/hc-sr04.json
```diff
@@ -0,0 +1,78 @@
...
+          {
+            "data_type": "boolean",
+            "default": false,
+            "description": "Change 'echo' meaning to be the system parameters needed to address the desired pin. Use it if you know what you are doing.",
+            "name": "echo_raw"
+          },
+          {
+            "data_type": "int",
+            "default": 30,
+            "description": "Polling time on 'echo' pin.",
```
glima: In miliseconds

src/modules/flow/hc-sr04/hc-sr04.c
```diff
@@ -0,0 +1,217 @@
...
+}
+
+static void
+hc_sr04_close(struct sol_flow_node *node, void *data)
+{
+    struct hc_sr04_data *mdata = data;
+
+    sol_gpio_close(mdata->trig_gpio);
+    sol_gpio_close(mdata->echo_gpio);
+    sol_timeout_del(mdata->timer);
```
glima: No checks for ``NULL`` timer?

src/modules/flow/hc-sr04/hc-sr04.json
```diff
@@ -0,0 +1,78 @@
...
+            "name": "echo_poll_timeout"
+          },
+          {
+            "data_type": "int",
+            "default": 400,
+            "description": "Offset between pulse on trigger end and pulse on echo starts, on microseconds. It's only required if the echo pulse is too short that when the first interruption is handled pulse ended already.",
```
glima: "Offset between pulse on trigger end and pulse on echo starts, on microseconds. It's only required if the echo pulse is too short that when the first interruption is handled pulse ended already." I'd rephrase as "An additional time offset between the pulse on trigger and the echo response events, on microseconds. It's only required if the echo response is so fast that turning off the trigger happens after that. It may happen on non real time systems, like Linux" or something like that.

bdilly:
Thanks @glima, I've changed points you commented on.

mbelluzzo:
+2, just fix checkstyle complaining of some spaces

PR #1973, opened 2016-05-04
build: Respect the CC variable from the environment

vcgomes:
This patch is necessary for using clang scan-build with Soletta, as it
works by forcing the compiler pointed by the CC variable to be one its
analyzer.


barbieri:
-1

this is troublesome, people will do mistakes. The proper solution is to define TARGETCC or HOSTCC

vcgomes:
I don't think that I understand. Anyway, see if this version is better.

btw, my idea with this is at least to have what is needed to run `scan-build` in Soletta, I don't care specially if this is not applied.

barbieri:
this one is better

barbieri:
+1

vcgomes:
Pushed with a minor fix in the commit message. Thanks

PR #1974, opened 2016-05-04
dtls: Fix dtls socket wrapper callbacks and parameters

otaviobp:
DTLS socket was not working after socket restructuring because it was
using socket_ip callbacks instead of using wrapped functions.


ceolin:
lgtm :+1:

otaviobp:
merged

PR #1975, opened 2016-05-04
oic: Use sol-cert API to load pre shared keys

otaviobp:
Remove dependencies from sol-util-file in sol-security and use proposed API from sol-certificate.
Sol-certificate only have implementation for linux, using files to store/load certificates, but implementations for OS without filesystem support are possible.

### DIFF
--- a/src/lib/common/include/sol-certificate.h
+++ b/src/lib/common/include/sol-certificate.h
@@ -18,6 +18,8 @@
 
 #pragma once
 
+#include "sol-buffer.h"
+
 #ifdef __cplusplus
 extern "C" {
 #endif
@@ -41,6 +43,15 @@ extern "C" {
  * This object is the abstraction of certificate.
  */
 struct sol_cert;
+
+/**
+ * @brief Create a certificate located in @a path.
+ *
+ * @param path Absolute path to the certificate file.
+ *
+ * @return sol_cert object on success, NULL otherwise
+ */
+struct sol_cert *sol_cert_new(const char *path);
 
 /**
  * @brief Load a certificate from a file
@@ -71,6 +82,36 @@ void sol_cert_unref(struct sol_cert *cert);
  * @return sol_cert object on success, NULL otherwise
  */
 const char *sol_cert_get_filename(const struct sol_cert *cert);
+
+/**
+ * @brief Read data from @a cert.
+ *
+ * @param cert Certificate object.
+ * @param buffer The buffer to write the read data.
+ *
+ * @return 0 on success or a negative error number on errors.
+ */
+int sol_cert_read_data(const struct sol_cert *cert, struct sol_buffer *buffer);
+
+/**
+ * @brief Write data to @a cert.
+ *
+ * @param cert Certificate object.
+ * @param buffer The buffer to be written to certificate.
+ *
+ * @return 0 on success or a negative error code on errors.
+ */
+int sol_cert_write_data(const struct sol_cert *cert, struct sol_buffer *buffer);
+
+/**
+ * @brief Get the size in bytes occupied by the certificate.
+ *
+ * @param cert The certificate object.
+ *
+ * @return The size in bytes occupied by this certificate or a negative error
+ *         code on errors.
+ */
+ssize_t sol_cert_size(const struct sol_cert *cert);
 
 /**
  * @}
--- a/src/lib/common/sol-certificate-impl-linux.c
+++ b/src/lib/common/sol-certificate-impl-linux.c
@@ -22,6 +22,7 @@
 #define SOL_LOG_DOMAIN &_sol_certificate_log_domain
 #include "sol-log-internal.h"
 #include "sol-util-internal.h"
+#include "sol-util-file.h"
 #include "sol-vector.h"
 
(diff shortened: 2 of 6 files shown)


src/lib/common/include/sol-certificate.h
```diff
@@ -73,6 +84,36 @@ void sol_cert_unref(struct sol_cert *cert);
...
+
+/**
+ * @brief Get the size in bytes occupied by the certificate.
+ *
+ * @param cert The certificate object.
+ *
+ * @return The size in bytes occupied by this certificate or a negative error
+ *         code on errors.
+ */
+ssize_t sol_cert_size(const struct sol_cert *cert);
```
glima: ``get_size()`` aligns better with our naming: https://github.com/solettaproject/soletta/wiki/API-design-conventions

src/shared/include/sol-util-file.h
```diff
@@ -78,9 +78,9 @@ enum sol_util_iterate_dir_reason {
 int sol_util_write_file(const char *path, const char *fmt, ...) SOL_ATTR_PRINTF(2, 3);
 
 /**
- * @brief Write the formatted string in the file pointed by @c path.
+ * @brief Write the formatted string in the file pointed by @a path.
  *
- * It is equivalent to @c sol_util_write_file except it receives @c
+ * It is equivalent to @a sol_util_write_file except it receives @a
```
glima: I'd remove this ``@a`` and add ``()`` at the end of the function

src/lib/common/include/sol-certificate.h
```diff
@@ -73,6 +84,36 @@ void sol_cert_unref(struct sol_cert *cert);
...
 
 /**
+ * @brief Read data from @a cert.
+ *
+ * @param cert Certificate object.
+ * @param buffer The buffer to write the read data.
+ *
+ * @return 0 on success or a negative error number on errors.
+ */
+int sol_cert_read_data(const struct sol_cert *cert, struct sol_buffer *buffer);
```
barbieri: no, we're working with blobs, see the api introduced by @ceolin. actually he offers this one already, just missing the "save/persist".

src/lib/common/include/sol-certificate.h
barbieri: this API shouldn't exist.

src/lib/common/sol-certificate-impl-linux.c
```diff
@@ -92,19 +92,20 @@ sol_cert_load_from_file(const char *filename)
...
-    }
-
     path = find_cert(filename, search_paths);
     if (path == NULL) {
         SOL_WRN("Certificate not found: %s", filename);
         return NULL;
     }
 
+    SOL_PTR_VECTOR_FOREACH_IDX (&storage, cert, idx) {
+        if (streq(cert->filename, path)) {
```
barbieri: I don't get this order change, the commit doesn't clarify much either. The idea, AFAIU, is to only load once, so first you check in the storage, then if not found, you load it.

src/lib/comms/sol-oic-security.c
```diff
@@ -266,11 +269,26 @@ creds_init(const void *data)
     creds->security = data;
     sol_vector_init(&creds->items, sizeof(struct cred_item));
 
-    file_data = sol_util_load_file_string(file_name, &length);
-    if (!file_data)
+    cert = sol_cert_load_from_file(file_name);
```
barbieri: this is not right as it implies a file. This logic should move away from sol-oic-security into sol-cert. For instance "sol_cert_new_for_XXX()", like "XXX" being  "system", "app" or "user", depends on the scope of it, that would result saving in "/etc/ssl", "~/.config/APPNAME" or "~/.ssl" (or something along these lines, need to check the XDG convention for applications and the openssl for system and user.

Then, internally, these will create a path and load or save it. In Small OS they would be the same function, since there are no separation.

src/lib/comms/sol-oic-security.c
```diff
@@ -266,11 +269,26 @@ creds_init(const void *data)
...
+
+    file_data = malloc((size_t)cert_size);
+    SOL_NULL_CHECK_GOTO(file_data, error);
+
+    sol_buffer_init_flags(&buf, file_data, (size_t)cert_size,
+        SOL_BUFFER_FLAGS_CLEAR_MEMORY | SOL_BUFFER_FLAGS_NO_NUL_BYTE |
+        SOL_BUFFER_FLAGS_FIXED_CAPACITY);
+
+    r = sol_cert_read_data(cert, &buf);
+    SOL_INT_CHECK_GOTO(r, < 0, error_buf);
```
barbieri: this is not the best approach, sync with @ceolin as he did a nicer implementation.

src/lib/comms/sol-oic-security.c
```diff
@@ -266,11 +269,26 @@ creds_init(const void *data)
...
+
+    sol_json_scanner_init(&scanner, buf.data, buf.used);
```
barbieri: this is also not right. A Certificate as a specific meaning, we can't store a certificate as a JSON. Ideally this needs to be segmented into proper certificate formats, like p12/pem/crt.

PR #1978, opened 2016-05-04
modules/hc-sr04: expose a single 'raw' option for both pins

bdilly:
No use cases for some pins on raw mode and other
on label mode.

So let's make it simpler to users.


mbelluzzo:
LGTM

PR #1980, opened 2016-05-05
IIO node name changes

glima:


bdilly:
Makes sense

src/modules/flow/iio/iio.json
```diff
@@ -201,7 +201,7 @@
         "close": "temperature_close",
         "open": "temperature_open"
       },
-      "name": "iio/temperature",
+      "name": "iio/temperature-sensor",
```
edersondisouza: Anything against thermometer?

edersondisouza:
looks ok

src/modules/flow/iio/iio.json
glima: Me gusta. Might as well change the others.

glima:
Merged

PR #1981, opened 2016-05-05
Ipm

edersondisouza:
Samples on soletta-small-oses-samples: https://github.com/solettaproject/soletta-small-oses-samples/pull/17
One thing I'm not very comfortable is that currently there's no queueing of messages received, so if another ISR happens before a message is dispatched, previous message is lost. I've tried an attempt of using Zephyr `sys_ring_buf`, but I've got weird bugs on ARC. If someone is curious about it, it's available at https://github.com/edersondisouza/soletta/tree/ipm-ring-buf.


src/lib/common/sol-mainloop.c
```diff
@@ -76,6 +76,10 @@ extern void sol_comms_shutdown(void);
 extern int sol_update_init(void);
 extern void sol_update_shutdown(void);
 #endif
+#ifdef USE_UPDATE
```
glima: Typo, I guess

src/lib/io/sol-ipm.c
```diff
@@ -0,0 +1,323 @@
...
+#include "sol-util.h"
+#include "sol-vector.h"
+
+SOL_LOG_INTERNAL_DECLARE_STATIC(_sol_ipm_log_domain, "ipm");
+
+#define SOL_IPM_ID_FREE_REMOTE 0
+
+/* Declare channels we are going to use.
+ * Channel 0 is Quark -> ARC
+ * Channel 1 is ARC -> Quark
```
glima: Quark or x86?

src/modules/flow/ipm/ipm.c
```diff
@@ -0,0 +1,599 @@
...
+#include "sol-flow/ipm.h"
+#include "sol-flow-internal.h"
+#include "sol-ipm.h"
+
+#include <sol-util.h>
+#include <errno.h>
+
+/* TODO would be lovely to have this available globally... create a new
+* blob from some variable - but duplicating it, not just wrapping it */
+#define CREATE_BLOB(_val) \
```
glima: In your usage you always create and update -- why not unify them?

src/lib/io/sol-ipm.c
```diff
@@ -0,0 +1,323 @@
...
+static struct device *ipm_sender, *ipm_receiver;
+static uint32_t ipm_max_id;
+
+static void
+free_ipm_blob(struct sol_blob *blob)
+{
+    struct sol_blob_ipm *ipm_blob = (struct sol_blob_ipm *)blob;
+    int r;
+
+    if (blob->type != &SOL_BLOB_TYPE_IPM) {
```
vcgomes: This condition is impossible, no? I mean, the only way `free_ipm_blob()` is called is when a `sol_blob` of type IPM is free'd, right?

glima:
Make doxygen is bitching too

src/lib/common/sol-mainloop.c
edersondisouza: Nope, really. It was wrong before ;)
Maybe I should do this on a separated commit.

src/lib/io/sol-ipm.c
edersondisouza: Fair. It was a last minute change to x86

src/modules/flow/ipm/ipm.c
edersondisouza: Because `CREATE_BLOB` declares variables, I wanted to make it on the beginning. And `UPDATE_BLOB` must happen after we got packet value.

src/lib/io/sol-ipm.c
edersondisouza: True that.

edersondisouza:
I'll look into make doxygen. I always forget. Lucky we have CI =D

src/lib/common/sol-mainloop.c
barbieri: I'm trying to move out of ifdef, then please do the ifdef to declare as extern OR a static inline that is empty. In the call code, no ifdefs are required.

src/lib/io/Kconfig
```diff
@@ -81,3 +81,12 @@ config USE_MEMMAP
         Users must provide a mapping - for instance, using a JSON
         file - that will be used to define where on memory
         entries will be saved. Examples of memory are NVRAM or EEPROM.
+
+config USE_IPM
+    bool "Inter-processor messaging"
+    depends on ZEPHYR
```
barbieri: @ibriano what's your take on this? `depends on ZEPHYR` or declare a feature that only zephyr defines?

ibriano: On Thu, 05 May 2016 14:37:16 -0700, Gustavo Sverzut Barbieri wrote:
> > @@ -81,3 +81,12 @@ config USE_MEMMAP
> >          Users must provide a mapping - for instance, using a JSON
> >          file - that will be used to define where on memory
> >          entries will be saved. Examples of memory are NVRAM or EEPROM.
> > +
> > +config USE_IPM
> > +    bool "Inter-processor messaging"
> > +    depends on ZEPHYR
> 
> @ibriano what's your take on this? `depends on ZEPHYR` or declare a feature that only zephyr defines?

It'd be nice to always use FEATURE_ for this kind of things.

src/lib/io/sol-ipm.c
```diff
@@ -0,0 +1,323 @@
...
+
+/* Declare channels we are going to use.
+ * Channel 0 is Quark -> ARC
+ * Channel 1 is ARC -> Quark
+ */
+#ifdef CONFIG_X86
+QUARK_SE_IPM_DEFINE(message_ipm_receiver, 1, QUARK_SE_IPM_INBOUND);
+QUARK_SE_IPM_DEFINE(message_ipm_sender, 0, QUARK_SE_IPM_OUTBOUND);
+#elif CONFIG_ARC
+#warning "Building for ARC"
```
barbieri: do not add warnings, will break -Werror

src/lib/io/sol-ipm.c
```diff
@@ -0,0 +1,323 @@
...
+QUARK_SE_IPM_DEFINE(message_ipm_receiver, 0, QUARK_SE_IPM_INBOUND);
+QUARK_SE_IPM_DEFINE(message_ipm_sender, 1, QUARK_SE_IPM_OUTBOUND);
+#endif
+
+/*XXX there MUST be a better way of doing so */
+#define AS_PTR(_ptr) *(uint64_t *)_ptr
```
barbieri: `uintptr_t`?

src/lib/io/sol-ipm.c
```diff
@@ -0,0 +1,323 @@
...
+    struct sol_blob_ipm *ipm_blob = (struct sol_blob_ipm *)blob;
+    int r;
+
+    if (blob->type != &SOL_BLOB_TYPE_IPM) {
+        SOL_WRN("Invalid blob type: expected SOL_BLOB_TYPE_IPM(%p), got: %p",
+            &SOL_BLOB_TYPE_IPM, blob->type);
+        return;
+    }
+
+    r = ipm_send(ipm_sender, 1, SOL_IPM_ID_FREE_REMOTE, &ipm_blob->remote,
```
barbieri: just send the blob, not the pointer to one... it's already a pointer! :-)

src/lib/io/sol-ipm.c
```diff
@@ -0,0 +1,323 @@
...
+        sizeof(ipm_blob->remote));
```
barbieri: sizeof(uintptr_t)

src/lib/io/sol-ipm.c
```diff
@@ -0,0 +1,323 @@
...
+    uint32_t id = (uint32_t)data;
+    struct sol_ipm_receiver_entry *receiver;
+
+    SOL_VECTOR_FOREACH_IDX (&ipm_receivers, receiver, idx) {
+        if (receiver->id == id) {
+            struct sol_blob_ipm *remote_blob;
+
+            if (!receiver->last_remote_blob) {
+                SOL_WRN("Remote blob for IPM id %" PRIu32 " disappered!", id);
+                return;
```
barbieri: what?

src/lib/io/sol-ipm.c
```diff
@@ -0,0 +1,323 @@
...
+            }
+
+            remote_blob = calloc(1, sizeof(struct sol_blob_ipm));
+            if (!remote_blob) {
+                SOL_WRN("Could not create blob to deliver IPM message");
+                receiver->last_remote_blob = NULL;
```
barbieri: you should send back it for free, like done in `free_ipm_blob()`

src/lib/io/sol-ipm.c
```diff
@@ -0,0 +1,323 @@
...
+                return;
+            }
+
+            sol_blob_setup((struct sol_blob *)remote_blob, &SOL_BLOB_TYPE_IPM,
```
barbieri: no cast, use `&remote_blob->base`

src/lib/io/sol-ipm.c
```diff
@@ -0,0 +1,323 @@
...
+
+            remote_blob->remote = receiver->last_remote_blob;
+            receiver->last_remote_blob = NULL;
+            receiver->receive_cb((void *)receiver->data, id,
+                (struct sol_blob *)remote_blob);
+            return;
+        }
+    }
+
+    SOL_WRN("Processed IPM id %" PRIu32 " but no receiver was found!", id);
```
barbieri: warning? or info? not sure :-/

barbieri: in any case, send it back for unref!

src/lib/io/sol-ipm.c
```diff
@@ -0,0 +1,323 @@
...
+        me.cb = ipm_reaper_process;
+        me.data = remote_blob;
+        sol_mainloop_event_post(&me);
+        return;
+    }
+
+    //TODO rethink this. What if we receive another message with same id
+    //before processing current one?
+    me.cb = ipm_receiver_process,
+    me.data = (void *)id; /* I could send `receiver` entry, and avoid yet
```
barbieri: answering "rethink this": create the wrapper blob before sending and set the id in it, you already extend the structure to contain the remote blob.

src/lib/io/sol-ipm.c
```diff
@@ -0,0 +1,323 @@
...
+                             another loop to find receiver. Although currently,
+                             Zephyr implementation does process events before
+                             timeouts, I'm not sure we can count on that
+                             `ipm_vectors` is not going to be changed in the
+                             meanwhile.*/
+
+    SOL_VECTOR_FOREACH_IDX (&ipm_receivers, receiver, idx) {
+        if (receiver->id == id) {
+            receiver->last_remote_blob = remote_blob;
+            int r = sol_mainloop_event_post(&me);
```
barbieri: I doubt this is passing check-style.py...

src/lib/io/sol-ipm.c
```diff
@@ -0,0 +1,323 @@
...
+            return;
+        }
+    }
+
+    /* Is it really OK/safe to print from isr? */
+    SOL_INF("Got message with id %" PRIu32 " but there's no receiver associated to it",
+        id);
```
barbieri: likely not, and you shouldn't do the loop or the print if use my suggestion

src/lib/io/sol-ipm.c
barbieri: also, handle errors by sending back the blob for unref

src/lib/io/sol-ipm.c
```diff
@@ -0,0 +1,323 @@
...
+                receiver->receive_cb = receive_cb;
+                return 0;
+            } else {
+                return sol_vector_del(&ipm_receivers, idx);
+            }
+        }
+    }
+
+    /* Not changing or deleting. Must have a receive_cb*/
+    SOL_NULL_CHECK(receive_cb, -EINVAL);
```
barbieri: ENOENT?

src/lib/io/sol-ipm.c
```diff
@@ -0,0 +1,323 @@
...
+    SOL_NULL_CHECK(message, -EINVAL);
+
+    sent = sol_vector_append(&ipm_sent_messages);
+    SOL_NULL_CHECK(sent, -ENOMEM);
+
+    r = ipm_send(ipm_sender, 1, id, &message, sizeof(message));
+    SOL_INT_CHECK_GOTO(r, > 0, fail);
+
+    sent->id = id;
+    sent->message_consumed_cb = message_consumed_cb;
```
barbieri: do we expect this to be done per-send? Or per id is enough?

src/lib/io/sol-ipm.c
```diff
@@ -0,0 +1,323 @@
...
+    sent->data = data;
+    sent->message = sol_blob_ref(message); /* Hold a reference to it so it doesn't die early*/
```
barbieri: my proposal as per above is here, instead of sending `message`, you create the wrapped blob here, it will contain the message and the id. If you can move `message_consumed_cb` outside, then this becomes a simple `sol_ptr_vector`.

src/lib/io/sol-ipm.c
```diff
@@ -0,0 +1,323 @@
...
+    return 0;
+
+fail:
+    sol_vector_del_last(&ipm_sent_messages);
+    return -r;
+}
+
+
+SOL_API uint32_t
+sol_ipm_max_id(void)
```
barbieri: get_max_id

src/lib/io/sol-ipm.c
```diff
@@ -0,0 +1,323 @@
...
+    if (max_data_size < sizeof(void *)) {
+        SOL_WRN("IPM max data size < sizeof(void *)");
+        return -EINVAL;
+    }
+
+    ipm_max_id = ipm_max_id_val_get(ipm_sender);
+
+    ipm_register_callback(ipm_receiver, ipm_receiver_cb, NULL);
+    r = ipm_set_enabled(ipm_receiver, 1);
+    if (r != 0) {
```
barbieri: I think you should only enable IPM if it's used, make the "lazy_init()" and shutdown pattern.

What do you think? Maybe it's not the case since one can disable the IPM, but I'd say enable by default and make its "non usage" cheap enough

src/modules/flow/ipm/ipm.c
barbieri: add these helpers to sol-blob.h, makes sense. But make a static inline function that does the work and a macro that does the sizeof() and calls it.

src/modules/flow/ipm/ipm.c
```diff
@@ -0,0 +1,599 @@
...
+    struct ipm_data *mdata = data;
+    bool in_value;
+    int r;
+
+    CREATE_BLOB(&in_value);
+
+    r = sol_flow_packet_get_boolean(packet, &in_value);
+    SOL_INT_CHECK(r, < 0, r);
+
+    UPDATE_BLOB(blob, &in_value);
```
barbieri: as @glima said, make the wrapper work on the final value. Then you declare `struct sol_blob *blob;` yourself and call: `blob = SOL_BLOB_NEW_DUP(in_value)`, with:

```c
static inline struct sol_blob *
sol_blob_new_dup(const void *mem, size_t size) {
    ...;
}
#define SOL_BLOB_NEW_DUP(x) sol_blob_new_dup((x), sizeof(x))
```

src/modules/flow/ipm/ipm.c
```diff
@@ -0,0 +1,599 @@
...
+    char *str;
+    int r;
+
+    r = sol_flow_packet_get_string(packet, &in_value);
+    SOL_INT_CHECK(r, < 0, r);
+
+    str = strdup(in_value);
+    SOL_NULL_CHECK(str, -ENOMEM);
+
+    blob = sol_blob_new(SOL_BLOB_TYPE_DEFAULT, NULL, str, strlen(str) + 1);
```
barbieri: also valid a `sol_blob_new_dup_str()`

src/lib/io/sol-ipm.c
edersondisouza: On 05-05-2016 18:40, Gustavo Sverzut Barbieri wrote:
> In src/lib/io/sol-ipm.c
> <https://github.com/solettaproject/soletta/pull/1981#discussion_r62260030>:
>
>> +#include "sol-util.h"
>> +#include "sol-vector.h"
>> +
>> +SOL_LOG_INTERNAL_DECLARE_STATIC(_sol_ipm_log_domain, "ipm");
>> +
>> +#define SOL_IPM_ID_FREE_REMOTE 0
>> +
>> +/* Declare channels we are going to use.
>> + * Channel 0 is Quark -> ARC
>> + * Channel 1 is ARC -> Quark
>> + */
>> +#ifdef CONFIG_X86
>> +QUARK_SE_IPM_DEFINE(message_ipm_receiver, 1, QUARK_SE_IPM_INBOUND);
>> +QUARK_SE_IPM_DEFINE(message_ipm_sender, 0, QUARK_SE_IPM_OUTBOUND);
>> +#elif CONFIG_ARC
>> +#warning "Building for ARC"

Ouch... this is debug stuff... =/

src/lib/io/sol-ipm.c
edersondisouza: On 05-05-2016 18:41, Gustavo Sverzut Barbieri wrote:
> In src/lib/io/sol-ipm.c
> <https://github.com/solettaproject/soletta/pull/1981#discussion_r62260174>:
>
>> +
>> +/* Declare channels we are going to use.
>> + * Channel 0 is Quark -> ARC
>> + * Channel 1 is ARC -> Quark
>> + */
>> +#ifdef CONFIG_X86
>> +QUARK_SE_IPM_DEFINE(message_ipm_receiver, 1, QUARK_SE_IPM_INBOUND);
>> +QUARK_SE_IPM_DEFINE(message_ipm_sender, 0, QUARK_SE_IPM_OUTBOUND);
>> +#elif CONFIG_ARC
>> +#warning "Building for ARC"
>> +QUARK_SE_IPM_DEFINE(message_ipm_receiver, 0, QUARK_SE_IPM_INBOUND);
>> +QUARK_SE_IPM_DEFINE(message_ipm_sender, 1, QUARK_SE_IPM_OUTBOUND);
>> +#endif
>> +
>> +/*XXX there MUST be a better way of doing so */
>> +#define AS_PTR(_ptr) *(uint64_t *)_ptr
>
> |uintptr_t|?

I swear I was trying to find such a type... tks!

src/lib/io/sol-ipm.c
edersondisouza: On 05-05-2016 18:53, Gustavo Sverzut Barbieri wrote:
> In src/lib/io/sol-ipm.c
> <https://github.com/solettaproject/soletta/pull/1981#discussion_r62261797>:
>
>> +    }
>> +
>> +    //TODO rethink this. What if we receive another message with same id
>> +    //before processing current one?
>> +    me.cb = ipm_receiver_process,
>> +    me.data = (void *)id; /* I could send `receiver` entry, and avoid yet
>> +                             another loop to find receiver. Although currently,
>> +                             Zephyr implementation does process events before
>> +                             timeouts, I'm not sure we can count on that
>> +                             `ipm_vectors` is not going to be changed in the
>> +                             meanwhile.*/
>> +
>> +    SOL_VECTOR_FOREACH_IDX (&ipm_receivers, receiver, idx) {
>> +        if (receiver->id == id) {
>> +            receiver->last_remote_blob = remote_blob;
>> +            int r = sol_mainloop_event_post(&me);
>
> I doubt this is passing check-style.py...

Well, it is... but again, debug leftover

src/lib/io/sol-ipm.c
edersondisouza: On 05-05-2016 18:53, Gustavo Sverzut Barbieri wrote:
> In src/lib/io/sol-ipm.c
> <https://github.com/solettaproject/soletta/pull/1981#discussion_r62261696>:
>
>> +    struct sol_ipm_receiver_entry *receiver;
>> +    struct mainloop_event me;
>> +
>> +    /* id == 0 messages. Those messages signal sender
>> +    * core that blob was consumed and may be freed */
>> +    if (id == SOL_IPM_ID_FREE_REMOTE) {
>> +        me.cb = ipm_reaper_process;
>> +        me.data = remote_blob;
>> +        sol_mainloop_event_post(&me);
>> +        return;
>> +    }
>> +
>> +    //TODO rethink this. What if we receive another message with same id
>> +    //before processing current one?
>> +    me.cb = ipm_receiver_process,
>> +    me.data = (void *)id; /* I could send `receiver` entry, and avoid yet
>
> answering "rethink this": create the wrapper blob before sending and set
> the id in it, you already extend the structure to contain the remote blob.

What do you mean? Adding an id field to `sol_blob_ipm` and wrapping sent 
blob in it?

src/lib/io/sol-ipm.c
edersondisouza: On 05-05-2016 18:57, Gustavo Sverzut Barbieri wrote:
> In src/lib/io/sol-ipm.c
> <https://github.com/solettaproject/soletta/pull/1981#discussion_r62262335>:
>
>> +{
>> +    int r;
>> +    struct sol_ipm_sent_message *sent;
>> +
>> +    SOL_INT_CHECK(id, > ipm_max_id, -EINVAL);
>> +    SOL_INT_CHECK(id, == 0, -EINVAL);
>> +    SOL_NULL_CHECK(message, -EINVAL);
>> +
>> +    sent = sol_vector_append(&ipm_sent_messages);
>> +    SOL_NULL_CHECK(sent, -ENOMEM);
>> +
>> +    r = ipm_send(ipm_sender, 1, id, &message, sizeof(message));
>> +    SOL_INT_CHECK_GOTO(r, > 0, fail);
>> +
>> +    sent->id = id;
>> +    sent->message_consumed_cb = message_consumed_cb;
>
> do we expect this to be done per-send? Or per id is enough?

If producer sends faster than consumer speed, doing per id would lead to 
leaks... don't you think?

edersondisouza: Oh, or do you mean something like:

```c
SOL_API sol_ipm_set_consumed_callback(uint8_t id, void (*message_consumed_cb)(void *cb_data, uint32_t id, struct sol_blob *message));
```

I thought of this API, but found it ugly in the beginning, but considering that flow always send `NULL` on callback stuff, it may be reasonable to be separated.

src/lib/io/sol-ipm.c
barbieri: yes, that way you keep things simpler. Also, at the local site (origin of send) you can create that blob with the original being "parent", so you create 2 ipm blob types: local and remote. Remote is the one you have now. Local is the new one, it contains an "id" and maybe the callback+data (if can't be per "id"), it uses the parent = sent blob, mem = sent->data, size = sent->size, this way when you come back, you unref this Local blob and it will unref the parent.

src/lib/io/sol-ipm.c
barbieri: indeed... and AFAIU one must "own the channel/id", right? Then this would even help. Just be sure you handle set(NULL) from the actual callback to address the "temporary owner" case: own/set -> send -> wait -> unset.

edersondisouza: Not sure if I understood the "temporary owner" stuff...

barbieri: **Persistent owner:** from `startup()`, set a callback for an id. Unset on `shutdown()`. During runtime, messages are sent when there is a need, always from the same "owner", like a flow node type. Trying to set another callback for the same id from another owner (node type), will failed `-EEXIST`.

**Temporary owner:** when a message must be sent and needs a callback, that callback is set (take ownership), the message is sent, from the callback it will unset (release ownership, thus temporary).

src/lib/io/sol-ipm.c
edersondisouza: Not sure if I got:
`ipm_blob->remote` has type `struct sol_blob *`. Let's say it's content is `0x40001234` - memory position of a `struct sol_blob`.
`ipm_send` takes a `void *` and a `size`, it will put what is pointed by, up to `size` bytes.
So, if send only `ipm_blob->remote`, it will look what is at position `0x40001234`, and get `size` (4) bytes there. But this is actually part of the struct itself.
That's why a send `&ipm_blob->remote`. This way, `0x40001234` will be on channel registers - and this value is then, on remote side, cast back to a pointer to `struct sol_blob`.

barbieri: ohhh... got it

src/lib/io/sol-ipm.c
edersondisouza: Ah, so do you mean we still have the possibility to have callbacks set per sent message? I was thinking on dropping this possibility. Do you think it's useful to have both? If so, all 'consumed' messages for some id would go to global handler, unless a specific handler is specified when sending a message?

barbieri: no, no callback per message, would do it outside:

```c
mycb(...) {
    // release ownership
    sol_ipm_set_consumed_callback(1, NULL, NULL);
}

myfun(...) {
   int r;

   // temporarily take ownership, if someone owned it, r == -EEXIST
   r = sol_ipm_set_consumed_callback(1, mycb, mydata);
   SOL_INT_CHECK(r, < 0, ...);
   
   r = sol_ipm_send(1, myblob);
   SOL_INT_CHECK(r, < 0, ...);
}

File: src/lib/io/sol-ipm.c (hunk 21 above)
Comment (edersondisouza) [inline src/lib/io/sol-ipm.c]: It can be done... not sure about how useful would it be... in fact, this enabling is quite lightweight: setting a bit on some register.

File: src/lib/io/sol-ipm.c (hunk 18 above)
Comment (edersondisouza) [inline src/lib/io/sol-ipm.c]: Ah, ok, got it.

File: src/lib/io/sol-ipm.c (hunk 21 above)
Comment (barbieri) [inline src/lib/io/sol-ipm.c]: but what about getting the ipm_receiver and sender devices?

Comment (edersondisouza) [inline src/lib/io/sol-ipm.c]: On 06-05-2016 14:39, Gustavo Sverzut Barbieri wrote:
> In src/lib/io/sol-ipm.c
> <https://github.com/solettaproject/soletta/pull/1981#discussion_r62363148>:
>
>> +    SOL_NULL_CHECK_MSG(ipm_receiver, -EINVAL, "Could not get IPM receiver channel");
>> +
>> +    ipm_sender = device_get_binding("message_ipm_sender");
>> +    SOL_NULL_CHECK_MSG(ipm_sender, -EINVAL, "Could not get IPM sender channel");
>> +
>> +    max_data_size = ipm_max_data_size_get(ipm_sender);
>> +    if (max_data_size < sizeof(void *)) {
>> +        SOL_WRN("IPM max data size < sizeof(void *)");
>> +        return -EINVAL;
>> +    }
>> +
>> +    ipm_max_id = ipm_max_id_val_get(ipm_sender);
>> +
>> +    ipm_register_callback(ipm_receiver, ipm_receiver_cb, NULL);
>> +    r = ipm_set_enabled(ipm_receiver, 1);
>> +    if (r != 0) {
>
> but what about getting the ipm_receiver and sender devices?

Hmmm... fair.

File: src/lib/io/sol-ipm.c [hunk 24]
```diff
@@ -0,0 +1,323 @@
...
+
+#define SOL_IPM_ID_FREE_REMOTE 0
+
+/* Declare channels we are going to use.
+ * Channel 0 is Quark -> ARC
+ * Channel 1 is ARC -> Quark
+ */
+#ifdef CONFIG_X86
+QUARK_SE_IPM_DEFINE(message_ipm_receiver, 1, QUARK_SE_IPM_INBOUND);
+QUARK_SE_IPM_DEFINE(message_ipm_sender, 0, QUARK_SE_IPM_OUTBOUND);
```
ibriano: Make the channels configurable, even if it's through Kconfig. We can't assume that something else in the OS will not have taken these already.

edersondisouza: On 06-05-2016 16:58, Iván Briano wrote:
> In src/lib/io/sol-ipm.c
> <https://github.com/solettaproject/soletta/pull/1981#discussion_r62382942>:
>
>> +#include "sol-mainloop-zephyr.h"
>> +#include "sol-types.h"
>> +#include "sol-util.h"
>> +#include "sol-vector.h"
>> +
>> +SOL_LOG_INTERNAL_DECLARE_STATIC(_sol_ipm_log_domain, "ipm");
>> +
>> +#define SOL_IPM_ID_FREE_REMOTE 0
>> +
>> +/* Declare channels we are going to use.
>> + * Channel 0 is Quark -> ARC
>> + * Channel 1 is ARC -> Quark
>> + */
>> +#ifdef CONFIG_X86
>> +QUARK_SE_IPM_DEFINE(message_ipm_receiver, 1, QUARK_SE_IPM_INBOUND);
>> +QUARK_SE_IPM_DEFINE(message_ipm_sender, 0, QUARK_SE_IPM_OUTBOUND);
>
> Make the channels configurable, even if it's through Kconfig. We can't
> assume that something else in the OS will not have taken these already.

Fair.

edersondisouza:
v2 on #1996

Issue #1971, opened 2016-05-04
Add support for proximity sensors into soletta iio node


lblim:
#### Task Description
Proximity sensor IIO node is not yet supported, can this be included for the upcoming Soletta release?

barbieri:
Coincidence or not, we just noticed that today. Bruno submitted a stop gap user space sensor at https://github.com/solettaproject/soletta/pull/1970

But we will work with ostro team to support IIO asap. So you have some specific sensor? Please share the part number.

lblim:
We are looking for SX9500, and APDS9930. The kernel driver for SX9500 is already available in Linux mainline, but the IIO driver for APDS9930 has not been upstreamed yet.

barbieri:
@yongli3 can you work with @lblim to introduce soletta node types to use that family? Also work in Ostro to backport such modules if possible.

yongli3:
@lblim please check the apds9960 and apds9300, are they similar with apds9930 ?

lblim:
Hi @yongli3 

APDS9960, APDS9300 and APDS9930 are all ALS sensor but APDS9960 and 9930 are combo sensors that has more features. A separate kernel driver is needed for all three since they all have different features and export different IIO entries.

And Yes, in_proximity_raw is used.

Here is the IIO entries for APDS9930
root@sofia-3gr:/sys/bus/iio/devices/iio:device4# cat in_illuminance_input
200

root@sofia-3gr:/sys/bus/iio/devices/iio:device4# cat in_intensity_ir_raw
408

root@sofia-3gr:/sys/bus/iio/devices/iio:device4# cat in_intensity_both_raw
5924

root@sofia-3gr:/sys/bus/iio/devices/iio:device4# cat in_proximity_raw
1023

root@sofia-3gr:/sys/bus/iio/devices/iio:device4# cat uevent
MAJOR=252
MINOR=4
DEVNAME=iio:device4
DEVTYPE=iio_device
OF_NAME=proxsensor
OF_FULLNAME=/xgold/noc/l2_noc/ahb_per@1/i2c_3/proxsensor
OF_COMPATIBLE_0=avago,apds9930
OF_COMPATIBLE_N=1

root@sofia-3gr:/sys/bus/iio/devices/iio:device4# ls
dev     in_illuminance_input   in_intensity_ir_raw  name   subsystem
events  in_intensity_both_raw  in_proximity_raw     power  uevent

yongli3:
[...]

PR #1986, opened 2016-05-06
flow:iio: Add Proximity sensor Category support into iio node

yongli3:

### DIFF
--- a/src/modules/flow/iio/iio.json
+++ b/src/modules/flow/iio/iio.json
@@ -803,6 +803,95 @@
         }
       ],
       "private_data_type": "light_data"
+     },
+     {
+      "category": "input/hw",
+      "description": "IIO based proximity input node. As any IIO device, it can use a buffer to get the readings. To use a buffer, define a size > 0 on buffer size. Readings are sent to the buffer via a trigger mechanism. Set iio_trigger_name to a valid iio trigger name in order to use buffer. If buffer is enabled but no iio_trigger_name was set, it will attempt to create a default trigger that will be activated when sending packets to TICK port.",
+      "in_ports": [
+        {
+          "data_type": "any",
+          "description": "Packets sent in here will trigger a proximity reading and produce packets on each of the the output ports. If buffer is enabled, current trigger must accept manual activation (default one, sysfs trigger, does).",
+          "name": "TICK",
+          "methods": {
+            "process": "proximity_tick"
+          }
+        }
+      ],
+      "methods": {
+        "close": "proximity_close",
+        "open": "proximity_open"
+      },
+      "name": "iio/proximity",
+      "options": {
+       "members": [
+         {
+           "data_type": "string",
+           "description": "IIO device identifier. It's a space separated list of commands. For commands, if it's an integer value, will be interpreted as IIO device id. If it's a string starting with '/', will be interpreted as absolute path of IIO device on sysfs. If it's on the form 'i2c/X-YYYY', will evaluate to an i2c device on sysfs, where X is the bus number and YYYY is the device number, eg, 7-0069, for device 0x69 on bus 7. If it's on the form 'create,i2c,<rel_path>,<devnumber>,<devname>', where rel_path is the path of bus relative to '/sys/devices', them it will attempt to create an IIO device on that i2c bus and use it.",
+           "name": "iio_device"
+         },
+         {
+           "data_type": "int",
+           "description": "IIO buffer size. If -1, buffering is disabled. If 0, will use default buffer size. If enabled (> 0), a trigger is necessary to perform readings. Set it using 'iio_trigger_name'; if no trigger name is set, it will attempt to create a default one, which is activated via TICK port.",
+           "default": 0,
+           "name": "buffer_size"
+         },
+         {
+           "data_type": "string",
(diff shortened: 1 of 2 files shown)


yongli3:
@lblim  @barbieri @edersondisouza please help to review, this PR is used for issue #1971

src/modules/flow/iio/iio.json
```diff
@@ -803,6 +803,95 @@
         }
       ],
       "private_data_type": "light_data"
+     },
+	 {
```
edersondisouza: This `{` looks mis-indented

src/modules/flow/iio/iio.json
```diff
@@ -803,6 +803,95 @@
...
+             "description": "Range of output packet. Usually, is the output range of used sensor. It'll be the 'min' and 'max' fields of drange packets sent on OUT port.",
+             "name": "out_range"
+         }
+       ],
+       "version": 1
+      },
+      "out_ports": [
+        {
+         "data_type": "float",
+         "description": "proximity data read in lux.",
```
edersondisouza: Proximity in lux?

src/modules/flow/iio/nodes.c
```diff
@@ -1429,4 +1429,156 @@ light_tick(struct sol_flow_node *node, void *data, uint16_t port, uint16_t conn_
...
+        .max = mdata->out_range.max,
+        .step = mdata->out_range.step
+    };
+    bool b;
+
+    b = sol_iio_read_channel_value(mdata->channel_val, &out.val);
+    if (!b) goto error;
+
+    sol_flow_send_drange_value_packet(node,
+        SOL_FLOW_NODE_TYPE_IIO_LIGHT__OUT__OUT, out.val);
```
edersondisouza: `SOL_FLOW_NODE_TYPE_IIO_LIGHT__OUT__OUT`? Probably you wanted proximity node port...

edersondisouza:
About the in_proximityX... how are they used? What does mean the data they read? A simple answer would be adding another option to node type, but I'd like to understand more about meaning of this data. Are they complementary? Do they measure different ranges?

yongli3:
Thanks @edersondisouza  for the review. About the in_proximity data reading, for common sensors such as APDS9960/APDS9930, users will use "cat in_proximity_raw" to get the data, this PR can support such  kind of sensors. But if there are more than one raw data, such as SX9500 in issue #1971, we need to get data from in_proximity2_raw. @lblim could you share some information about these in_proximityX? do we need to read all of them, or just in_proximity2_raw?

src/modules/flow/iio/iio.json

src/modules/flow/iio/iio.json
yongli3: The data is raw data

src/modules/flow/iio/nodes.c

yongli3:
@edersondisouza  @laykuanloon I modified this PR, it can support the default in_proximity_raw and in_proximity2_raw. 

If you want to support both these 4 channels, I think we can use another PR to implement it. @edersondisouza how do you think about it?

src/modules/flow/iio/nodes.c
```diff
@@ -1429,4 +1429,158 @@ light_tick(struct sol_flow_node *node, void *data, uint16_t port, uint16_t conn_
...
+    SOL_NULL_CHECK(mdata->device, false);
+
+#define ADD_CHANNEL(_val) \
+    if (!mdata->use_device_default_scale) \
+        channel_config.scale = mdata->scale; \
+    if (!mdata->use_device_default_offset) \
+        channel_config.offset = mdata->offset; \
+    mdata->channel_ ## _val = sol_iio_add_channel(mdata->device, "in_proximity", &channel_config); \
+    if (!mdata->channel_ ## _val) \
+        mdata->channel_ ## _val = sol_iio_add_channel(mdata->device, "in_proximity2", &channel_config); \
```
edersondisouza: You are adding two channels, but only saves one. So, you only read `in_proximity2` in fact, as it's the last one you saved. If a device happens to not have `in_proximity2`, you'll get NULL here and no readings after.
Nevertheless, I still would like to understand the differences between these channels. I suspect that we could model all channels on only one final information.

edersondisouza: Oh... disregard first comment, only now I've seen the `if (!mdata->channel_ ## _val)`... sorry.
But I'm kinda puzzled by the `in_proximity2`... AFAICU from sx9500 datasheet, there's no differentiation among channels...

yongli3: @edersondisouza As discussed in issue #1971, only channel 2 in sx9500 is needed. 
APDS9960/APDS9930 only have the in_proximity_raw, and the SX9500 has in_proximity0_raw to in_proximity3_raw, if we only use in_proximity2_raw, this PR can support all these three sensors. But if we want to add multi-channel support, we need a new implement

edersondisouza: Ok. I'll merge this PR, but we need to change nodes to address this multiple channel case.

edersondisouza:
Integrated. Tks!
@yongli3 I'll create a Github issue about multiple channels and assign it for you, ok?

yongli3:
No problem☺

PR #1972, opened 2016-05-04
Node.js bindings: Add sudo to invasive tests only if euid is not root

gabrielschulhof:
Signed-off-by: Gabriel Schulhof <gabriel.schulhof@intel.com>

PR #1987, opened 2016-05-06
Node.js bindings: Add sudo to invasive tests only if euid is not root

gabrielschulhof:

Replaces #1972 
Additionally fixes the situation where an async callback is being called twice.

### DIFF
--- a/bindings/nodejs/tests/invasive-tests/Main Loop Integration.js
+++ b/bindings/nodejs/tests/invasive-tests/Main Loop Integration.js
@@ -23,6 +23,7 @@ var soletta = require( require( "path" )
 var exec = require( "child_process" ).exec;
 var initialHostname = require( "os" ).hostname();
 var initialTimezone, destinationTimezone;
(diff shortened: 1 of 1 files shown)


gabrielschulhof:
@wanghongjuan I've now fixed the "Callback was already called" error as well. Unfortunately the environment in which you're trying to run the test does not seem to have proper time zone support, so the test will fail. At least now the error message will point to the root of the problem (the lack of time zone support). Can you please confirm that the error "Callback was already called" is replaced by an error that actually shows the problem (such as
```
Error: Test requested teardown: Command failed: /bin/sh -c timedatectl set-timezone 
Invalid number of arguments.
```
or
```
Error: Test requested teardown: Command failed: /bin/sh -c timedatectl set-timezone n/a
Failed to set time zone: Invalid time zone 'n/a'
```

wanghongjuan:
@gabrielschulhof the error was reproduced when I added `console.error( error.message );` in suite.js line 104.
```
Command failed: /bin/sh -c timedatectl set-timezone 
Invalid number of arguments.
```

PR #1990, opened 2016-05-06
Clang fixes

ceolin:
Fixes for problems spotted by Clang static analyzer.

### DIFF
--- a/src/lib/common/sol-bus.c
+++ b/src/lib/common/sol-bus.c
@@ -383,9 +383,8 @@ sol_bus_client_free(struct sol_bus_client *client)
     if (!client)
         return;
 
-    destroy_client(client);
-
     sol_ptr_vector_remove(&_ctx.clients, client);
+    destroy_client(client);
 }
 
 SOL_API const char *
--- a/src/lib/comms/sol-http-server-impl-microhttpd.c
+++ b/src/lib/comms/sol-http-server-impl-microhttpd.c
@@ -654,6 +654,8 @@ get_static_file(const struct static_dir *dir, const char *url, struct sol_buffer
     ret = sol_buffer_append_slice(path, sol_str_slice_from_str(real_path));
     free(real_path);
 
+    SOL_INT_CHECK(ret, < 0, ret);
+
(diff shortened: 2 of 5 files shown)


src/lib/common/sol-bus.c
```diff
@@ -383,9 +383,8 @@ sol_bus_client_free(struct sol_bus_client *client)
     if (!client)
         return;
 
-    destroy_client(client);
-
     sol_ptr_vector_remove(&_ctx.clients, client);
+    destroy_client(client);
```
barbieri: really strange, the ptr vector doesn't look into the pointed memory...

ibriano: I read somewhere (don't remember where) that even if you don't access the contents, using freed pointers is undefined behavior. A pointer's value after freed is no longer valid.

ceolin: @barbieri i thought the same

barbieri: makes no sense to me since nobody changes the numeric value, it's basically an "unsigned long int".

ibriano: On Fri, 06 May 2016 10:56:39 -0700, Gustavo Sverzut Barbieri wrote:
> >      sol_ptr_vector_remove(&_ctx.clients, client);
> > +    destroy_client(client);
> 
> makes no sense to me since nobody changes the numeric value, it's basically an "unsigned long int".

Yes, in practice, but there are all sort of things in the standard that
justify saying that after free(), the value of the pointer is invalid.
It's the same reason why arithmetic of pointers that don't all point to
the same object is invalid.

vcgomes:
Anyway, I am fine with these changes. The CI failure looks unrelated (this phrase is awfully common), but please confirm.

src/lib/common/sol-bus.c
barbieri: well, for that at least there is a reason: if you `a - b` and they are not in the same segment, then the result will not be stable.

ceolin:
@vcgomes you are right but the fixes for CI are already in.

ceolin:
thanks

PR #1991, opened 2016-05-06
bluetooth: Fix printf format warnings

ceolin:
Fixes:

./src/lib/comms/sol-bluetooth.c: In function 'sol_bt_uuid_to_str':
./src/lib/comms/sol-bluetooth.c:143:42: warning: format '%x' expects
argument of type 'unsigned int', but argument 3 has type 'uint32_t {aka
long unsigned int}' [-Wformat=]
     r = sol_buffer_append_printf(buffer,
     "%.8x-%.4x-%.4x-%.4x-%.8x%.4x",

### DIFF
--- a/src/lib/comms/sol-bluetooth.c
+++ b/src/lib/comms/sol-bluetooth.c
@@ -16,6 +16,7 @@
  * limitations under the License.
  */
 
+#include <inttypes.h>
 #include <stdio.h>
 
 #include <sol-bluetooth.h>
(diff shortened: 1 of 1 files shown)


ceolin:
@vcgomes could you check it please ?

vcgomes:
Looks ok. I wonder why I didn't get this warning (and neither does BlueZ, which has similar code).

ibriano:
On Fri, 06 May 2016 12:49:05 -0700, Vinicius Gomes wrote:
> Looks ok. I wonder why I didn't get this warning (and neither does BlueZ, which has similar code).

Maybe because when building on Linux (and especially, 64-bits), uint32_t
== unsigned int?

ceolin:
thank you guys

PR #1996, opened 2016-05-09
Ipm v2

edersondisouza:
v2:
Sent blob now keeps message id, so it's easier to reference them on receivers side;
Channel numbers are build configurable;
Consumed confirmation callback now per id, not per message;
New `sol_blob` functions to create blobs that duplicate memory, instead of only wrapping them;
Failure on receiving blobs shall send messages to sender that blob is consumed;
Using `uintptr_t`;
Hopefully all issues of v1 addressed;
*Note:* One more commit to fix build when flow is disabled.

Samples on soletta-small-oses-samples: solettaproject/soletta-small-oses-samples#18


src/lib/common/sol-mainloop.c
```diff
@@ -76,6 +76,10 @@ extern void sol_comms_shutdown(void);
 extern int sol_update_init(void);
 extern void sol_update_shutdown(void);
 #endif
+#ifdef USE_UPDATE
+extern int sol_ipm_init(void);
+extern void sol_ipm_shutdown(void);
+#endif
```
barbieri: what about my proposal to:

```c
#else
static inline int
sol_ipm_init(void)
{
    return 0;
}

static inline void
sol_ipm_shutdown(void)
{
}
#endif
```

src/lib/io/Kconfig
```diff
@@ -81,3 +81,28 @@ config USE_MEMMAP
...
+    depends on USE_IPM
+    default 0
+    help
+        Which IPM channel should be used to communication from x86 core
+        to ARC one.
+
+config IPM_CHANNEL_ARC_TO_X86
+    int "IPM Channel ARC -> x86"
+    depends on USE_IPM
+    default 1
```
barbieri: can't we use the same id for the other way around?

src/lib/io/include/sol-ipm.h
```diff
@@ -0,0 +1,133 @@
...
+ * If @c NULL, deletes a previously set callback.
+ * @param data user defined data that will be sent to callback @a cb
+ *
+ * @return 0 if receiver was set (or unset if @a receive_cb is @c NULL).
+ * Negative number otherwise.
+ *
+ * @note message received is a @c sol_blob. Only when all references to this
+ * blob are gone sender will receive its consumed confirmation.
+ *
+ * @attention Soletta reserves id 0 - do not use it.
```
barbieri: this needs to relate to the build-time option. You can mention by default it's 0. I guess we need a getter for that value as well :-/

src/lib/common/sol-mainloop.c
edersondisouza: On 09-05-2016 16:01, Gustavo Sverzut Barbieri wrote:
> In src/lib/common/sol-mainloop.c
> <https://github.com/solettaproject/soletta/pull/1996#discussion_r62553975>:
>
>> @@ -76,6 +76,10 @@ extern void sol_comms_shutdown(void);
>>  extern int sol_update_init(void);
>>  extern void sol_update_shutdown(void);
>>  #endif
>> +#ifdef USE_UPDATE
>> +extern int sol_ipm_init(void);
>> +extern void sol_ipm_shutdown(void);
>> +#endif
>
> what about my proposal to:
>
> #else
> static inline int
> sol_ipm_init(void)
> {
>     return 0;
> }
>
> static inline void
> sol_ipm_shutdown(void)
> {
> }
> #endif

Omg... I knew I was forgetting something =/

src/lib/io/Kconfig
edersondisouza: On 09-05-2016 16:03, Gustavo Sverzut Barbieri wrote:
> In src/lib/io/Kconfig
> <https://github.com/solettaproject/soletta/pull/1996#discussion_r62554193>:
>
>> +        Provide API to Inter-processor messaging. Some platforms
>> +        have more than one processor - this API provides communication
>> +        among them.
>> +
>> +config IPM_CHANNEL_X86_TO_ARC
>> +    int "IPM Channel x86 -> ARC"
>> +    depends on USE_IPM
>> +    default 0
>> +    help
>> +        Which IPM channel should be used to communication from x86 core
>> +        to ARC one.
>> +
>> +config IPM_CHANNEL_ARC_TO_X86
>> +    int "IPM Channel ARC -> x86"
>> +    depends on USE_IPM
>> +    default 1
>
> can't we use the same id for the other way around?

And use one channel? If so, both cores will be interrupted with 
information. Idea is to use one channel for each direction...

src/lib/io/include/sol-ipm.h
edersondisouza: On 09-05-2016 16:04, Gustavo Sverzut Barbieri wrote:
> In src/lib/io/include/sol-ipm.h
> <https://github.com/solettaproject/soletta/pull/1996#discussion_r62554382>:
>
>> + * Receives messages sent with id @a id on callback @a receive_cb. Message
>> + * content is on callback parameter @a message.
>> + *
>> + * @param id id of messages to listen to. Only messages sent with @a id
>> + * will be sent to this receiver.
>> + * @param receive_cb callback to be called with message sent with @a id.
>> + * If @c NULL, deletes a previously set callback.
>> + * @param data user defined data that will be sent to callback @a cb
>> + *
>> + * @return 0 if receiver was set (or unset if @a receive_cb is @c NULL).
>> + * Negative number otherwise.
>> + *
>> + * @note message received is a @c sol_blob. Only when all references to this
>> + * blob are gone sender will receive its consumed confirmation.
>> + *
>> + * @attention Soletta reserves id 0 - do not use it.
>
> this needs to relate to the build-time option. You can mention by
> default it's 0. I guess we need a getter for that value as well :-/

No, channel id is different from message id. Soletta uses always 0 as 
message id, not channel id. Build time options are for channels id.

src/lib/io/sol-ipm.c
```diff
@@ -0,0 +1,370 @@
...
+#include <stdlib.h>
+
+#ifndef SOL_LOG_DOMAIN
+#define SOL_LOG_DOMAIN &_sol_ipm_log_domain
+#endif
+
+#include "sol-log-internal.h"
+
+#include "ipm.h"
+#include "ipm/ipm_quark_se.h"
```
barbieri: Should we rename `sol-ipm.c` to `sol-ipm-impl-zephyr.c`? I've asked Guedes and Contiki port does not enable the ARC processor, but once someone enables that feature, Contiki could use QMSI to do IPM as well.

src/lib/io/sol-ipm.c
```diff
@@ -0,0 +1,370 @@
...
+    struct sol_blob_ipm_received *ipm_blob = (struct sol_blob_ipm_received *)blob;
+
+    free_remote_blob(ipm_blob->remote);
+
+    /* Free local shadow */
+    free(blob);
+}
+
+extern unsigned char *heap_base;
+extern unsigned int heap_sz, heap_max;
```
barbieri: debug left overs?

src/lib/io/sol-ipm.c
```diff
@@ -0,0 +1,370 @@
...
+
+static void
+ipm_receiver_process(void *data)
+{
+    int idx;
```
barbieri: vectors use `uint16_t`, you should use it here as well. not sure why it did no trigger a signed x unsigned comparison warning

src/lib/io/sol-ipm.c
```diff
@@ -0,0 +1,370 @@
...
+    struct sol_ipm_receiver_entry *receiver;
+
+    SOL_VECTOR_FOREACH_IDX (&ipm_receivers, receiver, idx) {
+        if (receiver->id == message->id) {
+            struct sol_blob_ipm_received *remote_blob;
+
+            /* Setup local shadow blob whose `mem` points to the same `mem` as remote one*/
+            remote_blob = calloc(1, sizeof(struct sol_blob_ipm_received));
+            if (!remote_blob) {
+                SOL_WRN("Could not create blob to deliver IPM message");
```
barbieri: more context to the log, like message address and message->id?

src/lib/io/sol-ipm.c
```diff
@@ -0,0 +1,370 @@
...
+    int idx;
+    struct sol_blob_ipm_sent *message = data;
+    struct sol_ipm_consumed_handler *handler;
+
+    SOL_VECTOR_FOREACH_IDX (&ipm_consumed_handlers, handler, idx) {
+        if (handler->id == message->id) {
+            handler->message_consumed_cb((void *)handler->data, message->id,
+                message->base.parent);
+            sol_blob_unref(&message->base);
+            free(message); /* This blob is SOL_BLOB_TYPE_NO_FREE, so we need to
```
barbieri: the `NO_FREE` doesn't free the pointed memory, but it does free the blob structure. Actually this is something I've asked @cabelitos to review, but AFAIU it was not done yet. See https://github.com/solettaproject/soletta/pull/1918#discussion_r60498120

src/lib/io/sol-ipm.c
```diff
@@ -0,0 +1,370 @@
...
+sol_ipm_set_receiver(uint32_t id,
+    void (*receive_cb)(void *cb_data, uint32_t id, struct sol_blob *message),
+    const void *data)
+{
+    int idx;
+    struct sol_ipm_receiver_entry *receiver;
+
+    if (!initialised) {
+        int r = init();
+        SOL_INT_CHECK(r, != 0, r);
```
barbieri: uncrustify will bitch about  this, it requires a space between the declaration and the usage.

src/lib/io/sol-ipm.c
```diff
@@ -0,0 +1,370 @@
...
+    }
+
+    SOL_INT_CHECK(id, > ipm_max_id, -EINVAL);
+    SOL_INT_CHECK(id, == 0, -EINVAL);
+
+    SOL_VECTOR_FOREACH_IDX (&ipm_receivers, receiver, idx) {
+        if (receiver->id == id) {
+            if (receive_cb) {
+                receiver->receive_cb = receive_cb;
```
barbieri: what about  the new data? You should set it as well.

However, I'm thinking if we should be strict here and check if `receiver->receiver_cb == NULL`, if not then fail with `-EEXIST`. This is to avoid accidental "steal of ownership", since there is no "get/check receiver for this id"

src/lib/io/sol-ipm.c
```diff
@@ -0,0 +1,370 @@
...
+                return 0;
+            } else {
+                return sol_vector_del(&ipm_receivers, idx);
+            }
+        }
+    }
+
+    /* NULL receive_cb means it's trying to remove a receiver, but none was found */
+    SOL_NULL_CHECK(receive_cb, -EEXIST);
```
barbieri: `-ENOENT`

src/lib/io/sol-ipm.c
```diff
@@ -0,0 +1,370 @@
...
+
+    blob = calloc(1, sizeof(struct sol_blob_ipm_sent));
+    SOL_NULL_CHECK(blob, -ENOMEM);
+
+    sol_blob_setup(&blob->base, SOL_BLOB_TYPE_NO_FREE, message->mem, message->size);
+    sol_blob_set_parent(&blob->base, message);
+    blob->id = id;
+
+    r = ipm_send(ipm_sender, 1, id, &blob, sizeof(uintptr_t));
+    SOL_INT_CHECK_GOTO(r, > 0, fail);
```
barbieri: greater than zero is error? or less?

src/lib/io/sol-ipm.c
barbieri: also, shouldn't you initialize only in the success case where you need to `sol_vector_append()`?  Otherwise on errors, you did that for no good reason.

src/lib/io/sol-ipm.c
barbieri: well, it's consistent with the `ipm_set_enabled()`, but there you check `!= 0`. Maybe do the same pattern?

src/lib/io/sol-ipm.c
```diff
@@ -0,0 +1,370 @@
...
+    }
+
+    SOL_INT_CHECK(id, > ipm_max_id, -EINVAL);
+    SOL_INT_CHECK(id, == 0, -EINVAL);
+
+    SOL_VECTOR_FOREACH_IDX (&ipm_consumed_handlers, handler, idx) {
+        if (handler->id == id && message_consumed_cb)
+            return -EEXIST;
+
+        return sol_vector_del(&ipm_consumed_handlers, idx);
```
barbieri: what? if you set a callback == NULL it will delete all other ids :-) `handler->id == id` false, then `return sol_vector_del(, idx)`

src/modules/flow/ipm/ipm.json
```diff
@@ -0,0 +1,475 @@
...
+      "category": "output/ipm",
+      "description": "Inter-Processor Messaging boolean writer. Writes a boolean packet with a given ID to a different core.",
+      "methods": {
+        "open": "boolean_writer_open"
+      },
+      "name": "ipm/boolean-writer",
+      "options": {
+        "members": [
+          {
+            "data_type": "int",
```
barbieri: you need an `uint32_t`, but it's not available. However I believe we should add it to the generator (and schema) instead of propagating the casts in the code. This looks like a reasonable type to add.

src/modules/flow/ipm/ipm.json
```diff
@@ -0,0 +1,475 @@
...
+      },
+      "in_ports": [
+        {
+          "data_type": "boolean",
+          "description": "Boolean packets sent here will be sent to another core.",
+          "methods": {
+            "process": "boolean_writer_process"
+          },
+          "name": "IN"
+        }
```
barbieri: why no `CONSUMED` (or `PROCESSED`) port? It could be used as a trigger to process further information and control the flow to not overflow the channel. Take a loop where the next iteration can be controlled by its own trigger (ie: accumulator), then you'd link that port to the accumulator increase.

tools/build/Makefile.targets
```diff
@@ -237,7 +237,11 @@ endif #neq($(USE_NODEJS),)
 
 DEFAULT_ALIAS_FILE := $(build_aliasesdir)/50-default.json
 
+ifeq (y,$(FLOW_SUPPORT))
 gen-aliases:
 	$(Q)$(MKDIR) -p $(build_aliasesdir)
 	$(Q)echo "     "GEN"   "$(DEFAULT_ALIAS_FILE)
 	$(Q)$(FLOW_NODE_TYPE_ALIAS) --output=$(DEFAULT_ALIAS_FILE) $(all-gens)
+else
+gen-aliases:
```
barbieri: there is no need to change the rule, only do not add it to `PRE_GEN`

edersondisouza: On 09-05-2016 17:06, Gustavo Sverzut Barbieri wrote:
> In tools/build/Makefile.targets
> <https://github.com/solettaproject/soletta/pull/1996#discussion_r62563559>:
>
>>  gen-aliases:
>>  	$(Q)$(MKDIR) -p $(build_aliasesdir)
>>  	$(Q)echo "     "GEN"   "$(DEFAULT_ALIAS_FILE)
>>  	$(Q)$(FLOW_NODE_TYPE_ALIAS) --output=$(DEFAULT_ALIAS_FILE) $(all-gens)
>> +else
>> +gen-aliases:
>
> there is no need to change the rule, only do not add it to |PRE_GEN|

This was properly addressed on Dilly's patch - I'll drop this one.

src/lib/io/sol-ipm.c
edersondisouza: On 09-05-2016 16:59, Gustavo Sverzut Barbieri wrote:
> In src/lib/io/sol-ipm.c
> <https://github.com/solettaproject/soletta/pull/1996#discussion_r62562414>:
>
>> +    int idx;
>> +    struct sol_ipm_consumed_handler *handler;
>> +
>> +    if (!initialised) {
>> +        int r = init();
>> +        SOL_INT_CHECK(r, != 0, r);
>> +    }
>> +
>> +    SOL_INT_CHECK(id, > ipm_max_id, -EINVAL);
>> +    SOL_INT_CHECK(id, == 0, -EINVAL);
>> +
>> +    SOL_VECTOR_FOREACH_IDX (&ipm_consumed_handlers, handler, idx) {
>> +        if (handler->id == id && message_consumed_cb)
>> +            return -EEXIST;
>> +
>> +        return sol_vector_del(&ipm_consumed_handlers, idx);
>
> what? if you set a callback == NULL it will delete all other ids :-)
> |handler->id == id| false, then |return sol_vector_del(, idx)|

It doesn't delete all,  only first, as it's a return =P
-ETOOMUCHDORGAS

src/lib/io/sol-ipm.c
edersondisouza: Maybe, but couldn't we leave this to when it's needed? Them some common code could be factored out to some `sol-ipm-common.c`.

src/modules/flow/ipm/ipm.json
edersondisouza: Shall this be done for this PR? I think it could be done after, looking for other places to use new type...

barbieri: you can do it afterwards, just do not forget about it :-)

edersondisouza:
v3 on #2006

PR #2001, opened 2016-05-09
scripts/dependency-resolver: check for exec versions

bdilly:
Add options to check for specific exec versions.

Require at least LTS version of node (v4.4.4).

On old versions, like v0.10.x it wasn't able to build,
it was re-triggering builds on loop.

@glima please review it when possible ^

### DIFF
--- a/data/scripts/dependency-resolver.py
+++ b/data/scripts/dependency-resolver.py
@@ -249,6 +249,12 @@ def handle_exec_check(args, conf, context):
     dep = conf.get("dependency")
     dep_sym = dep.upper()
     exe = conf.get("exec")
+    atleast_ver = conf.get("atleast-version")
+    max_ver = conf.get("max-version")
+    exact_ver = conf.get("exact-version")
+    cmd = conf.get("version-command")
+    ver_match = True
+    version = None
 
     if not exe:
         context.info("Could not parse dependency: %s, no exec was specified.", dep)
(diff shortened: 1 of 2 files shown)


data/scripts/dependency-resolver.py
```diff
@@ -249,6 +249,12 @@ def handle_exec_check(args, conf, context):
     dep = conf.get("dependency")
     dep_sym = dep.upper()
     exe = conf.get("exec")
+    atleast_ver = conf.get("atleast-version")
+    max_ver = conf.get("max-version")
+    exact_ver = conf.get("exact-version")
+    cmd = conf.get("command-version")
```
barbieri: nitpick, but would `version-command` match better the usual names we're using here, like `exact-version`  instead of `version-exact`

bdilly: makes sense, I've just sent a version with that fixed.

barbieri:
+1

data/scripts/dependency-resolver.py
```diff
@@ -257,20 +263,40 @@ def handle_exec_check(args, conf, context):
...
+            ver_match = False
+
+    if required:
+        if not path:
+            req_label = context.find_makefile_var("NOT_FOUND")
+            req_label += "executable: %s\\n" % exe
+            context.add_append_makefile_var("NOT_FOUND", req_label, True)
+        elif not ver_match:
+            req_label = context.find_makefile_var("NOT FOUND")
+            req_label += "%s version: %s\\n" % (exe, version)
```
glima: You could tell the user the expected version as well

glima:
It seems to work, but it beats me where those messages will go to

data/scripts/dependency-resolver.py
```diff
@@ -257,20 +263,40 @@ def handle_exec_check(args, conf, context):
...
+            ver_match = False
+        elif (exact_ver and result != exact_ver) or \
+            (max_ver and result > max_ver) or \
+            (atleast_ver and result < atleast_ver):
+            version = result
+            ver_match = False
+
+    if required:
+        if not path:
+            req_label = context.find_makefile_var("NOT_FOUND")
```
glima: NOT_FOUND or NOT FOUND?

bdilly:
@glima it logs it on Makefile.gen, but only if it's a requirement marked as "required", not the case.

So for the only use case so far, no message should be log in any case.

data/scripts/dependency-resolver.py
bdilly: It could, I'm just following the patter used for pkg-config and python checks.
It would need to check what is the kind of check and the expected version. As I said in other comment, it doesn't even log anything for any case we have so far. So I'm not much concerned about it. It's easy to find this information on dependencies file (as easy as to check on this log message).

glima:
Oh, I see. Ack.

data/scripts/dependency-resolver.py
bdilly: It should be "NOT_FOUND". Fixing

PR #2002, opened 2016-05-10
Bluetooth changes

ceolin:
- Add support for bluetooth address over all platforms
- Fix warnings
- Share common code
- Change transport values

### DIFF
--- a/src/lib/comms/include/sol-bluetooth.h
+++ b/src/lib/comms/include/sol-bluetooth.h
@@ -264,7 +264,7 @@ struct sol_bt_device_info {
  */
 enum sol_bt_transport {
     /** @brief Discover devices over the Bluetooth Low Energy transport */
-    SOL_BT_TRANSPORT_LE,
+    SOL_BT_TRANSPORT_LE = 1,
     /** @brief Discover devices over the Bluetooth Basic Rate transport */
     SOL_BT_TRANSPORT_BREDR,
     /** @brief Discover devices over All transports */
--- a/src/lib/comms/sol-network-util.h
+++ b/src/lib/comms/sol-network-util.h
@@ -55,10 +55,10 @@ sol_bluetooth_is_addr_str(const char *str)
     const char *p = str;
 
     while (*str) {
-        if (!isxdigit(*str++))
+        if (!isxdigit((int)*str++))
             return false;
 
-        if (!isxdigit(*str++))
+        if (!isxdigit((int)*str++))
             return false;
 
         if (*str == '0')
--- a/src/lib/comms/sol-network-util.h
+++ b/src/lib/comms/sol-network-util.h
@@ -18,7 +18,86 @@
 
 #pragma once
 
+#include <ctype.h>
+
+#include <sol-buffer.h>
 #include <sol-network.h>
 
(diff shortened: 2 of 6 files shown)


ceolin:
@vcgomes could you take a look please

src/lib/comms/sol-network-util.h
```diff
@@ -55,10 +55,10 @@ sol_bluetooth_is_addr_str(const char *str)
     const char *p = str;
 
     while (*str) {
-        if (!isxdigit(*str++))
+        if (!isxdigit((int)*str++))
             return false;
 
-        if (!isxdigit(*str++))
+        if (!isxdigit((int)*str++))
```
cabelitos: I would merge this one with  366e3a8

vcgomes: I personally prefer this as it is, but no strong feelings about it.

src/lib/comms/include/sol-bluetooth.h
```diff
@@ -264,7 +264,7 @@ struct sol_bt_device_info {
  */
 enum sol_bt_transport {
     /** @brief Discover devices over the Bluetooth Low Energy transport */
-    SOL_BT_TRANSPORT_LE,
+    SOL_BT_TRANSPORT_LE = 1,
```
vcgomes: Ugh, I wasn't thinking right when I changed this to be a bit mask.

src/lib/comms/sol-network-util.h
ibriano: On Tue, 10 May 2016 07:46:50 -0700, Vinicius Gomes wrote:
> >              return false;
> >  
> > -        if (!isxdigit(*str++))
> > +        if (!isxdigit((int)*str++))
> 
> I personally prefer this as it is, but no strong feelings about it.

These ctype macros tend to complain some times when the parameter passed
is a plain 'char'. At the very least, make sure to cast it to an
unsigned one.

vcgomes:
Looks good.

The only thing that I don't know is how valuable it is to add the address conversion stuff to RIOT, as there are no plans to support Bluetooth there.

src/lib/comms/sol-network-util.h
vcgomes: I didn't explain myself well, I was referring to @cabelitos comment about merging with an earlier commit.

ceolin: @cabelitos I prefer in a different commit. They have different meaning. If I ammend the reason of this change will be lost.

src/lib/comms/sol-network-util.h
```diff
@@ -18,7 +18,86 @@
...
+    r = sol_buffer_append_printf(buffer, "%2.2X:%2.2X:%2.2X:%2.2X:%2.2X:%2.2X",
+        ba[0], ba[1], ba[2], ba[3], ba[4], ba[5]);
+    if (r < 0)
+        return NULL;
+
+    return buffer->data;
+}
+
+static inline bool
+sol_bluetooth_is_family(int family)
```
glima: This does not seem to be exported, so fine. But the name is weird.

vcgomes:
After thinking a little more, I will lift my restriction that this might not make sense for RIOT.

PR #2004, opened 2016-05-10
Fix incorrect string empty check in validators

otaviobp:
Int and byte validators were not checking correctly for empty validator string.


barbieri:
+1

bdilly:
ok

PR #2005, opened 2016-05-10
oic: Fix memory leak when string fields are not changed in server update

otaviobp:
When a from_repr_vec method is called, the oic server resource contains
a string, but this string is not updated, we had a memory leak.

Also add a name to Light Server sample to have string fields tested.


glima:
Make ``make samples`` pass again and gg.

PR #2006, opened 2016-05-10
Ipm v3

edersondisouza:
v3:
'Consumed' and 'receive' callbacks can only be set from NULL, or to NULL. This avoids accidental overwriting them;
Improve `sol-mainloop.c` inits and shutdowns readability;
Remove debug leftovers;
New port 'CONSUMED' on node types, akin to consumed callback from C API.

v2:
Sent blob now keeps message id, so it's easier to reference them on receivers side;
Channel numbers are build configurable;
Consumed confirmation callback now per id, not per message;
New sol_blob functions to create blobs that duplicate memory, instead of only wrapping them;
Failure on receiving blobs shall send messages to sender that blob is consumed;
Using uintptr_t;
Hopefully all issues of v1 addressed;
Note: One more commit to fix build when flow is disabled.

Samples on soletta-small-oses-samples: solettaproject/soletta-small-oses-samples#19


src/lib/io/sol-ipm.c
```diff
@@ -0,0 +1,371 @@
...
+    .free = free_ipm_blob,
+};
+
+/*** IPM Stuff ***/
+
+struct sol_ipm_receiver_entry {
+    uint32_t id;
+    void (*receive_cb)(void *cb_data, uint32_t id, struct sol_blob *message);
+    const void *data;
+    struct sol_blob *last_remote_blob; /* Hackish... I store last blob received here, to avoid having yet another list to store them */
```
barbieri: told you... this nasty bit keeps lurking into commits :-)

barbieri:
+1, you can remove the struct member and merge it directly.

src/lib/io/sol-ipm.c
edersondisouza: :new_moon_with_face:

edersondisouza:
Integrated with due fix, tks!

PR #2007, opened 2016-05-10
Uart API changes v5

cabelitos:
Changes since v4:
* The config now contains tx_size and rx_size
* Fixed problems pointed by @barbieri and @edersondisouza

### DIFF
--- a/src/lib/io/include/sol-uart.h
+++ b/src/lib/io/include/sol-uart.h
@@ -99,6 +99,7 @@ enum sol_uart_stop_bits {
  * @brief A configuration struct used to set the UART paramenters.
  *
  * @see sol_uart_open()
+ * @note UART follows the Soletta stream guidelines, which can be found here: @ref streams
  */
 struct sol_uart_config {
 #ifndef SOL_NO_API_VERSION
@@ -278,6 +279,7 @@ void sol_uart_close(struct sol_uart *uart);
  * @param blob The blob to be written
  * @return 0 on success or negative errno on error
  * @see sol_uart_config::tx_cb()
+ * @note UART follows the Soletta stream guidelines, which can be found here: @ref streams
  *
  */
 int sol_uart_write(struct sol_uart *uart, struct sol_blob *blob);
--- a/src/lib/io/include/sol-uart.h
+++ b/src/lib/io/include/sol-uart.h
@@ -42,6 +42,13 @@ extern "C" {
  * @{
  */
 
+/**
+ * @struct sol_uart
+ * @brief A handle to a UART device.
+ * @see sol_uart_open()
+ * @see sol_uart_close()
+ * @see sol_uart_write()
+ */
 struct sol_uart;
 
 /**
@@ -58,13 +65,13 @@ enum sol_uart_baud_rate {
 };
 
 /**
- * @brief Amount of stop bits.
+ * @brief Amount of data bits.
  */
 enum sol_uart_data_bits {
-    SOL_UART_DATA_BITS_8 = 0,
-    SOL_UART_DATA_BITS_7,
-    SOL_UART_DATA_BITS_6,
-    SOL_UART_DATA_BITS_5
+    SOL_UART_DATA_BITS_8 = 0, /**< Use 8 data bits */
+    SOL_UART_DATA_BITS_7, /**< Use 7 data bits */
+    SOL_UART_DATA_BITS_6, /**< Use 6 data bits */
+    SOL_UART_DATA_BITS_5 /**< Use 5 data bits */
 };
 
 /**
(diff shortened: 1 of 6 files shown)


src/lib/io/include/sol-uart.h
```diff
@@ -81,22 +90,56 @@ enum sol_uart_parity {
...
+    /**
+     * @brief Callback containing data was read from UART
+     *
+     * @param data User data as provided in sol_uart_config::user_data
+     * @param uart The UART handle
+     * @param buf The UART data
+     *
+     * @return The number of bytes read from @c buf or negative errno on error.
+     */
+    ssize_t (*rx_cb)(void *data, struct sol_uart *uart, struct sol_buffer *buf);
```
barbieri: should buf be `const`?

barbieri: or maybe `sol_str_slice` (you can mention it is a binary data and thus may contain inner \0 or no \0 at all, thus `len` must be considered)

src/lib/io/sol-uart-impl-linux.c
```diff
@@ -32,49 +33,125 @@
...
-    if (active_flags & SOL_FD_FLAGS_IN) {
-        uint8_t c; /* Backing storage for next sol_buffer */
-        struct sol_buffer buf = SOL_BUFFER_INIT_FLAGS(&c, sizeof(c),
-            SOL_BUFFER_FLAGS_MEMORY_NOT_OWNED | SOL_BUFFER_FLAGS_NO_NUL_BYTE);
-        int status = sol_util_fill_buffer(uart->fd, &buf, 1);
-        if (status > 0)
-            uart->async.rx_cb((void *)uart->async.rx_user_data, uart, c);
+
+    if ((active_flags & SOL_FD_FLAGS_IN) && uart->rx_cb) {
+        size_t remeaning = uart->rx.capacity - uart->rx.used;
```
barbieri: typo, and I guess check-style will bitch about lack of newline.

src/lib/io/sol-uart-impl-linux.c
```diff
@@ -151,14 +229,35 @@ sol_uart_open(const char *port_name, const struct sol_uart_config *config)
...
+            SOL_ERR("Unable to add file descriptor to watch UART.");
+            goto fail;
+        }
+
+        uart->rx_cb = config->rx_cb;
+        rx_size = config->rx_size;
+        if (!rx_size)
+            rx_size = DEFAULT_BUFFER_SIZE;
+
+        rx_buf = calloc(1, rx_size);
```
barbieri: minor, but here you could use `malloc()` since the contents do not need to be zeroed.

src/lib/io/sol-uart-impl-linux.c
```diff
@@ -172,72 +271,72 @@ sol_uart_open(const char *port_name, const struct sol_uart_config *config)
...
-        goto error;
+    if (!uart->fd_handler) {
+        uart->fd_handler = sol_fd_add(uart->fd,
+            FD_ERROR_FLAGS | SOL_FD_FLAGS_OUT, uart_fd_handler_callback, uart);
+        SOL_NULL_CHECK(uart->fd_handler, -ENOMEM);
+        fd_created = true;
+    } else {
+        bool err;
+
+        err = sol_fd_add_flags(uart->fd_handler, SOL_FD_FLAGS_OUT);
```
barbieri: either here or in the `sol_fd_add_flags()` (https://github.com/solettaproject/soletta/blob/master/src/lib/common/sol-mainloop.c#L334) we'd need to check if the flag was already set, in that case avoid marking the "main loop" as dirty (ie: https://github.com/solettaproject/soletta/blob/master/src/lib/common/sol-mainloop-impl-posix.c#L753)

I guess on `sol-mainloop.c` it may benefit more people without adding complexity

src/lib/io/sol-uart-impl-linux.c
```diff
@@ -172,72 +271,72 @@ sol_uart_open(const char *port_name, const struct sol_uart_config *config)
...
-{
-    SOL_NULL_CHECK(uart, false);
-    SOL_EXP_CHECK(uart->async.tx_fd_handler != NULL, false);
+    r = sol_ptr_vector_append(&uart->pending_blobs, blob);
+    SOL_INT_CHECK_GOTO(r, < 0, err_append);
 
-    uart->async.tx_fd_handler = sol_fd_add(uart->fd,
-        FD_ERROR_FLAGS | SOL_FD_FLAGS_OUT, uart_tx_callback, uart);
-    SOL_NULL_CHECK(uart->async.tx_fd_handler, false);
+    uart->pending_tx += blob->size;
```
barbieri: why not just use `uart->pending_tx = total`, the sum was already calculated.

barbieri:
much nicer, minor nitpicks. Also noticed we do not have a uart fbp module with a reader and a writer, we should add those modeled on top of blob, a sample could be read from serial and stream via http or mqtt.

src/lib/io/sol-uart-impl-linux.c
cabelitos: Check style did not complain that the newline, but I will add it.

src/lib/io/sol-uart-impl-linux.c
cabelitos: Indeed.

src/lib/io/sol-uart-impl-linux.c
cabelitos: Ok.

src/lib/io/sol-uart-impl-linux.c
cabelitos: I will add that to `sol-mainloop.c`

src/lib/io/include/sol-uart.h
cabelitos: I will change to const

bdilly:
@barbieri months ago we discussed about it - providing generic node types for generic protocols that handle complex data, like SPI, UART and I2C - and our conclusion is that they wouldn't be useful.

Instead we should provide node types that use these protocols for specific purposes. And it's the approach we're following. It's the reason we don't have node types for any of these.

We certainly may reconsider it. Your use case seems interesting. Anyway I consider other tasks related to our C API stability have higher priority than adding such node types atm.

I'll create an issue for it so we can attack it later.

barbieri:
ok, let's leave node types to be implemented later. no types for streaming at this moment

src/lib/io/sol-uart-impl-linux.c
```diff
@@ -110,6 +110,7 @@ sol_uart_open(const char *port_name, const struct sol_uart_config *config)
     SOL_NULL_CHECK(port_name, NULL);
 
     r = snprintf(device, sizeof(device), "/dev/%s", port_name);
+    SOL_INT_CHECK(r, < 0, NULL);
     SOL_INT_CHECK(r, >= (int)sizeof(device), NULL);
```
ceolin: You could use `SOL_EXP_CHECK` here instead of two `SOL_INT_CHECK` if you want.

src/lib/io/include/sol-uart.h
edersondisouza: @barbieri Why `sol_str_slice`? I'm just curious...

cabelitos:
#2014

PR #2008, opened 2016-05-10
sol_mainloop_set_implementation() is called before sol_init(), no logs allowed.

barbieri:
We can't use sol-log.h primitives since sol_init() wasn't called, so
proceed without logs.


ceolin:
:+1:

PR #2012, opened 2016-05-11
vector: fix suboptimal sol_ptr_vector_del_element implementation v2

kaspersky:
Signed-off-by: Vladimir Cernov <cernov.vladimir@gmail.com>

### DIFF
--- a/src/lib/datatypes/sol-vector.c
+++ b/src/lib/datatypes/sol-vector.c
@@ -377,18 +377,23 @@ sol_vector_del_element(struct sol_vector *v, const void *elem)
 SOL_API int
 sol_ptr_vector_del_element(struct sol_ptr_vector *pv, const void *elem)
 {
-    int r;
-    uint16_t i, removed = 0;
-    void *cur;
+    uint16_t i = 0, offset = 0;
+    void **v = pv->base.data;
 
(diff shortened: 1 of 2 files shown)


kaspersky:
#1730 

I have avoided copying elements in the same place without using an if statement: by having 2 for loops (first one finds first occurence, second one removes elements).

edersondisouza:
+1

src/lib/datatypes/sol-vector.c
```diff
@@ -377,18 +377,23 @@ sol_vector_del_element(struct sol_vector *v, const void *elem)
...
-            SOL_INT_CHECK(r, < 0, r);
-            removed++;
-        }
+    uint16_t i = 0, offset = 0;
+    void **v = pv->base.data;
+
+    while (i < pv->base.len && v[i] != elem) {
+        i++;
+        offset++;
+    }
```
vcgomes: I am going to suggest the tiniest of optimizations, feel totally free to ignore: do `offset = i;` after the `while` loop.

vcgomes:
Looking good.

src/lib/datatypes/sol-vector.c
kaspersky: @vcgomes , good observation, will do. Wonder how I didn't think of it myself.

kaspersky:
#2019

PR #2013, opened 2016-05-11
WIP refactor code

yongli3:

### DIFF
--- a/src/lib/common/sol-pin-mux.c
+++ b/src/lib/common/sol-pin-mux.c
@@ -75,8 +75,8 @@ _load_mux(const char *name)
 
 #ifndef SOL_NO_API_VERSION
     if (p_sym->api_version != SOL_PIN_MUX_API_VERSION) {
-        SOL_WRN("Mux '%s' has incorrect api_version: %u expected %u", path, p_sym->api_version,
-            SOL_PIN_MUX_API_VERSION);
+        SOL_WRN("Mux '%s' has incorrect api_version: %" PRIu16 " expected %"
+            PRIu16, path, p_sym->api_version, SOL_PIN_MUX_API_VERSION);
         goto error;
     }
 #endif
--- a/src/lib/common/sol-update.c
+++ b/src/lib/common/sol-update.c
@@ -57,8 +57,8 @@ check_module(const char *path, const char *symbol_name, void *symbol)
 
 #ifndef SOL_NO_API_VERSION
     if ((*p_update)->api_version != SOL_UPDATE_API_VERSION) {
-        SOL_WRN("Module [%s] has incorrect api_version: %u expected %u", path,
-            (*p_update)->api_version, SOL_UPDATE_API_VERSION);
+        SOL_WRN("Module [%s] has incorrect api_version: %" PRIu16 " expected %"
+            PRIu16, path, (*p_update)->api_version, SOL_UPDATE_API_VERSION);
         return false;
     }
 #endif
--- a/src/lib/common/sol-worker-thread.c
+++ b/src/lib/common/sol-worker-thread.c
@@ -34,8 +34,8 @@ sol_worker_thread_new(const struct sol_worker_thread_config *config)
 
 #ifndef SOL_NO_API_VERSION
     if (SOL_UNLIKELY(config->api_version != SOL_WORKER_THREAD_CONFIG_API_VERSION)) {
-        SOL_WRN("Couldn't create worker thread with unsupported version '%u', "
-            "expected version is '%u'",
+        SOL_WRN("Couldn't create worker thread with unsupported version '%" PRIu16 "', "
+            "expected version is '%" PRIu16 "'",
             config->api_version, SOL_WORKER_THREAD_CONFIG_API_VERSION);
         return NULL;
     }
--- a/src/lib/comms/include/sol-http.h
+++ b/src/lib/comms/include/sol-http.h
@@ -23,6 +23,7 @@
 #include <sol-macros.h>
 #include <sol-str-slice.h>
 #include <sol-vector.h>
+#include <inttypes.h>
 
 #ifdef __cplusplus
 extern "C" {
@@ -213,7 +214,7 @@ struct sol_http_url {
 #define SOL_HTTP_RESPONSE_CHECK_API_VERSION(response_, ...) \
     if (SOL_UNLIKELY((response_)->api_version != \
         SOL_HTTP_RESPONSE_API_VERSION)) { \
-        SOL_ERR("Unexpected API version (response is %u, expected %u)", \
+        SOL_ERR("Unexpected API version (response is %" PRIu16 ", expected %" PRIu16 ")", \
             (response_)->api_version, SOL_HTTP_RESPONSE_API_VERSION); \
         return __VA_ARGS__; \
     }
@@ -247,7 +248,7 @@ struct sol_http_url {
 #define SOL_HTTP_RESPONSE_CHECK_API_VERSION_GOTO(response_, label) \
     if (SOL_UNLIKELY((response_)->api_version != \
         SOL_HTTP_RESPONSE_API_VERSION)) { \
-        SOL_ERR("Unexpected API version (response is %u, expected %u)", \
+        SOL_ERR("Unexpected API version (response is %" PRIu16 ", expected %" PRIu16 ")", \
             (response_)->api_version, SOL_HTTP_RESPONSE_API_VERSION); \
         goto label; \
     }
@@ -299,7 +300,7 @@ struct sol_http_url {
 #define SOL_HTTP_PARAMS_CHECK_API_VERSION(params_, ...) \
     if (SOL_UNLIKELY((params_)->api_version != \
         SOL_HTTP_PARAM_API_VERSION)) { \
-        SOL_ERR("Unexpected API version (response is %u, expected %u)", \
+        SOL_ERR("Unexpected API version (response is %" PRIu16 ", expected %" PRIu16 ")", \
             (params_)->api_version, SOL_HTTP_PARAM_API_VERSION); \
         return __VA_ARGS__; \
     }
@@ -317,7 +318,7 @@ struct sol_http_url {
 #define SOL_HTTP_PARAMS_CHECK_API_VERSION_GOTO(params_, label_) \
     if (SOL_UNLIKELY((params_)->api_version != \
         SOL_HTTP_PARAM_API_VERSION)) { \
-        SOL_ERR("Unexpected API version (params is %u, expected %u)", \
+        SOL_ERR("Unexpected API version (params is %" PRIu16 ", expected %" PRIu16 ")", \
             (params_)->api_version, SOL_HTTP_PARAM_API_VERSION); \
         goto label_; \
     }
--- a/src/lib/comms/include/sol-mqtt.h
+++ b/src/lib/comms/include/sol-mqtt.h
@@ -23,6 +23,7 @@
 #include <sol-common-buildopts.h>
 #include <stdbool.h>
 #include <stdlib.h>
+#include <inttypes.h>
 
 #ifdef __cplusplus
 extern "C" {
@@ -423,7 +424,7 @@ int sol_mqtt_subscribe(const struct sol_mqtt *mqtt, const char *topic, sol_mqtt_
 #define SOL_MQTT_MESSAGE_CHECK_API_VERSION(msg_, ...) \
     if (SOL_UNLIKELY((msg_)->api_version != \
         SOL_MQTT_MESSAGE_API_VERSION)) { \
-        SOL_ERR("Unexpected API version (message is %u, expected %u)", \
+        SOL_ERR("Unexpected API version (message is %" PRIu16 ", expected %" PRIu16 ")", \
             (msg_)->api_version, SOL_MQTT_MESSAGE_API_VERSION); \
         return __VA_ARGS__; \
     }
--- a/src/lib/comms/sol-coap.c
+++ b/src/lib/comms/sol-coap.c
@@ -57,7 +57,7 @@ SOL_LOG_INTERNAL_DECLARE(_sol_coap_log_domain, "coap");
         if (SOL_UNLIKELY(resource->api_version != \
             SOL_COAP_RESOURCE_API_VERSION)) { \
             SOL_WRN("Couldn't handle resource that has unsupported version " \
-                "'%u', expected version is '%u'", \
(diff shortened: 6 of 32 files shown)


yongli3:
@edersondisouza please help to review this PR for the code refactor, it is not finished yet, just want to share my thoughts as soon as possible. I would like to re-define the data struct for all different iio nodes to use the same one, and by changing the reader_cb function, all different nodes can re-use it

src/modules/flow/iio/nodes.c
```diff
@@ -25,15 +25,39 @@
...
     struct sol_direction_vector offset;
     struct sol_drange_spec out_range;
     struct sol_iio_device *device;
-    struct sol_iio_channel *channel_x;
-    struct sol_iio_channel *channel_y;
-    struct sol_iio_channel *channel_z;
+    struct sol_iio_channel *channels[MAX_CHANNEL];
+    enum iiodevice_types type;
+    uint16_t src_port;
+    int channel_count : 1;
```
edersondisouza: I think that only one bit won't keep any useful 'channel_count', maybe you wanted a `uint8_t`.

src/modules/flow/iio/nodes.c
```diff
@@ -25,15 +25,39 @@
...
     struct sol_direction_vector scale;
     struct sol_direction_vector offset;
     struct sol_drange_spec out_range;
     struct sol_iio_device *device;
-    struct sol_iio_channel *channel_x;
-    struct sol_iio_channel *channel_y;
-    struct sol_iio_channel *channel_z;
+    struct sol_iio_channel *channels[MAX_CHANNEL];
+    enum iiodevice_types type;
+    uint16_t src_port;
```
edersondisouza: `src_port` is a bit misleading, don't you think? This is actually an output port. Nevertheless, you may avoid having this per node instance, if you use node type data. See flow/update or flow/ipm to see samples where this is used. Look for 'node_type' entry on their JSON. Should you have any doubt, please don't hesitate in asking.

src/modules/flow/iio/nodes.c
```diff
@@ -44,24 +68,31 @@ reader_cb(void *data, struct sol_iio_device *device)
...
 
-    b = sol_iio_read_channel_value(mdata->channel_x, &out.x);
-    if (!b) goto error;
-
-    b = sol_iio_read_channel_value(mdata->channel_y, &out.y);
-    if (!b) goto error;
-
-    b = sol_iio_read_channel_value(mdata->channel_z, &out.z);
-    if (!b) goto error;
+    switch (mdata->type) {
```
edersondisouza: Don't forget to see checkstyle, Soletta switches have a different indentation.

src/modules/flow/iio/nodes.c
```diff
@@ -25,15 +25,39 @@
...
+IIO_PROXIMITY=10,
+};
+
+struct iiodevice_out {
+    union {
+        struct sol_direction_vector direction;
+        struct sol_drange drange;
+        struct sol_rgb rgb;
+    };
+};
```
edersondisouza: I'm not sure that this will be useful... why do you think it will? Maybe I'm not seeing something, when you add more node types to refactor I may see the truth...

src/modules/flow/iio/nodes.c
```diff
@@ -71,33 +102,47 @@ reader_cb(void *data, struct sol_iio_device *device)
...
+    SOL_NULL_CHECK_GOTO(mdata->channels[2], error);
 
-#undef ADD_CHANNEL
 
     sol_iio_device_start_buffer(mdata->device);
 
     return true;
 
 error:
-    SOL_WRN("Could not create iio/gyroscope node. Failed to open IIO device %d",
```
edersondisouza: It would still be interesting to know which type of node failed. I think you can get this information via `enum iiodevice_types type`.

src/modules/flow/iio/nodes.c
```diff
@@ -44,24 +68,31 @@ reader_cb(void *data, struct sol_iio_device *device)
...
-    if (!b) goto error;
-
-    b = sol_iio_read_channel_value(mdata->channel_y, &out.y);
-    if (!b) goto error;
-
-    b = sol_iio_read_channel_value(mdata->channel_z, &out.z);
-    if (!b) goto error;
+    switch (mdata->type) {
+        case IIO_GYROSCOPE:
+        case IIO_MAGNETOMETER:
```
edersondisouza: Don't forget `IIO_ACCELEROMETER`.

edersondisouza:
Looks interesting =D
I would look at node type data, like used on flow/update or flow/ipm (there are plenty of other samples of usage on flow): this way, you don't need to store which output port should be used on each _node instance_, but only on each _node type_. You could store IIO type there, too.
Don't forget to unify 'close' and 'tick' methods, I think you'll only need one of each for all node types.
Tip: name these PRs that are not intended for merging with 'WIP:', so people know that they aren't on final state. They would complain, for instance, at commit message, etc.
I'm looking forward for next iteration =D

src/modules/flow/iio/nodes.c
```diff
@@ -25,15 +25,39 @@
...
+    IIO_MAGNETOMETER=2,
+    IIO_THERMOMETER=3,
+    IIO_PRESSURE_SENSOR=4,
+    IIO_COLOR=5,
+    IIO_ACCELEROMETER=6,
+    IIO_HUMIDITY_SENSOR=7,
+    IIO_ADC=8,
+    IIO_LIGHT_SENSOR=9,
+IIO_PROXIMITY=10,
+};
```
cabelitos: There's no need to enumerate all of them.

yongli3: I would like to use these enum items to re-use these common functions, I just uploaded a new version, please help to review

src/modules/flow/iio/nodes.c
```diff
@@ -25,43 +25,117 @@
...
+            b = sol_iio_read_channel_value(mdata->channels[0], &out.drange.val);
+            if (!b) goto error;
+            break;
+        default:
+            break;
+    }
 
-    sol_flow_send_direction_vector_packet(node,
-        SOL_FLOW_NODE_TYPE_IIO_GYROSCOPE__OUT__OUT, &out);
+    sol_flow_send_direction_vector_packet(node, type->out_port, &out.direction);
```
edersondisouza: Why are you always sending a direction-vector packet, even when port is of a different packet type?

src/modules/flow/iio/nodes.c
```diff
@@ -25,43 +25,117 @@
...
-    struct sol_direction_vector offset;
+    union {
+        struct sol_direction_vector scale_vector;
+        double scales[MAX_CHANNEL];
+    };
+
+    union {
+        struct sol_direction_vector offset_vector;
+        double offsets[MAX_CHANNEL];
+    };
```
edersondisouza: I don't get this unions. Why? I'd say you only need `offsets` and `scales`. The only place where you set `offset_vector` or `scale_vector` is on gyroscope code, but you override it nonetheless when add channels...

src/modules/flow/iio/nodes.c
edersondisouza: Well, I still don't think this union is useful at all: you may just use a different variable for each packet type will be sent on read.

src/modules/flow/iio/nodes.c
```diff
@@ -71,24 +145,82 @@ reader_cb(void *data, struct sol_iio_device *device)
...
+create_device_channels(struct iiodevice_data *mdata, int device_id, enum iiodevice_types type)
 {
     struct sol_iio_channel_config channel_config = SOL_IIO_CHANNEL_CONFIG_INIT;
-
+    int i = 0;
+ 
     mdata->device = sol_iio_open(device_id, &mdata->config);
     SOL_NULL_CHECK(mdata->device, false);
 
 #define ADD_CHANNEL(_axis) \
```
edersondisouza: Why insn't `i` a parameter to this macro? I've found those `i++` kinda enigmatic when using this macro.

src/modules/flow/iio/nodes.c
```diff
@@ -97,7 +229,7 @@ create_device_channels(struct gyroscope_data *mdata, int device_id)
     return true;
 
 error:
-    SOL_WRN("Could not create iio/gyroscope node. Failed to open IIO device %d",
+    SOL_WRN("Could not create iio node. Failed to open IIO device %d",
```
edersondisouza: Use `mdata->type` to print which type of node failed.

src/modules/flow/iio/nodes.c
```diff
@@ -356,32 +460,6 @@ struct temperature_data {
     bool use_device_default_offset : 1;
 };
```
edersondisouza: You forgot to set `mdevice->type` to all nodes but gyroscope.

src/modules/flow/iio/nodes.c
```diff
@@ -1571,7 +1452,7 @@ proximity_tick(struct sol_flow_node *node, void *data, uint16_t port, uint16_t c
         if (!sol_iio_device_trigger_now(mdata->device))
             goto error;
     } else {
-        proximity_reader_cb(node, mdata->device);
+        reader_cb(node, mdata->device);
     }
```
edersondisouza: Note that all `tick` and `close` methods are equal: you may define a `common_tick` and `common_close` and set nodes to use them on json.

edersondisouza:
It's becoming nice =D
Note that you can use node type data (same you used to define out port) to define type (the enum you created). With this, I think you'll end up with a common open function as well, needing to differentiate only channel addition and reading.

src/modules/flow/iio/iio.json
```diff
@@ -24,6 +24,15 @@
         "close": "gyroscope_close",
         "open": "gyroscope_open"
       },
+      "node_type": {
+        "access": [
+          "base"
+        ],
+        "data_type": "struct iio_node_type",
+        "extra_methods": {
+          "out_port": "SOL_FLOW_NODE_TYPE_IIO_GYROSCOPE__OUT__OUT"
```
edersondisouza: Here you could define some `type` or `iio_type` and set to `IIO_GYROSCOPE`. Then you could have a `common_open` for all node types.

src/modules/flow/iio/nodes.c
```diff
@@ -25,43 +25,117 @@
...
+    IIO_MAGNETOMETER=2,
+    IIO_THERMOMETER=3,
+    IIO_PRESSURE_SENSOR=4,
+    IIO_COLOR=5,
+    IIO_ACCELEROMETER=6,
+    IIO_HUMIDITY_SENSOR=7,
+    IIO_ADC=8,
+    IIO_LIGHT_SENSOR=9,
+    IIO_PROXIMITY=10,
+};
```
edersondisouza: What @cabelitos meant was that enums integer value are incremented automatically, so you'd need to define only first one. No need to do `IIO_MAGNETOMETER=2`, only `IIO_MAGNETOMETER` should suffice.

edersondisouza:
@yongli3 
I'm guessing that next iteration can be a real PR, don't you think? If so, besides updating issues pointed here, remember to create a new PR (instead of updating this one), with a proper message and touching all nodes. If you still believe that next iteration is not ready yet, don't worry, keep with WIP PRs ;)
I'll be on vacations next week, but you won't be left without review =D

src/modules/flow/iio/nodes.c
```diff
@@ -25,43 +25,117 @@
...
+         case IIO_LIGHT_SENSOR:
+         case IIO_PROXIMITY:
+            out.drange.min = mdata->out_range.min;
+            out.drange.max = mdata->out_range.max;
+            out.drange.step = mdata->out_range.step;
+            b = sol_iio_read_channel_value(mdata->channels[0], &out.drange.val);
+            if (!b) goto error;
+            break;
+        default:
+            break;
```
edersondisouza: You need to send an error packet here. If a new IIO node is added on future and someone forget to change here, it should be easy to get the error.

src/modules/flow/iio/nodes.c
```diff
@@ -71,24 +145,82 @@ reader_cb(void *data, struct sol_iio_device *device)
...
+        case IIO_LIGHT_SENSOR:
+            i = 0;
+            ADD_CHANNEL(in_illuminance0);
+            break;            
+        case IIO_PROXIMITY:
+             i = 0;
+            ADD_CHANNEL(in_proximity);
+            break;           
+        default:
+            break;
```
edersondisouza: You should error here: again, if someone add a new IIO node on future and miss changing here, it should get a nice error.

src/modules/flow/iio/nodes.c
```diff
@@ -25,43 +25,117 @@
...
-    if (!b) goto error;
-
-    b = sol_iio_read_channel_value(mdata->channel_y, &out.y);
-    if (!b) goto error;
-
-    b = sol_iio_read_channel_value(mdata->channel_z, &out.z);
-    if (!b) goto error;
+    type = (struct iio_node_type *)sol_flow_node_get_type(node);
+
+    switch (mdata->type) {
```
barbieri: why don't you break this into multiple functions, one for each kind of type (ie: read_direction_vector to be used by gyroscope, magnetometer and accelerometer). Then you set this to `type` via `extra_methods` on your json.

src/modules/flow/iio/nodes.c
barbieri: indeed, an explicit `ADD_CHANNEL(in_anglvel_x, 0); ADD_CHANNEL(in_anglvel_y, 1); ...` would be shorter and cleaner.

src/modules/flow/iio/nodes.c
```diff
@@ -106,14 +238,16 @@ create_device_channels(struct gyroscope_data *mdata, int device_id)
...
-    struct gyroscope_data *mdata = data;
+    struct iiodevice_data *mdata = data;
     const struct sol_flow_node_type_iio_gyroscope_options *opts;
     int device_id;
 
     SOL_FLOW_NODE_OPTIONS_SUB_API_CHECK(options, SOL_FLOW_NODE_TYPE_IIO_GYROSCOPE_OPTIONS_API_VERSION,
         -EINVAL);
     opts = (const struct sol_flow_node_type_iio_gyroscope_options *)options;
 
+    mdata->type = IIO_GYROSCOPE;
```
barbieri: this is not an instance member, but a class one. All gyroscopes share that, thus add this to `extra_methods` as you did for the out_port.

barbieri:
see what @edersondisouza said and consider moving more stuff to `extra_methods` and different functions instead of using the enumeration. In my understanding the enumeration should go away.

PR #2014, opened 2016-05-11
Uart API changes v6

cabelitos:
Changes since v5:
* Added commits 9af1fb1 and  5aab983
* Minor problems pointed by @barbieri

### DIFF
--- a/src/lib/common/include/sol-mainloop.h
+++ b/src/lib/common/include/sol-mainloop.h
@@ -64,8 +64,266 @@ extern "C" {
  * @li @ref Update
  * @li @ref Utils
  * @li @ref WorkerThread
+ * @li @ref patterns
  *
  * Please see the @ref authors page for contact details.
+ */
+
+/**
+ * @page patterns Design Patterns
+ * @tableofcontents
+ *
+ * This page will explain the common design patterns employed by Soletta.
+ *
+ * @section streams IO Streams
+ *
+ * Soletta has its own pattern for dealing with input/output streams. This
+ * section will describe how the soletta pattern is implemented and how to use it.
+ *
+ * The aim of this section is to guide the developer the to provide a unified API for all kind of streams, making
+ * it easier for a third party developer to migrate its application to any stream related IO, if necessary.
+ *
+ * The pattern is divided in three categories: The stream configuration, the write API and the read API.
+ *
+ * @subsection config Configuration
+ *
+ * The stream should configured using a struct during its creation, this struct should provide the following fields.
+ *
+ * <ul>
+ * <li> @c tx_size: The write buffer size - if @c 0 it means that is no limit (Data for this buffer is not necessary allocated).
+ * <li> @c rx_size: The read buffer size - If @c 0 a default value will be set (platform specific).
+ * <li> @c tx_cb: The write callback - It will be called when data was fully written.
+ *    <ul>
+ *    <li> Respecting the interface:
+ *    @code{.c}
+ *    void (*tx_cb)(void *user_data, my_stream_api_handle *handle, struct sol_blob *blob, int status);
+ *    @endcode
+ *    </ul>
+ * <li> @c rx_cb: The read callback - It will be called when there's data available to read.
+ *    <ul>
+ *    <li> Respecting the interface:
+ *    @code{.c}
+ *    ssize_t (*rx_cb)(void *user_data, my_stream_api_handle *handle, const struct sol_buffer *buf);
+ *    @endcode
+ *    </ul>
+ * <li> @c user_data: Data to @c rx_cb and @c tx_cb.
+ * </ul>
+ *
+ * Below there's a example of a configuration struct.
+ * @code{.c}
+ *
+ * //my_stream_api.h
+ *
+ * struct my_stream_api_config {
+ *    const void *user_data;
+ *    void (*tx_cb)(void *user_data, my_stream_api_handle *handle, struct sol_blob *blob, int status);
+ *    ssize_t (*rx_cb)()void *user_data, my_stream_api_handle *handle, const struct sol_buffer *buf);
+ *    size_t tx_size;
+ *    size_t rx_size;
+ * };
+ *
+ * //Code...
+ *
+ * //The config struct should be provided during the stream creation.
+ * struct my_stream_api_handle *my_stream_api_new(struct my_stream_api_config *config);
+ *
+ * //More code...
+ * @endcode
+ *
+ * @subsection write Writing
+ *
+ * Write operations must be async and the data that will be transfered must be
+ * provided as a pointer to @ref sol_blob. One should follow the API below:
+ * @code{.c}
+ * //returns 0 on success, -ENOSPC if tx_buffer is FULL or negative errno.
+ * int my_stream_api_write(struct my_stream_api_handle *handle, struct sol_blob *blob);
+ * @endcode
+ *
+ * Every blob requested to be written, must be queued in order to be sent. If there's no more space
+ * to queue more blobs (sum of all queued blobs >= @c tx_size) the function write should return @c -ENOSPC.
+ *
+ * Everytime a blob is fully written the @c tx_cb must be called in order to inform the user that
+ * the write operation was completed. The @c tx_cb should have the following signature:
+ *
+ * @code{.c}
+ * void (*tx_cb)(void *user_data, my_stream_api_handle *handle, struct sol_blob *blob, int status);
+ * @endcode
+ *
+ * One should warn the user that it's not necessary to use @ref sol_blob_unref(), because the stream
+ * API will already do that. The @c status variable should be 0 on success and @c -errno on error.
+ * The @c tx_cb can also be used to send blobs that were reject, because the tx buffer was full due
+ * @c tx_size restrictions.
+ *
+ * Below there's an example of a write implementation:
+ * @code{.c}
+ *
+ * //my_stream_api.c
+ *
+ * struct my_stream_api_handle {
+ *    Device *dev;
+ *    const void *user_data;
+ *    void (*tx_cb)(void *user_data, my_stream_api_handle *handle, struct sol_blob *blob, int status);
+ *    struct sol_ptr_vector pending_blobs;
+ *    size_t tx_size;
+ *    size_t pending_bytes;
+ * };
+ *
+ *
+ * static void
+ * _write(void *data)
+ * {
+ *    struct sol_blob *blob;
+ *    struct my_stream_api_handle *handle = data;
+ *    int status;
+ *
+ *    //Write the blob
+ *    blob = sol_ptr_vector_get_no_check(&handle->pending_blobs, 0);
+ *    status = my_device_write(handle->dev, blob->mem, blob->size);
+ *
+ *    // Data was written, inform the user
+ *    if (handle->tx_cb)
+ *       handle->tx_cb((void *)handle->user_data, handle, blob, status);
+ *    handle->pending_bytes -= blob->size;
+ *    sol_blob_unref(blob); //NOTE: that we unref the blob, not the user!
+ *
+ *    //Do we still have more data ?
+ *    sol_ptr_vector_del(&handle->pending_blobs, 0);
+ *    if (sol_ptr_vector_get_len(&handle->pending_blobs))
+ *       my_device_schedule_write(handle->dev, _write, handle);
+ * }
+ *
+ * SOL_API int
+ * my_stream_api_write(struct my_stream_api_handle *handle, struct sol_blob *blob)
+ * {
+ *    size_t total = handle->pending_bytes + blob->size;
+ *
+ *    //Tx_size was set and the total must not exceed tx_size
+ *    if (tx_size && total >= tx_size)
+ *       return -ENOSPC;
+ *
+ *    //Store the blob and ref it, since it will written later.
+ *    r = sol_ptr_vector_append(&handle->pending_blobs, blob);
+ *    SOL_INT_CHECK(r, < 0, r);
+ *    sol_blob_ref(blob);
+ *    handle->pending_bytes = total;
+ *
+ *    //Schedule the write operation
+ *    my_device_schedule_write(handle->dev, _write, handle);
+ *    return 0;
+ * }
+ * @endcode
+ *
+ * @subsection read Reading
+ *
+ * Since this is a stream, there's no direct way for a third party developer to request a read operation.
+ * In order to be able to read from the stream, the third party developer must provide the @c rx_cb durtion the
+ * stream configuration/creation. The @c rx_cb is provided during the stream creation, one must inform the user every time
+ * data is avaible to read.
+ * The @c rx_cb has the following signature:
+ *
+ * @code{.c}
+ * ssize_t (*rx_cb)(void *user_data, my_stream_api_handle *handle, const struct sol_buffer *buf);
+ * @endcode
+ *
+ * Note that the @c rx_cb returns a @c ssize_t, the returned value must be @c -errno if an error happened or the number
+ * of bytes consumed from @c buf (0 is valid). The consumed bytes will be removed from the @c buf.
+ *
+ * Below there's an example of the read implementation:
+ *
+ * @code{.c}
+ * //my_stream_api.c
+ *
+ * struct my_stream_api_handle {
+ *    Device *dev;
+ *    const void *user_data;
+ *    void (*tx_cb)(void *user_data, my_stream_api_handle *handle, struct sol_blob *blob, int status);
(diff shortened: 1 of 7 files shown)


glima:
It seems you missed fixing the bindings usage of the uart API (CI)

cabelitos:
@glima I temporally disable UART bindings. @nagineni Could you please fix that for us as soon this series is merged?

src/lib/common/include/sol-mainloop.h
```diff
@@ -64,11 +64,269 @@ extern "C" {
...
+ * Everytime a blob is fully written the @c tx_cb must be called in order to inform the user that
+ * the write operation was completed. The @c tx_cb should have the following signature:
+ *
+ * @code{.c}
+ * void (*tx_cb)(void *user_data, my_stream_api_handle *handle, struct sol_blob *blob, int status);
+ * @endcode
+ *
+ * One should warn the user that it's not necessary to use @ref sol_blob_unref(), because the stream
+ * API will already do that. The @c status variable should be 0 on success and @c -errno on error.
+ * The @c tx_cb can also be used to send blobs that were reject, because the tx buffer was full due
```
barbieri: this is not clear to me. send blobs that were rejected? Why? you said and the code is returning -ENOSPC

src/lib/common/include/sol-mainloop.h
```diff
@@ -64,11 +64,269 @@ extern "C" {
...
+ *
+ * static void
+ * _write(void *data)
+ * {
+ *    struct sol_blob *blob;
+ *    struct my_stream_api_handle *handle = data;
+ *    int status;
+ *
+ *    //Write the blob
+ *    blob = sol_ptr_vector_get_no_check(&handle->pending_blobs, 0);
```
barbieri: don't document code with `no_check`, it will lead to people misusing that.

src/lib/common/include/sol-mainloop.h
```diff
@@ -64,11 +64,269 @@ extern "C" {
...
+ *    status = my_device_write(handle->dev, blob->mem, blob->size);
```
barbieri: go with write(2) like `my_device_write()` signature, and show in the example that the device may return less than proposed bytes, so we document the pending behavior

src/lib/common/include/sol-mainloop.h
```diff
@@ -64,11 +64,269 @@ extern "C" {
...
+ *       my_device_schedule_write(handle->dev, _write, handle);
+ * }
+ *
+ * SOL_API int
+ * my_stream_api_write(struct my_stream_api_handle *handle, struct sol_blob *blob)
+ * {
+ *    size_t total = handle->pending_bytes + blob->size;
+ *
+ *    //Tx_size was set and the total must not exceed tx_size
+ *    if (tx_size && total >= tx_size)
```
barbieri: `handle->tx_size`, missing `handle->`

src/lib/common/include/sol-mainloop.h
```diff
@@ -64,11 +64,269 @@ extern "C" {
...
+ *
+ *    //Write the blob
+ *    blob = sol_ptr_vector_get_no_check(&handle->pending_blobs, 0);
+ *    status = my_device_write(handle->dev, blob->mem, blob->size);
+ *
+ *    // Data was written, inform the user
+ *    if (handle->tx_cb)
+ *       handle->tx_cb((void *)handle->user_data, handle, blob, status);
+ *    handle->pending_bytes -= blob->size;
+ *    sol_blob_unref(blob); //NOTE: that we unref the blob, not the user!
```
barbieri: this is dangerous to document: if from `tx_cb` the user deletes the handle, the code will crash. Document the callback and following unref is to be the last thing to call.

src/lib/common/include/sol-mainloop.h
```diff
@@ -64,11 +64,269 @@ extern "C" {
...
+ *    //Schedule the write operation
+ *    my_device_schedule_write(handle->dev, _write, handle);
+ *    return 0;
+ * }
+ * @endcode
+ *
+ * @subsection read Reading
+ *
+ * Since this is a stream, there's no direct way for a third party developer to request a read operation.
+ * In order to be able to read from the stream, the third party developer must provide the @c rx_cb durtion the
```
barbieri: typo, would be nice to review the text with more care: `durtion`. (enable flyspell-mode in emacs)

src/lib/common/include/sol-mainloop.h
```diff
@@ -64,11 +64,269 @@ extern "C" {
...
+ *
+ * The pattern is divided in three categories: The stream configuration, the write API and the read API.
+ *
+ * @subsection config Configuration
+ *
+ * The stream should configured using a struct during its creation, this struct should provide the following fields.
+ *
+ * <ul>
+ * <li> @c tx_size: The write buffer size - if @c 0 it means that is no limit (Data for this buffer is not necessary allocated).
+ * <li> @c rx_size: The read buffer size - If @c 0 a default value will be set (platform specific).
```
barbieri: say that usually this buffer is allocated and the user will be called back with at **most** that amount (upper limit, may be less if less data was available)

src/lib/common/include/sol-mainloop.h
```diff
@@ -64,11 +64,269 @@ extern "C" {
...
+ * stream configuration/creation. The @c rx_cb is provided during the stream creation, one must inform the user every time
+ * data is avaible to read.
+ * The @c rx_cb has the following signature:
+ *
+ * @code{.c}
+ * ssize_t (*rx_cb)(void *user_data, my_stream_api_handle *handle, const struct sol_buffer *buf);
+ * @endcode
+ *
+ * Note that the @c rx_cb returns a @c ssize_t, the returned value must be @c -errno if an error happened or the number
+ * of bytes consumed from @c buf (0 is valid). The consumed bytes will be removed from the @c buf.
```
barbieri: I'd say that the buffer shouldn't be modified and no further references to it or its data should be held after the callback returns.

src/lib/common/include/sol-mainloop.h
```diff
@@ -64,11 +64,269 @@ extern "C" {
...
+ * static bool
+ * _inform_user(void *data)
+ * {
+ *    struct my_stream_api_handle *handle = data;
+ *    ssize_t r;
+ *
+ *    //Inform the user
+ *    r = handle->rx_cb((void *)handle->user_data, handle, handle->rx);
+ *    //Remove the data.
+ *    if (r < 0)
```
barbieri: same as above, this can lead to crashes if the user closes/deletes the handle from the read callback.

Unlike above, the solution is not that simple since you need the return from user callback to change the handle if not deleted. Then add a `bool delete_me, in_use`. Before calling the user you `in_use = true`, then you check if `delete_me == true` and do the actual delayed deletion. On `delete` you check `in_use == true` and just mark `delete_me` and return.

src/lib/common/include/sol-mainloop.h
```diff
@@ -64,11 +64,269 @@ extern "C" {
...
+ *       SOL_ERR("Something wrong happened %zd", r);
+ *   else
+ *       sol_buffer_remove_data(&uart->rx, 0, r);
```
barbieri: suddenly `handle` becomes `uart`, review with care and try to validate if these code excerpts are valid, there is nothing worse than examples that do not work.

I'd actually suggest we use an actual C file in `src/samples/design-pattern/stream-reader.c` and use doxygen commands to include it here. You can use just an excerpt so the bookkeeping is not displayed. The sample must compile as usual.

src/lib/common/include/sol-mainloop.h
```diff
@@ -64,11 +64,269 @@ extern "C" {
...
+ * static void
+ * _can_read(void *data)
+ * {
+ *    struct my_stream_api_handle *handle = data;
+ *    size_r read;
+ *    ssize_t r;
+ *
+ *    //Append data to the buffer
+ *    read = my_device_read(handle->dev, sol_buffer_at_end(&uart->rx), uart->rx.capacity - uart->rx.used);
+ *    uart->rx.used += read;
```
barbieri: uart and handle mixed...

src/lib/common/include/sol-mainloop.h
```diff
@@ -64,11 +64,269 @@ extern "C" {
...
+ * @endcode
+ *
+ * @subsection implementations Implementations
+ *
+ * Soletta currently implements streams for the following modules:
+ * <ul>
+ * <li> @ref UART
+ *    <ul>
+ *    <li> sol_uart_write()
+ *    <li> @ref sol_uart_config
```
barbieri: I believe if you write `struct sol_uart_config`, that is going to auto-link. Please test.

src/lib/io/include/sol-uart.h
```diff
@@ -81,22 +90,57 @@ enum sol_uart_parity {
...
+    SOL_UART_STOP_BITS_ONE = 0, /**< Use one stop bit*/
+    SOL_UART_STOP_BITS_TWO /**< Use two stop bits */
 };
 
+/**
+ * @struct sol_uart_config
+ * @brief A configuration struct used to set the UART paramenters.
+ *
+ * @see sol_uart_open()
+ * @note UART follows the Soletta stream guidelines, which can be found here: @ref streams
```
barbieri: s/guidelines/design pattern/

src/lib/io/include/sol-uart.h
```diff
@@ -81,22 +90,57 @@ enum sol_uart_parity {
...
+     * @param uart The UART handle
+     * @param blob The blob that was written
+     * @param status 0 on success or negative errno on error
+     * @note There is no need to call sol_blob_unref().
+     * @see sol_uart_write()
+     */
+    void (*tx_cb)(void *data, struct sol_uart *uart, struct sol_blob *blob, int status);
+    const void *user_data; /**< User data to sol_uart_config::tx_cb() and  sol_uart_config::rx_cb() */
+    /**
+     * @brief The tx buffer max size. The value @c 0 means unlimited data.
```
barbieri: I'd add a note saying that the blobs will be referenced and kept alive, so usually no extra buffers are allocated here in most implementations. Also note that if set, writes will return `-ENOSPC` if pending bytes exceed this amount and user should control the flow until the `tx_cb` is called.

src/lib/io/include/sol-uart.h
```diff
@@ -81,22 +90,57 @@ enum sol_uart_parity {
...
+     */
+    size_t tx_size;
+    /**
+     * @brief The rx buffer max size. The value @c 0 means to use the default size. On Linux the default is 4096 and on Riot is 128
```
barbieri: I'd replace Riot with "Small OSes".

also move this out from `@brief` and grow the documentation to include mentions this will allocate the buffer and this will be the upper limit of calls to `rx_cb`.

barbieri: actually, the upper limit may be misleading, some libraries may round that value to offer better performance and the number may be slightly different. We can write this and say the actual available amount is `buf->used`

src/lib/io/include/sol-uart.h
```diff
@@ -224,21 +268,21 @@ struct sol_uart *sol_uart_open(const char *port_name, const struct sol_uart_conf
...
- * an error happen during the transmission
- * @param data the first parameter of tx_cb
- * @return true if transfer was started
- *
- * @note Caller should guarantee that tx buffer will not be freed until
- * callback is called.
- */
-bool sol_uart_write(struct sol_uart *uart, const uint8_t *tx, unsigned int length, void (*tx_cb)(void *data, struct sol_uart *uart, uint8_t *tx, int status), const void *data);
+ * @param blob The blob to be written
+ * @return 0 on success or negative errno on error
```
barbieri: I'd explicitly mention `-ENOSPC` if `tx_size` is non-zero.

src/lib/io/sol-uart-impl-linux.c
```diff
@@ -32,49 +33,126 @@
...
+read_timeout(void *data)
 {
     struct sol_uart *uart = data;
+    ssize_t r;
+    int err;
+
+    r = uart->rx_cb((void *)uart->user_data, uart, &uart->rx);
+    SOL_INT_CHECK_GOTO(r, < 0, err_exit);
+
+    err = sol_buffer_remove_data(&uart->rx, 0, r);
```
barbieri: as reviewed in the sample code, this will blow if user delete the uart from read callback, like in "I want to read N bytes and then I can close the uart".

src/lib/io/sol-uart-impl-linux.c
```diff
@@ -32,49 +33,126 @@
...
+uart_fd_handler_callback(void *data, int fd, uint32_t active_flags)
+{
+    struct sol_uart *uart = data;
+    struct sol_blob *blob;
+    int status = 0;
+    ssize_t r;
 
     if (active_flags & FD_ERROR_FLAGS) {
         SOL_ERR("Error flag was set on UART file descriptor %d.", fd);
         return true;
```
barbieri: I'm wondering if on these errors we should empty the pending blobs, reporting them all as failed and close the uart fd, removing the watches and failing any further write.

src/lib/io/sol-uart-impl-linux.c
```diff
@@ -172,72 +269,74 @@ sol_uart_open(const char *port_name, const struct sol_uart_config *config)
...
+        uart->rx_cb((void *)uart->user_data, uart, &uart->rx);
+
+    if (uart->fd_handler)
+        sol_fd_del(uart->fd_handler);
+
+    if (uart->read_timeout)
+        sol_timeout_del(uart->read_timeout);
+
+    SOL_PTR_VECTOR_FOREACH_IDX (&uart->pending_blobs, blob, i)
+        sol_blob_unref(blob);
```
barbieri: I'd say we need to callback and report they were not dispatched, like ECANCELLED

src/lib/common/include/sol-mainloop.h
```diff
@@ -64,11 +64,269 @@ extern "C" {
...
+ * @li @ref patterns
  *
  * Please see the @ref authors page for contact details.
  */
 
 /**
+ * @page patterns Design Patterns
+ * @tableofcontents
+ *
+ * This page will explain the common design patterns employed by Soletta.
```
barbieri: the documentation is good but this is the "implementation side", good for people that develop soletta itself or those that want to make compatible API.

There is a more important case that is missing: our users :-) That one can be in a different section (or move the current text to https://github.com/solettaproject/soletta/wiki/API-design-conventions and only document the user usage here. Cases to explicitly address:
 - simple `rx_cb` and `tx_cb` usages
 - delete from callbacks (say it's allowed)
 - reading partial contents (ie: return `< buf->used` from `rx_cb` is allowed), say what can happen if reaches `rx_size` (you can relate that to `buf->capacity`, say that in that case user must make room for more data, either by keeping a local copy or discard, in both cases return `> 0` to allow more data to be read.
 - saturating the write channel (ie: `returns -ENOSPC` with a `tx_size`), you should control your pipeline to stop producing data.

src/lib/common/include/sol-mainloop.h
cabelitos: ok

src/lib/common/include/sol-mainloop.h
cabelitos: oops

src/lib/io/include/sol-uart.h
cabelitos: Ok

src/lib/io/include/sol-uart.h
cabelitos: ok

src/lib/common/include/sol-mainloop.h
cabelitos: ok

src/lib/common/include/sol-mainloop.h
cabelitos: ok

src/lib/common/include/sol-mainloop.h
cabelitos: Indeed

nagineni:
@glima @cabelitos I will update UART JS bindings and enable back once these changes are merged.

src/lib/common/include/sol-mainloop.h
cabelitos: Ok

src/lib/common/include/sol-mainloop.h
cabelitos: Will do

src/lib/common/include/sol-mainloop.h
cabelitos: Ok

src/lib/common/include/sol-mainloop.h
cabelitos: ok

cabelitos:
#2028

PR #2015, opened 2016-05-11
HTTP: sol_http_progressive_response_feed() should receive a blob. v2

cabelitos:
Changes since v1:
* Added commits 02738b8 and 537ebe6
* Added config struct for the server sse counter part.

### DIFF
--- a/src/lib/comms/include/sol-http-server.h
+++ b/src/lib/comms/include/sol-http-server.h
@@ -278,6 +278,35 @@ sol_http_response_set_sse_headers(struct sol_http_response *response)
 }
 
 /**
+ * @brief Progressive server response configuration.
+ * @struct sol_http_server_progressive_config
+ * @see sol_http_server_send_progressive_response()
+ */
+struct sol_http_server_progressive_config {
+    /**
+     * @brief Callback used to inform that a struct @ref sol_blob was sent
+     * @param data The user data
+     * @param progressive The progressive response
+     * @param blob The blob that was transfered
+     * @note The blob will unref for you automatically
+     */
+    void (*on_feed)(void *data, const struct sol_http_progressive_response *progressive, struct sol_blob *blob);
+    /**
+     * @brief Callback used to inform that the client has closed the connection.
+     *
+     * @param data The user data
+     * @param progressive The progressive response
+     */
+    void (*on_close)(void *data, const struct sol_http_progressive_response *progressive);
+    const void *user_data; /*<< User data to @c on_feed and @c on_close*/
+    /**
+     * The output buffer size - 0 means unlimited data.
+     * @see sol_http_progressive_response_feed()
+     */
+    size_t max_bytes;
+};
+
+/**
  * @brief Send the response and keep connection alive to request given in the
  * callback registered on sol_http_server_register_handler().
  *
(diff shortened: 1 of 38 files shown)


cabelitos:
@edersondisouza Check the patch  02738b8, I think @barbieri would like that you use this kind of blob in your IPM PR

src/lib/comms/include/sol-http-server.h
```diff
@@ -278,6 +278,35 @@ sol_http_response_set_sse_headers(struct sol_http_response *response)
 }
 
 /**
+ * @brief Progressive server response configuration.
+ * @struct sol_http_server_progressive_config
+ * @see sol_http_server_send_progressive_response()
+ */
+struct sol_http_server_progressive_config {
+    /**
```
barbieri: missing api version thing

src/lib/comms/include/sol-http-server.h
```diff
@@ -278,6 +278,35 @@ sol_http_response_set_sse_headers(struct sol_http_response *response)
...
+     * @param data The user data
+     * @param progressive The progressive response
+     */
+    void (*on_close)(void *data, const struct sol_http_progressive_response *progressive);
+    const void *user_data; /*<< User data to @c on_feed and @c on_close*/
+    /**
+     * The output buffer size - 0 means unlimited data.
+     * @see sol_http_progressive_response_feed()
+     */
+    size_t max_bytes;
```
barbieri: do you want to rename this to `tx_size` now or later when the other PR documenting the design pattern goes in?

src/lib/comms/include/sol-http-server.h
cabelitos: Indeed. Adding.

src/lib/comms/include/sol-http-server.h
cabelitos: I will rename it now.

src/lib/comms/sol-http-client-impl-curl.c
```diff
@@ -574,7 +578,18 @@ perform_multi(CURL *curl, struct curl_slist *headers,
...
+        recv_size = RECV_DEFAULT_SIZE;
+
+        buf = malloc(recv_size);
+        SOL_NULL_CHECK_GOTO(buf, err_buf);
+
+        sol_buffer_init_flags(&connection->buffer, buf, recv_size,
+            SOL_BUFFER_FLAGS_NO_NUL_BYTE | SOL_BUFFER_FLAGS_FIXED_CAPACITY);
+    } else
+        sol_buffer_init_flags(&connection->buffer, NULL, 0,
+            SOL_BUFFER_FLAGS_NO_NUL_BYTE | SOL_BUFFER_FLAGS_DEFAULT);
```
barbieri: this behavior diverges a bit from what we wrote regarding `rx_size`, we better pick one approach and follow it. If we say `4096` then use that instead of allocating using a dynamic buffer.

barbieri:
I'd merge the two first patches related to blob asap and keep working on the sse part later

cabelitos:
Merging the first too commits.

src/lib/comms/sol-http-client-impl-curl.c
cabelitos: I guess I will set the behaviour just like tx_size. 0 means unlimited.

edersondisouza:
@cabelitos I'd say that you missed sol-ipm.c when changing blob. But now that you merged it, I'll fix sol-ipm. Regarding what @barbieri, I'd say that it was more of a confusion of mine there: ipm should use `SOL_BLOB_TYPE_NO_FREE_DATA`.

cabelitos:
@edersondisouza :[

cabelitos:
@2029

PR #2016, opened 2016-05-11
oic: Use sol-cert API to load pre shared keys - v2

otaviobp:
Replaces #1975

### DIFF
--- a/src/lib/common/include/sol-certificate.h
+++ b/src/lib/common/include/sol-certificate.h
@@ -65,6 +65,8 @@ struct sol_cert *sol_cert_load_from_file(const char *filename);
  */
 void sol_cert_unref(struct sol_cert *cert);
 
+#ifdef FEATURE_FILESYSTEM
+
 /**
  * @brief Get the full path to the certificate in the filesystem
  *
@@ -73,6 +75,8 @@ void sol_cert_unref(struct sol_cert *cert);
  * @return sol_cert object on success, NULL otherwise
  */
 const char *sol_cert_get_filename(const struct sol_cert *cert);
+
+#endif /*FEATURE_FILESYSTEM*/
 
 /**
  * @brief Get the certificate contents
--- a/src/lib/common/include/sol-certificate.h
+++ b/src/lib/common/include/sol-certificate.h
@@ -45,14 +45,14 @@ extern "C" {
 struct sol_cert;
 
 /**
- * @brief Load a certificate from a file
+ * @brief Load a certificate from a file name.
  *
- * @param file_name Path to the certificate file
+ * @param file_name Name of the certificate to be loaded.
  *
- * The path to the certificate file does not need to be absolute. The file
- * can also be in the default system folders ($SYSCONF/ssl/certs,
- * $SYSCONF/ssl/private, $SYSCONF/tls/certs and $SYSCONF/tls/private), or
- * in a directory specified by SSL_CERT_DIR
+ * The @a file_name can be the full path to the certificate file when file system
+ * is supported. If not absolute the file will be looked at the default system
+ * folders ($SYSCONF/ssl/certs, $SYSCONF/ssl/private, $SYSCONF/tls/certs and
+ * $SYSCONF/tls/private), or in a directory specified by SSL_CERT_DIR.
  *
  * @return sol_cert object on success, NULL otherwise
  */
@@ -86,6 +86,20 @@ const char *sol_cert_get_file_name(const struct sol_cert *cert);
  * @return the contents of the certificate @a cert on success, NULL otherwise
  */
 struct sol_blob *sol_cert_get_contents(const struct sol_cert *cert);
+
+/**
+ * @brief Write @a contents to @a cert.
+ *
+ * Certificates are always saved in user context.
+ *
+ * @param file_name The name of the certificate file. The certificate file name
+ *        must be relative. File name with full path is not supported.
+ * @param contents A blob containing the contents to be written to the
+ *        certificate.
+ *
+ * @return 0 on success or a negative error number on errors.
+ */
+int sol_cert_write_contents(const char *file_name, const struct sol_blob *contents);
 
 /**
  * @}
--- a/src/lib/common/include/sol-certificate.h
+++ b/src/lib/common/include/sol-certificate.h
@@ -47,7 +47,7 @@ struct sol_cert;
 /**
  * @brief Load a certificate from a file
  *
- * @param filename Path to the certificate file
+ * @param file_name Path to the certificate file
  *
  * The path to the certificate file does not need to be absolute. The file
  * can also be in the default system folders ($SYSCONF/ssl/certs,
@@ -56,7 +56,7 @@ struct sol_cert;
  *
  * @return sol_cert object on success, NULL otherwise
  */
-struct sol_cert *sol_cert_load_from_file(const char *filename);
+struct sol_cert *sol_cert_load_from_file(const char *file_name);
 
 /**
  * @brief Free the resources a sol_cert object and the object itself
@@ -74,7 +74,7 @@ void sol_cert_unref(struct sol_cert *cert);
  *
  * @return sol_cert object on success, NULL otherwise
  */
-const char *sol_cert_get_filename(const struct sol_cert *cert);
+const char *sol_cert_get_file_name(const struct sol_cert *cert);
 
 #endif /*FEATURE_FILESYSTEM*/
 
--- a/src/lib/common/sol-certificate-impl-linux.c
+++ b/src/lib/common/sol-certificate-impl-linux.c
@@ -31,6 +31,7 @@ struct sol_cert {
     int refcnt;
 
     char *filename;
+    char *basename;
 };
 
 static struct sol_ptr_vector storage = SOL_PTR_VECTOR_INIT;
@@ -83,6 +84,18 @@ find_cert(const char *filename, const char *const paths[])
     return NULL;
 }
 
+static char *
+get_basename(char *str)
+{
+    char *basename;
+
+    basename = strrchr(str, '/');
+    if (!basename || basename[1] == 0)
+        return str;
+
+    return basename + 1;
+}
+
 SOL_API struct sol_cert *
 sol_cert_load_from_file(const char *filename)
 {
@@ -93,10 +106,20 @@ sol_cert_load_from_file(const char *filename)
 
     SOL_NULL_CHECK(filename, NULL);
 
-    SOL_PTR_VECTOR_FOREACH_IDX (&storage, cert, idx) {
-        if (streq(cert->filename, filename)) {
-            cert->refcnt++;
-            return cert;
+    if (filename[0] == '/') {
+        SOL_PTR_VECTOR_FOREACH_IDX (&storage, cert, idx) {
+            if (streq(cert->filename, filename)) {
+                cert->refcnt++;
+                return cert;
+            }
+        }
+    } else {
+        SOL_PTR_VECTOR_FOREACH_IDX (&storage, cert, idx) {
+            SOL_WRN("%s %s", cert->basename, filename);
+            if (streq(cert->basename, filename)) {
+                cert->refcnt++;
+                return cert;
+            }
         }
     }
 
@@ -111,6 +134,7 @@ sol_cert_load_from_file(const char *filename)
 
     cert->refcnt++;
     cert->filename = path;
+    cert->basename = get_basename(path);
 
     r = sol_ptr_vector_append(&storage, cert);
     SOL_INT_CHECK_GOTO(r, != 0, insert_error);
--- a/src/lib/common/sol-certificate-impl-linux.c
+++ b/src/lib/common/sol-certificate-impl-linux.c
@@ -45,17 +45,35 @@ static const char *const search_paths[] = {
     NULL,
 };
 
+static int
+get_home_config_dir(struct sol_buffer *buffer)
+{
+    int r;
+
+    r = sol_util_get_user_config_dir(buffer);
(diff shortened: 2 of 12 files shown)


src/lib/common/include/sol-certificate.h
```diff
@@ -65,14 +65,18 @@ struct sol_cert *sol_cert_load_from_file(const char *filename);
  */
 void sol_cert_unref(struct sol_cert *cert);
 
+#ifdef FEATURE_FILESYSTEM
```
barbieri: noooo... see there is no `SOL_` prefix in that symbol, this is a hint this symbol is not exported in our headers. You need to see in the generated `sol-common-buildopts.h` (from `sol-common-buildopts.h.in`) if there is a symbol, if not add it.

src/lib/common/include/sol-certificate.h
```diff
@@ -84,6 +88,20 @@ const char *sol_cert_get_filename(const struct sol_cert *cert);
...
+ * Certificates are always saved in user context.
+ *
+ * @param file_name The name of the certificate file. The certificate file name
+ *        must be relative. File name with full path is not supported.
+ * @param contents A blob containing the contents to be written to the
+ *        certificate.
+ *
+ * @return 0 on success or a negative error number on errors.
+ */
+int sol_cert_write_contents(const char *file_name, const struct sol_blob *contents);
```
barbieri: remove the `const` from blob, otherwise you can't take a reference to it. If there is no need to keep a reference in any of the API, then using a `struct sol_str_slice contents` is enough.

src/lib/common/sol-certificate-impl-linux.c
```diff
@@ -43,36 +45,62 @@ static const char *const search_paths[] = {
...
+
+    return sol_buffer_append_printf(buffer, "/certs");
+}
+
+static inline bool
+is_cert(const char *file)
+{
+    struct stat st;
+
+    return stat(file, &st) == 0 && S_ISREG(st.st_mode) && st.st_mode & S_IRUSR;
```
barbieri: why checking mode is required? What if no user permission but group allows that? eventually it's better to use `access(file, R_OK)`?

src/lib/common/include/sol-certificate.h
```diff
@@ -45,18 +45,18 @@ extern "C" {
...
+ * @param file_name Name of the certificate to be loaded.
  *
- * The path to the certificate file does not need to be absolute. The file
- * can also be in the default system folders ($SYSCONF/ssl/certs,
- * $SYSCONF/ssl/private, $SYSCONF/tls/certs and $SYSCONF/tls/private), or
- * in a directory specified by SSL_CERT_DIR
+ * The @a file_name can be the full path to the certificate file when file system
+ * is supported. If not absolute the file will be looked at the default system
+ * folders ($SYSCONF/ssl/certs, $SYSCONF/ssl/private, $SYSCONF/tls/certs and
+ * $SYSCONF/tls/private), or in a directory specified by SSL_CERT_DIR.
```
barbieri: no mentions to the home user

src/lib/common/sol-certificate-impl-linux.c
```diff
@@ -31,6 +32,7 @@ struct sol_cert {
     int refcnt;
 
     char *filename;
+    char *basename;
```
barbieri: to make it clear, add `const` to basename, so we're clear that shouldn't be touched or freed, since it's inside filename

src/lib/comms/sol-oic-security.c
```diff
@@ -53,34 +51,51 @@ struct cred_item {
...
 struct sol_socket *sol_coap_server_get_socket(const struct sol_coap_server *server);
 
 static ssize_t
 creds_get_psk(const void *data, struct sol_str_slice id,
     char *psk, size_t psk_len)
 {
-    const struct creds *creds = data;
-    struct cred_item *iter;
-    uint16_t idx;
+    SOL_BUFFER_DECLARE_STATIC(buf, PATH_MAX);
```
barbieri: instead of `PATH_MAX`, use `NAME_MAX` since this is just the base name, not the full path. It will help keep stack smaller and remember this will be enabled on the small OSes as well.

src/lib/comms/sol-oic-security.c
```diff
@@ -53,34 +51,51 @@ struct cred_item {
...
+    struct sol_blob *contents;
+    int r;
 
     SOL_DBG("Looking for PSK with ID=%.*s", (int)id.len, id.data);
 
-    SOL_VECTOR_FOREACH_IDX (&creds->items, iter, idx) {
-        if (sol_str_slice_eq(id, iter->id.slice)) {
-            if (iter->id.slice.len > psk_len)
-                return -ENOBUFS;
+    r = sol_buffer_append_printf(&buf, "%s", OIC_CRED_FILE_PREFIX);
```
barbieri: avoid printf when it's possible, use `sol_buffer_append_slice()` and use the `sol_str_slice_from_str()`, or use the `OIC_CRED_FILE_PREFIX` as as slice declared using `SOL_STR_SLICE_LITERAL()`.

src/lib/comms/sol-oic-security.c
```diff
@@ -53,34 +51,51 @@ struct cred_item {
...
 {
-    const struct creds *creds = data;
-    struct cred_item *iter;
-    uint16_t idx;
+    SOL_BUFFER_DECLARE_STATIC(buf, PATH_MAX);
+    struct sol_cert *cert;
+    struct sol_blob *contents;
+    int r;
 
     SOL_DBG("Looking for PSK with ID=%.*s", (int)id.len, id.data);
```
barbieri: `SOL_STR_SLICE_PRINT()` should be used (I know you didn't change this, but since you're touching the function, improve that)

src/lib/comms/sol-oic-security.c
```diff
@@ -53,34 +51,51 @@ struct cred_item {
...
 
-    SOL_VECTOR_FOREACH_IDX (&creds->items, iter, idx) {
-        if (sol_str_slice_eq(id, iter->id.slice)) {
-            if (iter->id.slice.len > psk_len)
-                return -ENOBUFS;
+    r = sol_buffer_append_printf(&buf, "%s", OIC_CRED_FILE_PREFIX);
+    SOL_INT_CHECK(r, < 0, -ENOBUFS);
+    r = sol_buffer_append_as_base64(&buf, id, SOL_BASE64_MAP);
```
barbieri: could you clarify why? I'm wondering if we should do an escape for filenames, like we did for http/url, this way most of the chars would be preserved and we have a faster process by doing memcpy() instead of the conversion to base64. Usually the convention is to do like sytemd-escape (https://www.freedesktop.org/software/systemd/man/systemd-escape.html). See https://github.com/solettaproject/soletta/blob/master/src/lib/comms/sol-http-common.c#L140, we could use a variation of that.

src/lib/comms/sol-oic-security.c
```diff
@@ -53,34 +51,51 @@ struct cred_item {
...
+    SOL_INT_CHECK(r, < 0, -ENOBUFS);
+    r = sol_buffer_append_printf(&buf, "%s", OIC_CRED_FILE_SUFIX);
+    SOL_INT_CHECK(r, < 0, -ENOBUFS);
+
+    cert = sol_cert_load_from_file(buf.data);
```
barbieri: this wouldn't work for small os like zephyr or riot. Actually `buf` can be handled as an `id`, so what about: `sol_cert_load_from_id(scope, buf.data)`, with the linux implementation using id as file name and using `sol_cert_load_from_file()`. Scope can be something like "user", "system", "both" (enum, string or we can use different functions).

src/lib/comms/sol-oic-security.c
```diff
@@ -53,34 +51,51 @@ struct cred_item {
...
 
-            memcpy(psk, iter->psk.data, iter->id.slice.len);
-            return (ssize_t)iter->psk.slice.len;
-        }
+    sol_buffer_reset(&buf);
+
+    contents = sol_cert_get_contents(cert);
+    sol_cert_unref(cert);
+
+    if (contents->size < PSK_BASE64_LEN) {
```
barbieri: `SOL_WRN()` explaining what's wrong?

src/lib/comms/sol-oic-security.c
```diff
@@ -53,34 +51,51 @@ struct cred_item {
...
-    return -ENOENT;
+    r = sol_buffer_append_from_base64(&buf,
+        SOL_STR_SLICE_STR(contents->mem, PSK_BASE64_LEN), SOL_BASE64_MAP);
+    sol_blob_unref(contents);
+    SOL_INT_CHECK(r, < 0, -ENOBUFS);
+
+    if (buf.used > psk_len)
+        return -ENOBUFS;
+
+    memcpy(psk, buf.data, buf.used);
```
barbieri: why are you storing as base64 and decoding it here?

Also, please keep the blob as further as possible to reduce memory pressure. On Linux, the blob will `mmap()`, on small OSes they can point directly to the storage in flash or RAM. Duplicating is bad.

src/lib/comms/sol-socket-dtls.h
```diff
@@ -37,7 +37,7 @@ enum sol_socket_dtls_cipher {
 struct sol_socket_dtls_credential_cb {
     const void *data;
 
-    void *(*init)(const void *data);
+    int (*init)(const void *data);
```
barbieri: better take the opportunity to fix the signature to include a size, or `struct sol_str_slice` (slices can contain binary data, including \0, as they specify the size).

src/shared/sol-file-reader.c
```diff
@@ -42,6 +42,9 @@ sol_file_reader_open(const char *filename)
     struct sol_file_reader *fr;
     int fd;
 
+    if (!filename || *filename == '\0')
```
barbieri: why not `SOL_NULL_CHECK()`? I'd do logging here as well.

src/shared/sol-file-reader.c
```diff
@@ -42,6 +42,9 @@ sol_file_reader_open(const char *filename)
...
+        return NULL;
+
     fd = open(filename, O_RDONLY | O_CLOEXEC);
     if (fd < 0)
```
barbieri: also convert this to `SOL_INT_CHECK(fd, < 0, NULL)`

src/shared/sol-file-reader.c
```diff
@@ -106,6 +111,11 @@ sol_file_reader_close(struct sol_file_reader *fr)
 struct sol_str_slice
 sol_file_reader_get_all(const struct sol_file_reader *fr)
 {
+    SOL_NULL_CHECK(fr, (struct sol_str_slice) {
```
barbieri: `SOL_STR_SLICE_STR(NULL, 0)` would be shorter.

src/shared/sol-util-file.c
```diff
@@ -593,3 +593,73 @@ sol_util_busy_wait_file(const char *path, uint64_t nanoseconds)
...
+}
+
+SOL_API int
+sol_util_get_user_config_dir(struct sol_buffer *buffer)
+{
+    const char *dir;
+
+    dir = getenv("XDG_CONFIG_HOME");
+    if (dir)
+        return sol_buffer_append_printf(buffer, "%s", dir);
```
barbieri: missing `~/.config/${APPNAME}`, where `APPNAME` should be what `sol_args()[0]` provides, or `soletta` if `sol_argc() == 0`

src/shared/sol-util-file.c
```diff
@@ -593,3 +593,73 @@ sol_util_busy_wait_file(const char *path, uint64_t nanoseconds)
...
+
+    dir = getenv("HOME");
+    if (!dir)
+        return -EINVAL;
+
+    return sol_buffer_append_printf(buffer, "%s/%s", dir, ".config/soletta/");
```
barbieri: same here, use `APPNAME`

src/lib/common/include/sol-certificate.h
otaviobp: I used same approach used in sol-http-server.h. I'll fix it in both.

src/lib/common/include/sol-certificate.h
otaviobp: no reference is needed. I just used blob for compatibility, but I don't mind using str_slice

src/lib/common/sol-certificate-impl-linux.c
otaviobp: yes. access is better

src/lib/comms/sol-oic-security.c
otaviobp: Actually I am using base64 because it is the way ids/psks were persisted in json files. But a "escape for filenames" would solve it here. And I could save the bytes without any encoding to the file too. not a problem.

src/lib/comms/sol-oic-security.c
otaviobp: ok.

src/lib/comms/sol-oic-security.c
otaviobp: I could avoid base64 here.
But there is no way to avoid copying the blob, because the psk buffer is allocated by tinydtls.

src/lib/comms/sol-socket-dtls.h
otaviobp: I don't get it. We don't need a size and the data here is a pointer to user data. It is initializing a psk/key store. init and clear are not being used anymore, but I kept them because they make sense.

barbieri: oh, got it wrong, understood it was the psk data to initialize.

PR #2017, opened 2016-05-11
Flow option uint

edersondisouza:
Currently, we are using `int` data_type on flow node options for some options whose range is naturally unsigned. This PR presents a `uint` data_type - mapped to `uint32_t`.
A bonus is that it also checks range, so no need for node types check `if (opts.x > 0)`.
In this PR, I've changed options of two node types: IPM and Form. This is more of a sample of usage than an extensive change.

### DIFF
--- a/data/schemas/node-type-genspec.schema
+++ b/data/schemas/node-type-genspec.schema
@@ -9,6 +9,13 @@
       ]
     },
 
+    "uint_or_limit": {
+      "oneOf": [
+        { "type": "integer" },
+        { "enum": [ "UINT32_MAX" ] }
+      ]
+    },
+
     "float_or_limit": {
       "oneOf": [
         { "type": "number" },
@@ -67,7 +74,8 @@
         "int",
         "irange-spec",
         "rgb",
-        "string"
+        "string",
+        "uint"
       ]
     },
 
@@ -322,6 +330,13 @@
       }
     },
 
+    "options_members_uint": {
+      "properties": {
+        "data_type": { "enum": [ "uint" ] },
+        "default": { "$ref": "#/definitions/uint_or_limit" }
+      }
+    },
+
     "options_members_by_type": {
       "allOf": [
         { "$ref": "#/definitions/options_members_common" },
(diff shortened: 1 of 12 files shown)


data/schemas/node-type-genspec.schema
```diff
@@ -9,6 +9,13 @@
       ]
     },
 
+    "uint_or_limit": {
+      "oneOf": [
+        { "type": "integer" },
```
barbieri: is there any way to say ">=0"?

src/lib/flow/sol-flow-node-options.c
```diff
@@ -511,6 +513,31 @@ int_parse(const char *value, struct sol_flow_node_named_options_member *m)
     return 0;
 }
 
+static int
+uint_parse(const char *value, struct sol_flow_node_named_options_member *m)
+{
+    char *endptr;
+    int64_t v;
+
+    if (!strcmp(UINT_MAX_STR, value)) {
```
barbieri: streq()

data/schemas/node-type-genspec.schema
edersondisouza: On 11-05-2016 17:12, Gustavo Sverzut Barbieri wrote:
> In data/schemas/node-type-genspec.schema
> <https://github.com/solettaproject/soletta/pull/2017#discussion_r62917901>:
>
>> @@ -9,6 +9,13 @@
>>        ]
>>      },
>>
>> +    "uint_or_limit": {
>> +      "oneOf": [
>> +        { "type": "integer" },
>
> is there any way to say ">=0"?

Hmmm... now that you told about it, I think there's a minimum attribute. 
Will check it.

barbieri:
+1, but please also add `uint` to timer, gpio, pwm, wallclock... maybe a grep with `o[a-z]*->[a-zA-Z0-9]\+ *<` in src/modules/flow are likely to show some spots to convert.

edersondisouza:
On 11-05-2016 17:22, Gustavo Sverzut Barbieri wrote:
> +1, but please also add |uint| to timer, gpio, pwm, wallclock... maybe a
> grep with |o[a-z]*->[a-zA-Z0-9]\+ *<| in src/modules/flow are likely to
> show some spots to convert.

Ack.

bdilly:
+0

One more type of option, more complexity for our flows (from users pov)

If we want to avoid issues with option values out of range, we could go further and properly define limits when declaring options and check it as part of framework.

For example a timer doesn't work with interval < 1, thingspeak with interval < 1000...

edersondisouza:
I'd say that for users, there's not that much complexity - `uint` should be a fairly simple type to understand. And it simplifies node type development, that don't need to worry about invalid negative options. I'll send a v2 with wider changes on node types - it is more clear that we are simplifying flows.

edersondisouza:
v2 on #2026

PR #2018, opened 2016-05-11
ipm: Use SOL_BLOB_TYPE_NO_FREE_DATA on sent blob wrapper

edersondisouza:

It was missed on blob type changes.


barbieri:
+1, this is a build fix, no need to PR

PR #2020, opened 2016-05-11
oic: Add option to make OIC server compatible with OIC 1.0 spec

otaviobp:
Some changes were made in 1.1 spec in server payload and those changes were
already implemented in Soletta. Adding now an option to force OIC server
to create payloads as specified in OIC 1.0 spec. This is important
because clients that implements 1.0 spec can't communicate with 1.1
servers. Note that 1.1 clients support both, 1.0 and 1.1 servers.

### DIFF
--- a/src/lib/comms/Makefile
+++ b/src/lib/comms/Makefile
@@ -92,6 +92,10 @@ obj-networking-$(OIC)-extra-cflags += \
     -DOIC_SUPPORT_URL=$(OIC_SERVER_SUPPORT_URL) \
     -DOIC_DEVICE_NAME=$(OIC_SERVER_DEVICE_NAME)
 
+ifeq (y,$(OIC_SERVER_COMPAT_1_0))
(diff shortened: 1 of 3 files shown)


src/lib/comms/Makefile
```diff
@@ -92,6 +92,10 @@ obj-networking-$(OIC)-extra-cflags += \
     -DOIC_SUPPORT_URL=$(OIC_SERVER_SUPPORT_URL) \
     -DOIC_DEVICE_NAME=$(OIC_SERVER_DEVICE_NAME)
 
+ifeq (y,$(OIC_SERVER_COMPAT_1_0))
+COMMON_CFLAGS += -DOIC_SERVER_COMPAT_1_0
```
barbieri: is this needed? AFAIU this goes into sol-generated.h or something like that, which is auto-included in all builds

otaviobp: That's right. we don't need that.

otaviobp:
fixed and merged

PR #2021, opened 2016-05-11
docs: include C API examples

bdilly:


barbieri:
Can we auto generate this list with "find" so we don't need to maintain it manually?

bdilly:
Do you mean using a src/lib/common/include/sol-mainloop.h.in to generate a final sol-mainloop.h when we build?

barbieri:
It seems we have docs/doxygen/examples.dox that is empty. We could move the Doxyfile to include build/stage/docs/doxygen/examples.dox (or something like that) from:

mkdir -p build/stage/docs/doxygen
cp docs/doxygen/examples-header.dox build/stage/docs/doxygen/examples.dox
find src/samples -type f -printf "@example %h/%f\n" >> build/stage/docs/doxygen/examples.dox

bdilly:
merged with proposed changes

PR #2022, opened 2016-05-11
pin-mux: Pass the full gpio config to pin-mux

mbelluzzo:
Since 'drive' mode is back as part of our GPIO API (and config)
we can use it again to set the driver mode during pin mux execution.


mbelluzzo:
Merged, thanks!

Issue #1720, opened 2016-03-23
API Changes: COAP


cabelitos:
#### Task Description
- [x] Remove typedefs
- [x] Replace all `CONTENTTYPE`s with `CONTENT_TYPE`. Example `SOL_COAP_CONTENTTYPE_NONE` should be `SOL_COAP_CONTENT_TYPE_NONE`
- [x] `LINKFORMAT` -> `LINK_FORMAT`
- [x] `MAKE_RSPCODE` -> `sol_coap_make_response_code()`
- [x] Replace `RSPCODE` with `RESPONSE_CODE`
- [x] `responsecode` -> `response_code`
- [x] `SOL_COAP_TYPE_NONCON` -> `SOL_COAP_TYPE_NON_CON`
- [x] Use `size_t` for sizes instead of `uint16_t` - Examples - `sol_coap_add_option()` argument `len` should be a size_t
- [x] `sol_coap_header_set_ver()` -> `sol_coap_header_set_version()`
- [x] `sol_coap_header_get_ver()` -> `sol_coap_header_get_version()`
- [x] `sol_coap_packet_notification_new()` -> `sol_coap_packet_new_notification()`
- [x] `sol_coap_packet_request_new()` -> `sol_coap_packet_new_request()`
- [x] `sol_coap_unobserve_server()` -> `sol_coap_unobserve_by_token()`
- [x] `sol_coap_uri_path_to_buf()` -> `sol_coap_path_to_buffer()`
- [ ] Remove `sol_coap_secure_server_new()` and `sol_coap_server_new()` should receive a `bool` argument that flags if the server is secure or not

#### Dependencies
None.

otaviobp:
most of tasks fixed in #2024 
Assigning to @ceolin because he is working in last issue

cabelitos:
We're leaving the sizes as `uint16_t`
The last issue (`sol_coap_secure_server`) is fixed in https://github.com/solettaproject/soletta/pull/2185

cabelitos:
merged

PR #2024, opened 2016-05-12
CoAP API style changes

otaviobp:
Fix issue #1720

### DIFF
--- a/bindings/nodejs/src/functions/oic-client-resource-ops.cc
+++ b/bindings/nodejs/src/functions/oic-client-resource-ops.cc
@@ -44,7 +44,7 @@ public:
     bool isOneShot;
 };
 
-static void requestAnswered(void *data, sol_coap_responsecode_t code,
+static void requestAnswered(void *data, sol_coap_response_code_t code,
     struct sol_oic_client *client,
     const struct sol_network_link_addr *address,
     const struct sol_oic_map_reader *response) {
@@ -160,7 +160,7 @@ typedef int (*ObserveAPI)(
     struct sol_oic_resource *res,
     void(*callback)(
         void *data,
-        sol_coap_responsecode_t response_code,
+        sol_coap_response_code_t response_code,
         struct sol_oic_client *cli,
         const struct sol_network_link_addr *addr,
         const struct sol_oic_map_reader *repr_map),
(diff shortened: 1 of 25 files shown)


src/lib/comms/sol-coap.c
```diff
@@ -1072,7 +1072,7 @@ sol_coap_packet_new_request(sol_coap_method_t method, sol_coap_msgtype_t type)
 }
 
 SOL_API int
-sol_coap_add_option(struct sol_coap_packet *pkt, uint16_t code, const void *value, uint16_t len)
+sol_coap_add_option(struct sol_coap_packet *pkt, uint16_t code, const void *value, size_t len)
```
vcgomes: The use of `uint16_t` documents that the option format only allows an option length that can be represented in 16 bits (the actual max would be 65804 IIRC, but I didn't see much value in supporting that).

If you really want to change this, also update `coap_option_encode()` and `coap_parse_option()` to take this into account.

src/lib/comms/include/sol-coap.h
```diff
@@ -78,7 +78,7 @@ extern "C" {
  *
  * Refer to RFC 7252, section 12.2 for more information.
  */
-typedef enum {
+enum sol_coap_option_num_t {
```
vcgomes: Please remove the `_t` suffix as well, I don't think it's used for enums.

glima: ugh, -1 as well. if you were really following our patterns you'd remove the typedef (no _t)

src/lib/comms/sol-coap.c
glima: I also prefer to document on .h's for those with 16bit limits then to change the type

src/lib/comms/include/sol-coap.h
glima: And now I see that the enum part was done :)

bindings/nodejs/src/functions/oic-client-resource-ops.cc
```diff
@@ -44,7 +44,7 @@ class OicClientRequest: public OicCallbackData {
     bool isOneShot;
 };
 
-static void requestAnswered(void *data, sol_coap_responsecode_t code,
+static void requestAnswered(void *data, sol_coap_response_code_t code,
```
cabelitos: I would also remove the `_t` suffix.

src/lib/comms/sol-coap.c
barbieri: keep `uint16_t`

otaviobp:
merged with fixes

PR #2025, opened 2016-05-12
Export SOL_FEATURE_FILESYSTEM flag

otaviobp:


glima:
Ack

src/lib/common/include/sol-certificate.h
```diff
@@ -74,6 +77,8 @@ void sol_cert_unref(struct sol_cert *cert);
  */
 const char *sol_cert_get_filename(const struct sol_cert *cert);
 
+#endif /*SOL_FEATURE_FILESYSTEM*/
```
edersondisouza: It may be a noob question, but shouldn't you also ifdef implementation?

otaviobp: We only have a implementation for linux so far, that is using filesystem for all methods.

edersondisouza: On 12-05-2016 13:31, Otávio wrote:
> In src/lib/common/include/sol-certificate.h
> <https://github.com/solettaproject/soletta/pull/2025#discussion_r63052225>:
>
>> @@ -74,6 +77,8 @@ void sol_cert_unref(struct sol_cert *cert);
>>   */
>>  const char *sol_cert_get_filename(const struct sol_cert *cert);
>>
>> +#endif /*SOL_FEATURE_FILESYSTEM*/
>
> We only have a implementation for linux so far, that is using filesystem
> for all methods.

Fair then =D

otaviobp:
merged

PR #2026, opened 2016-05-12
Flow option uint v2

edersondisouza:
v2: 
Many more places were changed to use new type.
Small fix for PWM node type

Currently, we are using int data_type on flow node options for some options whose range is naturally unsigned. This PR presents a uint data_type - mapped to uint32_t.
A bonus is that it also checks range, so no need for node types check if (opts.x > 0).

### DIFF
--- a/data/schemas/node-type-genspec.schema
+++ b/data/schemas/node-type-genspec.schema
@@ -9,6 +9,16 @@
       ]
     },
 
+    "uint_or_limit": {
+      "oneOf": [
+        {
+          "type": "integer",
+          "minimum": 0
+        },
+        { "enum": [ "UINT32_MAX" ] }
+      ]
+    },
+
     "float_or_limit": {
       "oneOf": [
         { "type": "number" },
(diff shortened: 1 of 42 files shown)


edersondisouza:
Checkstyle is complaining on code I didn't touch =/
(But touched the files)

bdilly:
We don't have this concept of unsigned variables on flow (default data types nor options).
So you're adding more complexity (a new type of option) and won't solve any issue.

I still believe it's not the way to go. If the idea is to solve issues related to values out of range in node types, you should support declaring ranges when defining the options. As I said, < 0 is only one of many different checks node types do.

barbieri:
while @bdilly has a point, currently we fail in other areas as well: like timeout takes uint32_t, if we use int32_t>0, then we're half of the allowed values for timeout :-/

so range checking/spec could be used, just not sure about the other problem.

and not to forget that some ranges are dynamic, like IPM maximum id.

bdilly:
closed after discussion with ederson

PR #2029, opened 2016-05-12
HTTP: sol_http_progressive_response_feed() should receive a blob. v3

cabelitos:

CHanges since v2:
* Minor fixes pointed by @barbieri 
* The user is now able to delete the `sol_http_client_connection` from the callbacks `send_cb` `response_cb` `recv_cb`

### DIFF
--- a/src/lib/comms/sol-http-client-impl-curl.c
+++ b/src/lib/comms/sol-http-client-impl-curl.c
@@ -70,6 +70,7 @@ struct sol_http_client_connection {
     CURL *curl;
     struct curl_slist *headers;
     struct curl_httppost *formpost;
+    struct sol_timeout *del_timeout;
     struct sol_buffer buffer;
     struct sol_vector watches;
     struct sol_http_params response_params;
@@ -77,7 +78,21 @@ struct sol_http_client_connection {
     const void *data;
 
     bool error;
+    bool in_use;
 };
+
+static void destroy_connection(struct sol_http_client_connection *c);
+
+static bool
+schedule_del(void *data)
+{
+    struct sol_http_client_connection *c = data;
+
+    c->del_timeout = NULL;
+    sol_ptr_vector_remove(&global.connections, c);
+    destroy_connection(c);
+    return false;
+}
 
 static void
 destroy_connection(struct sol_http_client_connection *c)
@@ -98,6 +113,9 @@ destroy_connection(struct sol_http_client_connection *c)
         sol_fd_del(cwatch->watch);
 
     sol_vector_clear(&c->watches);
+
+    if (c->del_timeout)
+        sol_timeout_del(c->del_timeout);
 
     free(c);
     sol_http_client_shutdown_lazy();
@@ -170,13 +188,19 @@ call_connection_finish_cb(struct sol_http_client_connection *connection)
 
     response->param = connection->response_params;
 
-    if (connection->interface.response_cb)
+    if (connection->interface.response_cb) {
+        connection->in_use = true;
         connection->interface.response_cb((void *)connection->data, connection, response);
+        connection->in_use = false;
+    }
     goto end;
 
 err:
-    if (connection->interface.response_cb)
+    if (connection->interface.response_cb) {
+        connection->in_use = true;
         connection->interface.response_cb((void *)connection->data, connection, NULL);
+        connection->in_use = false;
+    }
 end:
     sol_buffer_fini(&response->content);
     destroy_connection(connection);
@@ -199,7 +223,9 @@ write_cb(char *data, size_t size, size_t nmemb, void *connp)
     if (connection->interface.recv_cb) {
         ssize_t ret;
 
+        connection->in_use = true;
         ret = connection->interface.recv_cb((void *)connection->data, connection, &connection->buffer);
+        connection->in_use = false;
         if (ret < 0)
             return 0;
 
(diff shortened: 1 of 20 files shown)


cabelitos:
style-check will complain, however I did not touch those code sections

src/lib/comms/sol-http-server-impl-microhttpd.c
```diff
@@ -458,39 +462,55 @@ build_mhd_response(const struct sol_http_response *response, time_t last_modifie
...
 {
+    struct sol_blob *blob;
+    uint16_t i;
     struct sol_http_progressive_response *progressive = data;
 
-    if (progressive->on_del)
-        progressive->on_del((void *)progressive->cb_data, progressive);
+    SOL_PTR_VECTOR_FOREACH_IDX (&progressive->pending_blobs, blob, i) {
+        if (progressive->on_feed)
+            progressive->on_feed((void *)progressive->cb_data, progressive, blob, -ECANCELED);
```
edersondisouza: What if user calls `sol_http_progressive_response_feed` from this callback?

src/lib/comms/sol-http-server-impl-microhttpd.c
```diff
@@ -1307,21 +1335,33 @@ sol_http_progressive_response_del(struct sol_http_progressive_response *progress
...
+    SOL_INT_CHECK(ret, < 0, ret);
+
+    sol_blob_ref(blob);
+
     if (progressive->request->suspended) {
         progressive->request->suspended = false;
         MHD_resume_connection(progressive->request->connection);
     }
 
+    progressive->accumulated_bytes += total;
```
edersondisouza: You add `progressive->accumulated_bytes` and `blob->size` to get `total`. Then you add `total` to `progressive->accumulated_bytes`... is it right? Shouldn't `progressive->accumulated_bytes` just receive `total`? The way you do, if `progressive->accumulated_bytes`is `10` and `blob->size` is `5`, `total` will be `15`. And after this line, `progressive->accumulated_bytes` will be `25`...

src/samples/http/server-sse.c
```diff
@@ -72,14 +72,28 @@ static struct sol_ptr_vector responses = SOL_PTR_VECTOR_INIT;
...
+};
+
+static struct sol_blob sse_suffix = {
+    .type = &SOL_BLOB_TYPE_NO_FREE,
+    .parent = NULL,
+    .mem = (void *)"\n\n",
+    .size = sizeof("\n\n") - 1,
+    .refcnt = 1
+};
+
```
edersondisouza: I saw this code some times on this patch... couldn't it be Soletta responsibility to deal with these prefix and suffix stuff?

cabelitos: Maybe we could create a function `sol_http_...._send_sse_prefix()`
@ceolin @barbieri what do you guys think? To be honest I don't like the current approach

src/lib/comms/sol-http-server-impl-microhttpd.c
cabelitos: We're fine here. the `delete_me` variable will be true and `sol_http_progressive_response_feed` will fail

src/lib/comms/sol-http-server-impl-microhttpd.c
cabelitos: My bad. it should be `progressive->accumulated_bytes = total`
Fixed.

src/lib/comms/sol-http-client-impl-curl.c
```diff
@@ -1059,6 +1099,12 @@ sol_http_client_connection_cancel(struct sol_http_client_connection *pending)
 {
     SOL_NULL_CHECK(pending);
 
+    if (pending->in_use) {
+        pending->del_timeout = sol_timeout_add(0, schedule_del, pending);
+        SOL_NULL_CHECK(pending->del_timeout);
```
barbieri: this is not good since in cases where the main loop is finalized/quit, the timer may not run.

Why not use a `delete_me` as you did in other code? Maybe check crypto/message-digest and use the flag with a refcount?

src/modules/flow/http-server/http-server.c
```diff
@@ -313,43 +329,60 @@ sse_conn_closed(void *data, const struct sol_http_progressive_response *sse)
...
+
+    r = sol_http_progressive_response_feed(client, &sse_prefix);
+    SOL_INT_CHECK(r, < 0, r);
+
+    r = sol_http_progressive_response_feed(client, blob);
+    SOL_INT_CHECK(r, < 0, r);
+
+    r = sol_http_progressive_response_feed(client, &sse_suffix);
+    SOL_INT_CHECK(r, < 0, r);
+
```
barbieri: as someone said elsewhere, this function can be move to http server, something like `sol_http_progressive_response_feed_sse()`, then you don't need to replicate this here and in web-inspector.

src/samples/http/server-sse.c
barbieri: said above, we should create a `sol_http_progressive_response_send_sse()` or `sse_data()`, where it will take a blob with the actual data, will queue the prefix and suffix for you.

cabelitos: I'm calling `sol_http_progressive_response_sse_feed()`

cabelitos:
#2049

PR #2030, opened 2016-05-12
WIP: Introduces a new design for web-inspector

ronaldosilva:
This is an UI implementation to visualyze flow data that
is being generated by Soletta.

This patch adds the basic features/images of the
tool.

Signed-off-by: Ronaldo Pereira <ronaldo.silva@intel.com>


ronaldosilva:
This is a WIP patch, please do not merge it already. It misses some needed functionalities.

Could you guys please start reviewing this patch ?

src/bin/sol-fbp-runner/Port.js
```diff
@@ -0,0 +1,122 @@
...
+	'use strict'
+
+	function Port (params) {
+
+		//data properties
+		var instance = this; //returns the __proto__ of the object
+
+		this.name = params["name"];
+		this.value = params["value"];
+		this.type = params["type"] || "in";
```
barbieri: are you receiving empty type? That shouldn't happen

src/bin/sol-fbp-runner/Port.js
```diff
@@ -0,0 +1,122 @@
...
+	Port.prototype.getElement = function() {
+		return this.el;
+	}
+
+	Port.prototype.getParentWidget = function() {
+		return this.widget;
+	}
+
+	Port.prototype.getPortType = function() {
+		return (this.type === "in") ? "in" : "out";
```
barbieri: you already ensure it's a string when you set it, no need to check it here

src/bin/sol-fbp-runner/Port.js
```diff
@@ -0,0 +1,122 @@
...
+	}
+
+	Port.prototype.select = function() {
+		this.emitt = true;
```
barbieri: why the double "t"? Also, why not call it "selected"?

src/bin/sol-fbp-runner/Widget.js
```diff
@@ -0,0 +1,222 @@
...
+
+	'use strict'
+
+	function Widget(payload, key) {
+
+		//caching a secure reference to the instance
+		var instance = this;
+		//var selectedPort;
+
+		Widget.prototype.portsCount = 0;
```
barbieri: you shouldn't set a class property from an instance constructor. Maybe move this outside this function

src/bin/sol-fbp-runner/Widget.js
```diff
@@ -0,0 +1,222 @@
...
+
+		//unique id for the widget and other unique properties
+		this.raw = JSON.stringify(payload, null, 4);
```
barbieri: keeping this is not that good, is this a stop-gap until we can render this in a nice format/html?

bdilly:
@ronaldosilva regarding licenses, please add it on all files, just like you have done with style.css

It would be nice to have a license for image assets as well. Maybe CC or whatever that makes sense for you. I suppose you're doing the same on Soletta Dev-App.

src/bin/sol-fbp-runner/Widget.js
```diff
@@ -0,0 +1,222 @@
...
+		this.key = key;
+		this.widgetName = payload.id; //change to 'name ?'
+		this.uid = payload.path[payload.path.length-1];
```
barbieri: ok, but you should also save `payload.path` somewhere and have some parenting/container. This will happen when you have flows that use sub flows. I'm not sure how to represent that in a nice graphical way. To the parent flow, the internal is represented as a single node with input and output ports. We can either "click to enter/expand" or show inlined with some graphical delimiter.

src/bin/sol-fbp-runner/Widget.js
```diff
@@ -0,0 +1,222 @@
...
+		$("<div/>").addClass("widget-display").appendTo(this.el).addClass("widget-data").append(this.displayValue);
+		this.widgetPorts = $("<div/>").addClass("widget-ports").appendTo(this.el);
+
+		//putting the ports in the correct container
+		this.lports = $("<div/>").addClass("port-container").appendTo("#"+this.uid + " .widget-ports");
+		this.rports = $("<div/>").addClass("port-container").appendTo("#"+this.uid + " .widget-ports");
+
+		//reading all in ports
+		var port;
+		payload.ports_in.forEach(function(element, index) {
```
barbieri: this will be empty as when we start, there are no connections yet. Or is this deferred?  If so, until when? In theory connections and disconnections could happen all the time (dynamic), but in practice it's done once at startup (static)

src/bin/sol-fbp-runner/Widget.js
```diff
@@ -0,0 +1,222 @@
...
+				data_type:element["data_type"],
+				required:element["required"],
+				widget:instance,
+				key:instance.key + "_" + Widget.prototype.portsCount
+			});
+
+			instance.portsIn.push(port);
+			instance.totalPorts++;
+			port.el.on("port-select", instance.onPortSelect);
+			Widget.prototype.portsCount ++;
```
barbieri: this is affecting the class, not the instance. Did you want to mean "this"?

src/bin/sol-fbp-runner/Widget.js
```diff
@@ -0,0 +1,222 @@
...
+			port = new Port(
+			{
+				name:element["name"],
+				value:undefined,
+				type:"in",
+				port_index:element["base_port_idx"],
+				data_type:element["data_type"],
+				required:element["required"],
+				widget:instance,
+				key:instance.key + "_" + Widget.prototype.portsCount
```
barbieri: you can use the node uid + port index as the key. The node id is unique during the flow lifetime. The port is also unique within a node. So `this.uid + "-" + index` is enough.

src/bin/sol-fbp-runner/Widget.js
```diff
@@ -0,0 +1,222 @@
...
+			});
+			instance.portsOut.push(port);
+			instance.totalPorts++;
+			port.el.on("port-select", instance.onPortSelect);
+			Widget.prototype.portsCount ++;
+		});
+
+		//Auto Adjusting the size of the Widget Container
+		var h = Math.max(this.portsIn.length, this.portsOut.length) * 26;
```
barbieri: can we get `26` from some existing element? Or at least make it a constant that we can remember about it later :-)

src/bin/sol-fbp-runner/Widget.js
```diff
@@ -0,0 +1,222 @@
...
+	};
+
+	//UPDATES THE VALUES OF ALL PORTS
+	Widget.prototype.update = function(type, data) {
+
+		//getting the child ports of the widget
+		var port = (type === "deliver") ? this.portsIn[data.port_idx] : this.portsOut[data.port_idx];
+
+		//if the ports have no pair, skip
+		if(port.connections.length < 0) return;
```
barbieri: for Inspector, it may be useful to know the packet being delivered even if there are no connections. I'd not do this early return

src/bin/sol-fbp-runner/Widget.js
```diff
@@ -0,0 +1,222 @@
...
+
+		if(packetType === "empty"){
+			port.update("empty");
+		}else if(packetType === "int" || packetType === "float"){
+			port.update(packet.payload.value)
+		}else if(packetType === "byte"){
+			port.update(packet.payload);
+		}else if(packetType === "boolean"){
+			port.update(packet.payload.toString());
+		}
```
barbieri: see my original web-inspector javascript, it's better to move this to a function that renders a packet to html, possibly with some CSS classes to allow identification of type. Things like rgb can display a color preview, etc. This may be reused in other places

src/bin/sol-fbp-runner/Widget.js
```diff
@@ -0,0 +1,222 @@
...
+
+		//clearing last selected port
+		if(this.selectedPort != null) {
+			this.selectedPort.unselect();
+			this.unsubscribe(port);
+		}
+
+		//selecting new port
+		this.clearDisplay().append(value);
```
barbieri: why take a `value`? just use `port.value`, no?

src/bin/sol-fbp-runner/Widget.js
```diff
@@ -0,0 +1,222 @@
...
+		this.selectedPort = port;
+		this.selectedPort.select();
+		this.subscribe(port);
+	};
+
+	Widget.prototype.onPortValueChange = function(event) {
+		event.widget.clearDisplay().append(event.value);
```
barbieri: this signature doesn't look right. Why not find the port and specify it here, then call the function on the `Port` so it could update itself.

src/bin/sol-fbp-runner/Widget.js
```diff
@@ -0,0 +1,222 @@
...
+		return this.el;
+	};
+
+	Widget.prototype.getLeft = function() {
+		return this.lports;
+	};
+
+	Widget.prototype.getRight = function() {
+		return this.rports;
+	};
```
barbieri: left and right come from nowhere. What if we use "in" and "out"? If they happen to be at the left, right, top, bottom... is up to the theme/graphics

src/bin/sol-fbp-runner/web-inspector.js
```diff
@@ -16,451 +16,178 @@
...
-    } else {
-        var keys = [];
-        for (var key in payload) {
-            if (!payload.hasOwnProperty(key))
-                continue;
-            keys.push(key);
+        try {
+            feed(e.data);
+        } catch (err){
+            var fixed = e.data.toString().replaceAll(":inf,",':"inf",');
```
barbieri: care to say why is that? This looks like an error in the sent data, which shouldn't happen and must be fixed in our C code

src/bin/sol-fbp-runner/web-inspector.js
```diff
@@ -16,451 +16,178 @@
...
-        html += '<dt class="fbp-node-type">type</dt>' +
-            '<dd class="fbp-node-type">' + payload.type + '</dd>';
-    }
+    if (data.event == "open") {
 
-    if (payload.category) {
-        html += '<dt class="fbp-node-category">category</dt>' +
-            '<dd class="fbp-node-category">' + payload.category + '</dd>';
-    }
+        if(data.payload.id.indexOf(".fbp") > -1) return;
```
barbieri: this is because it's the root node, I'd say do not use `data.payload.id` but `data.payload.path.length == 1`.

src/bin/sol-fbp-runner/web-inspector.js
```diff
@@ -16,451 +16,178 @@
...
-
-    return html;
-};
-
-FBPInspector.prototype.renderNodePorts = function (node, direction, descriptions) {
-    var n = this.nodes[node];
-    var html = "";
+        nodeId = data.payload.node;
+        widgets.forEach(function(widget, index) {
+            if (widget.getUID() === nodeId) {
```
barbieri: it's better to keep a map `{nodeId: widget}`

src/bin/sol-fbp-runner/web-inspector.html
```diff
@@ -1,57 +1,21 @@
...
-          <th>Packet</th>
-        </thead>
-        <tbody id="fbp-log-delivered" >
-        </tbody>
-      </table>
-    </div>
-  </body>
+	<head>
+		<meta charset="utf-8" />
+		<link rel="stylesheet" href="css/style.css" />
```
barbieri: it's better to use no directory, just `style.css` since to install to a different directory will be more painful in our build system (for a final version you'll need to edit Makefile)

barbieri:
@ronaldosilva nice work, man! I did many comments (as usual in my reviews) and I believe the engineers can help a bit to improve the data structures in the JS, but that's minor. Keep working and once you're happy with the inspector UX we can help with that.

src/bin/sol-fbp-runner/web-inspector.js
ronaldosilva: Yes, we got an error when trying to read a port with the "inf" value. The value itself is lacking the double quotes around it, so we are receiving inf instead of "inf" and it is messing with the JSON Parser. I added this try/catch to temporarily solve this issue.

src/bin/sol-fbp-runner/Modal.js
```diff
@@ -0,0 +1,50 @@
+(function(window) {
+
+	'use strict'
+
+	function Modal(){
```
brunobottazzini: split the { from the Modal() 

```
function Modal() {
```

src/bin/sol-fbp-runner/Modal.js
```diff
@@ -0,0 +1,50 @@
...
+		this.el = $("<div/>").addClass("modal").appendTo("#modal-container");
+		this.header = $("<div/>").addClass("modal-header").appendTo(this.el);
+		this.title = $("<span/>").appendTo(this.header).addClass("modal-title");
+		this.content = $("<pre/>").appendTo(this.el);
+
+		this.el.hide();
+		this.shield = $("<div/>").attr("id","shield");
+
+		this.closeButton = $("<img/>").attr("src","images/ico_close.png").addClass("close").appendTo(this.header)
+		.click(function(){
```
brunobottazzini: Same split here

```
.click(function() {
```

src/bin/sol-fbp-runner/Modal.js
```diff
@@ -0,0 +1,50 @@
...
+			instance.hide();
+		});
+
+
```
brunobottazzini: try not putting so much empty lines between braces

src/bin/sol-fbp-runner/Port.js
```diff
@@ -0,0 +1,122 @@
...
+
+	Port.prototype.unselect = function() {
+		this.emitt = false;
+		this.el.removeClass("selected");
+	}
+
+	Port.prototype.connect = function (port) {
+
+		//ERROR TRYING TO CONNECT PORTs!
+		if(this.connections.indexOf(port.key) == -1)
```
brunobottazzini: If the port.key is always the same type you should use three equals.

Also, put a space between if and (.  :)

```
if (this.connections.indexOf(port.key) === -1)
```

src/bin/sol-fbp-runner/Modal.js
```diff
@@ -0,0 +1,50 @@
+(function(window) {
+
+	'use strict'
```
brunobottazzini: Missing semicolon here...

src/bin/sol-fbp-runner/Modal.js
```diff
@@ -0,0 +1,50 @@
...
+
+		this.el = $("<div/>").addClass("modal").appendTo("#modal-container");
+		this.header = $("<div/>").addClass("modal-header").appendTo(this.el);
+		this.title = $("<span/>").appendTo(this.header).addClass("modal-title");
+		this.content = $("<pre/>").appendTo(this.el);
+
+		this.el.hide();
+		this.shield = $("<div/>").attr("id","shield");
+
+		this.closeButton = $("<img/>").attr("src","images/ico_close.png").addClass("close").appendTo(this.header)
```
brunobottazzini: Missing semicolon here...

src/bin/sol-fbp-runner/Port.js
```diff
@@ -0,0 +1,122 @@
+(function(window) {
+
+	'use strict'
```
brunobottazzini: Missing semicolon here...

src/bin/sol-fbp-runner/Port.js
```diff
@@ -0,0 +1,122 @@
...
+		var instance = this; //returns the __proto__ of the object
+
+		this.name = params["name"];
+		this.value = params["value"];
+		this.type = params["type"] || "in";
+		this.data_type = params["data_type"];
+		this.widget = params["widget"];
+		this.port_index = params["port_index"];
+		this.required = params["required"] || true;
+		this.key = "port_" + this.type + "_" + params["key"];
```
brunobottazzini: All these parameters should be written in dot notation instead of using ['key']

It would be something like this:
```
this.required = params.required;
```

You can see more info about this here:
https://jslinterrors.com/a-is-better-written-in-dot-notation

src/bin/sol-fbp-runner/Port.js
```diff
@@ -0,0 +1,122 @@
...
+		var instance = this.instance;
+
+		if(this.emitt) this.el.trigger(
+		{
+			type:"port-update",
+			widget:widget,
+			port:instance,
+			value:value,
+		});
+	}
```
brunobottazzini: Missing semicolon here...

src/bin/sol-fbp-runner/Port.js
```diff
@@ -0,0 +1,122 @@
...
+		this.el.unbind();
+	}
+
+	Port.prototype.showPort = function() {
+		//TODO
+	}
+
+	window.Port = Port;
+
+})(window)
```
brunobottazzini: Missing semicolon here...

src/bin/sol-fbp-runner/Port.js
```diff
@@ -0,0 +1,122 @@
...
+
+	Port.prototype.hidePort = function() {
+		this.el.addClass("hidden");
+		this.el.addClass("non-selectable");
+		this.el.unbind();
+	}
+
+	Port.prototype.showPort = function() {
+		//TODO
+	}
```
brunobottazzini: All Port prototype should be ended with semicolon.

src/bin/sol-fbp-runner/Widget.js
```diff
@@ -0,0 +1,222 @@
+(function(window) {
+
+	'use strict'
```
brunobottazzini: semicolon here :)

src/bin/sol-fbp-runner/Widget.js
```diff
@@ -0,0 +1,222 @@
...
+		//reading all out ports
+		payload.ports_out.forEach(function(element, index) {
+			port = new Port(
+			{
+				name:element["name"],
+				value:undefined,
+				type:"out",
+				port_index:element["base_port_idx"],
+				data_type:element["data_type"],
+				required:element["required"],
```
brunobottazzini: Use dot notation instead of using ['key'] here too

src/bin/sol-fbp-runner/Widget.js
```diff
@@ -0,0 +1,222 @@
...
+		var port;
+		payload.ports_in.forEach(function(element, index) {
+			port = new Port(
+			{
+				name:element["name"],
+				value:undefined,
+				type:"in",
+				port_index:element["base_port_idx"],
+				data_type:element["data_type"],
+				required:element["required"],
```
brunobottazzini: Use dot notation instead of using ['key'] here too

src/bin/sol-fbp-runner/Widget.js
```diff
@@ -0,0 +1,222 @@
...
+		if(port.connections.length < 0) return;
+
+		//found paired ports, update its value
+		var packet = data.packet;
+		var packetType = packet.packet_type;
+
+		if(packetType === "empty"){
+			port.update("empty");
+		}else if(packetType === "int" || packetType === "float"){
+			port.update(packet.payload.value)
```
brunobottazzini: Missing semicolon here... 

also in these line of codes try not letting everything together, just for organization
Such as 

```
} else if (packetType === "int" || packetType === "float") {
```

src/bin/sol-fbp-runner/Widget.js
```diff
@@ -0,0 +1,222 @@
...
+		},delay);
+	};
+
+	Widget.prototype.clearPorts = function() {
+
+		this.portsIn.forEach(function(p, i){
+			if(p.connections.length === 0){
+				p.hidePort();
+			}
+		})
```
brunobottazzini: missing semicolon here

src/bin/sol-fbp-runner/Widget.js
```diff
@@ -0,0 +1,222 @@
...
+
+		this.portsOut.forEach(function(p, i){
+			if(p.connections === 0){
+				p.hidePort();
+			}
+		})
```
brunobottazzini: missing semicolon here

src/bin/sol-fbp-runner/Widget.js
```diff
@@ -0,0 +1,222 @@
...
+	};
+
+	Widget.prototype.autoSelectPort = function(port) {
+		if(this.selectedPort == null){
```
brunobottazzini: Here you should always use three === to compare NULL.

You can also put like this:

```
if (!this.selectedPort) {
```

src/bin/sol-fbp-runner/Widget.js
```diff
@@ -0,0 +1,222 @@
...
+		//dealing with repeated port click
+		if(port === this.selectedPort) {
+			port.unselect();
+			this.unsubscribe(port);
+			this.selectedPort = null;
+			return;
+		}
+
+		//clearing last selected port
+		if(this.selectedPort != null) {
```
brunobottazzini: change != to !==

or

```
if (this.selectedPort) {
```

src/bin/sol-fbp-runner/Widget.js
```diff
@@ -0,0 +1,222 @@
...
+		return this.portsIn[index];
+	};
+
+	Widget.prototype.getOutByIndex = function(index) {
+		return this.portsOut[index];
+	};
+
+	window.Widget = Widget;
+
+})(window)
```
brunobottazzini: Add semicolon here

src/bin/sol-fbp-runner/web-inspector.js
```diff
@@ -16,451 +16,178 @@
...
-            '<dd class="fbp-node-license">' + payload.license + '</dd>';
-    }
+    } else if (data.event == "disconnect") {
+        //TODO
 
-    if (payload.version) {
-        html += '<dt class="fbp-node-version">version</dt>' +
-            '<dd class="fbp-node-version">' + payload.version + '</dd>';
-    }
+    } else if (data.event == "send") {
```
brunobottazzini: If data.event is always a string use ===

src/bin/sol-fbp-runner/web-inspector.js
```diff
@@ -16,451 +16,178 @@
...
-                '<dt>required</dt><dd>'+ o.required + '</dd>' +
-                '<dt>type</dt><dd>' + o.data_type + '</dd>' +
-                '<dt>default value</dt><dd>' + this.renderOptionValue(o.data_type, o.defvalue) + '</dd>' +
-                '<dt>description</dt><dd>' + o.description + '</dd>' +
-                '</dl></div></td>' +
-                '</tr>';
+        if (!isCleared) {
+            widgets.forEach(function(item, index) {
+                item.clearPorts();
+            })
```
brunobottazzini: Add missing semicolon here

src/bin/sol-fbp-runner/web-inspector.js
```diff
@@ -16,451 +16,178 @@
...
+        widgets.forEach(function(widget, index) {
 
-        // TODO: show
-
-    } else if (ev.event == "send") {
-        var packetHtml = this.renderPacket(payload.packet);
+            if (widget.getUID() === src.node) {
+                src.port = widget.getOutByIndex(src.idx);
+            } else if (widget.getUID() === dest.node) {
+                dest.port = widget.getInByIndex(dest.idx)
```
brunobottazzini: Missing semicolon

brunobottazzini:
@ronaldosilva, I just reviewed and your patch is awesome, I just mentioned minor things...

Thanks!

PR #2033, opened 2016-05-13
OIC: Change errno value.

cabelitos:
The value ECOMM is not defined by Posix, this can lead to problems on small OSs.


otaviobp:
+1

src/lib/comms/sol-oic-client.c
```diff
@@ -996,7 +996,7 @@ _resource_request(struct sol_oic_client_request *request,
 
 cbor_error:
     SOL_ERR("Could not encode CBOR representation: %s", cbor_error_string(err));
-    r = -ECOMM;
+    r = -ECANCELED;
```
ceolin: I'm wondering if this error is the best for this case. Maybe `EBADMSG` fits better.

cabelitos:
Merged with ceolin's suggestion

PR #2038, opened 2016-05-13
tests: increase tests timeout

bdilly:
Increase a second for all tests and double it for HTTP,
since we often face fails on semaphoreci and jenkins
for these tests.


glima:
Fine by me

bdilly:
merged, thanks

PR #2041, opened 2016-05-13
Crypto: It should follow the stream design pattern.

cabelitos:

### DIFF
--- a/src/lib/crypto/include/sol-message-digest.h
+++ b/src/lib/crypto/include/sol-message-digest.h
@@ -151,15 +151,27 @@ struct sol_message_digest_config {
      * @li @c handle the handle used to push data.
      * @li @c input the input data originally pushed/fed. There is no
      *     need to sol_blob_unref() it is done automatically.
+     * @li @c status 0 on success -errno on error.
      *
      * @note it is safe to delete the message digest handle from
      *       inside this callback.
      */
-    void (*on_feed_done)(void *data, struct sol_message_digest *handle, struct sol_blob *input);
+    void (*on_feed_done)(void *data, struct sol_message_digest *handle, struct sol_blob *input, int status);
     /**
      * The context data to give to all callbacks.
      */
     const void *data;
+
+    /**
+     * The tx buffer max size. The value @c 0 means unlimited data.
+     * Since sol_message_digest_feed() works with blobs, no extra buffers will be allocated in order
+     * to store @c tx_size bytes. All the blobs that are schedule to be written will be referenced
+     * and the sum of all queued blobs must not be equal or exceed @c tx_size.
+     * If it happens sol_message_digest_feed() will return @c -ENOSPC and one must start to control the
+     * writing flow until @c on_feed_done is called.
+     * @see sol_message_digest_feed()
+     */
+    size_t tx_size;
 };
 
 /**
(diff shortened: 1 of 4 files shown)


src/lib/crypto/include/sol-message-digest.h
```diff
@@ -151,15 +151,27 @@ struct sol_message_digest_config {
...
+     * The tx buffer max size. The value @c 0 means unlimited data.
+     * Since sol_message_digest_feed() works with blobs, no extra buffers will be allocated in order
+     * to store @c tx_size bytes. All the blobs that are schedule to be written will be referenced
+     * and the sum of all queued blobs must not be equal or exceed @c tx_size.
+     * If it happens sol_message_digest_feed() will return @c -ENOSPC and one must start to control the
+     * writing flow until @c on_feed_done is called.
+     * @see sol_message_digest_feed()
+     */
+    size_t tx_size;
 };
```
barbieri: you should follow uart and mention the pattern in the sol_message_digest_config structure.

I'm unsure if we rename `tx_size` to `feed_size` or `on_feed_done` to `tx_cb`, ask others to chime in

src/lib/crypto/include/sol-message-digest.h
```diff
@@ -218,8 +230,8 @@ void sol_message_digest_del(struct sol_message_digest *handle);
  *        @c config->on_digest_ready() is called with the resulting digest
  *        as a blob.
  *
- * @return 0 on success, -errno otherwise. If error, then the input reference
- *         is not taken.
+ * @return 0 on success, @c -ENOSPC if sol_message_digest_config::tx_size is not zero and there's
+ * no more space left or -errno on error. If error or -ENOPSC, If error, then the input reference is not taken.
```
barbieri: link to the design pattern page, as done with UART

src/lib/crypto/include/sol-message-digest.h
barbieri: tx will be strange, keep it `feed`, but maybe mention this is an alias in the design pattern.

barbieri:
review the code comparing to your written design pattern. Maybe we'll need to change the design pattern to mention 'feed' instead of `tx`, so it may be applicable in more scenarios.

Also note that the `on_feed_done()` doesn't respect the "delete from callback", we'll need the delete_me/in_use booleans to do that.

src/lib/crypto/include/sol-message-digest.h
cabelitos: I did not mention because the docs are not part of this PR. I don't want to break the docs generation

cabelitos:
@barbieri I did not worry about the delete_me/in_use because the docs says ` * @note it is safe to delete the message digest handle from  inside this callback. `

src/lib/crypto/include/sol-message-digest.h
cabelitos: To be honest I don't like `tx_....` prefix. I guess we should change it to `feed`.

barbieri:
ah, that's right, there is a refcnt in the handle and it's increased before calling the user.

barbieri:
btw, to follow your new design pattern, the _sol_message_digest_free() must be changed to call `on_feed_done(... , -ECANCELLED)` for pending blobs.

cabelitos:
@barbieri It's calling with `-ECANCELLED`

src/lib/crypto/sol-message-digest-common.c
```diff
@@ -317,6 +320,8 @@ _sol_message_digest_free(struct sol_message_digest *handle)
     _sol_message_digest_thread_fini(handle);
 
     SOL_VECTOR_FOREACH_IDX (&handle->pending_feed, pf, i) {
+        if (handle->on_feed_done)
+            handle->on_feed_done((void *)handle->data, handle, pf->blob, -ECANCELED);
```
cabelitos: Right here

barbieri: hum... but that just accounts what is to be processed. Look into `_sol_message_digest_thread_fini()`, what was processed but not dispatched, will still be in `handle->pending_dispatch`.

This list is consumed from `_sol_message_digest_thread_feedback()`, where the vector is stolen to a local variable and re-initialized to zero (took me few minutes to get that, and I was the one who wrote the code... ouch!)

cabelitos: You're right. I will add that

cabelitos:
#2048

PR #2046, opened 2016-05-16
Fix a memory issues in coap and oic.

otaviobp:
- Fix a invalid memory access in Oic server
 - Fix memory leaks in CoAP
 - Rename notification functions in OIC and CoAP to a proper name


vcgomes:
I didn't test, but the changes look good.

One thing that this made me think is whether it makes sense to keep those two `sol_coap_server`s (one for plaintext connections and one for DTLS connections) in the OIC side.

I am wondering if it makes sense to have one `sol_coap_server` with multiple server sockets.

PR #2048, opened 2016-05-16
Crypto: It should follow the stream design pattern. V2

cabelitos:
Changes since v1:
* Renamed `tx_size` to `feed_size`
* The function `_sol_message_digest_thread_fini` now callback `on_feed_done` with `-ECANCELED`


cabelitos:
merged

PR #2049, opened 2016-05-16
HTTP: sol_http_progressive_response_feed() should receive a blob. v4

cabelitos:
Changes since v3:
* Added commit 93c425a
* Added function `sol_http_progressive_response_sse_feed()`
* Renamed the stream callbacks in order to match the design pattern.
* Fixed a bug that caused `progressive->accumulated_bytes` to be wrongly updated


cabelitos:
merged

PR #2051, opened 2016-05-16
The Resurrection of: Lsm9ds0

mbelluzzo:
Dispatcher works \o/, but...

... It won't work with our original API because I can't change our read and write APIs to queue on the dispatcher because we set addr on different call. So the will compete and break everything.


barbieri:
why gyroscope is its own node and there is a strange "xmt" with the rest? I'd try to stick to the IIO pattern where there is one node per interface, so would have one for accelerometer, another for the magnetometer and so on. Internally they can share the same device.

mbelluzzo:
Because you cant split does, its all the same device, if you try to split
than the configuration part will be a mess with three different nodes
trying to set things (config) in the device. If you do split it may work if
all nodes "agree" on the operation mode, but is a compatibility that will
be maintained by hand.

mbelluzzo:
And if the question for all our I2C problems should be "use IIO" (what in
this case there isn't a driver yet but ok), its better to drop the i2c api.
Its less about this particular sensor and more about our i2c api.

src/lib/io/sol-i2c-common.c
```diff
@@ -23,20 +23,259 @@
...
+    if (!op_set)
+        return; // We finished for now
+
+proceed:
+    if (dispatcher->timer)
+        SOL_ERR("Dispatcher timer should be NULL in this point.");
+    dispatcher->timer = sol_timeout_add(op_set->delay, i2c_dispatcher_exec_op, shared);
+    if (!dispatcher->timer) {
+        SOL_ERR("Failed to schedule I2C operation.");
+        status = -1;
```
glima: We could have all sol_timeout_add() implementations set errno properly instead

src/lib/io/sol-i2c-common.c
```diff
@@ -23,20 +23,259 @@
...
+        return true;
+    }
+
+    op = sol_vector_get(op_set->set, dispatcher->set_idx);
+
+    dispatcher->pending = (op->type == SOL_I2C_WRITE) ?
+        sol_i2c_write_register(shared->i2c, op->reg, &op->value, 1,
+        i2c_dispatcher_op_done, shared) :
+        sol_i2c_read_register(shared->i2c, op->reg, &op->value, 1,
+        i2c_dispatcher_op_done, shared);
```
glima: You should state that this API is solely for 1 byte read/writes or, maybe, extend it to support more bytes as well

glima:
I also don't get the nodes separation. Will the two nodes function in the same flow, as they are right now? If the answer is yes, they could very well be a single node.

mbelluzzo:
You can do both. With 2 nodes we keep consistency with our nodes API

src/modules/flow/lsm9ds0/lsm9ds0.json
```diff
@@ -0,0 +1,127 @@
...
+  "$schema": "http://solettaproject.github.io/soletta/schemas/node-type-genspec.schema",
+  "name": "lsm9ds0",
+  "meta": {
+    "author": "Intel Corporation",
+    "license": "Apache-2.0",
+    "version": "1"
+  },
+  "types": [
+    {
+      "name": "lsm9ds0/gyroscope",
```
glima: you should alias it back to gyroscope/lsm9ds0

src/modules/flow/lsm9ds0/lsm9ds0.json
```diff
@@ -0,0 +1,127 @@
...
+          "name": "OUT",
+          "data_type": "direction-vector",
+          "description": "Gyroscope in all X/Y/Z axes, in a scale of degrees per second."
+        }
+      ],
+      "private_data_type": "lsm9ds0_gyro_data",
+      "url": "http://solettaproject.org/doc/latest/node_types/lsm9ds0/gyroscope.html"
+    },
+    {
+      "name": "lsm9ds0/xmt",
```
glima: are the usages independent? I mean, can I use it only as thermometer? We could then split the init commands into three sets and have 3 nodes, no? This would really match our expected/pristine nodes API.

src/modules/flow/lsm9ds0/lsm9ds0.json
```diff
@@ -0,0 +1,127 @@
...
+      "category": "input/hw",
+      "description": "Accelerometer (X), Magnetometer (M) and Temperature (T) node for the lsm9ds0 chipset.",
```
glima: s/temperature/thermometer

src/modules/flow/lsm9ds0/lsm9ds0.json
mbelluzzo: Usages could be independent, but check the config issue I mentioned in the pull request itself. That is the real issue in split those features in different nodes. We can only split until different slave device level.

src/modules/flow/lsm9ds0/lsm9ds0.json
mbelluzzo: :+1:

src/modules/flow/lsm9ds0/lsm9ds0.json
mbelluzzo: :+1:

barbieri:
you're missing my point.

 * IIO: indeed this is the recommended path for sensors like that and all, we shouldn't need to ship these modules and etc, really making the i2c/spi and other low level IO hidden or unnecessary.  *However* that wasn't my point. My point is that for IIO nodes, even if a driver exposes multiple interfaces (ie: luminosity + presence), we are not creating all combinations of them. Instead the approach is (will be, since not implemented yet) to have one node per interface, all using the same actual device.

 * The FBP is all about the user and being able to reuse flows with different components (ie: `sol-flow.json`). If you use completely different port names and types per nodes, they will not be reusable. The actual implementation of the node will deal with the sharing, so in your case that I2C communication would be shared among node instances. IIO will do that as well.

@edersondisouza is talking to @yongli3 about that, the idea is to keep a list of open parameters and once it's opened, the device is queried in the list, if exists and the configuration is the same, then it will reuse. If did not exist, it will create. If exists and differs, will refuse to `open()`.

src/lib/io/sol-i2c-common.c
```diff
@@ -23,20 +23,259 @@
...
+    uint8_t i;
+
+    if (dispatcher->timer)
+        sol_timeout_del(dispatcher->timer);
+
+    if (dispatcher->pending)
+        sol_i2c_pending_cancel(i2c, dispatcher->pending);
+
+    SOL_PTR_VECTOR_FOREACH_IDX (&dispatcher->queue, op_set, i) {
+        free(op_set);
```
ibriano: What about the vector inside the set?

mbelluzzo: Is owned by the user and can be static some times.

src/modules/flow/lsm9ds0/lsm9ds0.c
```diff
@@ -0,0 +1,498 @@
...
+    struct lsm9ds0_gyro_data *mdata = data;
+
+    if (mdata->i2c) {
+        if (mdata->pending)
+            sol_i2c_dispatcher_remove_op_set(mdata->i2c, mdata->pending);
+        sol_i2c_close(mdata->i2c);
+    }
+
+    sol_vector_clear(&mdata->init_q);
+    sol_vector_clear(&mdata->read_q);
```
ibriano: Why not keep one queue alone? You use one for init, and clear it on the done callback, then use the same vector for all the read queues you may need during the lifetime of the node. Same for the other one below.

mbelluzzo: :+1: done

src/lib/io/sol-i2c-common.c
mbelluzzo: add on the struct doc

PR #2053, opened 2016-05-17
Add the Bluetooth agent API (for BlueZ)

vcgomes:
This adds the concept of an agent, that will serve to receive input from the user during the pairing process primarily.

The API is pretty similar to what Zephyr provides, which is based on the BlueZ D-Bus API, so this should map pretty well to both.

When reviewing, please see if the approach of the `sol_bt_agent_reply_*` family of functions is easy to understand.

### DIFF
--- a/src/lib/comms/sol-bluetooth-impl-bluez.c
+++ b/src/lib/comms/sol-bluetooth-impl-bluez.c
@@ -29,6 +29,8 @@
 
 #include "sol-bluetooth-impl-bluez.h"
 
+#define AGENT_PATH "/org/solettaproject/agent"
+
 struct sol_bt_scan_pending {
     sd_bus_slot *slot;
     void (*callback)(void *user_data, const struct sol_bt_device_info *device);
@@ -292,6 +294,23 @@ destroy_device(struct device_info *device)
 
     free(device->path);
     free(device);
+}
+
+static void
+destroy_pairing(struct context *ctx, bool success)
+{
+    if (ctx->pair_cb)
+        ctx->pair_cb((void *)ctx->pair_user_data, success, ctx->auth_conn);
+
+    ctx->pair_cb = NULL;
+
+    if (ctx->auth_conn)
+        sol_bt_conn_unref(ctx->auth_conn);
+
+    ctx->auth_conn = NULL;
+
+    ctx->pair_slot = sd_bus_slot_unref(ctx->pair_slot);
+    ctx->agent_msg = sd_bus_message_unref(ctx->agent_msg);
 }
 
 static struct device_info *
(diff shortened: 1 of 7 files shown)


src/lib/comms/sol-bluetooth-impl-none.c
```diff
@@ -78,3 +78,70 @@ sol_bt_conn_get_addr(const struct sol_bt_conn *conn)
...
 
+SOL_API const struct sol_bt_device_info *
+sol_bt_conn_get_device_info(const struct sol_bt_conn *conn)
+{
+    return NULL;
+}
+
+SOL_API int
+sol_bt_conn_pair(struct sol_bt_conn *conn,
+    void (*cb)(void *user_data, bool success, struct sol_bt_conn *conn),
```
glima: I wonder if we can have an integer success/status variable to indicate a broader range of errors

src/lib/comms/sol-bluetooth-impl-bluez.c
```diff
@@ -29,6 +29,8 @@
 
 #include "sol-bluetooth-impl-bluez.h"
 
+#define AGENT_PATH "/org/solettaproject/agent"
```
glima: OIC started to expose those things as Kconfig entries. Should we here?

src/lib/comms/sol-bluetooth-impl-bluez.c
```diff
@@ -1581,6 +1614,574 @@ sol_bt_stop_scan(struct sol_bt_scan_pending *scan)
...
+
+    ctx->agent_msg = sd_bus_message_unref(ctx->agent_msg);
+
+    r = sd_bus_send(NULL, reply, NULL);
+
+    sd_bus_message_unref(reply);
+
+    SOL_INT_CHECK(r, < 0, r);
+
+    return -ENOSYS;
```
glima: You don't return 0?

src/samples/bluetooth/simple-pair.c
```diff
@@ -0,0 +1,184 @@
...
+
+static void
+startup(void)
+{
+    if (sol_argc() > 1) {
+        const struct sol_network_link_addr *addr;
+
+        addr = sol_network_link_addr_from_str(&pair_addr, sol_argv()[1]);
+        SOL_NULL_CHECK(addr);
+    }
```
glima: else you should not proceed, no?

src/lib/comms/sol-bluetooth-impl-bluez.c
vcgomes: Ugh. Will fix.

src/samples/bluetooth/simple-pair.c
vcgomes: The idea here is that without any arguments this `simple-pair` tool will list all devices, so users usually will call it first without arguments until the desired device appears, then with the device's address.

src/lib/comms/sol-bluetooth-impl-bluez.c
vcgomes: I don't see any value, and this only makes sense for the BlueZ implementation, but perhaps I am missing something.

src/lib/comms/sol-bluetooth-impl-none.c
vcgomes: From what I could gather from the Zephyr API, there's no way to give more information than that.

vcgomes:
The new (and improved) version is at https://github.com/solettaproject/soletta/pull/2074, closing.

Issue #2054, opened 2016-05-18
OIC: Cancelling the pending call from the find resource callback causes a double free


gabrielschulhof:
#### Issue Description
```C
#include <stdio.h>
#include <string.h>
#include <sol-oic-client.h>
#include <soletta.h>

static struct sol_oic_pending *pending = NULL;

#define MIN(x, y) (((x) < (y)) ? (x) : (y))

static bool resourceFound(void *nothing, struct sol_oic_client *client,
		struct sol_oic_resource *resource) {
	printf("resourceFound: %.*s\n", SOL_STR_SLICE_PRINT(resource->path));
	if (sol_argc() > 1 && !strncmp(sol_argv()[1], resource->path.data,
			MIN(strlen(sol_argv()[1]), resource->path.len))) {
		sol_oic_pending_cancel(pending);
		resource = NULL;
	}

	return !!resource;
}

static void startup() {
	struct sol_oic_client *client = sol_oic_client_new();
	struct sol_network_link_addr address = {
		.addr = {
			.in6 = { 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0 }
		},
		.family = SOL_NETWORK_FAMILY_INET,
		.port = 5683
	};
	if (!sol_network_link_addr_from_str(&address, "224.0.1.187")) {
		fprintf(stderr, "Failed to retrieve address");
		return;
	}
	pending = sol_oic_client_find_resources(client, &address, "", "",
		resourceFound, NULL);
}

static void shutdown() {}

SOL_MAIN_DEFAULT(startup, shutdown);
```

#### Steps to reproduce

 1. Start an OIC server that exposes a resource at path *path*
 2. Compile the above code into an executable *binary*
 3. Run *binary* *path*

At this point it will die with a double-free, because the context within which the callback is called is unconditionally freed in response to the false return value given by the callback, but the context has already been freed by the explicit call to ```sol_oic_pending_cancel()```.

#### Board information

Host OS and architecture:

Fedora 23 x86_64

PR #2055, opened 2016-05-18

gabrielschulhof:
Fixes gh-2054

One point remains for discussion: Is there ever an instance where the context is set to FREE_ME after the callback returns, but it is not freed because the callback did not return false?

### DIFF
--- a/src/lib/comms/sol-coap.c
+++ b/src/lib/comms/sol-coap.c
@@ -92,7 +92,14 @@ struct resource_observer {
     uint8_t token[0];
 };
 
+enum pending_reply_freshness {
+    PENDING_REPLY_QUIET,
+    PENDING_REPLY_PENDING,
+    PENDING_REPLY_FREE_ME
+};
+
 struct pending_reply {
+    enum pending_reply_freshness freshness;
     struct sol_coap_server *server;
     struct sol_timeout *timeout;
     bool (*cb)(void *data, struct sol_coap_server *server, struct sol_coap_packet *req,
@@ -633,6 +640,11 @@ pending_reply_free(struct pending_reply *reply)
     if (!reply)
         return;
 
+    if (reply->freshness == PENDING_REPLY_PENDING) {
+        reply->freshness = PENDING_REPLY_FREE_ME;
+        return;
+    }
+
     if (reply->observing)
         free(reply->path);
 
@@ -892,6 +904,7 @@ sol_coap_send_packet_with_reply(struct sol_coap_server *server, struct sol_coap_
     }
 
     sol_coap_header_get_id(pkt, &reply->id);
+    reply->freshness = PENDING_REPLY_QUIET;
     reply->cb = reply_cb;
     reply->data = data;
     reply->observing = observing;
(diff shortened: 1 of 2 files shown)


gabrielschulhof:
Yes, there is: The callback returns true, but it also cancels the context.

gabrielschulhof:
@otaviobp @glima could you please have a look?

src/lib/comms/sol-coap.c
```diff
@@ -662,8 +662,11 @@ pending_timeout_cb(void *data)
...
 
-    if (reply->cb((void *)reply->data, server, NULL, NULL))
-        return true;
+    if (reply->cb((void *)reply->data, server, NULL, NULL)) {
+        if (reply->freshness != PENDING_REPLY_FREE_ME) {
+            return true;
+        }
+    }
 
     sol_ptr_vector_remove(&server->pending, reply);
```
glima: This will fail if the reply is cancelled in the callback

src/lib/comms/sol-oic-client.c
```diff
@@ -1056,7 +1113,7 @@ request_new(enum sol_coap_method method, enum sol_coap_msgtype type, struct sol_
     }
 
     r = _set_token_and_mid(request->base.pkt, &request->token);
-    SOL_INT_CHECK_GOTO(r, < 0, error;);
+    SOL_INT_CHECK_GOTO(r, < 0, error; );
```
glima: We should fix this ``;``

gabrielschulhof: Dang! I thought I fixed that. I'll force-push.

gabrielschulhof: The new version of uncrustify must want this, because this fails my git pre-commit hook if the space is not there. Whatever, I'll still force-push it.

src/lib/comms/sol-coap.c
gabrielschulhof: No, because if you call ```pending_reply_free()``` while ```reply->freshness``` is set to ```PENDING_REPLY_PENDING```, it won't free, it'll just set ```reply->freshness``` to ```PENDING_REPLY_FREE_ME```. Have a [look](https://github.com/solettaproject/soletta/pull/2055/files#diff-71a300ca520dd6eed36cfceb353e8312R643).

gabrielschulhof: ... and it is expected that the reply will be freed by whoever changed ```reply->freshness``` from ```PENDING_REPLY_QUIET``` to ```PENDING_REPLY_PENDING```.

gabrielschulhof: ... after the stack unwinds.

glima:
I wonder if status is not better than "freshness".

glima:
@vcgomes might want to take a look as well

src/lib/comms/sol-coap.c
gabrielschulhof: ... or, more precisely, after the outermost callback returns.

glima: I mean by ``sol_coap_cancel_send_packet()`` and friends, what you say about ``pending_reply_free()`` is true, but what about the also do ``sol_ptr_vector_del(&server->pending, i);``?

gabrielschulhof: Well, in order for the ```pending_reply``` memory to be freed, you have to call ```pending_reply_free()```, right? You don't ```free()``` directly, ever, right? So, if ```pending_reply_free()``` refuses to free, you're OK, right?

gabrielschulhof:
@glima - sure, ```status``` works for me. Whenever I get the :+1: I'll land it with ```status```.

src/lib/comms/sol-coap.c
glima: You still haven't looked at the sol_ptr_vector_del() done *outside* pending_reply_free() :) That's minor, though, as we don't even check the return value of the former.

cabelitos:
@gabrielschulhof I think you can resolve this problem in using simpler way using two boolean variables. Could you please take a look at `sol-uart-impl-linux.c`. Pay attention to the variables `delete_me` and `in_use`

cabelitos:
@gabrielschulhof The HEAD version

Issue #1716, opened 2016-03-23
API Changes: Sol-network


cabelitos:
#### Task Description
- [ ] Rename `SOL_INET_ADDR_STRLEN` to `SOL_NETWORK_INET_ADDR_STR_LEN`
- [ ] Rename `sol_network_hostname_handle` to `sol_network_hostname_pending`
- [ ] Rename `sol_network_cancel_get_hostname_address_info()` to `sol_network_hostname_pending_cancel()`
- [ ] Functions `sol_network_link_down()`, `sol_network_link_up()`, `sol_network_unsubscribe_events()` and `sol_network_subscribe_events()` should return an `int` instead of `bool`. Where `0` (zero) means success  and on error `-errno` should be returned.

#### Dependencies
None.

amandawrcoelho:
@ceolin we can close this task

PR #2057, opened 2016-05-18
network: Fix api name

amandawrcoelho:
issue #1716


bdilly:
That's nice, @amandawrcoelho 
But I would suggest you to make all API changes related to this module / issue at once.

cabelitos:
I agree with @bdilly.
Btw the patch looks good.

PR #2059, opened 2016-05-19
OIC client and coap: Make contexts re-entrant to prevent double free (v2)

gabrielschulhof:
Fixes gh-2054

Changes:
  * Factored out the reentrant stuff into a header so it can be used in both sol-coap.c and sol-oic-client.c
  * Used the two-variable approach from @cabelitos
  * Moved the relevant ```sol_ptr_vector_remove()``` into ```pending_reply_free()```

### DIFF
--- a/src/lib/common/include/sol-reentrant.h
+++ b/src/lib/common/include/sol-reentrant.h
@@ -1,0 +1,109 @@
+/*
+ * This file is part of the Soletta Project
+ *
+ * Copyright (C) 2015 Intel Corporation. All rights reserved.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+#pragma once
+
+#ifdef __cplusplus
+extern "C" {
+#endif
+
+/**
+ * @file
+ * @brief Structure and macros for making structures reentrancy-proof
+ */
+
+/**
+ * @defgroup Reentrant Reentrant
+ *
+ * @{
+ */
+
+/**
+ * @struct sol_reentrant
+ *
+ * @brief Structure containing the flags for safely freeing a larger structure
+ *
+ * This structure is meant to be used as the base of larger structures which
+ * are affected by calls to external callbacks that in turn end up calling
+ * library APIs. The possibility of a double free is particularly likely in
+ * such cases.
+ */
+struct sol_reentrant {
+    /**
+     * @brief Structure is in use
+     */
+     bool in_use;
+     /**
+      * @brief Structure is stale and should be freed as soon as possible
+      */
+      bool is_stale;
+};
+
+/**
+ * @brief Perform a function call to an external callback
+ *
+ * @param reentrant The reentrant affected by the function call
+ *
+ * @param call_statement The function call statement
+ *
+ * Wraps a call to a function by first recording whether the reentrant
+ * structure was in use and, after the function call, restoring the state of
+ * the reentrant structure from the recorded value.
+ */
+#define SOL_REENTRANT_CALL(reentrant, call_statement) \
+    do { \
+        struct sol_reentrant *local = ((struct sol_reentrant *)(reentrant)); \
+        bool was_in_use = local->in_use; \
+        local->in_use = true; \
+        call_statement; \
+        local->in_use = was_in_use; \
+    } while(0)
+
+/**
+ * @brief Conditionally free a reentrant structure
+ *
+ * @param reentrant The reentrant to free
+ *
+ * @param free_function The free function to call
+ *
+ * Calls the function specified by free_function if the reentrant is not marked
+ * as being in use. Otherwise, it marks the reentrant as stale but does not
+ * free it.
+ */
+#define SOL_REENTRANT_FREE(reentrant, free_function) \
+    do { \
+        struct sol_reentrant *local = ((struct sol_reentrant *)(reentrant)); \
+        local->is_stale = true; \
+        if (!(local->in_use)) { \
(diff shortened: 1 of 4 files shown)


gabrielschulhof:
Replaces #2055.

src/lib/comms/sol-coap.c
```diff
@@ -1503,19 +1512,23 @@ respond_packet(struct sol_coap_server *server, struct sol_coap_packet *req,
     /* If it isn't a request. */
     if (code & ~SOL_COAP_REQUEST_MASK) {
         bool found_reply = false;
+        bool reply_callback_result = false;
         SOL_PTR_VECTOR_FOREACH_REVERSE_IDX (&server->pending, reply, i) {
             if (!match_reply(reply, req))
                 continue;
 
-            if (!reply->cb((void *)reply->data, server, req, cliaddr)) {
-                sol_ptr_vector_del(&server->pending, i);
```
gabrielschulhof: Is it important that the reply is deleted from the vector *before* the unobserve packet is sent? I ask, because this modification would call ```send_unobserve_packet()``` before removing the pending reply from the vector in case an observation is in place.

src/lib/comms/sol-coap.c
```diff
@@ -1669,9 +1682,9 @@ sol_coap_server_destroy(struct sol_coap_server *server)
     }
 
     SOL_PTR_VECTOR_FOREACH_REVERSE_IDX (&server->pending, reply, i) {
-        sol_ptr_vector_del(&server->pending, i);
```
gabrielschulhof: Again, as [above](https://github.com/solettaproject/soletta/pull/2059/files#r63861131), the order
  0. remove-from-vector
  0. call-external-callback

is being reversed by this patch.

src/lib/comms/sol-coap.c
```diff
@@ -2010,13 +2022,14 @@ sol_coap_unobserve_by_token(struct sol_coap_server *server, const struct sol_net
         if (!match_observe_reply(reply, token, tkl))
             continue;
 
-        reply->cb((void *)reply->data, server, NULL, NULL);
-        sol_ptr_vector_del(&server->pending, i);
```
gabrielschulhof: ... and again, the unobserve packet will henceforth be sent while the pending_reply is still in the vector. Hope that's OK.

gabrielschulhof:
@glima, @cabelitos, @vcgomes got a new version ready.

src/lib/comms/sol-coap.c
glima: I don't see why this order would impact anything, from a 1st look

glima:
Kinda liking it. @cabelitos, this would fit uart code as well, right?

cabelitos:
@glima Me too.
We should replicate this sol-reetrant in other parts as well!

src/lib/common/include/sol-reentrant.h
```diff
@@ -0,0 +1,109 @@
...
+ */
+struct sol_reentrant {
+    /**
+     * @brief Structure is in use
+     */
+     bool in_use;
+     /**
+      * @brief Structure is stale and should be freed as soon as possible
+      */
+      bool is_stale;
```
vcgomes: cosmetic: the indentation looks wrong.

gabrielschulhof: Oh, yeah. Weird. I'll fix that.

src/lib/comms/sol-coap.c
```diff
@@ -1519,19 +1528,23 @@ respond_packet(struct sol_coap_server *server, struct sol_coap_packet *req,
     /* If it isn't a request. */
     if (code & ~SOL_COAP_REQUEST_MASK) {
         bool found_reply = false;
+        bool reply_callback_result = false;
```
vcgomes: Move this declaration one level down, inside the loop block. Seems that you don't need it outside the loop.

src/lib/comms/sol-coap.c
```diff
@@ -899,6 +906,8 @@ sol_coap_send_packet_with_reply(struct sol_coap_server *server, struct sol_coap_
     }
 
     sol_coap_header_get_id(pkt, &reply->id);
+    reply->reentrant.in_use = false;
+    reply->reentrant.is_stale = false;
```
vcgomes: optional: As `reply` is allocated with `calloc()` you don't need this. But if you think that this increases clarity, keep it.

src/lib/common/include/sol-reentrant.h
```diff
@@ -0,0 +1,109 @@
...
+ *
+ * @{
+ */
+
+/**
+ * @struct sol_reentrant
+ *
+ * @brief Structure containing the flags for safely freeing a larger structure
+ *
+ * This structure is meant to be used as the base of larger structures which
```
glima: I'd enforce that "base" means the 1st field as well

src/lib/common/include/sol-reentrant.h
```diff
@@ -0,0 +1,109 @@
...
+ *
+ * @param reentrant The reentrant affected by the function call
+ *
+ * @param call_statement The function call statement
+ *
+ * Wraps a call to a function by first recording whether the reentrant
+ * structure was in use and, after the function call, restoring the state of
+ * the reentrant structure from the recorded value.
+ */
+#define SOL_REENTRANT_CALL(reentrant, call_statement) \
```
vcgomes: I got the feeling that this has some overlap with `sol-monitors`, so just a question: did you consider using it?

If did consider and you found it lacking for this use case, perhaps it could be useful to document here in which use cases this is a better fit.

vcgomes:
I liked as well, just a couple of very minor stuff, and a question.

barbieri:
liked the approach a lot, thanks for taking the time to do it :-)

My main comment is about name: reentrant usually is assigned with threads and this is not thread-safe. I'm bad with names, so no suggestions on my side, sorry :-/

A secondary comment is to follow some of the sol-monitors and other names for members, so `delete_me` instead of `stale`. But that's minor.

If possible to avoid the cast to "reentrant", we can be safer and allow the reentrant to be anywhere in the structure, the booleans are usually best placed at the end due alignment needs being 1 byte (and if we pack both booleans using `:1`, the whole thing is one byte. Then the usage would be `MACRO(hande, call)` if the macro uses `&(handle)->reentrant` (fixed name, you can use another macro to declare it in the structure) or `MACRO(&handle->reentrant, call)` (flexible name, user gives the actual pointer).

PR #2060, opened 2016-05-19
Node.js bindings: Adapt UART JS bindings to the latest API changes

nagineni:
Adapt UART JS bindings to the latest API changes and enable them by default.  Also, add a sample that demonstrates the usage of bindings for reading and writing.

### DIFF
--- a/bindings/nodejs/src/functions/uart.cc
+++ b/bindings/nodejs/src/functions/uart.cc
@@ -16,6 +16,8 @@
  * limitations under the License.
  */
 
+#include <sol-buffer.h>
+#include <sol-str-slice.h>
 #include <sol-uart.h>
 #include <node.h>
 #include <nan.h>
@@ -32,19 +34,63 @@ public:
     static const char *jsClassName() { return "SolUART"; }
 };
 
-static void sol_uart_read_callback(void *user_data, struct sol_uart *uart,
-    unsigned char byte_read) {
+static ssize_t sol_uart_on_data_callback(void *user_data,
+    struct sol_uart *uart, const struct sol_buffer *buf) {
     Nan::HandleScope scope;
     sol_uart_data *uart_data = (sol_uart_data *)user_data;
-    Nan::Callback *callback = uart_data->rx_cb;
+    if (!uart_data)
+        return 0;
+
+    Nan::Callback *callback = uart_data->on_data_cb;
     if (!callback)
-        return;
+        return 0;
+
+    struct sol_str_slice slice;
+    slice = sol_buffer_get_slice(buf);
+    Local<Value> buffer =
+        Nan::CopyBuffer(slice.data, slice.len).ToLocalChecked();
 
     Local<Value> arguments[1] = {
-        Nan::New(byte_read)
+        buffer
+    };
+    callback->Call(1, arguments);
+
+    return slice.len;
+}
+
+static void sol_uart_on_feed_done_callback(void *user_data,
+    struct sol_uart *uart, struct sol_blob *blob, int status) {
+    Nan::HandleScope scope;
+    Local<Value> buffer;
+    sol_uart_data *uart_data = (sol_uart_data *)user_data;
+    if (!uart_data)
+        return;
+
+    if (status < 0) {
+        buffer = Nan::Null();
+    } else {
+        struct sol_str_slice slice;
+        slice = sol_str_slice_from_blob(blob);
+        buffer = Nan::CopyBuffer(slice.data, slice.len).ToLocalChecked();
+    }
+
+    Local<Value> arguments[2] = {
+        buffer,
+        Nan::New(status)
     };
 
-    callback->Call(1, arguments);
+    Nan::Callback *on_feed_done_cb = uart_data->on_feed_done_cb;
+    if (on_feed_done_cb)
+        on_feed_done_cb->Call(2, arguments);
+
+   // Retrieve the JS feed callback and call it
+   auto iter = uart_data->feed_callbacks_map.find(blob);
+   if (iter != uart_data->feed_callbacks_map.end()) {
+       Nan::Callback *feed_callback = iter->second;
+       feed_callback->Call(2, arguments);
+       uart_data->feed_callbacks_map.erase(blob);
+       delete feed_callback;
+   }
 }
 
 NAN_METHOD(bind_sol_uart_open) {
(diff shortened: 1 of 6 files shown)


nagineni:
@cabelitos Could you please have a look?

bindings/nodejs/src/functions/uart.cc
```diff
@@ -32,19 +34,63 @@ class SolUART : public JSHandle<SolUART> {
...
+
+    Nan::Callback *callback = uart_data->on_data_cb;
     if (!callback)
-        return;
+        return 0;
+
+    struct sol_str_slice slice;
+    slice = sol_buffer_get_slice(buf);
+    Local<Value> buffer =
+        Nan::CopyBuffer(slice.data, slice.len).ToLocalChecked();
```
barbieri: the idea of passing a buffer is to improve memory usage and copies. I know the user may take a reference to that buffer and thus you'd need to copy. But can you do that on demand? Like "CopyOnReference", that would only copy the given slice if there is a reference taken when the function returns.

As this is a wrapper object, you can easily replace its internals, so a way to do that would be to check after `callback->Call(1, arguments)`, if `buffer` reference is 1 and would die. In this case you'd introduce something like `buffer->ForceCopy()`

We'll use this pattern in other places, like the HTTP progressive response (likely not be wrapped in nodejs, since you already have that as a native module), Crypto/message-digest and so on.

bindings/nodejs/src/functions/uart.cc
```diff
@@ -32,19 +34,63 @@ class SolUART : public JSHandle<SolUART> {
...
+    sol_uart_data *uart_data = (sol_uart_data *)user_data;
+    if (!uart_data)
+        return;
+
+    if (status < 0) {
+        buffer = Nan::Null();
+    } else {
+        struct sol_str_slice slice;
+        slice = sol_str_slice_from_blob(blob);
+        buffer = Nan::CopyBuffer(slice.data, slice.len).ToLocalChecked();
```
barbieri: not the way to go. I'll mention later, but you can create a blob for a JS Object, then no copies are need :-)

Here you'd get the original JS object to use.

bindings/nodejs/src/functions/uart.cc
```diff
@@ -32,19 +34,63 @@ class SolUART : public JSHandle<SolUART> {
...
+    }
+
+    Local<Value> arguments[2] = {
+        buffer,
+        Nan::New(status)
```
barbieri: should you convert this `0` or `-errno` to some error instance? I don't think that the integer would make much sense to the JS users.

bindings/nodejs/src/functions/uart.cc
```diff
@@ -140,36 +168,38 @@ NAN_METHOD(bind_sol_uart_write) {
...
-
-    outputBuffer = (unsigned char *) malloc(length * sizeof(unsigned char));
+    void *outputBuffer = malloc(length);
     if (!outputBuffer) {
         Nan::ThrowError("Failed to allocate memory for output buffer");
         return;
     }
 
     memcpy(outputBuffer, node::Buffer::Data(info[1]), length);
+    blob = sol_blob_new(&SOL_BLOB_TYPE_DEFAULT, NULL, outputBuffer, length);
```
barbieri: Here is where the new pattern and used types will shine by reducing copies and memory usage.

The purpose of `struct sol_buffer` is to allow users to define read-only memory that has reference counting and parenting, as well as a custom type that provides its own "free".

That said, if you hava JS object you should create a blob for it, increase its reference count (since the blob will keep it), and when the blob is deleted, you unref the JS object. You can do this using two ways, pick whatever is the best:
 1. create a blob type for `SOL_BLOB_TYPE_JS_OBJECT`, use `data = object, len = whatever`, on `SOL_BLOB_TYPE_JS_OBJECT->free()` you unref `data`. In this case, you create a second blob of type `SOL_BLOB_TYPE_NO_FREE_DATA` for the actual buffer, `data = buffer, len = buffer_length, parent = jsobject_blob`. Then you feed this child blob. It will take a reference to the parent, when it not used anymore, the parent will be freed and the JS Object will be unreferenced;
 2. create a blob type for `SOL_BLOB_TYPE_JS_BUFFER` with a new extension/subclass structure: `struct sol_blob_js_buffer { struct sol_blob base; JSBuffer *js_buffer; }`. In this case you use `base.data = buffer, base.len = buffer_length; js_buffer = js_buffer->ref()`. Your new `SOL_BLOB_TYPE_JS_BUFFER->free()` will only `blob->js_buffer->unref(); free(blob);`

They are marginally different and provide the same net effect: no copies of your JS buffer. In your current code it would save 2 copies, if you imagine a huge buffer it would make a big difference :-)

The first approach I've mentioned is easier to extend to different objects, they could all use the same parent type. However it will allocate one more `struct sol_blob`. The second approach works only for buffer, but is slightly more efficient. Since the expected users are to be "buffer" kind, I don't see a problem going with the second approach.

IPM was recently introduced using this pattern, see example code at https://github.com/solettaproject/soletta/blob/master/src/lib/io/sol-ipm.c#L59 (ignore what is IPM, just see the blob declared in this line)

src/samples/nodejs/uart_sample.js
```diff
@@ -0,0 +1,130 @@
...
+const usage = "Usage: node uart_sample.js -p <producerUART> -c <consumerUART>\n" +
+    "options: \n" +
+    "  -h, --help \n" +
+    "  -p, --producer <producerUART>\n" +
+    "  -c, --consumer <consumerUART>\n";
+
+for ( var i = 0; i < args.length; i++ ) {
+    var arg = args[ i ];
+
+    switch ( arg ) {
```
barbieri: no getopt-like modules for nodejs?

brunobottazzini: There is getopt for nodejs here: https://www.npmjs.com/package/getopt-c

Maybe it can be used...

nagineni: There are several modules available for parsing the options. I'm not using them here just to avoid the dependency.

bindings/nodejs/src/functions/uart.cc
nagineni: Handling this already in the low level JS bindings.We don't pass the integer to JS users. We use this in the low level bindings to resolve the promise (i.e fulfill/reject with error) based on the status.

bindings/nodejs/src/functions/uart.cc
nagineni: Instead of creating a new blob type, I will make the JS buffer as persistent and pass the buffer data to sol_blob_new(). Later when the feed callback is done, I will free the persistent JS buffer. With this simple approach I can avoid the copies of the JS buffer.

bindings/nodejs/src/functions/uart.cc
nagineni: The buffer will have the reference always until it undergoes the garbage collection, so I would say we are copying it on the reference(not on the demand for sure). NewBuffer() is the most efficient way of creating a buffer, but that takes the ownership of the data memory. If I'm allowed to steal the data from the sol_buffer here, I can use the NewBuffer() instead of CopyBuffer() and that would make this much nicer.

bindings/nodejs/src/functions/uart.cc
nagineni: As per the spec, we don't need to pass the original JS object/buffer back, so I will change this a bit to pass only the callback status that can be used to resolve the promise of write call.

nagineni:
Replaced by #2081

bindings/nodejs/src/functions/uart.cc
barbieri: No, you aren't supposed to steal the buffer memory. But as I said, with a simple hack you can avoid copies on the most common case, that is to get the buffer and convert to something else (i.e.: JSON, split lines...) before returning

nagineni: Okay, I will look into this and make the changes for copy on demand in the followup patch.

Issue #1677, opened 2016-03-18
API Changes: Flow


bdilly:
#### Task Description

 - [x] use unit8_t instead of unsigned char for byte representation
 - [x] send_location_packet miss function receiving components, just like rgb and direction_vector does
 - [x] rename `sol_flow_builder_export_in_port()` to `sol_flow_builder_export_port_in()` (the same applies for export_out_port
 - [x] rename `sol_flow_packet_get_packet_type_as_string()` to `sol_flow_get_packet_type_name()`
 - [ ] function read_file() on struct sol_flow_parser_client should use sol_buffer
 - [ ] `sol_flow_parse_buffer_metatype()` should use sol_buffer too
 - [x] `sol_flow_simplectype_event` struct should be renamed to `sol_flow_simple_c_type_event`
 - [x] `sol_flow_simplectype_*` should be `sol_flow_simple_c_type` in the whole API
 - [x] `sol_flow_single_port_in_connect()` should be `sol_flow_single_connect_port_in()` (same applies for disconnect and the remaining of API).

nirmalsinghania2008:
is it uint8_t or unit8_t?
unit8_t is included in stdint.h
i am working on it assuming it is uint8_t

kaspersky:
#2061

kaspersky:
#2155

ceolin:
done

PR #2063, opened 2016-05-19
Flow apichanges v2

kaspersky:
#1677 

#2061


src/lib/flow/sol-flow-packet.c
```diff
@@ -694,13 +694,44 @@ sol_flow_packet_new_location(const struct sol_location *location)
...
 
+SOL_API int
+sol_flow_packet_get_location_components(const struct sol_flow_packet *packet, double *lat, double *lon, double *alt)
+{
+    struct sol_location ret;
+    int ret_val;
+
+    SOL_FLOW_PACKET_CHECK(packet, SOL_FLOW_PACKET_TYPE_LOCATION, -EINVAL);
+    ret_val = sol_flow_packet_get(packet, &ret);
+    if (ret_val == 0) {
```
cabelitos: I prefer `SOL_INT_CHECK(ret_val, < 0, ret_val);`

src/bin/sol-fbp-generator/main.c
```diff
@@ -233,8 +232,7 @@ read_meta_type_file(const struct sol_flow_metatype_context *ctx,
     }
 
     contents = sol_file_reader_get_all(reader);
-    *buf = contents.data;
-    *size = contents.len;
+    *buf = SOL_BUFFER_INIT_CONST((char *)contents.data, contents.len);
```
cabelitos: Not good. Data should be appended to the buffer

cabelitos: I don't like this approach. I prefer that we append that to the buffer

src/bin/sol-fbp-runner/runner.c
```diff
@@ -121,8 +121,7 @@ read_file(void *data, const char *name, const char **buf, size_t *size)
 
     free(path);
     slice = sol_file_reader_get_all(fr);
-    *buf = slice.data;
-    *size = slice.len;
+    *buf = SOL_BUFFER_INIT_CONST((char *)slice.data, slice.len);
```
cabelitos: I don't like this approach. I prefer that we append that to the buffer

src/lib/flow/include/sol-flow-metatype.h
```diff
@@ -95,12 +95,12 @@ struct sol_flow_metatype_context {
      * @param ctx The metatype context.
      * @param name The file path.
      * @param buf Where the file contents should be stored.
```
cabelitos: Explicitly say that the buffer must be already initialized and data will be appended. If the data does not fit, an error is returned

src/lib/flow/include/sol-flow-parser.h
```diff
@@ -72,11 +73,10 @@ struct sol_flow_parser_client {
      * @param data Client's data
      * @param name File name to read
      * @param buf Where to store the file
-     * @param size Buffer's size
      *
      * @return @c 0 on success, error code (always negative) otherwise
      */
-    int (*read_file)(void *data, const char *name, const char **buf, size_t *size);
+    int (*read_file)(void *data, const char *name, struct sol_buffer *buf);
```
cabelitos: Explicitly say that the buffer must be already initialized and data will be appended. If the data does not fit, an error is returned

src/bin/sol-fbp-generator/main.c
cabelitos: I'm not write sure here. I think the best option is append the data and not init the buffer.
@barbieri What do you think?

src/bin/sol-fbp-runner/runner.c
cabelitos: Same thing

src/lib/flow/include/sol-flow-metatype.h
```diff
@@ -95,12 +95,12 @@ struct sol_flow_metatype_context {
...
-     * @param size The size of the buffer.
```
cabelitos: Explicitly say that the buffer must be already initialized and data will be appended. If the data does not fit, an error is returned

src/lib/flow/include/sol-flow-parser.h
```diff
@@ -72,11 +73,10 @@ struct sol_flow_parser_client {
      * @param data Client's data
      * @param name File name to read
      * @param buf Where to store the file
```
cabelitos: Explicitly say that the buffer must be already initialized and data will be appended. If the data does not fit, an error is returned

src/lib/flow/include/sol-flow-parser.h
```diff
@@ -108,15 +108,13 @@ int sol_flow_parser_del(
  *
  * @param parser The Flow Parser handle
  * @param buf Flow description to be parsed
```
cabelitos: Same thing about buf initialization...

cabelitos:
In general the buffer commit looks good. I just don't know if the init pattern is the best one to go. Most parts of soletta that receives a buffer as arguments the data will always be appended!

src/lib/flow/include/sol-flow-metatype.h
kaspersky: You mean in the case that the _read_file_ would append data to the buffer? As the patch currently is, it is the read_file responsibility to init the sol_buffer

kaspersky:
@cabelitos Thanks for the review. I was not aware of the append to buffer pattern. Related to appending data to the received buffer parameter, it is a bit tricky because we have to make sure the appended data should be read-only (originally there was a const char * passed around).

src/bin/sol-fbp-generator/main.c
ceolin: append the data will make you duplicate the data. I'm ok with this since you have sure about the life-cycle of this memory, i mean, the buffer will not be used after the sol_file_reader be released.

@kaspersky this is not c++ the cast is not necessary i think.

src/bin/sol-fbp-runner/runner.c
ceolin: same too

src/lib/flow/include/sol-flow-metatype.h
cabelitos: I meant, in case we change to append data to the buffer you should say what I stated. Otherwise you should also add a note saying thet the buffer will be initialized by the function and all previous data will be lost/leaked

src/bin/sol-fbp-generator/main.c
barbieri: the actual problem is the function signature, this `buf` is completely non-sense, likely it's a left over.

All that is using `buffer` could be modified to take a `sol_str_slice`, thus the `read_file` and `sol_flow_parse_buffer()`. Or even better, if we moved this to `struct sol_blob`, then one can even remove the shadow list `file_readers` by using `sol_file_reader_to_blob()`. Users should take a reference of the blob, use the `sol_str_slice_from_blob()` when using the parsing, then `sol_blob_unref()` when done, so the file can be closed.

The code is ancient and predates things like blobs, could use a review.  Can you do it @kaspersky ?

kaspersky: @barbieri, Actually, using ```sol_str_slice``` was exactly what I wanted to do in the beginning. I will look into this ```sol_blob``` structure, and see how it can be used in this context. For now, I will take out this commit out of this PR so the other commits could be integrated.

barbieri: Ok. To save you some time in the understanding: the purpose of `sol_blob` is to reference count read only memory and use a custom function to release it. It may have a parent to keep "views" of data, in this case the child keeps a reference to parent. Releasing the last reference of the child will unref its parent.

With that in mind we created blobs with various free functions (or no_data_free/no_free to not free the data -- to be used in views -- and not free the blob at all, like when the blob itself is part of something else). One of them is a blob to a file reader, that keeps the file opened and the memory mmap()ed, so the file is no allocated and can be evicted from RAM even if no swap is present.

Then if the `read_file` is changed to return a blob, once that blob is unreferenced, it would munmap()/close() right when it's not used anymore. There is no need to keep the list of opened files to be closed once the program is closed.

Functions that are processing "text" shouldn't take a blob, but a str_slice. Unless they need to keep the memory alive after they return.

kaspersky:
#2155

PR #2066, opened 2016-05-19
Fix some compiler warnings

otaviobp:


src/samples/design_patterns/stream_sample.c
```diff
@@ -245,7 +245,7 @@ struct my_stream_api_handle *
 my_stream_api_new(const struct my_stream_api_config *config, int dev)
 {
     struct my_stream_api_handle *handle;
-    size_t data_buffer_size;
+    size_t data_buffer_size = 0;
```
cabelitos: I really don't know why gcc did not warn me about this one

src/modules/flow/process/output.c
```diff
@@ -76,7 +76,7 @@ output_write(struct output_data *output)
         } else if (r < 0) {
             if (errno == EINTR)
                 continue;
-            else if (errno == EAGAIN || errno == EWOULDBLOCK)
+            else if (errno == EAGAIN)
```
cabelitos: Well, what are we going to do on those "non supported systems" if the values are different?

ibriano: On Thu, 19 May 2016 10:35:02 -0700, Guilherme Íscaro wrote:
> > @@ -76,7 +76,7 @@ output_write(struct output_data *output)
> >          } else if (r < 0) {
> >              if (errno == EINTR)
> >                  continue;
> > -            else if (errno == EAGAIN || errno == EWOULDBLOCK)
> > +            else if (errno == EAGAIN)
> 
> Well, what are we going to do on those "non supported systems" if the values are different?

Plan on porting Soletta for AIX?

cabelitos: You never know....

otaviobp:
merged.

PR #2067, opened 2016-05-19
networkv2: Fix api names and functions

amandawrcoelho:
issue #1716

### DIFF
--- a/src/lib/comms/sol-network-impl-contiki.c
+++ b/src/lib/comms/sol-network-impl-contiki.c
@@ -46,7 +46,7 @@ sol_network_link_addr_to_str(const struct sol_network_link_addr *addr,
     if (addr->family != SOL_NETWORK_FAMILY_INET6)
         return NULL;
 
-    if (buf->capacity - buf->used < SOL_INET_ADDR_STRLEN)
+    if (buf->capacity - buf->used < SOL_NETWORK_INET_ADDR_STR_LEN)
         return NULL;
 
     p = sol_buffer_at_end(buf);
@@ -161,20 +161,20 @@ sol_network_shutdown(void)
     return;
 }
 
-SOL_API bool
+SOL_API int
 sol_network_subscribe_events(void (*cb)(void *data, const struct sol_network_link *link,
     enum sol_network_event event),
     const void *data)
 {
-    return false;
+    return -errno;
 }
 
-SOL_API bool
+SOL_API int
 sol_network_unsubscribe_events(void (*cb)(void *data, const struct sol_network_link *link,
     enum sol_network_event event),
     const void *data)
 {
-    return false;
+    return -EFAULT;
 }
 
 SOL_API const struct sol_vector *
--- a/src/lib/comms/sol-network-impl-linux.c
+++ b/src/lib/comms/sol-network-impl-linux.c
@@ -51,7 +51,7 @@ struct callback {
     const void *data;
 };
 
-struct sol_network_hostname_handle {
+struct sol_network_hostname_pending {
     char *hostname;
     enum sol_network_family family;
     void (*cb)(void *data, const struct sol_str_slice hostname,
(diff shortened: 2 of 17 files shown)


src/lib/comms/sol-network-impl-contiki.c
```diff
@@ -161,20 +161,20 @@ sol_network_shutdown(void)
...
 }
 
-SOL_API bool
+SOL_API int
 sol_network_subscribe_events(void (*cb)(void *data, const struct sol_network_link *link,
     enum sol_network_event event),
     const void *data)
 {
-    return false;
+    return -errno;
```
cabelitos: This is not quite correctly. You must return the errno value not the value of the errno variable.
What I mean, is that you need to do something like: `return -ENOSYS;`

src/lib/comms/sol-network-impl-contiki.c
```diff
@@ -161,20 +161,20 @@ sol_network_shutdown(void)
...
 }
 
-SOL_API bool
+SOL_API int
 sol_network_unsubscribe_events(void (*cb)(void *data, const struct sol_network_link *link,
     enum sol_network_event event),
     const void *data)
 {
-    return false;
+    return -EFAULT;
```
cabelitos: Use -ENOSYS here

src/lib/comms/sol-network-impl-linux.c
```diff
@@ -464,24 +464,24 @@ sol_network_subscribe_events(void (*cb)(void *data, const struct sol_network_lin
 
     SOL_NULL_CHECK(cb, false);
 
-    if (sol_network_start_netlink() < 0) return false;
+    if (sol_network_start_netlink() < 0) return -errno;
```
cabelitos: Again, return the `errno` value is not correct in this case
I suggest that you do the following:
```C
int r;
r = sol_network_start_netlink();
SOL_INT_CHECK(r, < 0, r);
```

src/lib/comms/sol-network-impl-linux.c
```diff
@@ -464,24 +464,24 @@ sol_network_subscribe_events(void (*cb)(void *data, const struct sol_network_lin
...
 
     callback = sol_vector_append(&network->callbacks);
     SOL_NULL_CHECK(callback, false);
```
cabelitos: You must also change this return value. Replace `false` with `-ENOMEM`
Remenber to check all return values and replace `true` and `false` with the proper values

src/lib/comms/sol-network-impl-linux.c
```diff
@@ -580,13 +580,13 @@ sol_network_link_set_status(uint16_t link_index, unsigned int changes, unsigned
     return true;
 }
 
-SOL_API bool
+SOL_API int
 sol_network_link_up(uint16_t link_index)
 {
     return sol_network_link_set_status(link_index, IFF_UP, IFF_UP);
```
cabelitos: This is wrong. `sol_network_link_set_status()` returns a `bool` and not an `int`

src/lib/comms/sol-network-impl-linux.c
```diff
@@ -580,13 +580,13 @@ sol_network_link_set_status(uint16_t link_index, unsigned int changes, unsigned
...
 }
 
-SOL_API bool
+SOL_API int
 sol_network_link_down(uint16_t link_index)
 {
     return sol_network_link_set_status(link_index, IFF_UP, ~IFF_UP);
```
cabelitos: Same as above

src/lib/comms/sol-network-impl-riot.c
```diff
@@ -173,20 +173,20 @@ sol_network_shutdown(void)
...
 }
 
-SOL_API bool
+SOL_API int
 sol_network_subscribe_events(void (*cb)(void *data, const struct sol_network_link *link,
     enum sol_network_event event),
     const void *data)
 {
-    return false;
+    return -errno;
```
cabelitos: replace `-errno` with `-ENOSYS`

src/lib/comms/sol-network-impl-contiki.c
ceolin: yes, you do not even know what is the value in the errno variable

src/lib/comms/sol-network-impl-linux.c
amandawrcoelho: In this case, should I change the sol_network_link_status( ) or should I keep it as the same way and change sol_network_link_up( ) and change sol_network_link_down( )?

cabelitos: @amandawrcoelho You should change the functions `sol_network_link_set_status()` return value type

PR #2069, opened 2016-05-20
Add netctl fbp node type and samples for list services , Connect or Disconnect AP

guchaojie:


src/samples/flow/netctl/network-connect.fbp
```diff
@@ -0,0 +1,29 @@
...
+# limitations under the License.
+
+# This program will print to console a list of all network services
+#
+# Examples:
+#  * connect or disconnect network services
+#    $ ./network-connect.fbp Guest
+
+ap(app/argv:index=1)
+ap OUT -> CONNECT conn(netctl/connection) OUT -> IN _(console)
```
barbieri: the node should be called `netctl/service`, not `connection`

src/samples/flow/netctl/network-connect.fbp
```diff
@@ -0,0 +1,29 @@
...
+
+ap ERROR -> IN _(converter/empty-to-string:output_value="List all the network services") OUT -> IN Info(console)
```
barbieri: what's the point in this?

src/modules/flow/netctl/netctl.json
```diff
@@ -0,0 +1,80 @@
...
+	"close": "close_service_connection"
+      },
+      "name": "netctl/connection",
+      "options": {
+        "members": [
+          {
+            "data_type": "string",
+            "default": null,
+            "description": "Name of AP to be used to connect. If 'null', no filter is applied.",
+            "name": "connect_ap"
```
barbieri: "AP" is not right. It should be "service". If the service is an access point (in wifi terms) or something else, it's not visible to the user here.

src/modules/flow/netctl/netctl.json
```diff
@@ -0,0 +1,80 @@
...
+          "description": "Name of network AP.",
+          "methods": {
+            "process": "set_disconnect_ap_name"
+          },
+          "name": "DISCONNECT"
+        }
+      ],
+      "methods": {
+        "open": "open_service_connection",
+	"close": "close_service_connection"
```
barbieri: indentation error

src/modules/flow/netctl/netctl.json
```diff
@@ -0,0 +1,80 @@
...
+      "out_ports": [
+        {
+          "data_type": "string",
+          "description": "Name of network services supported by system.",
+          "name": "OUT"
+        }
+      ],
+      "private_data_type": "list_service_data"
+    },
+    {
```
barbieri: all in all, `netctl/connection` should be `netctl/service` and be close to https://github.com/solettaproject/soletta/blob/master/src/modules/flow/platform/platform.json#L573 (you can use `CONNECT` instead of `START` and `DISCONNECT` instead of `STOP`), but try to keep the overall usage, like `NAME` to set the service name OR use it from options. Also use a string port `OUT` as you did, but you may add helper `ONLINE` or `CONNECTED` boolean ports to notify in an easy way if that service is usable.

PR #2070, opened 2016-05-20
flow:iio: Code refactor for process/close functions

yongli3:
Modify the data structure, to use the process/close functions for a node types

### DIFF
--- a/src/modules/flow/iio/iio.json
+++ b/src/modules/flow/iio/iio.json
@@ -16,12 +16,12 @@
           "description": "Packets sent in here will trigger a gyroscope reading and produce packets on each of the the output ports. If buffer is enabled, current trigger must accept manual activation (default one, sysfs trigger, does).",
           "name": "TICK",
           "methods": {
-            "process": "gyroscope_tick"
+            "process": "iiodevice_tick"
           }
         }
       ],
       "methods": {
-        "close": "gyroscope_close",
+        "close": "iiodevice_close",
         "open": "gyroscope_open"
       },
       "name": "iio/gyroscope",
@@ -96,7 +96,7 @@
          "name": "OUT"
         }
       ],
-      "private_data_type": "gyroscope_data"
+      "private_data_type": "iiodevice_data"
     },
     {
       "category": "input/hw",
@@ -107,12 +107,12 @@
           "description": "Packets sent in here will trigger a magnet reading and produce packets on each of the the output ports. If buffer is enabled, current trigger must accept manual activation (default one, sysfs trigger, does).",
           "name": "TICK",
           "methods": {
-            "process": "magnet_tick"
+            "process": "iiodevice_tick"
           }
         }
       ],
       "methods": {
-        "close": "magnet_close",
+        "close": "iiodevice_close",
         "open": "magnet_open"
       },
       "name": "iio/magnetometer",
@@ -188,7 +188,7 @@
          "name": "OUT"
         }
       ],
-      "private_data_type": "magnet_data"
+      "private_data_type": "iiodevice_data"
     },
     {
       "category": "input/hw",
@@ -199,12 +199,12 @@
           "description": "Packets sent in here will trigger a temperature reading and produce packets on each of the the output ports. If buffer is enabled, current trigger must accept manual activation (default one, sysfs trigger, does).",
           "name": "TICK",
           "methods": {
-            "process": "temperature_tick"
+            "process": "iiodevice_tick"
           }
         }
       ],
       "methods": {
-        "close": "temperature_close",
+        "close": "iiodevice_close",
         "open": "temperature_open"
       },
       "name": "iio/thermometer",
@@ -280,7 +280,7 @@
          "name": "OUT"
         }
       ],
-      "private_data_type": "temperature_data"
+      "private_data_type": "iiodevice_data"
     },
     {
       "category": "input/hw",
@@ -291,12 +291,12 @@
           "description": "Packets sent in here will trigger a pressure reading and produce packets on each of the the output ports. If buffer is enabled, current trigger must accept manual activation (default one, sysfs trigger, does).",
           "name": "TICK",
           "methods": {
-            "process": "pressure_tick"
+            "process": "iiodevice_tick"
           }
         }
       ],
       "methods": {
-        "close": "pressure_close",
+        "close": "iiodevice_close",
         "open": "pressure_open"
       },
       "name": "iio/pressure-sensor",
@@ -372,7 +372,7 @@
          "name": "OUT"
         }
       ],
-      "private_data_type": "pressure_data"
+      "private_data_type": "iiodevice_data"
     },
     {
       "category": "input/hw",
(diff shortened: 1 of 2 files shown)


yongli3:
barbieri  @edersondisouza  @neofang @peijiajames This PR is based on these comments on PR  #2013. Please help to review
This is the first step for the code refactor, I think there are about another 2 changes are needed.
By separating into several PRs, I think it is easy for review/test

src/modules/flow/iio/iio.json
```diff
@@ -421,8 +421,8 @@
            "name": "use_device_default_scale"
          },
          {
-             "data_type": "float",
-             "default": 0,
+             "data_type": "direction-vector",
+             "default": {},
```
barbieri: this is not right, what does a direction vector has to do here? Also for other nodes?

src/modules/flow/iio/nodes.c
```diff
@@ -25,39 +25,41 @@
...
-    struct sol_iio_channel *channel_x;
-    struct sol_iio_channel *channel_y;
-    struct sol_iio_channel *channel_z;
+    uint8_t channel_count;
     bool buffer_enabled : 1;
     bool use_device_default_scale : 1;
     bool use_device_default_offset : 1;
+    struct sol_iio_channel *channels[MAX_CHANNEL];
+    double scales[MAX_CHANNEL];
+    double offsets[MAX_CHANNEL];
```
barbieri: why not using a `struct sol_vector` with a `struct { struct sol_iio_channel *channel; double scale; double offset; }`?

barbieri: also move types with larger alginment needs up in the structure to avoid holes. Boolean/uint8_t alignment is 1, those pointers are 4/8.

src/modules/flow/iio/iio.json
yongli3: @barbieri the color sensor has three channel values: red/green/blue. So we need to use three different values for scale and offset

src/modules/flow/iio/nodes.c
```diff
@@ -25,39 +25,41 @@
...
 
-    b = sol_iio_read_channel_value(mdata->channel_y, &out.y);
+    b = sol_iio_read_channel_value(mdata->channels[1], &out.y);
     if (!b) goto error;
 
-    b = sol_iio_read_channel_value(mdata->channel_z, &out.z);
+    b = sol_iio_read_channel_value(mdata->channels[2], &out.z);
     if (!b) goto error;
 
     sol_flow_send_direction_vector_packet(node,
```
barbieri: as @edersondisouza pointed out, this is not right as a supposedly "general" `read_cb` is assuming x/y/z and a direction-vector. The port is also being assumed to be `SOL_FLOW_NODE_TYPE_IIO_GYROSCOPE__OUT_OUT`.

As I wrote in my review, there should be multiple `read_cb`, one per interface, like `read_direction_vector_cb`, `read_color_cb`... These should be set in your extended type (as done with port index) with `"extra_methods": {"read_cb": "read_direction_vector_cb"}` in JSON. Instead of registering `read_cb` as you do now, you register `type->read_cb`.

Each `read_XXX_cb` should use `type->port_out` to use the port index, also defined in `extra_methods`.

src/modules/flow/iio/nodes.c
```diff
@@ -71,24 +73,24 @@ reader_cb(void *data, struct sol_iio_device *device)
...
+        channel_config.offset = mdata->offsets[_index]; \
+    mdata->channels[_index] = sol_iio_add_channel(mdata->device, "in_anglvel_" # _axis, &channel_config); \
+    SOL_NULL_CHECK_GOTO(mdata->channels[_index], error);
 
-    ADD_CHANNEL(x);
-    ADD_CHANNEL(y);
-    ADD_CHANNEL(z);
+    ADD_CHANNEL(x, 0);
+    ADD_CHANNEL(y, 1);
+    ADD_CHANNEL(z, 2);
```
barbieri: also not correct since this is only meaningful for direction vector.

src/modules/flow/iio/nodes.c
```diff
@@ -106,10 +108,12 @@ create_device_channels(struct gyroscope_data *mdata, int device_id)
 static int
 gyroscope_open(struct sol_flow_node *node, void *data, const struct sol_flow_node_options *options)
 {
-    struct gyroscope_data *mdata = data;
+    struct iiodevice_data *mdata = data;
     const struct sol_flow_node_type_iio_gyroscope_options *opts;
     int device_id;
 
+    SOL_WRN("%s-%s:%s\n", __func__, __DATE__, __TIME__);
```
barbieri: never use `__DATE__` and `__TIME__` in logs. The logs already include the function name and `\n`.

src/modules/flow/iio/nodes.c
```diff
@@ -153,27 +160,30 @@ gyroscope_open(struct sol_flow_node *node, void *data, const struct sol_flow_nod
...
 
 static int
-gyroscope_tick(struct sol_flow_node *node, void *data, uint16_t port, uint16_t conn_id, const struct sol_flow_packet *packet)
+iiodevice_tick(struct sol_flow_node *node, void *data, uint16_t port, uint16_t conn_id, const struct sol_flow_packet *packet)
 {
     static const char *errmsg = "Could not read channel values";
-    struct gyroscope_data *mdata = data;
+    struct iiodevice_data *mdata = data;
+
+    SOL_WRN("%s mdata=%p \n", __func__, mdata);
```
barbieri: as above, this should not exist.

src/modules/flow/iio/nodes.c
```diff
@@ -153,27 +160,30 @@ gyroscope_open(struct sol_flow_node *node, void *data, const struct sol_flow_nod
...
 
     if (mdata->buffer_enabled) {
         if (!sol_iio_device_trigger_now(mdata->device))
             goto error;
     } else {
-        reader_cb(node, mdata->device);
+        mdata->config.sol_iio_reader_cb(node, mdata->device);
```
barbieri:
```c
const struct iio_node_type *type = (const struct iio_node_type *)sol_flow_node_get_type(node);
type->read_cb(node, mdata->device);
```

This will work once you define your `iio_node_type` extending `sol_flow_node_type`. See how it's done in http-server:
 - https://github.com/solettaproject/soletta/blob/master/src/modules/flow/http-server/http-server.c#L74 (struct declaration)
 - https://github.com/solettaproject/soletta/blob/master/src/modules/flow/http-server/http-server.json#L22 (json declaration)
 - https://github.com/solettaproject/soletta/blob/master/src/modules/flow/http-server/http-server.c#L234 (fetch and cast)

src/modules/flow/iio/iio.json
barbieri: that is right, but it's not a direction vector. I think it's better to go with separate options, one per channel.

bdilly: @yongli3 @barbieri I would recommend to use a single "rgb" option.
We have an option data type specific for this purpose.

barbieri: @bdilly AFAIU he wants a scale, like "When I read RED, multiply it by 0.3".

In that sense I wonder if we do really need this, given that the output RGB packet can hold the maximum value of components, then we can do the scale/map afterwards.

edersondisouza: @barbieri I didn't get the 'afterwards' part. This option is actually to define in which scale(or offset) device is operating...
Anyway, as `sol_rgb` doesn't hold doubles, using three separated options should be the way to go. Using `direction-vector` is misleading...

yongli3:
Close this one, replaced with #2102

PR #2073, opened 2016-05-20
Network - v3 - fixed all the appointments

amandawrcoelho:
Changed all the -errno returns.

issue #1716

### DIFF
--- a/src/lib/comms/sol-network-impl-linux.c
+++ b/src/lib/comms/sol-network-impl-linux.c
@@ -51,7 +51,7 @@ struct callback {
     const void *data;
 };
 
-struct sol_network_hostname_handle {
+struct sol_network_hostname_pending {
     char *hostname;
     enum sol_network_family family;
     void (*cb)(void *data, const struct sol_str_slice hostname,
@@ -96,7 +96,7 @@ sol_network_link_addr_to_str(const struct sol_network_link_addr *addr,
         if (r || (!r && errno != ENOSPC))
             break;
 
-        err = sol_buffer_expand(buf, SOL_INET_ADDR_STRLEN);
+        err = sol_buffer_expand(buf, SOL_NETWORK_INET_ADDR_STR_LEN);
         SOL_INT_CHECK(err, < 0, NULL);
     }
 
@@ -353,7 +353,7 @@ sol_network_start_netlink(void)
     network->nl_socket = socket(AF_NETLINK, SOCK_RAW | SOCK_CLOEXEC | SOCK_NONBLOCK, NETLINK_ROUTE);
     if (network->nl_socket < 0) {
         SOL_WRN("failed to create netlink socket, cannot listen to events or manage the links!");
-        return -errno;
+        return -EBADMSG;
     }
 
     network->nl_addr.nl_family = AF_NETLINK;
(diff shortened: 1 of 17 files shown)


src/lib/comms/sol-network-impl-linux.c
```diff
@@ -574,19 +577,19 @@ sol_network_link_set_status(uint16_t link_index, unsigned int changes, unsigned
 
     if (sendmsg(network->nl_socket, (struct msghdr *)&msg, 0) <= 0) {
         SOL_WRN("Failed on send message to set link up");
-        return false;
+        return -errno;
```
amandawrcoelho: Change it

src/lib/comms/sol-network-impl-linux.c
```diff
@@ -455,42 +455,45 @@ sol_network_shutdown(void)
...
 {
+    int r;
+
     struct callback *callback;
 
-    SOL_NULL_CHECK(cb, false);
+    SOL_NULL_CHECK(cb, -EINVAL);
 
-    if (sol_network_start_netlink() < 0) return false;
+    r = sol_network_start_netlink;
```
cabelitos: Did you compile this? I guess you forgot to add `()`

src/lib/comms/sol-network-impl-linux.c
```diff
@@ -532,7 +535,7 @@ sol_network_link_set_status(uint16_t link_index, unsigned int changes, unsigned
     uint8_t mode = IN6_ADDR_GEN_MODE_EUI64;
     size_t buf_size = sizeof(buf);;
 
-    if (sol_network_start_netlink() < 0) return false;
+    if (sol_network_start_netlink() < 0) return -ENOMEM;
```
cabelitos: No no, you can't simple return `-ENOMEM` here. The function `sol_network_start_netlink()` already returns the error value. You should do the following:

```
int r;
...
r = sol_network_start_netlink();
SOL_INT_CHECK(r, < 0, r);
```

src/lib/comms/sol-network-impl-linux.c
```diff
@@ -353,7 +353,7 @@ sol_network_start_netlink(void)
     network->nl_socket = socket(AF_NETLINK, SOCK_RAW | SOCK_CLOEXEC | SOCK_NONBLOCK, NETLINK_ROUTE);
     if (network->nl_socket < 0) {
         SOL_WRN("failed to create netlink socket, cannot listen to events or manage the links!");
-        return -errno;
+        return -EBADMSG;
```
cabelitos: In this case returning the `-errno` is valid. The `socket()` will set the errno with the correctly value if something goes wrong

cabelitos: Remenber to check the man pages.

src/lib/comms/sol-network-impl-linux.c
cabelitos: What do you mean?

ibriano: Careful with those, even the if the function uses errno to set the error, SOL_WRN (and related) may re-set errno to some other value. In these cases you need to store errno to a temporary variable before you call anything else, and then return (or set back) that value.

cabelitos: I checked. SOL_WRN will not mess with the errno value

PR #2074, opened 2016-05-20
Add the Bluetooth agent API (for BlueZ) v2

vcgomes:
This adds the concept of an agent, that will serve to receive input from the user during the pairing process primarily. 

The API is pretty similar to what Zephyr provides, which is based on the BlueZ D-Bus API, so this should map pretty well to both.  When reviewing, please see if the approach of the `sol_bt_agent_reply_*` family of functions is easy to understand.

Changes from last version:
- BlueZ D-Bus API has changed (but still marked as experimental, so no API compatibility is promised), this version uses the current upstream API;
- More helpful information when browsing for devices, i.e. the device name is printed;
- With the new D-Bus API, it is now possible to obtain the connection in which a GATT operation is happening.

### DIFF
--- a/src/lib/common/sol-bus.h
+++ b/src/lib/common/sol-bus.h
@@ -272,3 +272,50 @@ int sol_bus_remove_interfaces_watch(struct sol_bus_client *client,
  */
 int sol_bus_log_callback(sd_bus_message *reply, void *userdata,
     sd_bus_error *ret_error);
+
+/**
+ * @brief Represents an entry in a D-Bus dictionary
+ *
+ * A common idiom in D-Bus is having an array of type 'a{sv}', representing
+ * a dictionary, this struct and sol_bus_parse_dict() help parsing those
+ * data structures.
+ */
+struct sol_bus_dict_entry {
+    /**
+     * @brief The name of the dictionary entry
+     */
+    const char *name;
+    /**
+     * @brief The type of value of the property in the dictionary
+     *
+     * It can be any of the basic types, or 'v' (SD_BUS_TYPE_VARIANT), in
+     * case the type is not basic, in this case, the function parse_variant() will
+     * be called.
+     */
+    char type;
+    /**
+     * @brief In which place to store the parsed value
+     */
+    void *value;
+    /**
+     * @brief Function to be called in case the value being parsed is not basic.
+     */
+    int (*parse_variant)(sd_bus_message *m, void *value);
+};
+
+/**
+ * @brief Represents the final entry in a dictionary parser.
+ */
+#define SOL_BUS_DICT_ENTRY_FINAL { .name = NULL }
+
+/**
+ * @brief Helper to parse D-Bus dictionaries
+ *
+ * @see #sol_bus_dict_entry
+ *
+ * @param m D-Bus message containing the D-Bus dictionary array
+ * @param dict Dictionary representation, informing how to parse the dictionary.
+ *
+ * @return 0 on success, -errno otherwise.
+ */
+int sol_bus_parse_dict(sd_bus_message *m, const struct sol_bus_dict_entry *dict);
--- a/src/lib/comms/include/sol-bluetooth.h
+++ b/src/lib/comms/include/sol-bluetooth.h
@@ -155,6 +155,15 @@ const struct sol_network_link_addr *sol_bt_conn_get_addr(
     const struct sol_bt_conn *conn);
 
 /**
+ * @brief Returns the device info associated with a connection.
+ *
+ * @param conn The reference to a connection.
+ *
+ * @return Information about the device connected
+ */
+const struct sol_bt_device_info *sol_bt_conn_get_device_info(const struct sol_bt_conn *conn);
+
+/**
  * @brief Attempts to establish a connection with a remote device.
  *
  * @param addr The network link address of the remote device,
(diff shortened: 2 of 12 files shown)


src/lib/comms/include/sol-bluetooth.h
```diff
@@ -328,6 +337,212 @@ struct sol_bt_scan_pending *sol_bt_start_scan(
...
+/**
+ * @brief Registers an agent for the system
+ *
+ * It is only possible to have one agent at a time.
+ *
+ * @param agent The agent to be registered
+ * @param data The user data to be passed to each agent callback
+ * @return 0 on sucess, -errno otherwise
+ */
+int sol_bt_register_agent(const struct sol_bt_agent *agent, void *data);
```
barbieri: minor, but `const void *data` and also state if `agent` is copied or not, if not mention the memory must be kept alive during the agent lifetime (ie: no stack!)

Also, why not use the pattern "null to unregister"? we use it elsewhere. In this case to prevent mistakes, only accept a non-NULL agent if none was set. See https://github.com/solettaproject/soletta/blob/master/src/lib/io/sol-ipm.c#L235

src/lib/comms/include/sol-bluetooth.h
```diff
@@ -328,6 +337,212 @@ struct sol_bt_scan_pending *sol_bt_start_scan(
...
+/**
+ * @brief Confirms the pairing attempt
+ *
+ * This should be called after pairing_confirm() indicates a pairing attempt.
+ *
+ * @param conn The connection to be authenticated
+ * @param passkey The passkey entered by the user
+ * @return 0 on sucess, -errno otherwise
+ */
+int sol_bt_agent_reply_pairing_confirm(struct sol_bt_conn *conn);
```
barbieri: I wonder if `reply` is the best term. Maye `dismiss` or even better `finish`, since it may fit better all cases (`reply_cancel()` is kinda confusing)

src/lib/comms/include/sol-bluetooth.h
```diff
@@ -328,6 +337,212 @@ struct sol_bt_scan_pending *sol_bt_start_scan(
...
+ * @brief Replies to a request to the user to enter a pincode
+ *
+ * This should be called after passkey_entry() is called with the passkey
+ * entered by the user.
+ *
+ * @param conn The connection to be authenticated
+ * @param passkey The passkey entered by the user
+ * @return 0 on sucess, -errno otherwise
+ */
+int sol_bt_agent_reply_pincode_entry(struct sol_bt_conn *conn, const char *pin);
```
barbieri: pin code as a string?

src/lib/comms/sol-bluetooth-impl-bluez.c
```diff
@@ -294,6 +296,23 @@ destroy_device(struct device_info *device)
     free(device);
 }
 
+static void
+destroy_pairing(struct context *ctx, bool success)
+{
+    if (ctx->pair_cb)
+        ctx->pair_cb((void *)ctx->pair_user_data, success, ctx->auth_conn);
```
barbieri: Is there any chance `ctx` can be destroyed from the user callback?

src/lib/comms/sol-bluetooth-impl-bluez.c
```diff
@@ -1581,6 +1640,613 @@ sol_bt_stop_scan(struct sol_bt_scan_pending *scan)
...
+    sd_bus *bus = sol_bus_client_get_bus(ctx->bluez);
+    const char *service = sol_bus_client_get_service(ctx->bluez);
+    const char *path = AGENT_PATH;
+    const char *capabilities;
+    int r;
+
+    SOL_NULL_CHECK(agent, -EINVAL);
+
+    if (ctx->agent)
+        return -EALREADY;
```
barbieri: In `sol-ipm` we're using `-EEXIST` for this error, see https://github.com/solettaproject/soletta/blob/master/src/lib/io/sol-ipm.c#L254

src/samples/bluetooth/simple-pair.c
```diff
@@ -0,0 +1,184 @@
...
+        const struct sol_network_link_addr *addr;
+
+        addr = sol_network_link_addr_from_str(&pair_addr, sol_argv()[1]);
+        SOL_NULL_CHECK(addr);
+    }
+
+    session = sol_bt_enable(enabled, NULL);
+    if (!session) {
+        SOL_WRN("Couldn't create a Bluetooth session");
+        sol_quit_with_code(-ENOMEM);
```
barbieri: this is the exit code of the application, like `EXIT_FAILURE`, not a negative number

src/samples/bluetooth/simple-pair.c
```diff
@@ -0,0 +1,184 @@
...
+}
+
+static void
+startup(void)
+{
+    if (sol_argc() > 1) {
+        const struct sol_network_link_addr *addr;
+
+        addr = sol_network_link_addr_from_str(&pair_addr, sol_argv()[1]);
+        SOL_NULL_CHECK(addr);
```
barbieri: this would not quit the application when an invalid address is used, is that ok?

src/samples/bluetooth/simple-pair.c
```diff
@@ -0,0 +1,184 @@
...
+        return;
+
+    SOL_INF("Bluetooth Adapter enabled");
+
+    scan = sol_bt_start_scan(SOL_BT_TRANSPORT_ALL, found_device, NULL);
+    SOL_NULL_CHECK(scan);
+
+    r = sol_bt_register_agent(&agent, NULL);
+    if (r < 0) {
+        SOL_WRN("r (%d) < 0", r);
```
barbieri: use the `strerror()` to convert the error to string?

src/samples/bluetooth/simple-pair.c
```diff
@@ -0,0 +1,184 @@
...
+
+static struct sol_bt_conn *auth_conn;
+static struct sol_network_link_addr pair_addr;
+
+static struct sol_timeout *timeout;
+
+static void
+on_error(void *user_data, int error)
+{
+    SOL_DBG("error %d", error);
```
barbieri: our samples use regular `printf()` to stdout for what would be the program UI. The logs are used only for our internal code and modules, the user is free to use them, of course, but in the samples we don't .

src/samples/bluetooth/simple-pair.c
```diff
@@ -0,0 +1,184 @@
...
+
+    sol_bt_agent_reply_pairing_confirm(conn);
+}
+
+static struct sol_bt_agent agent = {
+    .pairing_confirm = pairing_confirm,
+};
+
+static void
+enabled(void *data, bool powered)
```
barbieri: many callbacks with different naming schema. `enabled`, `timeout_cb` and `on_connected`. Let's pick one :-)

src/lib/common/sol-bus.h
```diff
@@ -272,3 +272,50 @@ int sol_bus_remove_interfaces_watch(struct sol_bus_client *client,
...
+    char type;
+    /**
+     * @brief In which place to store the parsed value
+     */
+    void *value;
+    /**
+     * @brief Function to be called in case the value being parsed is not basic.
+     */
+    int (*parse_variant)(sd_bus_message *m, void *value);
+};
```
barbieri: I like the idea, but I'm trying to unify these things with `sol-memdesc`. Do you think we should try it here, or wouldn't help?

src/lib/comms/sol-gatt-impl-bluez.c
```diff
@@ -147,52 +159,77 @@ sol_gatt_pending_reply(struct sol_gatt_pending *pending, int error,
...
     return r;
 }
 
 static int
+parse_operation_dict(sd_bus_message *m, uint16_t *offset, const char **path)
+{
+    const struct sol_bus_dict_entry dict[] = {
+        { .name = "offset", .type = 'q', .value = offset },
+        { .name = "device", .type = 'o', .value = path },
+        SOL_BUS_DICT_ENTRY_FINAL
```
barbieri: we're using an empty `{}` to be the sentinel.

barbieri:
overall looks ok, minor API alignments and the sample could be improved.

I see lots of potential to have the agent exposed in a node type, our demo could use the form and lcd-string to do a nice pairing interface. Talk to @edersondisouza and @glima since they know these well.

src/lib/comms/include/sol-bluetooth.h
vcgomes: I like `finish`, let's see how it looks.

src/lib/comms/include/sol-bluetooth.h
vcgomes: In practice this will only be used when pairing with legacy keyboards, when you type the same thing in the devices being paired, some keyboards allow you to type anything, not just numbers...

src/lib/comms/sol-bluetooth-impl-bluez.c
vcgomes: Not really, the `context` keeps track of the bluetoothd service name.

src/lib/comms/sol-bluetooth-impl-bluez.c
vcgomes: Will fix.

src/samples/bluetooth/simple-pair.c
vcgomes: Ugh. Will fix.

src/samples/bluetooth/simple-pair.c
vcgomes: Yeah, I think so, in case an invalid address is passed, this sample will do a scan and list all available devices.

src/samples/bluetooth/simple-pair.c
vcgomes: Yeah. Will fix.

src/samples/bluetooth/simple-pair.c
vcgomes: Will fix.

src/samples/bluetooth/simple-pair.c
vcgomes: `on_something` seems more common, will fix.

src/lib/comms/sol-gatt-impl-bluez.c
vcgomes: Cool. Will fix.

src/lib/common/sol-bus.h
vcgomes: Will take a look at `sol-memdesc`.

vcgomes: At this level, I don't see much advantage, what I can imagine being useful is something at the level of `sd_bus_message`, i.e. a D-Bus message representation based on `sol-memdesc`.

Am I not considering something?

src/lib/comms/include/sol-bluetooth.h
vcgomes: Fixed both issues.

vcgomes:
An updated version is at https://github.com/solettaproject/soletta/pull/2107. Closing this. Thanks.

PR #2077, opened 2016-05-20
coap: Fix build

ceolin:
It fixes the build when SOL_API_VERSION is disabled.


otaviobp:
+1

PR #2078, opened 2016-05-20
Memmap: add zephyr backend [v3]

glima:
Sample at small-os-samples, as well.

Changes since v1:
- Makefile issue fixed
- MAX_RW from Zephyr now
- Leak fixed

Changes since v2:
- sscanf() on Linux path
- path now for Zephyr as well, specifying mem specs


glima:
https://github.com/solettaproject/soletta-small-oses-samples/pull/21 as well

src/lib/io/Makefile
```diff
@@ -88,6 +88,10 @@ obj-io-storage-$(USE_EFIVARS) += \
     sol-efivarfs-storage.o
 obj-io-storage-$(USE_MEMMAP) += \
     sol-memmap-storage.o
+obj-io-storage-$(LINUX) += \
+    sol-memmap-storage-impl-linux.o
+obj-io-storage-$(ZEPHYR) += \
+    sol-memmap-storage-impl-zephyr.o
```
barbieri: should these be inside `if` blocks? This would include the zephyr or linux objects even if no `USE_STORAGE` is defined.

src/lib/io/sol-memmap-storage-impl-zephyr.c
```diff
@@ -0,0 +1,285 @@
...
+#include "sol-str-slice.h"
+#include "sol-str-table.h"
+#include "sol-util-internal.h"
+#include "sol-vector.h"
+
+static struct device *flash_dev = NULL;
+
+#define ERASE_ALIGN (4096) // see sol_memmap_impl_write_raw()
+#define MAX_RW (256) //every read/write operation is limited to that
+                     //buffer size in Zephyr
```
barbieri: could we use this from some kconfig from zephyr? I'm not sure these will always be these values.

src/lib/io/sol-memmap-storage-impl-zephyr.c
```diff
@@ -0,0 +1,285 @@
...
+#include "sol-memmap-storage-impl.h"
+
+#include "sol-buffer.h"
+#include "sol-mainloop.h"
+#include "sol-str-slice.h"
+#include "sol-str-table.h"
+#include "sol-util-internal.h"
+#include "sol-vector.h"
+
+static struct device *flash_dev = NULL;
```
barbieri: why a single device? Can't we have multiple flash devices attached? Or there is a limitation in zephyr that inhibits that.

src/lib/io/sol-memmap-storage-impl-zephyr.c
```diff
@@ -0,0 +1,285 @@
...
+    uint64_t mask,
+    struct sol_blob *blob,
+    void (*cb)(void *data, const char *name, struct sol_blob *blob, int status),
+    const void *data)
+{
+    int r = 0;
+    uint8_t *back_mem;
+    size_t offset, mod, back_sz, cnt;
+
+    if (!sol_blob_ref(blob))
```
barbieri: any reason to ref/unref in this block? If you're not going back to main loop, there is no reason. And you're leaking it on error chcks (`SOL_NULL_CHECK(...)`)

src/lib/io/sol-memmap-storage-impl-zephyr.c
```diff
@@ -0,0 +1,285 @@
...
+
+        memcpy(&old_value, back_mem + mod, entry->size);
+
+        value <<= entry->bit_offset;
+        value &= mask;
+        value |= (old_value & ~mask);
+
+        memcpy(back_mem + mod, &value, entry->size);
+
+        r = flash_write_protection_set(flash_dev, false);
```
barbieri: it's already false, you did it before `if (mask)...`

src/lib/io/sol-memmap-storage-impl-zephyr.c
```diff
@@ -0,0 +1,285 @@
...
+
+        for (cnt = 0; cnt < back_sz; cnt += MAX_RW) {
+            r = flash_write(flash_dev, offset - mod + cnt, back_mem + cnt,
+                MAX_RW);
+            if (r < 0) {
+                SOL_WRN("Flash write failed");
+                goto error;
+            }
+            r = flash_write_protection_set(flash_dev, false);
+            SOL_INT_CHECK_GOTO(r, < 0, error);
```
barbieri: again, why reset to false?

src/lib/io/sol-memmap-storage-impl-zephyr.c
```diff
@@ -0,0 +1,285 @@
...
+    return false;
+}
+
+int
+sol_memmap_impl_init(void)
+{
+    if (flash_dev)
+        return 0;
+
+    flash_dev = device_get_binding("W25QXXDV");
```
barbieri: shouldn't this be configurable? Or at least specified in the memmap address field?

src/lib/io/sol-memmap-storage-impl-zephyr.c
```diff
@@ -0,0 +1,285 @@
...
+
+    tmp_vector = map_internal->pending_writes;
+    sol_vector_init(&map_internal->pending_writes,
+        sizeof(struct pending_write_data));
+
+    SOL_VECTOR_FOREACH_IDX (&tmp_vector, pending, i) {
+        sol_memmap_impl_write_raw((struct map_internal *)map_internal,
+            pending->name, pending->entry, pending->mask, pending->blob,
+            pending->cb, pending->data);
+        free(pending->name);
```
barbieri: we need to see why is this `pending->name` is being allocated and try to avoid that.

src/lib/io/sol-efivarfs-storage.c
```diff
@@ -214,11 +214,7 @@ sol_efivars_read_raw(const char *name, struct sol_buffer *buffer)
         goto end;
     }
 
-    if (buffer->capacity) {
-        r = sol_util_fill_buffer_exactly(fd, buffer, buffer->capacity);
-    } else {
-        r = sol_util_load_file_fd_buffer(fd, buffer);
-    }
+    r = sol_util_load_file_fd_buffer(fd, buffer);
```
barbieri: why is this change? I don't recall the details, but I remember once I tried to do this I ran into problems. Did you try with both fixed capacity and growing buffers?

src/lib/io/sol-memmap-storage-impl-linux.c
```diff
@@ -0,0 +1,373 @@
...
+resolve_i2c_path(const char *path, char **resolved_path)
+{
+    char *rel_path = NULL, *dev_number_s = NULL, *dev_name = NULL, *end_ptr;
+    struct sol_str_slice command = sol_str_slice_from_str(path);
+    struct sol_buffer result_path = SOL_BUFFER_INIT_EMPTY;
+    struct sol_vector instructions;
+    unsigned int dev_number;
+    int ret = -EINVAL;
+
+    instructions = sol_str_slice_split(command, ",", 5);
```
barbieri: wondering if this would be better if replaced with `sscanf()` using `%[ˆ,]` to parse `create`, `i2c`, `rel_path`...

barbieri: also `create,i2c,` is checked before calling this function, so can be ignored from `path`

src/lib/io/Makefile
glima: See the changes on the usages further

src/lib/io/sol-memmap-storage-impl-zephyr.c
glima: I'll check

src/lib/io/sol-memmap-storage-impl-zephyr.c
glima: Okay, maybe I was lazy here

src/lib/io/sol-memmap-storage-impl-zephyr.c
glima: This is legacy, I'll check

src/lib/io/sol-memmap-storage-impl-zephyr.c
glima: Mmm, thanks, I'll double check

src/lib/io/sol-memmap-storage-impl-zephyr.c
glima: After each write it protects itself again. Deal with it.

src/lib/io/sol-memmap-storage-impl-zephyr.c
glima: We're not exposing driver choice configuration on Soletta-Zephyr yet. Should we?

src/lib/io/sol-memmap-storage-impl-zephyr.c
glima: I'll check.

src/lib/io/sol-efivarfs-storage.c
glima: All them are growing buffers now (at least from the sole user, persistence.c).

src/lib/io/sol-memmap-storage-impl-linux.c
glima: This is legacy code, but I'll check.

src/lib/io/sol-memmap-storage-impl-zephyr.c
ibriano: On Mon, 23 May 2016 07:01:23 -0700, Gustavo Lima Chaves wrote:
> > +        value |= (old_value & ~mask);
> > +
> > +        memcpy(back_mem + mod, &value, entry->size);
> > +
> > +        r = flash_write_protection_set(flash_dev, false);
> > +        SOL_INT_CHECK_GOTO(r, < 0, error);
> > +
> > +        for (cnt = 0; cnt < back_sz; cnt += MAX_RW) {
> > +            r = flash_write(flash_dev, offset - mod + cnt, back_mem + cnt,
> > +                MAX_RW);
> > +            if (r < 0) {
> > +                SOL_WRN("Flash write failed");
> > +                goto error;
> > +            }
> > +            r = flash_write_protection_set(flash_dev, false);
> > +            SOL_INT_CHECK_GOTO(r, < 0, error);
> 
> After each write it protects itself again. Deal with it.

The Zephyr API does that? Or is a thing of that particular controller
and the Zephyr API is too close to it instead of trying to abstract
that away?

src/lib/io/sol-efivarfs-storage.c
barbieri: even fixed sizes like booleans, integers and so on?

src/lib/io/Makefile
glima: Okay, now I see your point, I'll change

src/lib/io/sol-memmap-storage-impl-zephyr.c
glima: The second was there, the 1st comes from the memory datasheet (it's the instruction used on erases), not much to do.

src/lib/io/sol-memmap-storage-impl-zephyr.c
glima: Thinking better, for the Arduino 101 SoC, which I target here, there's no sense on having more than one attached (it won't change). If we're having broader support later, we could fix this accordingly.

src/lib/io/sol-memmap-storage-impl-zephyr.c
glima: I guess it's to be safe around the user cb(). I'll make sure we don't leak the ref but still keep it, OK?

src/lib/io/sol-memmap-storage-impl-zephyr.c
glima: I did an erase in between, see next comment.

src/lib/io/sol-memmap-storage-impl-zephyr.c
glima: It's a particular thing of the flash mem itself. Yeah, Zephyr does not abstract any particularities of those memories.

ibriano: On Mon, 23 May 2016 10:01:17 -0700, Gustavo Lima Chaves wrote:
> > +        value |= (old_value & ~mask);
> > +
> > +        memcpy(back_mem + mod, &value, entry->size);
> > +
> > +        r = flash_write_protection_set(flash_dev, false);
> > +        SOL_INT_CHECK_GOTO(r, < 0, error);
> > +
> > +        for (cnt = 0; cnt < back_sz; cnt += MAX_RW) {
> > +            r = flash_write(flash_dev, offset - mod + cnt, back_mem + cnt,
> > +                MAX_RW);
> > +            if (r < 0) {
> > +                SOL_WRN("Flash write failed");
> > +                goto error;
> > +            }
> > +            r = flash_write_protection_set(flash_dev, false);
> > +            SOL_INT_CHECK_GOTO(r, < 0, error);
> 
> It's a particular thing of the flash mem itself. Yeah, Zephyr does not abstract any particularities of those memories.

Might be a good idea to report that to them in that case.

src/lib/io/sol-memmap-storage-impl-zephyr.c
glima: If the sole user of the storage API was the flow, we'd be safe: it's the "name" option of those nodes, that lives while the node itself lives. But since ``sol_memmap_write_raw()`` is a public C API as well, I wonder if we should really be safe WRT that name.

src/lib/io/sol-efivarfs-storage.c
glima: Yeah, to accommodate the case where they may lie on a bit field. On my sample I made sure this case was tested:
```
    {
     "name": "bool_test",
     "offset": 3,
     "size": 4,
     "bit_size": 1,
     "bit_offset": 2
    },
```,

a 4-byte size reserved space where the only relevant data is one bit at (bit) offset 2. We must read the whole 4-byte chunk, regardless of the position of the useful data in it (we extract it later).

File: src/lib/io/sol-memmap-storage-impl-linux.c (hunk 10 above)
Comment (glima) [inline src/lib/io/sol-memmap-storage-impl-linux.c]: sscanf() would have to have buffers to hold the scanned values, right? I prefer the current approach, then. The double check on create,i2c is indeed bad, what I could do is to also use splits to find that earlier and pass the str slice forward to the function in question, with an offset.

File: src/lib/io/sol-memmap-storage-impl-zephyr.c (hunk 6 above)
Comment (glima) [inline src/lib/io/sol-memmap-storage-impl-zephyr.c]: Done.

File: src/lib/io/sol-memmap-storage-impl-linux.c (hunk 10 above)
Comment (barbieri) [inline src/lib/io/sol-memmap-storage-impl-linux.c]: `%[^,]` can be used with `s` or `m`, the later will allocate the memory for you.

File: src/lib/io/sol-memmap-storage-impl-zephyr.c (hunk 4 above)
Comment (barbieri) [inline src/lib/io/sol-memmap-storage-impl-zephyr.c]: could be, but the comment to protect makes no sense. Anyway, getting the refcnt right is ok.

File: src/lib/io/sol-memmap-storage-impl-zephyr.c (hunk 8 above)
Comment (barbieri) [inline src/lib/io/sol-memmap-storage-impl-zephyr.c]: I have to admit I didn't look in detail, but I wonder if we should look that in the memmap table and use that name instead, then just point to an existing string instead of strdup one

File: src/lib/io/sol-memmap-storage-impl-zephyr.c (hunk 7 above)
Comment (barbieri) [inline src/lib/io/sol-memmap-storage-impl-zephyr.c]: all in all I think the driver and parameters are either specified in the kconfig (not so good) or in the memmap itself. We can use the path as something like `<driver>,<erase_size>,<max_rw>,<offset>`, so in this case you'd use `W25QXXDV,4096,256,0`. You could make some of these optional, or even use key=value pairs, like `driver=W25QXXV,erase_size=4096,max_rw=256`. Then you keep these in the context of memmap, not as a global, pass the handle and get the information from there.

I'd talk to the Zephyr guys if these informations (erase, max_rw) could be queried on their side, in this case avoid these options on our side altogether.

@ibriano how does that sound to you?

File: src/lib/io/sol-memmap-storage-impl-zephyr.c [hunk 11]
```diff
@@ -0,0 +1,291 @@
...
+sol_memmap_impl_map_new(const struct sol_memmap_map *map)
+{
+    struct map_internal *map_internal;
+
+    SOL_NULL_CHECK_ERRNO(map, EINVAL, NULL);
+
+    map_internal = calloc(1, sizeof(*map_internal));
+    SOL_NULL_CHECK_ERRNO(map, ENOMEM, NULL);
+
+    map_internal->map = map;
```
barbieri: as a note, you'd move the flash_dev and what I mentioned here.

src/lib/io/sol-memmap-storage-impl-zephyr.c
ibriano: On Mon, 23 May 2016 13:21:42 -0700, Gustavo Sverzut Barbieri wrote:
> > +        sol_blob_unref(pending->blob);
> > +    }
> > +    sol_vector_clear(&tmp_vector);
> > +
> > +    SOL_DBG("Performed pending writes");
> > +
> > +    return false;
> > +}
> > +
> > +int
> > +sol_memmap_impl_init(void)
> > +{
> > +    if (flash_dev)
> > +        return 0;
> > +
> > +    flash_dev = device_get_binding("W25QXXDV");
> 
> all in all I think the driver and parameters are either specified in the kconfig (not so good) or in the memmap itself. We can use the path as something like `<driver>,<erase_size>,<max_rw>,<offset>`, so in this case you'd use `W25QXXDV,4096,256,0`. You could make some of these optional, or even use key=value pairs, like `driver=W25QXXV,erase_size=4096,max_rw=256`. Then you keep these in the context of memmap, not as a global, pass the handle and get the information from there.
> 
> I'd talk to the Zephyr guys if these informations (erase, max_rw) could be queried on their side, in this case avoid these options on our side altogether.
> 
> @ibriano how does that sound to you?

Looking at the flash implementation for QMSI, I say they should provide
some of this information themselves. On QMSI, page sizes are different
from this other chip and the sizes you use to write need to be aligned
too.
I know very little about flash controllers to say if there's a pattern
that could be easily followed, but it does feel as if the Zephyr API for
it is not offering enough.

barbieri: that's part of what I commented. The other part is: support multiple devices, using the memmap path to provide information.

ibriano: That would work now, that device_get_binding() is still the way to get the devices. That may change in the relatively-near future, but until then I don't see any problem with your idea.

src/lib/io/sol-memmap-storage-impl-linux.c
glima: OK, done for Linux

src/lib/io/sol-memmap-storage-impl-zephyr.c
glima: I seems it's possible, trying so

glima:
@barbieri ^

src/lib/io/sol-memmap-storage.c
```diff
@@ -16,58 +16,37 @@
...
 static bool
-get_entry_metadata_on_map(const char *name, const struct sol_memmap_map *map, const struct sol_memmap_entry **entry, uint64_t *mask)
+get_entry_metadata_on_map(const char *name,
+    const char **out_name,
+    const struct sol_memmap_map *map,
+    const struct sol_memmap_entry **entry,
+    uint64_t *mask)
 {
     uint32_t bit_size;
+    const struct sol_str_table_ptr *table_entry = sol_str_table_ptr_entry_lookup
```
barbieri: why did you introduce this function if all you're looking for is the `table_entry->val`?

note that you can use the `errno` to detect if `NULL` is returned due problems or an empty pointer in the value.

barbieri:
+0, I'd merge the extra patches to buffer and boolean/toggle right now and wait for @edersondisouza to check the persistence stuff, particularly that comment related to always using a dynamic buffer to read contents even for the fixed sized types.

src/lib/io/sol-memmap-storage.c
```diff
@@ -16,58 +16,37 @@
...
+            (map->entries, sol_str_slice_from_str(name));
 
-    if (sol_str_table_ptr_lookup(map->entries, sol_str_slice_from_str(name), entry)) {
+    if (table_entry) {
+        *entry = table_entry->val;
+        if (out_name)
+            *out_name = table_entry->key;
```
glima: ``entry->key`` is needed here.

barbieri: it's the one you used, or you need the exact pointer?

if you really need to keep this, then add the matching function to other table variants.

glima: OK, I'll do that (and make the others depend on this, maybe)

src/modules/flow/persistence/persistence.c
```diff
@@ -320,23 +321,31 @@ persist_open(struct sol_flow_node *node,
...
     if (mdata->packet_data_size) {
-        struct sol_buffer buf = SOL_BUFFER_INIT_FLAGS(mdata->value_ptr,
-            mdata->packet_data_size,
-            SOL_BUFFER_FLAGS_MEMORY_NOT_OWNED | SOL_BUFFER_FLAGS_NO_NUL_BYTE);
-
-        r = storage_read(mdata, &buf);
+        if (r >= 0) {
+            /* entry's total size may be bigger that actual
+             * packet_data_size (think bit fields). the useful data
+             * with be the leading bytes, on all cases */
```
edersondisouza: I didn't get this... it shouldn't be the other way round: entry size smaller than packet size?

src/lib/io/sol-memmap-storage-impl-zephyr.c
```diff
@@ -0,0 +1,412 @@
...
+        for (i = 0, j = 0; i < entry->size; i++, j += 8)
+            value |= (uint64_t)((uint8_t *)buffer->data)[i] << j;
+
+        value &= mask;
+        value >>= entry->bit_offset;
+
+        memset(buffer->data, 0, buffer->capacity);
+        for (i = 0; i < entry->size; i++, value >>= 8)
+            ((uint8_t *)buffer->data)[i] = value & 0xff;
+    }
```
edersondisouza: World could be a better place if it was possible to share this code...

src/lib/io/sol-memmap-storage-impl-zephyr.c
```diff
@@ -0,0 +1,412 @@
...
+
+    *driver_name = sol_str_slice_to_string(
+        *(const struct sol_str_slice *)sol_vector_get(&instructions,
+        DRIVER_NAME_IDX));
+    SOL_NULL_CHECK_GOTO(*driver_name, err_parse);
+
+    min_erase_size_s = sol_str_slice_to_string(
+        *(const struct sol_str_slice *)sol_vector_get(&instructions,
+        MIN_ERASE_SZ_IDX));
+    SOL_NULL_CHECK_GOTO(*min_erase_size_s, err_parse);
```
edersondisouza: This kind of 'instruction mapping' code is appearing more and more. Couldn't we unify them? On Linux you changed it to `sscanf`, just to use `sol_str_slice` here again...

src/lib/datatypes/sol-buffer.c
```diff
@@ -142,8 +142,12 @@ sol_buffer_append_bytes(struct sol_buffer *buf, const uint8_t *bytes, size_t siz
...
 
     SOL_NULL_CHECK(buf, -EINVAL);
-    if (size == 0)
-        return 0;
+    if (size == 0) {
+        if (SOL_BUFFER_NEEDS_NUL_BYTE(buf))
+            return sol_buffer_ensure_nul_byte(buf);
+        else
+            return 0;
+    }
```
edersondisouza: Why?

src/lib/io/sol-efivarfs-storage.c
edersondisouza: As I said below - and maybe it's because I've just returned from vacations - I didn't get this... why does buffer need to grow for bitfields?

src/lib/io/sol-memmap-storage-impl-zephyr.c
glima: Ok, I'll try

src/lib/io/sol-memmap-storage-impl-zephyr.c
glima: sscanf was a desire by @barbieri, I was OK with the previous approach as well. At least now on Linux we're more flexible and allow spaces in between the commas.

src/modules/flow/persistence/persistence.c
glima: Look at the previous comment for src/lib/io/sol-efivarfs-storage.c

src/lib/datatypes/sol-buffer.c
glima: Not related to this patch (it was an attempt to fix the FIXME I introduce, but it's not enough). But the logic is sound -- if we get nul_bytes, any append/resize/whatever has to ensure it.

glima:
Thanks, everyone. Finally fixing those and merging.

PR #2080, opened 2016-05-23

gabrielschulhof:
Fixes gh-2054

Replaces #2059

Adressed review comments.

### DIFF
--- a/src/lib/common/include/sol-reentrant.h
+++ b/src/lib/common/include/sol-reentrant.h
@@ -1,0 +1,112 @@
+/*
+ * This file is part of the Soletta Project
+ *
+ * Copyright (C) 2015 Intel Corporation. All rights reserved.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+#pragma once
+
+#ifdef __cplusplus
+extern "C" {
+#endif
+
+/**
+ * @file
+ * @brief Structure and macros for making structures reentrancy-proof
+ */
+
+/**
+ * @defgroup Reentrant Reentrant
+ *
+ * @{
+ */
+
+/**
+ * @struct sol_reentrant
+ *
+ * @brief Structure containing the flags for safely freeing a larger structure
+ *
+ * This structure is meant to be used as the first field of larger structures
+ * which are affected by calls to external callbacks that in turn end up
+ * calling library APIs. The possibility of a double free is particularly
+ * likely in such cases.
+ */
+struct sol_reentrant {
+    /**
+     * @brief Structure is in use
+     */
+    bool in_use;
+    /**
+     * @brief Structure is stale and should be freed as soon as possible
+     */
+    bool is_stale;
+};
+
+/**
+ * @brief Perform a function call to an external callback
+ *
+ * @param reentrant The reentrant affected by the function call
+ *
+ * @param call_statement The function call statement
+ *
+ * Wraps a call to a function by first recording whether the reentrant
+ * structure was in use and, after the function call, restoring the state of
+ * the reentrant structure from the recorded value. Unlike a monitor, which
+ * protects the list of callbacks itself, this wrapper protects the library's
+ * structure from being deleted as part of the external callback, if it's still
+ * in use after the callback completes.
+ */
+#define SOL_REENTRANT_CALL(reentrant, call_statement) \
+    do { \
+        struct sol_reentrant *local = ((struct sol_reentrant *)(reentrant)); \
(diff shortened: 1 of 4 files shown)


src/lib/common/include/sol-reentrant.h
```diff
@@ -0,0 +1,112 @@
...
+ * Wraps a call to a function by first recording whether the reentrant
+ * structure was in use and, after the function call, restoring the state of
+ * the reentrant structure from the recorded value. Unlike a monitor, which
+ * protects the list of callbacks itself, this wrapper protects the library's
+ * structure from being deleted as part of the external callback, if it's still
+ * in use after the callback completes.
+ */
+#define SOL_REENTRANT_CALL(reentrant, call_statement) \
+    do { \
+        struct sol_reentrant *local = ((struct sol_reentrant *)(reentrant)); \
```
barbieri: I still dislike requiring the reentrant structure to be the first member and do this cast. We could easily ask the user to pass the reentrant pointer. At least for structure alignment, moving the 2 booleans to the end would be better to avoid holes.

src/lib/common/include/sol-reentrant.h
```diff
@@ -0,0 +1,112 @@
...
+ * Calls the function specified by free_function if the reentrant is not marked
+ * as being in use. Otherwise, it marks the reentrant as stale but does not
+ * free it.
+ */
+#define SOL_REENTRANT_FREE(reentrant, free_function) \
+    do { \
+        struct sol_reentrant *local = ((struct sol_reentrant *)(reentrant)); \
+        local->is_stale = true; \
+        if (!(local->in_use)) { \
+            free_function((reentrant)); \
```
barbieri: why in `_CALL()` you give a statement, here you give a function? Let's keep it consistent?

src/lib/common/include/sol-reentrant.h
```diff
@@ -0,0 +1,112 @@
...
+ */
+struct sol_reentrant {
+    /**
+     * @brief Structure is in use
+     */
+    bool in_use;
+    /**
+     * @brief Structure is stale and should be freed as soon as possible
+     */
+    bool is_stale;
```
barbieri: `delete_me` is what we use in other contexts

src/lib/common/include/sol-reentrant.h
barbieri: also, after looking at the usage I'm wondering if we do something similar to `SOL_MONITORS_WALK()` https://github.com/solettaproject/soletta/blob/master/src/shared/sol-monitors.h#L127 Example:

```c
#define SOL_REENTRANT_CALL(handle) \
    for (bool executed = false; !executed; ) \
        for (bool was_in_use = (handle)->reentrant.in_use; !executed; (handle)->reentrant.in_use = was_in_use)\
            for ((handle)->reentrant.in_use = true;                    \
                 !executed;                                            \
                 executed = true)

int main() {
    struct myst {
        int i;
        struct sol_reentrant reentrant;
    } h = {.i = 10};

    SOL_REENTRANT_CALL(&h) {
        printf("h=%p, i=%d\n", &h, h.i);
    }
```

src/lib/common/include/sol-reentrant.h
gabrielschulhof: Because the free function always has the same signature, whereas the callback could return void, or it could return a value, which you need to save. Consider the OIC request callback vs. the OIC discovery callback. So, the parameter to ```_CALL()``` has to be a statement.

gabrielschulhof: ... although I'm thinking about using the for-loop trick to actually wrap the call. It allows for truly arbitrary statements.

src/lib/common/include/sol-reentrant.h
gabrielschulhof: The only similarity this has to monitors is that it shares the goal of safe deletion. However, in the case of monitors, it is the callback that safely needs to be deleted from the list of callbacks, whereas in this case, it's the data structure which serves as the context for the callback and which needs to be accessed both before and after the callback that needs to be protected from the actions of the callback.

gabrielschulhof: I'll let the reentrant structure reside anywhere and thus require a separate pointer.

gabrielschulhof: Also, unlike a monitor, the context is not one provided along with the callback. The context is a structure internal to soletta which the callback may destroy by calling a soletta API. The place where I came across this was (in a hybrid JS/C pseudocode):
```
pending = sol_oic_client_find_resources( function( resource ) {
  sol_oic_pending_cancel( pending );
  return NULL;
} );
```

So, the context was being destroyed twice: Once during the action of the callback because it was explicitly freeing the context via ```sol_oic_pending_cancel()```, and again in response to the NULL return value, indicating that the context is no longer necessary. The pending is an internal Soletta structure, rather than one specified along with the discovery response callback.

gabrielschulhof: Also, the only use case for monitors seems to be to have a static structure at the top of the file. Thus, there's never an issue with the monitor structure itself being destroyed, whereas in this case, there is no global place where to store the structure to protect it from being removed. Thus, we store the information in the structure itself.

barbieri: I wasn't talking about using monitors, I get the differences, I guess it was someone else that suggested it.

What I said is to use the `for()` statements to create a block that sets/unsets the `in_use`, providing a real block and not using statements inside a macro parameter.

PR #2081, opened 2016-05-23
Node.js bindings: Adapt UART JS bindings to the latest API changes

nagineni:
Adapt UART JS bindings to the latest API changes and enable them by default. Also, add a sample that demonstrates the usage of bindings for reading and writing.

Replaces #2060


nagineni:
@barbieri Could you please have a look? Now we have zero copy in the uart feed calls.

barbieri:
+1

LGTM!

PR #2068, opened 2016-05-20
sol-connman: Add sol-connman basic API

wzhen12:
so-connman is for implementing connection manager features.
The APIs should be implemented in Zephyr, connman and so on.

The patches are for implementing sol-connman basic APIs to connman.
the connman D-Bus APIs are wrapped in the so-connman APIs.
The user can use the so-connman APIs to control connman.

Signed-off-by: Wu Zheng wu.zheng@intel.com

PR #2087, opened 2016-05-24
comms: Add so-netctl API

wzhen12:
so-netctl is for implementing connection manager features.
The APIs should be implemented in Zephyr, connman and so on.
To connman, the connman D-Bus APIs are wrapped in the so-netctl
APIs. The user can use the so-netctl APIs to control connman.

To invoke the D-Bus method of connman, the async call is used,
the callback function is set to get the results of async call.

The basic so-netctl APIs are in the patch.
It provides the APis of getting the name, ip address... of services.
The service connect/disconnect APIs and set/get offline
are implemented too.

The further sol-netctl APIs are being planned.

Signed-off-by: Wu Zheng <wu.zheng@intel.com>


wzhen12:
1. Modify the all connman to netctl
2. Modify the help comments.

Please check it again. Thanks.

wzhen12:
According to the comments of #2068,
the related issues have been fixed.
It is the version 5 of sol-netctl.

1. Modify the all connman to netctl
2. Modify the help comments.

Please check it. Thanks.

src/modules/flow/netctl/netctl.json
```diff
@@ -0,0 +1,80 @@
...
+          "methods": {
+            "process": "set_connect_ap_name"
+          },
+          "name": "CONNECT"
+        },
+        {
+          "data_type": "string",
+          "description": "Name of network AP.",
+          "methods": {
+            "process": "set_disconnect_ap_name"
```
barbieri: no reason why keep `_ap_` here, refer all as `service` for consistency.

src/modules/flow/netctl/netctl.json
```diff
@@ -0,0 +1,80 @@
...
+          "data_type": "string",
+          "description": "Name of network AP.",
+          "methods": {
+            "process": "set_connect_ap_name"
+          },
+          "name": "CONNECT"
+        },
+        {
+          "data_type": "string",
+          "description": "Name of network AP.",
```
barbieri: also adjust the description, replace `AP` with `service`,  also refer to `netctl/list-services` on how to get the service name.

src/modules/flow/netctl/netctl.json
```diff
@@ -0,0 +1,80 @@
...
+            "process": "set_disconnect_ap_name"
+          },
+          "name": "DISCONNECT"
+        }
+      ],
+      "methods": {
+        "open": "open_service_connection",
+	"close": "close_service_connection"
+      },
+      "name": "netctl/connection",
```
barbieri: call this `netctl/service`

src/modules/flow/netctl/netctl.json
```diff
@@ -0,0 +1,80 @@
...
+      "options": {
+        "members": [
+          {
+            "data_type": "string",
+            "default": null,
+            "description": "Name of AP to be used to connect. If 'null', no filter is applied.",
+            "name": "connect_ap"
```
barbieri: replace `connect_ap` with `name`, that refers to the service name.

src/modules/flow/netctl/netctl.json
barbieri: also, these ports should behave like https://github.com/solettaproject/soletta/blob/master/src/modules/flow/platform/platform.json and both `CONNECT` and `DISCONNECT` should be of `any` packet type. The service name is given as option or input port `NAME`

src/modules/flow/netctl/netctl.json
```diff
@@ -0,0 +1,80 @@
...
+          },
+          {
+            "data_type": "string",
+            "default": null,
+            "description": "Name of AP to be used to disconnect. If 'null', no filter is applied.",
+            "name": "disconnect_ap"
+          }
```
barbieri: as per above, no `disconnect_ap`,  just a single option `name`. Then connect/disconnect with an `any` packet type to connect or disconnect to it.

src/samples/flow/netctl/network-connect.fbp
```diff
@@ -0,0 +1,29 @@
...
+# This program will print to console a list of all network services
+#
+# Examples:
+#  * connect or disconnect network services
+#    $ ./network-connect.fbp Guest
+
+ap(app/argv:index=1)
+ap OUT -> CONNECT conn(netctl/connection) OUT -> IN _(console)
+
+ap ERROR -> IN _(converter/empty-to-string:output_value="List all the network services") OUT -> IN Info(console)
```
barbieri: why error is converted to a fixed string?

src/samples/flow/netctl/network-disconnect.fbp
```diff
@@ -0,0 +1,29 @@
...
+# This program will print to console a list of all network services
+#
+# Examples:
+#  * connect or disconnect network services
+#    $ ./network-disconnect.fbp Guest
+
+ap(app/argv:index=1)
+ap OUT -> DISCONNECT conn(netctl/connection) OUT -> IN _(console)
+
+ap ERROR -> IN _(converter/empty-to-string:output_value="List all the network services") OUT -> IN Info(console)
```
barbieri: same comment for error, why is it being converted to a fixed string?

src/modules/flow/netctl/netctl.c
```diff
@@ -0,0 +1,227 @@
...
+
+    name = sol_netctl_service_get_name(service);
+    r = sol_flow_send_string_packet(node,
+        SOL_FLOW_NODE_TYPE_NETCTL_LIST_SERVICES__OUT__OUT,
+        name);
+    if (r < 0)
+        SOL_WRN("Failed to send network service name: %s", name);
+
+    sol_ptr_vector_append(&mdata->service_list, service);
+}
```
barbieri: I think we don't need to keep this mirror list `service_list`, since that's the same as using the `sol_netctl_get_services()`. However, that is not as dynamic as I'd expect.

What if we change `netclt/list-services` to be dynamic, on start it will list all existing services (`sol_netctl_get_services()`) and output then on `ADDED` port (instead of `OUT` port). Then when service is removed, you do emit its name in the `REMOVED` port, if a new service appears, then it would also go in the `ADDED` port.

Even in this case, there is no need to keep the mirror/shadow `mdata->services_list`.

What do you think? @bdilly @vcgomes @wzhen12

src/modules/flow/netctl/netctl.c
```diff
@@ -0,0 +1,227 @@
...
+            sol_flow_send_string_packet(node,
+                SOL_FLOW_NODE_TYPE_NETCTL_CONNECTION__OUT__OUT,
+                "Connect AP!");
+        } else if ((state == SOL_NETCTL_SERVICE_STATE_READY) && (!mdata->connect)) {
+            sol_netctl_service_disconnect(service);
+            sol_flow_send_string_packet(node,
+                SOL_FLOW_NODE_TYPE_NETCTL_CONNECTION__OUT__OUT,
+                "Disconnect AP!");
+        }
+}
```
barbieri: This function must be reworked if changing `CONNECT` and `DISCONNECT` ports to be a trigger (takes `any` packet type). This monitor would only be used to change the `OUT` port, that is to output the string representing the `sol_netctl_service_get_state()`, not `"Connect AP!"`  or `"Disconnect AP!"` (you'll need a `sol_netctl_service_state_to_str()`). As I've mentioned earlier, you may also have an `ONLINE` (boolean) port that will emit true/false if state is `SOL_NETCTL_STATE_ONLINE`. This is a nice helper.

In the `CONNECT` and `DISCONNECT` port process functions you'd call `sol_netctl_get_services()`, iterate over all services and see `streq(sol_netctl_service_get_name(itr), mdata->name)`. In this case you call `sol_netctl_service_connect(itr)` (or disconnect). If the iteration did not find any services, then it must emit an error packet `error = ENOENT, message = "Did not found service name %s", mdata->name`.

src/modules/flow/netctl/netctl.c
```diff
@@ -0,0 +1,227 @@
...
+    static const char *errmsg = "Could not add service monitor";
+    struct list_service_data *mdata = data;
+
+    mdata->node = node;
+
+    sol_ptr_vector_init(&mdata->service_list);
+
+    if (sol_netctl_add_service_monitor(network_service_cb, mdata))
+        goto error;
+
```
barbieri: here you'd get `sol_netctl_get_services()` and emit them all in the `ADDED` port. (see my other comment)

src/modules/flow/netctl/netctl.json
```diff
@@ -0,0 +1,80 @@
...
+      "methods": {
+        "open": "open_list_services",
+	"close": "close_list_services"
+      },
+      "name": "netctl/list-services",
+      "out_ports": [
+        {
+          "data_type": "string",
+          "description": "Name of network services supported by system.",
+          "name": "OUT"
```
barbieri: change this to `ADDED` port, then add a `REMOVED` port. This way you keep the list-services dynamic. Also document this on its `description`, mentioning that on open, existing services will be dispatched at `ADDED`, then when new services appear, they will be dispatched on that same port, while services that are gone (ie: access points, technologies that were turned off) go in `REMOVED` port.

src/modules/flow/netctl/netctl.json
```diff
@@ -0,0 +1,80 @@
...
+      "out_ports": [
+        {
+          "data_type": "string",
+          "description": "Output network services connection status.",
+          "name": "OUT"
+        }
+      ],
+      "private_data_type": "connection_service_data"
+    }
+  ]
```
barbieri: I miss also other node types, like:

 * `netctl/state`:
   - input ports:
      - `TRIGGER` (any): emits an output packet with the current state `sol_netctl_get_state()`. This is only used to force the emission, the node is dynamic and will output new packets when it changes and on open.
   - output ports:
      - `OUT` (string): the string with the current state, such as `idle`, `online`, `ready`... (Document all!). 
      - `ONLINE` (boolean):  emits true if the device has a verified connection to the internet, false otherwise.
 * `netctl/radios`:
   - input ports:
      - `ENABLED` (bool): turn on or off the radios.
   - output ports:
      - `OUT` (bool): the actual state of the radios. If true, radios are online.

Please let others review the proposed ports. @glima, @ceolin, @bdilly, @cabelitos

src/modules/flow/netctl/netctl.json
```diff
@@ -0,0 +1,80 @@
...
+            "name": "disconnect_ap"
+          }
+        ],
+        "version": 1
+      },
+      "out_ports": [
+        {
+          "data_type": "string",
+          "description": "Output network services connection status.",
+          "name": "OUT"
```
barbieri: maybe rename `OUT` with `STATE` (string) and also add other ports with extra information, like strength, type, IP address?

barbieri:
+1 on the C API, it's ready to be merged.

The flow needs more work. You can chose to merge the C API and keep working only in the FBP, or wait until FBP is ready to merge them all together.

src/lib/comms/Kconfig
```diff
@@ -43,6 +43,18 @@ config BLUETOOTH_NONE
 
 endchoice
 
+config NETCTL
+        bool "connection manager interfaces"
+        default y
+        depends on FEATURE_NETCTL && NETWORK
+        help
+           This provides the connection manager interfaces API.
+           The user can use the connection manager APIs to contorl
```
ceolin: s/contorl/control/

src/lib/comms/include/sol-netctl.h
```diff
@@ -0,0 +1,476 @@
...
+/**
+ * @brief Service monitor callback used to inform a service changed
+ *
+ * @param data the user data
+ * @param service the monitor service
+ */
+typedef void (*sol_netctl_service_monitor_cb)
+    (void *data,
+    const struct sol_netctl_service *service);
+
```
ceolin: we are not using typedef, if i'm not mistaken the places where it was being used is already changed.

src/lib/comms/include/sol-netctl.h
```diff
@@ -0,0 +1,476 @@
...
+ *
+ * This function gets the service strength for the netctl service,
+ * after the service is to give by service monitor.
+ *
+ * @param service The service structure which the service strength is
+ * desired
+ *
+ * @return 0-100 on success, -errno on failure.
+ */
+int32_t sol_netctl_service_get_strength(
```
ceolin: why you are using `int32_t` (instead of using `int`) here ? The returned range fits in any size of int

src/lib/comms/sol-comms.c
```diff
@@ -74,8 +90,14 @@ sol_comms_init(void)
     if (ret != 0)
         goto http_server_error;
 
+    ret = sol_netctl_init();
+    if (ret != 0)
+        goto netctl_error;
```
ceolin: if the `init()` fail you do not need call the `shutdown()`, isn't ?

src/lib/comms/sol-netctl-impl-connman.c
```diff
@@ -0,0 +1,1171 @@
...
+void sol_netctl_shutdown(void);
+
+struct sol_netctl_service {
+    sd_bus_slot *slot;
+    char *path;
+    char *name;
+    enum sol_netctl_service_state state;
+    char *type;
+    struct sol_network_link link;
+    int32_t strength;
```
ceolin: re-order the fields to avoid holes in the struct

src/lib/comms/include/sol-netctl.h
barbieri: I've commented about this before and he remembered me that this is needed because of the sol_monitors. So it is correct

src/lib/comms/include/sol-netctl.h
barbieri: I've asked to keep the int32_t, in the beginning I thought it could go over 100, then fixing the 32 bits in size was the way to go. Now it's clear it's not needed. But doesn't hurt to keep that.

src/lib/comms/sol-netctl-impl-connman.c
```diff
@@ -0,0 +1,1171 @@
...
+}
+
+SOL_API int
+sol_netctl_add_service_monitor(
+    sol_netctl_service_monitor_cb cb, const void *data)
+{
+    int r;
+
+    r = sol_netctl_init_lazy();
+    SOL_INT_CHECK_GOTO(r, < 0, fail_init);
```
ceolin: The same here, if the `init()` fails, do you really need call the `shutdown()`  ?

ceolin:
the C api is good for me as well

src/lib/comms/sol-netctl-impl-connman.c
wzhen12: I will check it.

src/lib/comms/sol-netctl-impl-connman.c
wzhen12: Yes. Some global variables need to be cleared.

src/lib/comms/sol-comms.c
wzhen12: Some global variables need to be cleared.

src/lib/comms/Kconfig
wzhen12: Thanks.

src/lib/comms/sol-comms.c
ceolin: to my mind it would be better fix it inside the `init()`. I mean, the `init()` itself clear the variable in case of failing.

glima:
The proposed nodes and ports LGTM

src/lib/comms/sol-comms.c
wzhen12: After checking it, yes. It is not necessary.
I will remove it in the next version.

wzhen12:
The next version is https://github.com/solettaproject/soletta/pull/2108.

wzhen12:
Please check https://github.com/solettaproject/soletta/pull/2108.

guchaojie:
Sorry for late reply, I am busy with QA connectivity test transfer work , I  work on it and sumbit PR ASAP

src/modules/flow/netctl/netctl.json
guchaojie: Can you give me one related example for ADDED or REMOVED port . I cannot find how to use ADDED or REMOVED doc in soletta. And all the services would get from service_callback in sol_netctl_add_service_monitor,  calling sol_netctl_get_services maybe result in that all the services are not ready yet. For example, if there are 5 services in the network environment, monitor will  asyn get service to upper layer one by one, so calling sol_netctl_get_service would get wrong result.  Because I am not familar with ADDED/REMOVED port, these port can resolve this problem?

PR #2089, opened 2016-05-24

gabrielschulhof:
Fixes gh-2054

Changes: Addressed review comments
Replaces #2080

### DIFF
--- a/src/lib/comms/sol-coap.c
+++ b/src/lib/comms/sol-coap.c
@@ -27,6 +27,7 @@
 #include "sol-macros.h"
 #include "sol-mainloop.h"
 #include "sol-network.h"
+#include "sol-reentrant.h"
 #include "sol-socket.h"
 #include "sol-str-slice.h"
 #include "sol-util-internal.h"
@@ -104,6 +105,7 @@ struct resource_observer {
 };
 
 struct pending_reply {
+    struct sol_reentrant reentrant;
     struct sol_coap_server *server;
     struct sol_timeout *timeout;
     bool (*cb)(void *data, struct sol_coap_server *server, struct sol_coap_packet *req,
@@ -650,6 +652,8 @@ pending_reply_free(struct pending_reply *reply)
     if (reply->timeout)
         sol_timeout_del(reply->timeout);
 
+    sol_ptr_vector_remove(&reply->server->pending, reply);
+
     free(reply);
 }
 
@@ -660,13 +664,19 @@ pending_timeout_cb(void *data)
 {
     struct pending_reply *reply = data;
     struct sol_coap_server *server = reply->server;
+    bool callback_result;
 
-    if (reply->cb((void *)reply->data, server, NULL, NULL))
+    SOL_REENTRANT_CALL(reply->reentrant) {
+        callback_result = reply->cb((void *)reply->data, server, NULL, NULL);
+    }
+
+    if (callback_result && !SOL_REENTRANT_IS_STALE(reply->reentrant)) {
         return true;
+    }
 
-    sol_ptr_vector_remove(&server->pending, reply);
-
-    pending_reply_free(reply);
+    SOL_REENTRANT_FREE(reply->reentrant) {
+        pending_reply_free(reply);
+    }
 
     return false;
 }
@@ -900,6 +910,8 @@ sol_coap_send_packet_with_reply(struct sol_coap_server *server, struct sol_coap_
     }
 
     sol_coap_header_get_id(pkt, &reply->id);
+    reply->reentrant.in_use = false;
+    reply->reentrant.delete_me = false;
     reply->cb = reply_cb;
     reply->data = data;
     reply->observing = observing;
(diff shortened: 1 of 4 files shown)


src/lib/comms/sol-coap.c
```diff
@@ -104,6 +105,7 @@ struct resource_observer {
 };
 
 struct pending_reply {
+    struct sol_reentrant reentrant;
```
barbieri: you made it work without requiring it as the first member, so for alignment purposes move this as last member.

src/lib/comms/sol-coap.c
```diff
@@ -660,13 +664,19 @@ pending_timeout_cb(void *data)
 {
     struct pending_reply *reply = data;
     struct sol_coap_server *server = reply->server;
+    bool callback_result;
 
-    if (reply->cb((void *)reply->data, server, NULL, NULL))
-        return true;
+    SOL_REENTRANT_CALL(reply->reentrant) {
+        callback_result = reply->cb((void *)reply->data, server, NULL, NULL);
+    }
```
barbieri: feels much better, I like it!

src/lib/comms/sol-coap.c
```diff
@@ -660,13 +664,19 @@ pending_timeout_cb(void *data)
...
 
-    sol_ptr_vector_remove(&server->pending, reply);
+    if (callback_result && !SOL_REENTRANT_IS_STALE(reply->reentrant)) {
+        return true;
+    }
 
-    pending_reply_free(reply);
+    SOL_REENTRANT_FREE(reply->reentrant) {
+        pending_reply_free(reply);
+    }
```
barbieri: do you think we still need these two macros? Your proposal would be in plain C:

```c
if (callback_result && !reply->reentrant.delete_me) { // shorter and still clear
    return true;
}

if (reply->reentrant.delete_me) { // likewise, small and still clear
    pending_reply_free(reply);
}
```

src/lib/comms/sol-coap.c
```diff
@@ -900,6 +910,8 @@ sol_coap_send_packet_with_reply(struct sol_coap_server *server, struct sol_coap_
     }
 
     sol_coap_header_get_id(pkt, &reply->id);
+    reply->reentrant.in_use = false;
+    reply->reentrant.delete_me = false;
```
barbieri: do you think this would be useful:

```c
static inline void
sol_reentrant_init(struct sol_reentrant *r)
{
    r->in_use = false;
    r->delete_me = false;
}

...
    sol_reentrant_init(&reply->reentrant);
```

src/lib/comms/sol-oic-client.c
```diff
@@ -72,6 +73,7 @@ struct sol_oic_client {
 };
 
 struct ctx {
+    struct sol_reentrant reentrant;
```
barbieri: move last

src/lib/comms/sol-oic-client.c
```diff
@@ -477,6 +491,8 @@ client_get_info(struct sol_oic_client *client,
     int r = 0;
 
     ctx = sol_util_memdup(&(struct server_info_ctx) {
+            .base.reentrant.in_use = false,
+            .base.reentrant.delete_me = false,
```
barbieri: hum... my proposal to `sol_reentrant_init()` couldn't be used here. However, by default non-specified members default to zero, so false.

src/lib/comms/sol-coap.c
barbieri: sorry, I've checked the `_FREE()` is bit more than checking the flag, the macro is good!

src/shared/include/sol-reentrant.h
```diff
@@ -0,0 +1,114 @@
...
+
+/**
+ * @brief Determine whether a reentrant structure is in need of deletion
+ *
+ * @param reentrant
+ *
+ * Evaluates to true if the reentrant must be deleted, and false otherwise
+ */
+#define SOL_REENTRANT_IS_STALE(reentrant) \
+    ((reentrant).delete_me)
```
barbieri: I don't think this macro should remain, it's not better than simply checking `delete_me`

barbieri:
just minor nitpicks, the patch looks good overall :-)

gabrielschulhof:
@barbieri Thanks for reviewing!

src/lib/comms/sol-coap.c
gabrielschulhof: I don't think I can move it to the last place, because the ```pending_reply``` is a variable length structure. I'll move it after ```bool observing```. OK?

barbieri: ok

gabrielschulhof: @barbieri Actually I'll move it before ```bool observing```. What say you?

src/lib/comms/sol-coap.c
gabrielschulhof: Actually, if we can assume that members are always initialized to 0, we don't need this.

barbieri: the code uses `calloc()`, so 0 is guaranteed.

Issue #1820, opened 2016-04-05
API Changes: Review Comms API looking for convergence


bdilly:
#### Task Description

It would be nice to review all communication API and try to make they all look the same.

They should be async, callbacks should pass data pointer as first parameter, etc

But also they should look similar. E.g:
 * if in a protocol you add elements one by one using _add(*element) and other add an array of elements you should go with the same for both - if possible
 * if you initialized some modules instances passing a config and in others set config with a specific call

Go through all headers on `src/lib/comms/include/` and be happy

Missing items:

- [x] On sol-http-client.h (and maybe others), use the new ``SOL_REENTRANT()`` thing to avoid comments in the doc like "don't do X at context Y".
- [x] On sol-lwm2m.h, maybe use blob on ``sol_lwm2m_resource_init()`` and friends, so that we're more flexible WRT the life cicle of the resource's payload memory (we would unref the blob internally)
- [x] Also on LWM2M, maybe the contexts involving ``sol_lwm2m_parse_tlv()`` should deal with a buffer instead of a slice

cabelitos:
The http-client code is already safe. Ignore the first missing item. I'm only creating a PR that update the docs.

https://github.com/solettaproject/soletta/pull/2191

cabelitos:
fixed

PR #2094, opened 2016-05-24
Comms sanity [v2] partially addresses https://github.com/solettaproject/soletta/issues/1820

glima:
Partially because I edited it to contain some bullet points with more massive changes I found on my way and did not attack yet (someone else might do that afterwards).

The OIC changes regarding ``struct sol_oic_resource`` need love on the review. Is this way OK? Ideas?

Changes since v1:
- now ``struct sol_oic_resource`` has const members and the callbacks still have non-const modifiers for them.

### DIFF
--- a/data/scripts/sol-oic-gen.py
+++ b/data/scripts/sol-oic-gen.py
@@ -1744,7 +1744,7 @@ check_updated_boolean(const bool a, const bool b)
 static inline bool
 check_updated_number(const double a, const double b)
 {
-    return !sol_util_double_equal(a, b);
+    return !sol_util_double_eq(a, b);
 }
 
 static inline int
--- a/data/scripts/sol-oic-gen.py
+++ b/data/scripts/sol-oic-gen.py
@@ -430,7 +430,7 @@ def generate_object_from_repr_vec_fn_common_c(name, props):
     const struct sol_oic_map_reader *repr_vec, uint32_t decode_mask)
 {
     struct sol_oic_repr_field field;
-    enum sol_oic_map_loop_status end_status;
+    enum sol_oic_map_loop_reason end_reason;
     struct sol_oic_map_reader iterator;
     struct %(struct_name)s fields = {
 %(fields_init)s
@@ -438,10 +438,10 @@ def generate_object_from_repr_vec_fn_common_c(name, props):
     bool updated = false;
     int ret = 0;
 
-    SOL_OIC_MAP_LOOP(repr_vec, &field, &iterator, end_status) {
+    SOL_OIC_MAP_LOOP(repr_vec, &field, &iterator, end_reason) {
 %(fields)s
     }
-    if (end_status != SOL_OIC_MAP_LOOP_OK)
+    if (end_reason != SOL_OIC_MAP_LOOP_OK)
         goto out;
 
 %(update_state)s
--- a/src/lib/comms/include/sol-bluetooth.h
+++ b/src/lib/comms/include/sol-bluetooth.h
@@ -114,7 +114,7 @@ int sol_bt_uuid_to_str(const struct sol_bt_uuid *uuid, struct sol_buffer *buffer
  *
  * @return True if UUIDs are equal, False otherwise.
  */
-bool sol_bt_uuid_equal(const struct sol_bt_uuid *u1, const struct sol_bt_uuid *u2);
+bool sol_bt_uuid_eq(const struct sol_bt_uuid *u1, const struct sol_bt_uuid *u2);
 
 /**
  * @brief Represents an active connection to a Bluetooth device.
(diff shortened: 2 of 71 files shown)


src/lib/comms/include/sol-coap.h
```diff
@@ -660,7 +662,8 @@ int sol_coap_packet_add_uri_path_option(struct sol_coap_packet *pkt, const char
  * @brief Finds the first apeearence of the specified option in a packet.
  *
  * @param pkt The packet holding the options.
- * @param code The option code to look for.
+ * @param code The option code to look for. Pass of the
```
vcgomes: "Pass of" doesn't look right, "One of" sounds better.

src/lib/comms/include/sol-coap.h
```diff
@@ -671,7 +674,8 @@ const void *sol_coap_find_first_option(const struct sol_coap_packet *pkt, uint16
  * @brief Gets a number of specified option in a packet.
  *
  * @param pkt The packet holding the options.
- * @param code The option code to look for.
+ * @param code The option code to look for. Pass of the
```
vcgomes: Same here.

data/scripts/sol-oic-gen.py
```diff
@@ -430,18 +430,18 @@ def generate_object_from_repr_vec_fn_common_c(name, props):
     const struct sol_oic_map_reader *repr_vec, uint32_t decode_mask)
 {
     struct sol_oic_repr_field field;
-    enum sol_oic_map_loop_status end_status;
```
otaviobp: We have discussed that we should use status instead of reason. I did those changes some weeks ago to close issue #1721. Have we change that again?

data/scripts/sol-oic-gen.py
```diff
@@ -1220,7 +1220,7 @@ def master_c_as_string(generated, oic_gen_c, oic_gen_h):
     }
 
     SOL_INF("Found resource matching device_id");
-    resource->resource = sol_oic_resource_ref(oic_res);
+    resource->resource = sol_oic_resource_ref((struct sol_oic_resource *)oic_res);
```
otaviobp: we could hide the counter in sol_oic_resource and make sol_oic_resource_ref receiving a const sol_oic_resource, so casts won't be necessary. It is hard for users to figure out when they can cast from const to non const.

otaviobp: I just saw that you made refcount private. Is it a problem to change make sol_oic_resource_ref  to receive a non-const sol_oic_resource?

src/lib/comms/include/sol-coap.h
glima: Thanks, fixing.

src/lib/comms/include/sol-bluetooth.h
```diff
@@ -67,7 +67,7 @@ enum sol_bt_uuid_type {
 /**
  * @brief Representation of a Bluetooth UUID.
  *
- * In Bluetooth, an UUID represents the type of an entity, for example,
+ * In Bluetooth, a UUID represents the type of an entity, for example,
```
otaviobp: Isn't An UUID correct?

otaviobp:
+1 for everything else

data/scripts/sol-oic-gen.py
glima: But reason outnumbers status throughout soletta for these loop conditions. Also it differentiates from other enums and such that really carry status for handles. What do we do here? Change all the other reason entries?

src/lib/comms/include/sol-bluetooth.h
glima: https://owl.english.purdue.edu/owl/resource/591/01/

data/scripts/sol-oic-gen.py
glima: It's already like that. You mean a const for ref as well?

data/scripts/sol-oic-gen.py
otaviobp: Reason outnumbers status because most of the places were not fixed after our API discussion. I don't have a strong opinion about that. I just think we need to choose one and fix it everywhere. I don't remember why we chose status. @cabelitos do you remember that?

data/scripts/sol-oic-gen.py
otaviobp: I would add const to all fields inside struct sol_oic_resource and let usages of struct sol_oic_resource without const modifier.
It makes no sense to remove const in calls to unref functions, as I proposed before.

glima: I might tend to the const on fields way as well. Waiting for more input before continuing.

data/scripts/sol-oic-gen.py
glima: Since nobody else commented on it, I'm still following my take here.

glima:
@otaviobp @cabelitos ^

glima:
@otaviobp I'll merge at the end of this day, OK?

otaviobp:
go ahead

bdilly:
it seems ok

glima:
Nice, merging

PR #2097, opened 2016-05-25
build: Reorder setting of COMMON_CFLAGS

ibriano:
The idea is to allow CFLAGS that come from the user override those used
by Soletta, at least where it makes sense.
Things done include:
 - Always use -ffloat-store. The way it was checked if it was necessary
   before is not compatible with cross-compilers, and we want the
   effect of that flag always in any case.
 - Only use LTO flags in release mode. If the user wants to enable
   debug information, they should either set Soletta to a debug build
   or make sure to include -fno-lto in their CFLAGS
 - Set all Soletta CFLAGS first, then include the ones that come from
   the user/OS, so it allows users to override other flags or set
   optimization levels that match the target OS.
 - Set -std=gnu99 at the end of everything else. Unfortunately, some
   systems build with a different standard set that may enable
   __STRICT_ANSI__ and breaks building Soletta, so we need to always
   override this value to the one we expect.

### DIFF
--- a/data/jsons/dependencies.json
+++ b/data/jsons/dependencies.json
@@ -496,11 +496,6 @@
       "fragment": "strftime_l(0, 0, 0, 0, 0);"
     },
     {
-      "dependency": "getconf",
-      "type": "exec",
-      "exec": "getconf"
-    },
-    {
       "dependency": "get_current_dir_name",
       "type": "ccode",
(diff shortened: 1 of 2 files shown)


barbieri:
+1

ibriano:
Merged

PR #2098, opened 2016-05-25
oic: Fix null pointer exception

otaviobp:
When sol_coap_send_packet_with_reply is called with a NULL reply
callback, reply element won't be created and then we don't need to free
it on errors.


barbieri:
+1

otaviobp:
merged

PR #2099, opened 2016-05-25
Use sol-reentrant in UART and stream sample

cabelitos:


barbieri:
+1

src/lib/io/sol-uart-impl-linux.c
```diff
@@ -87,10 +86,10 @@ read_timeout(void *data)
     bool keep_running = true;
     int err;
 
-    uart->in_use = true;
-    r = uart->on_data((void *)uart->user_data, uart, &uart->rx);
-    uart->in_use = false;
-    SOL_INT_CHECK_GOTO(r, < 0, exit);
+    SOL_REENTRANT_CALL(uart->reentrant) {
+        r = uart->on_data((void *)uart->user_data, uart, &uart->rx);
+        SOL_INT_CHECK_GOTO(r, < 0, exit);
```
gabrielschulhof: This is bad because it won't restore ```->in_use```

gabrielschulhof: You can't jump out of the body of a ```SOL_REENTRANT_CALL()```, because it has an epilogue which must not be skipped, or you'll leak the structure it's protecting.

gabrielschulhof: It's sufficient if you move the ```SOL_INT_CHECK_GOTO()``` outside the block, since ```r``` is available there.

src/lib/io/sol-uart-impl-riot.c
```diff
@@ -102,21 +101,25 @@ rx_timeout(void *data)
     ssize_t r;
     int err;
 
-    uart->in_use = true;
-    r = uart->on_data((void *)uart->user_data, uart, &uart->rx);
-    uart->in_use = false;
-    SOL_INT_CHECK(r, < 0, true);
+    SOL_REENTRANT_CALL(uart->reentrant) {
+        r = uart->on_data((void *)uart->user_data, uart, &uart->rx);
+        SOL_INT_CHECK(r, < 0, true);
```
gabrielschulhof: Same here.

src/lib/io/sol-uart-impl-linux.c
cabelitos: Oooops, Indeed. Thank you

src/lib/io/sol-uart-impl-riot.c
glima: We ought to document this fact harder, I guess :)

cabelitos:
merged

PR #2102, opened 2016-05-26
WIP: iio: code refactor for nodes.c

yongli3:
These sensors are using direction-vector as output data type,
Change the data struct to share the same for them

### DIFF
--- a/src/modules/flow/iio/iio.json
+++ b/src/modules/flow/iio/iio.json
@@ -16,13 +16,23 @@
           "description": "Packets sent in here will trigger a gyroscope reading and produce packets on each of the the output ports. If buffer is enabled, current trigger must accept manual activation (default one, sysfs trigger, does).",
           "name": "TICK",
           "methods": {
-            "process": "gyroscope_tick"
+            "process": "iio_common_tick"
           }
         }
       ],
       "methods": {
-        "close": "gyroscope_close",
+        "close": "iio_common_close",
         "open": "gyroscope_open"
+      },
+      "node_type": {
+        "access": [
+            "base"
+        ],
+        "data_type": "struct iio_node_type",
+        "extra_methods": {
+            "out_port": "SOL_FLOW_NODE_TYPE_IIO_GYROSCOPE__OUT__OUT",
+            "reader_cb": "iio_direction_vector_reader_cb"
+       }
       },
       "name": "iio/gyroscope",
       "aliases": [
@@ -96,7 +106,7 @@
          "name": "OUT"
         }
       ],
-      "private_data_type": "gyroscope_data"
+      "private_data_type": "iio_direction_vector_data"
     },
     {
       "category": "input/hw",
@@ -107,13 +117,23 @@
           "description": "Packets sent in here will trigger a magnet reading and produce packets on each of the the output ports. If buffer is enabled, current trigger must accept manual activation (default one, sysfs trigger, does).",
           "name": "TICK",
           "methods": {
-            "process": "magnet_tick"
+            "process": "iio_common_tick"
           }
         }
       ],
       "methods": {
-        "close": "magnet_close",
+        "close": "iio_common_close",
         "open": "magnet_open"
+      },
+      "node_type": {
+        "access": [
+            "base"
+        ],
+        "data_type": "struct iio_node_type",
+        "extra_methods": {
+            "out_port": "SOL_FLOW_NODE_TYPE_IIO_MAGNETOMETER__OUT__OUT",
+            "reader_cb": "iio_direction_vector_reader_cb"
+       }
       },
       "name": "iio/magnetometer",
       "aliases": [
@@ -188,7 +208,7 @@
          "name": "OUT"
         }
       ],
-      "private_data_type": "magnet_data"
+      "private_data_type": "iio_direction_vector_data"
     },
     {
       "category": "input/hw",
@@ -199,13 +219,23 @@
           "description": "Packets sent in here will trigger a temperature reading and produce packets on each of the the output ports. If buffer is enabled, current trigger must accept manual activation (default one, sysfs trigger, does).",
           "name": "TICK",
           "methods": {
-            "process": "temperature_tick"
+            "process": "iio_common_tick"
           }
         }
       ],
       "methods": {
-        "close": "temperature_close",
+        "close": "iio_common_close",
         "open": "temperature_open"
+      },
+      "node_type": {
+        "access": [
+            "base"
+        ],
+        "data_type": "struct iio_node_type",
+        "extra_methods": {
+            "out_port": "SOL_FLOW_NODE_TYPE_IIO_THERMOMETER__OUT__OUT",
+            "reader_cb": "iio_double_reader_cb"
+       }
       },
       "name": "iio/thermometer",
       "aliases": [
@@ -280,7 +310,7 @@
          "name": "OUT"
         }
       ],
-      "private_data_type": "temperature_data"
+      "private_data_type": "iio_double_data"
     },
     {
       "category": "input/hw",
@@ -291,13 +321,23 @@
           "description": "Packets sent in here will trigger a pressure reading and produce packets on each of the the output ports. If buffer is enabled, current trigger must accept manual activation (default one, sysfs trigger, does).",
           "name": "TICK",
           "methods": {
-            "process": "pressure_tick"
+            "process": "iio_common_tick"
           }
         }
       ],
       "methods": {
-        "close": "pressure_close",
+        "close": "iio_common_close",
         "open": "pressure_open"
+      },
+      "node_type": {
+        "access": [
+            "base"
+        ],
+        "data_type": "struct iio_node_type",
+        "extra_methods": {
+            "out_port": "SOL_FLOW_NODE_TYPE_IIO_PRESSURE_SENSOR__OUT__OUT",
+            "reader_cb": "iio_double_reader_cb"
+       }
       },
       "name": "iio/pressure-sensor",
       "aliases": [
@@ -372,7 +412,7 @@
          "name": "OUT"
         }
       ],
-      "private_data_type": "pressure_data"
+      "private_data_type": "iio_double_data"
     },
     {
       "category": "input/hw",
@@ -383,13 +423,23 @@
           "description": "Packets sent in here will trigger a color reading and produce packets on each of the the output ports. If buffer is enabled, current trigger must accept manual activation (default one, sysfs trigger, does).",
           "name": "TICK",
           "methods": {
-            "process": "color_tick"
+            "process": "iio_common_tick"
           }
         }
       ],
       "methods": {
-        "close": "color_close",
+        "close": "iio_common_close",
         "open": "color_open"
+      },
+      "node_type": {
+        "access": [
+            "base"
+        ],
+        "data_type": "struct iio_node_type",
+        "extra_methods": {
+            "out_port": "SOL_FLOW_NODE_TYPE_IIO_COLOR_SENSOR__OUT__OUT",
+            "reader_cb": "iio_color_reader_cb"
+       }
       },
       "name": "iio/color-sensor",
       "aliases": [
@@ -423,8 +473,20 @@
          {
              "data_type": "float",
              "default": 0,
-             "description": "Scale to applied to device raw readings",
-             "name": "scale"
+             "description": "Scale to applied to device Red raw readings",
+             "name": "scale_red"
+         },
+         {
+             "data_type": "float",
+             "default": 0,
+             "description": "Scale to applied to device Green raw readings",
+             "name": "scale_green"
+         },
+         {
+             "data_type": "float",
+             "default": 0,
+             "description": "Scale to applied to device Blue raw readings",
+             "name": "scale_blue"
          },
          {
            "data_type": "boolean",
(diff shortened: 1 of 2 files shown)


yongli3:
@barbieri @edersondisouza @neofang @peijiajames Based on comments on PR #2070, I changed the data define for three sensors using direction-vector, please help to review

cabelitos:
To be honest I did not like the name,  it's can be confusing. I prefer that you change to something like iio_node_data or similar

yongli3:
@cabelitos Thanks for review. I think the direction_vector_data name can be changed to iio_node_direction_vector_data. Since next PR will combine all other on-channle senor data(temp/light/press...) into another one such as iio_node_double_data

edersondisouza:
Looks OK to me. But there are two things I'm confused:
 - You could define a 'base' struct and made the others, specific ones, extend it;
 - Nevertheless, isn't this solution actually conflicting with #2070?

yongli3:
@edersondisouza Thanks for your comments. For this code refactor, I am thinking using multiple PRs, to change the code one by one. The old PR changed a lot, it is hard to review/modify. I would like to change the easy part first, then add a new base struct define or using sol_vector, and finally we can achieve the same as #2070/#2013. Using this method, I think the submit process should be faster, how do you think about?

edersondisouza:
Ah, ok, so you are going to close #2070 then? And create a new one on top of this one? If that's so, ok. But please, update this PR sharing a base struct among private data structs.
That said, as I consider this refactor rather small, self-contained (only on `flow/iio/nodes.c`), maybe it would be easier to do them on only one PR, with different commits, each commit addressing a different task (one commit to share private data, another to share common methods and so on). This way we could see the direction it's going. But that's your call =D

yongli3:
@edersondisouza I added several new data structures and modified the iio color, please help to review

src/modules/flow/iio/nodes.c
```diff
@@ -25,34 +25,42 @@
...
     bool use_device_default_scale : 1;
     bool use_device_default_offset : 1;
 };
 
+struct color_data {
+    struct iio_channel_config channel_red;
+    struct iio_channel_config channel_green;
+    struct iio_channel_config channel_blue;
+    struct iio_device_config iio_config;
+};
```
edersondisouza: I'm rather confused: why `color_data` is composed of `iio_channel_config` but `iio_double_data` and `iio_direction_vector_data` are not?

src/modules/flow/iio/nodes.c
```diff
@@ -25,34 +25,42 @@
...
     bool use_device_default_offset : 1;
 };
 
-struct color_data {
-    struct sol_iio_config config;
+struct iio_channel_config {
     double scale;
     double offset;
+    struct sol_iio_channel *iio_channel;
+};
```
edersondisouza: Are we going to have scale and offset defined per channel? Is it necessary? It can make node types really bloated. Do we have any case in which this is necessary (except the future `channelX` stuff, that will be handled by multiple nodes)?

edersondisouza: Errr... now I've noticed that this is already the case, when using direction vector. Color is the odd one out.

src/modules/flow/iio/iio.json
```diff
@@ -421,8 +421,8 @@
            "name": "use_device_default_scale"
          },
          {
-             "data_type": "float",
-             "default": 0,
+             "data_type": "direction-vector",
```
edersondisouza: If we can use different scale and offset for each R, G or B, we'll need to have three options for each =/
Use direction-vector is convenient, but is also very counter-intuitive.

src/modules/flow/iio/nodes.c
edersondisouza: Guess I've got it: because there's no 'RGB' packet to set scale and offset. But then`iio_channel_config` is a misleading name, maybe `color_channel_config`?

src/modules/flow/iio/nodes.c
```diff
@@ -645,22 +653,22 @@ color_reader_cb(void *data, struct sol_iio_device *device)
     struct sol_flow_node *node = data;
     struct color_data *mdata = sol_flow_node_get_private_data(node);
     struct sol_rgb out = {
-        .red_max = mdata->out_range.max,
-        .green_max = mdata->out_range.max,
-        .blue_max = mdata->out_range.max
+        .red_max = mdata->iio_config.out_range.max,
+        .green_max = mdata->iio_config.out_range.max,
+        .blue_max = mdata->iio_config.out_range.max
     };
```
edersondisouza: You can simplify use of `color_data`, `double_data` and `direction_vector_data` if you do all of them extending a `base_data` - you called this one of `iio_device_config`. What I mean is:

```c
struct base_data {
    struct sol_iio_config config;
    struct sol_drange_spec out_range; // Not sure if it can be shared with color node, but currently is
    struct sol_iio_device *device;
    bool buffer_enabled : 1;
    bool use_device_default_scale : 1;
}

struct double_data {
    struct base_data base;
    struct sol_iio_channel *channel_val;
    double scale;
    double offset;
}
  
struct direction_vector_data {
    struct base_data base;
    struct sol_direction_vector scale;
    struct sol_direction_vector offset;
    struct sol_iio_channel *channel_x;
    struct sol_iio_channel *channel_y;
    struct sol_iio_channel *channel_z;
}

struct colour_data {
    struct base_data base;
    struct sol_iio_channel *channel_red;
    struct sol_iio_channel *channel_green;
    struct sol_iio_channel *channel_blue;
    double red_scale; // Or you can group `scale` and `offset` on a struct (and maybe even `channel`)
    double red_offset;// like you did. If you go that way, remember to name such struct as I said on
    double green_scale; // another comment
    double green_offset;
    double blue_scale;
    double blue_offset;
 }
```
Then you can unify `close` methods:

```c
static void
common_close(struct sol_flow_node *node, void *data)
{
    struct base_data *mdata = data;

    free((char *)mdata->config.trigger_name);
    if (mdata->device)
        sol_iio_close(mdata->device);
}
```

And on methods where you need the 'extended' struct, you use like

```c
   struct double_data *mdata = data;
   mdata->base.buffer_enabled = opts->buffer_size > -1;
```

Actually, it'll simplify uniting `tick` methods too.
What do you think?

yongli3: Thanks @edersondisouza ! I just changed the color data realted, and did not finish the modifications on other sensors. I will apply your suggestions: Using double for the RGB scale/offset

yongli3:
@edersondisouza I just upaded this PR, please help to review, thanks !

src/modules/flow/iio/nodes.c
```diff
@@ -696,48 +694,56 @@ color_create_channels(struct color_data *mdata, int device_id)
...
-    mdata->use_device_default_scale = opts->use_device_default_scale;
-    mdata->use_device_default_offset = opts->use_device_default_offset;
-    mdata->scale = opts->scale;
-    mdata->offset = opts->offset;
-    mdata->out_range = opts->out_range;
+    mdata->iio_base.use_device_default_scale = opts->use_device_default_scale;
+    mdata->iio_base.use_device_default_offset = opts->use_device_default_offset;
+
+    mdata->scale_red = opts->scale.x;
+    mdata->offset_red = opts->offset.x;
```
edersondisouza: Why `x` is `red`? There's no correlation between these values. Please don't use direction vector for colours. Use options for this. So you'd had something like:
```
mdata->scale_red = opts->scale_red;
mdata->offset_red = opts->offset_red;
```

Got it?

src/modules/flow/iio/nodes.c
```diff
@@ -184,9 +182,9 @@ gyroscope_close(struct sol_flow_node *node, void *data)
 {
     struct iio_direction_vector_data *mdata = data;
 
-    free((char *)mdata->config.trigger_name);
-    if (mdata->device)
-        sol_iio_close(mdata->device);
+    free((char *)mdata->iio_base.config.trigger_name);
+    if (mdata->iio_base.device)
+        sol_iio_close(mdata->iio_base.device);
```
edersondisouza: If you put `iio_base` as the _first_ element of each structure, as I showed you, you won't need a different `close` or `tick` for each node type. You can use:
```
struct iio_device_config *mdata = data;

free((char *)mdata->config.trigger_name);
if (mdata->device)
    sol_iio_close(mdata->device);
```
In a `common_close` method. Then, on JSON, you use only this `common_close` for all node types.

src/modules/flow/iio/nodes.c
yongli3: Got it, thanks!

src/modules/flow/iio/nodes.c
yongli3: Thanks @edersondisouza, just cast the void * to the base struct. I will change the close/tick functions as you said

yongli3:
@edersondisouza I updated the PR for tick function change, please help to review

src/modules/flow/iio/iio.json
```diff
@@ -424,7 +474,19 @@
              "data_type": "float",
              "default": 0,
              "description": "Scale to applied to device raw readings",
```
edersondisouza: Please mention on description to which colour is scale/offset being applied.

src/modules/flow/iio/iio.json
```diff
@@ -840,14 +954,24 @@
...
         "open": "proximity_open"
       },
+      "node_type": {
+        "access": [
+          "base"
+        ],
+        "data_type": "struct iio_node_type",
+        "extra_methods": {
+          "out_port": "SOL_FLOW_NODE_TYPE_IIO_PROXIMITY_SENSOR__OUT__OUT",
+		  "reader_cb": "iio_double_reader_cb"
```
edersondisouza: Please check these indentation issues on JSON.

edersondisouza:
Looking cool, nice job! I'll review more carefully later in the day, but overall looking good =D

src/modules/flow/iio/iio.json

src/modules/flow/iio/iio.json

src/modules/flow/iio/nodes.c
```diff
@@ -25,57 +25,101 @@
...
+    struct iio_device_config *mdata = data;
+    struct iio_node_type *type;
+
+    type = (struct iio_node_type *)sol_flow_node_get_type(node);
+
+    if (mdata->buffer_enabled) {
+        if (sol_iio_device_trigger(mdata->device) < 0)
+            goto error;
+    } else {
+        SOL_NULL_CHECK_GOTO(type->reader_cb, error);
```
edersondisouza: I don't think you need this check: it's you that controls this `type->reader_cb`, so it should never be `NULL`.

src/modules/flow/iio/nodes.c
```diff
@@ -101,15 +211,15 @@ create_device_channels(struct iio_direction_vector_data *mdata, int device_id)
 {
```
edersondisouza: For the sake of consistence, I'd change this method name to something like `gyro_create_channels`.

src/modules/flow/iio/nodes.c
```diff
@@ -101,15 +211,15 @@ create_device_channels(struct iio_direction_vector_data *mdata, int device_id)
...
     struct sol_iio_channel_config channel_config = SOL_IIO_CHANNEL_CONFIG_INIT;
 
-    mdata->device = sol_iio_open(device_id, &mdata->config);
-    SOL_NULL_CHECK(mdata->device, false);
+    mdata->iio_base.device = sol_iio_open(device_id, &mdata->iio_base.config);
+    SOL_NULL_CHECK(mdata->iio_base.device, false);
 
 #define ADD_CHANNEL(_axis) \
```
edersondisouza: Not for this patch, but maybe we could find a way to unify this `ADD_CHANNEL` stuff... don't you think? But no need to worry about it now, it's another future improvement =D

edersondisouza: I mean, there are lots of repeated macros, sometimes being used only for one channel...

src/modules/flow/iio/nodes.c
```diff
@@ -696,143 +569,91 @@ color_create_channels(struct color_data *mdata, int device_id)
...
-error:
-    sol_flow_send_error_packet(node, EIO, "%s", errmsg);
-    SOL_WRN("%s", errmsg);
+    mdata->scale_green = opts->scale_green;
+    mdata->offset_green = opts->offset_green;
 
-    return -EIO;
-}
+    mdata->scale_blue = opts->scale_green;
+    mdata->offset_blue = opts->offset_green;
```
edersondisouza: I think you wanted `opts->scale_blue` and `opts->offset_blue`.

edersondisouza:
Please, squash the two commits, only one is needed.

src/modules/flow/iio/nodes.c
yongli3: Good catch! thanks

src/modules/flow/iio/nodes.c
yongli3: I agree with you! We can modify the open and device_channle functions too. Just the same as reader_cb, using a common function

src/modules/flow/iio/nodes.c

src/modules/flow/iio/nodes.c

yongli3:
@edersondisouza I combined the two commits into one, please help to review

edersondisouza:
Integrated with a minor addition on commit message, stating `Node type`. Many thanks for you effort!

PR #2103, opened 2016-05-26
Node.js bindings: Organize node.js system constants like errno

gabrielschulhof:
Also reverse-maps them for use in the bindings

We need these constants to return more legible errors from the bindings.


nagineni:
+1
thanks!

PR #2105, opened 2016-05-27
LWM2M: Use struct sol_blob for string and opaque resource types.

cabelitos:

### DIFF
--- a/src/lib/comms/sol-lwm2m.c
+++ b/src/lib/comms/sol-lwm2m.c
@@ -1243,7 +1243,7 @@ get_resource_len(const struct sol_lwm2m_resource *resource, uint16_t index,
     switch (resource->data_type) {
     case SOL_LWM2M_RESOURCE_DATA_TYPE_STRING:
     case SOL_LWM2M_RESOURCE_DATA_TYPE_OPAQUE:
-        *len = resource->data[index].bytes.len;
+        *len = resource->data[index].blob->size;
         return 0;
     case SOL_LWM2M_RESOURCE_DATA_TYPE_INT:
     case SOL_LWM2M_RESOURCE_DATA_TYPE_TIME:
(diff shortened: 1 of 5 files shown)


src/lib/comms/sol-lwm2m.c
```diff
@@ -1780,13 +1784,25 @@ sol_lwm2m_resource_init(struct sol_lwm2m_resource *resource,
...
-    if (r < 0)
-        free(resource->data);
+    va_end(ap);
+    return 0;
 
+err_ref:
+    r = -EOVERFLOW;
+err_exit:
+    if (data_type == SOL_LWM2M_RESOURCE_DATA_TYPE_OPAQUE ||
+        data_type == SOL_LWM2M_RESOURCE_DATA_TYPE_OPAQUE) {
```
otaviobp: Or is comparing equal const values

otaviobp: I think one should be SOL_LWM2M_RESOURCE_DATA_TYPE_STRING

src/samples/coap/lwm2m-client.c
```diff
@@ -262,8 +273,7 @@ read_security_server_obj(void *instance_data, void *user_data,
     switch (res_id) {
     case SECURITY_SERVER_SERVER_URI_RES_ID:
         SOL_LWM2M_RESOURCE_INIT(r, res, 0, 1,
-            SOL_LWM2M_RESOURCE_DATA_TYPE_STRING,
-            sol_str_slice_from_str("coap://localhost:5683"));
+            SOL_LWM2M_RESOURCE_DATA_TYPE_STRING, &addr);
```
otaviobp: I am assuming you are using blob instead of str_slice because of memory control (ref/unref), right?
In this case, can't you do a helper to add a string literal? Because it is a bit annoying to have to create a blob in all string usages

bdilly:
@otaviobp @cabelitos I've merged it with the change on line 1799 (regarding constants).

The helper may be added later as another patch.

thanks

PR #2106, opened 2016-05-27
Build: Prevent node.js build from re-running during install

gabrielschulhof:

@nagineni does this fix the problem with the integration?


gabrielschulhof:
@nagineni I have now fixed the problem with ```make check```. Can you PTAL?

nagineni:
:+1: Thanks for fixing this!

PR #2109, opened 2016-05-30
Node.js bindings: Fix error in spi JS bindings

nagineni:
Do not create a new Buffer instance for rxData since it is already a Buffer instance.


nagineni:
@gabrielschulhof Could you please have a look?

gabrielschulhof:
:+1:

PR #2104, opened 2016-05-27
IO API changes v3

cabelitos:
Changes since v2:
* Added  ec3a3a9

PR #2112, opened 2016-05-30
fixes on top of 2104

bdilly:
Fixing some issues I found when applying pr #2104

### DIFF
--- a/src/lib/common/include/sol-pin-mux-modules.h
+++ b/src/lib/common/include/sol-pin-mux-modules.h
@@ -93,7 +93,7 @@ struct sol_pin_mux {
      *
      * @return @c 0 on success, error code (always negative) otherwise.
      */
-    int (*aio)(const int device, const int pin);
+    int (*aio)(int device, int pin);
 
     /**
      * @brief Callback to setup the given pin to operate in the given GPIO configuration.
@@ -106,7 +106,7 @@ struct sol_pin_mux {
      *
      * @return @c 0 on success, error code (always negative) otherwise.
      */
-    int (*gpio)(const uint32_t pin, const struct sol_gpio_config *config);
+    int (*gpio)(uint32_t pin, const struct sol_gpio_config *config);
 
     /**
      * @brief Callback to setup the pins used of the given i2c bus number to operate in I2C mode.
@@ -118,7 +118,7 @@ struct sol_pin_mux {
      *
      * @return @c 0 on success, error code (always negative) otherwise.
      */
-    int (*i2c)(const uint8_t bus);
+    int (*i2c)(uint8_t bus);
 
     /**
      * @brief Callback to setup the given pin to operate as PWM.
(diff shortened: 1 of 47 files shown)


otaviobp:
lgtm

src/lib/comms/include/sol-http-client.h
```diff
@@ -92,7 +92,7 @@ struct sol_http_request_interface {
      *       inside this callback.
      */
     ssize_t (*on_data)(void *userdata, struct sol_http_client_connection *connection,
-        const struct sol_buffer *buffer);
+        struct sol_buffer *buffer);
```
ceolin: Taking a better look I think it's better let it const. My points are:

- The memory inside buffer is `const` (allocated by curl), so the user will not be able to change it.
- Without the `const` is not clear who is responsible to call `sol_buffer_fini`. Now the caller is calling the `sol_buffer_fini` after the callback.
- Functions to get/see the buffer contents already receive a `const struct sol_buffer`. i.e `sol_buffer_at`, `sol_buffer_get_slice` ...

If you still think it's better remove the const, it will be necessary change ./src/samples/http/download.c too.

src/lib/io/sol-i2c-common.c
```diff
@@ -138,10 +139,19 @@ i2c_dispatcher_exec_op(void *data)
...
+            if (++dispatcher->retry >= SOL_I2C_MAX_RETRIES) {
+                SOL_ERR("Failed to schedule I2C operation.");
+                goto exit;
+            }
+        }
+
+        SOL_ERR("Failed to set slave address 0x%02x on I2C bus: %d.",
+            op_set->addr, sol_i2c_bus_get(shared->i2c));
         i2c_dispatcher_end_set(dispatcher, -1);
+
```
ceolin: in case of `r < 0` it shouldn't return `true` only if `r == -EBUSY` and the number of retry is less than `SOL_I2C_MAX_RETRIES` ?

src/lib/common/include/sol-pin-mux-modules.h
```diff
@@ -93,7 +93,7 @@ struct sol_pin_mux {
      *
      * @return @c 0 on success, error code (always negative) otherwise.
      */
-    int (*aio)(const int device, const int pin);
+    int (*aio)(int device, int pin);
```
ceolin: I remember have asked @mbelluzzo about those consts but not remember his answer :(

@mbelluzzo do you remember why do you put those consts ?

bdilly: @ceolin we talked about it already in other PRs / IRC. There is no reason.

src/lib/io/sol-i2c-common.c
bdilly: you're right. it should return true on this case.

bdilly:
merged with proposed changes

PR #2122, opened 2016-06-02
build: Change OS selection order to prioritize small OSes

ibriano:
Depending on how an OS builds for native, Soletta may still pass the
Linux detection test. Leaving Linux as the last OS to check for in
Kconfig allows any other OS to be used regardless of that test.


glima:
We could do better than that later, but +1 for now

bdilly:
ack

PR #2123, opened 2016-06-02
oic: Use sol-cert API to load pre shared keys - v3

otaviobp:
Replaces #2016

Changelog:
 - Don't use base64 for file names. Use file name scaping.
 - Save and load certificates in an user context config dir, inside an app specific directory.
 - Remove already merged commits
 - Other minor fixes

### DIFF
--- a/src/lib/common/sol-certificate-impl-linux.c
+++ b/src/lib/common/sol-certificate-impl-linux.c
@@ -17,6 +17,7 @@
  */
 
 #include <sys/stat.h>
+#include <unistd.h>
 
 #include "sol-certificate.h"
 #define SOL_LOG_DOMAIN &_sol_certificate_log_domain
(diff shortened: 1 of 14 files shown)


src/lib/common/sol-certificate-impl-linux.c
```diff
@@ -173,3 +198,30 @@ sol_cert_get_contents(const struct sol_cert *cert)
 
     return sol_file_reader_to_blob(fr);
 }
+
+SOL_API int
```
ceolin: why are you not using `ssize_t` ?

src/shared/sol-util-file.c
```diff
@@ -615,3 +616,81 @@ sol_util_file_get_basename(struct sol_str_slice path)
...
+SOL_API ssize_t
+sol_util_write_file_slice(const char *path, struct sol_str_slice slice)
+{
+    FILE *fp;
+    size_t bytes;
+    int r;
+
+    SOL_NULL_CHECK(path, -EINVAL);
+
+    errno = 0;
```
ceolin: it's not necessary since if `fopen` fails it will return `NULL` and sets the `errno`

src/shared/sol-util-file.c
```diff
@@ -615,3 +616,81 @@ sol_util_file_get_basename(struct sol_str_slice path)
...
+    fp = fopen(path, "we");
+    if (!fp)
+        return -errno;
+
+    bytes = fwrite(slice.data, 1, slice.len, fp);
+
+    errno = 0;
```
ceolin: ditto

src/shared/sol-util-file.c
```diff
@@ -615,3 +616,81 @@ sol_util_file_get_basename(struct sol_str_slice path)
...
+
+    dir = getenv("XDG_CONFIG_HOME");
+    if (dir) {
+        r = sol_buffer_append_slice(buffer, sol_str_slice_from_str(dir));
+        SOL_INT_CHECK(r, < 0, r);
+        r = sol_buffer_append_char(buffer, '/');
+        SOL_INT_CHECK(r, < 0, r);
+        r = sol_buffer_append_slice(buffer, sol_platform_get_appname());
+        SOL_INT_CHECK(r, < 0, r);
+        return sol_buffer_append_char(buffer, '/');
```
ceolin: maybe would be clear use:
```
r = sol_buffer_append_printf(buffer, "%s/%.*s/", dir,
    SOL_STR_SLICE_PRINT(sol_platform_get_appname()));
SOL_INT_CHECK(r, < 0, r);
```

ceolin:
pr looks awesome, just minor comments. Please check if they are valid.

Issue #1536, opened 2016-03-04
build: "root' leftovers after 'sudo make install' in the build folder


mbelluzzo:
If you use 'sudo' to install soletta on your system, some 'root' owned files are left in your local build folder obliging you to use 'sudo make clean' instead of just 'make clean' which is the expected behavior.

#### Steps to reproduce

 1. make alldefconfig
 2. make -j8
 3. sudo make install
 4. make clean

output:
```
   CLEAN   ./build/
rm: cannot remove './build/soletta_sysroot/usr/share/soletta/boards/50-default.json': Permission denied
rm: cannot remove './build/soletta_sysroot/usr/share/soletta/flow/schemas/node-type-genspec.schema': Permission denied
rm: cannot remove './build/soletta_sysroot/usr/share/gdb/auto-load/libsoletta.so.0.0.1-gdb.py': Permission denied
tools/build/Makefile.common:37: recipe for target 'clean' failed
make: *** [clean] Error 1
```

PR #2124, opened 2016-06-02
build: avoid files moved to build_dir on installing phase

bdilly:
Otherwise it may lead to situations where the users used
by make and make clean diverge from make install and
clean can't actually remove the whole build directory.

Close gh-1536


otaviobp:
+1

Issue #1675, opened 2016-03-18
API Changes: Sol-platform


cabelitos:
#### Task Description

- [ ] Remove macros `CHUNK_READ_SIZE`, `CHUNK_READ_MAX` and `CHUNK_MAX_TIME_NS` from the header. They should not be exported
- [ ] Rename `SOL_PLATFORM_TARGET_POWEROFF` to `SOL_PLATFORM_TARGET_POWER_OFF`
- [ ] The function `sol_platform_get_state()` should return an `enum sol_platform_state` and not an `int`

#### Dependencies
None.

amandawrcoelho:
@ceolin we can close this :)

PR #2127, opened 2016-06-02
API Changes: Sol-platform

amandawrcoelho:
Move CHUNK_READ_SIZE, CHUNK_READ_MAX and CHUNK_MAX_TIME_NS to
sol-util-internal.h

Rename function

Now sol_platform_get_state() return an enum

issue #1675


ceolin:
looks ok

src/lib/common/sol-platform.c
```diff
@@ -206,7 +207,7 @@ sol_platform_get_board_name(void)
     return board_name;
 }
 
-SOL_API int
+SOL_API enum sol_platform_state
 sol_platform_get_state(void)
 {
     return sol_platform_impl_get_state();
```
otaviobp: I would change the return type from sol_platform_impl_get_state to enum too.

otaviobp: And you will need to fix the implementation of contiki and zephyr.

amandawrcoelho: After a brief talk, its better dont change the type of this function, because enum type its not the best way.

PR #2129, opened 2016-06-03
Node.js bindings: Adapt JS bindings to the latest API changes

nagineni:
Adapt AIO,SPI and I2C JS bindings to the latest API changes.

### DIFF
--- a/bindings/nodejs/lib/i2c.js
+++ b/bindings/nodejs/lib/i2c.js
@@ -16,7 +16,7 @@
  * limitations under the License.
  */
 
-var soletta = require( 'bindings' )( 'soletta' ),
+var soletta = require( './lowlevel'),
     _ = require( 'lodash' );
 
 exports.open = function( init ) {
@@ -42,16 +42,19 @@ exports.open = function( init ) {
 var I2CBus = function( i2cbus ) {
     if ( !this._isI2CBus )
         return new I2CBus ( i2cbus );
+    this._construct();
     this._i2cbus = i2cbus;
 }
 
 _.extend( I2CBus.prototype, {
     _isI2CBus: true,
     _pending: null,
-    busy: {
-        get: function() {
-            return soletta.sol_i2c_busy( this._i2cbus );
-        }
+    _construct: function() {
+        Object.defineProperty(this, 'busy', {
+            get: function() {
+                return this._pending === "EBUSY";
+            }
+        } );
     },
 
     read: function( device, size, register, repetitions ) {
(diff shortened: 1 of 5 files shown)


nagineni:
@gabrielschulhof, Could you please take a look?

bindings/nodejs/lib/i2c.js
```diff
@@ -62,26 +65,26 @@ _.extend( I2CBus.prototype, {
...
                     this._i2cbus, register, size,
-                    function( register, data, status ) {
+                    _.bind( function( register, data, status ) {
                     this._pending = null;
                     fulfill( data );
-                });
+                }, this ) );
             }
-            if (!this._pending)
+            if ( !this._pending || typeof this._pending === "string" )
```
gabrielschulhof: It would be useful to add the string value of ```this._pending``` to the error message.

bindings/nodejs/lib/i2c.js
```diff
@@ -98,19 +101,19 @@ _.extend( I2CBus.prototype, {
...
-                    function( register, data, status ) {
+                    _.bind( function( register, data, status ) {
                     this._pending = null;
                     fulfill();
-                } );
+                }, this ) );
             }
-            if (!this._pending)
+            if ( !this._pending || typeof this._pending === "string" )
                 reject( new Error( "I2C write failed" ) );
```
gabrielschulhof: Same here ...

bindings/nodejs/lib/spi.js
```diff
@@ -64,10 +64,12 @@ _.extend( SPIBus.prototype, {
                function( txData, rxData, count ) {
                    if ( rxData !== null ) {
                        fulfill( rxData );
+                   } else {
+                       reject( new Error( "SPI transmission failed" ) );
                    }
            });
 
-           if ( ( typeof returnStatus === 'undefined' ) || !returnStatus ) {
+           if ( ( typeof returnStatus === 'undefined' ) || returnStatus < 0) {
```
gabrielschulhof: You don't need to worry about a ```returnStatus``` of ```undefined```, because that can only happen if an exception is thrown, in which case JS execution will jump to the ```reject()``` handler and ```returnStatus``` will not be assigned at all, and this code will not be reached.

bindings/nodejs/src/functions/aio.cc
```diff
@@ -155,5 +155,4 @@ NAN_METHOD(bind_sol_aio_pending_cancel)
     sol_aio_pending *aio_pending = (sol_aio_pending *)SolAio::Resolve(jsAioPending);
```
gabrielschulhof: You should have a null-check here.

bindings/nodejs/src/functions/i2c.cc
```diff
@@ -297,6 +304,7 @@ NAN_METHOD(bind_sol_i2c_write_quick)
     if (!i2c_pending) {
         delete callback;
         hijack_unref();
+        info.GetReturnValue().Set(ReverseLookupConstant("E", abs(errno)));
```
gabrielschulhof: ```errno``` should never be negative. This is a bug on the native side, confusing a return value (which should be negative) with setting the global ```errno```, which should be positive.

bindings/nodejs/lib/i2c.js
nagineni: Make sense. I will fix.

bindings/nodejs/lib/spi.js
nagineni: Okay

bindings/nodejs/src/functions/aio.cc
nagineni: Okay

bindings/nodejs/src/functions/i2c.cc
nagineni: I will fix this in the native side by setting a positive value to `errno`. @cabelitos hope it is okay with you.

Issue #1742, opened 2016-03-28
API Changes: IIO


bdilly:
#### Task Description

Some changes are required to match [Soletta's API design conventions](https://github.com/solettaproject/soletta/wiki/API-design-conventions)

 - [ ] Rename `sol_iio_device_trigger_now()` to `sol_iio_device_trigger()`
 - [ ] Change return type of functions that may fail to return error code instead of boolean `false`. Example: `sol_iio_read_channel_value()`

amandawrcoelho:
We can close this :D

ceolin:
certainly !!! Thanks

PR #2133, opened 2016-06-03
API Changes: IIO

amandawrcoelho:
Rename sol_iio_device_trigger_now() to sol_iio_device_trigger()

Change the type of iio_read_buffer_channel_value(), sol_iio_read_channel_value()
sol_iio_device_trigger() and sol_iio_device_start_buffer. Now they return an int,
in case of fail, they return a negative error code.

issue #1742

### DIFF
--- a/src/lib/io/sol-iio.c
+++ b/src/lib/io/sol-iio.c
@@ -971,7 +971,7 @@ error:
     return NULL;
 }
 
-static bool
+static int
 iio_read_buffer_channel_value(struct sol_iio_channel *channel, double *value)
 {
     uint64_t data = 0;
@@ -982,18 +982,18 @@ iio_read_buffer_channel_value(struct sol_iio_channel *channel, double *value)
     struct sol_iio_device *device = channel->device;
     uint8_t *buffer = device->buffer.data;
 
-    SOL_NULL_CHECK(buffer, false);
+    SOL_NULL_CHECK(buffer, -EINVAL);
 
     if (channel->storagebits > 64) {
         SOL_WRN("Could not read channel [%s] value - more than 64 bits of"
             " storage - found %d. Use sol_iio_read_channel_raw_buffer() instead",
             channel->name, channel->storagebits);
-        return false;
+        return -EBADMSG;
     }
 
     if (channel->offset_in_buffer + channel->storagebits > device->buffer_size * 8) {
         SOL_WRN("Invalid read on buffer.");
-        return false;
+        return -EBADMSG;
     }
 
     offset_bytes = channel->offset_in_buffer / 8;
@@ -1027,10 +1027,10 @@ iio_read_buffer_channel_value(struct sol_iio_channel *channel, double *value)
     if (!channel->processed)
         *value = (*value + channel->offset) * channel->scale;
 
-    return true;
+    return 0;
 }
 
-SOL_API bool
+SOL_API int
 sol_iio_read_channel_value(struct sol_iio_channel *channel, double *value)
 {
     int len;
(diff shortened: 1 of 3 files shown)


src/modules/flow/iio/nodes.c
```diff
@@ -49,16 +49,16 @@ reader_cb(void *data, struct sol_iio_device *device)
         .min = mdata->out_range.min,
         .max = mdata->out_range.max
     };
-    bool b;
+    int b;
```
otaviobp: use r instead of b here. People that are reading the code may think it is a boolean because it is called b.

src/modules/flow/iio/nodes.c
```diff
@@ -209,16 +209,16 @@ magnet_reader_cb(void *data, struct sol_iio_device *device)
...
```
otaviobp: same here

src/modules/flow/iio/nodes.c
```diff
@@ -49,16 +49,16 @@ reader_cb(void *data, struct sol_iio_device *device)
...
 
     b = sol_iio_read_channel_value(mdata->channel_x, &out.x);
-    if (!b) goto error;
+    if (b < 0) goto error;
```
otaviobp: I would use SOL_INT_CHECK_GOTO(r, <0, error) here, so it will log on errors.

otaviobp:
other than that, the patch looks good to me

src/lib/io/sol-iio.c
```diff
@@ -1054,21 +1054,21 @@ sol_iio_read_channel_value(struct sol_iio_channel *channel, double *value)
     if (!r) {
         SOL_WRN("Could not read channel [%s] in device%d", channel->name,
             device->device_id);
-        return false;
+        return -EBADMSG;
```
edersondisouza: Why `EBADMSG`? If `craft_filename_path` failed is more likely `ENOMEM` or `ENOENT`... but maybe there's some new guideline...

src/lib/io/sol-iio.c
```diff
@@ -1054,21 +1054,21 @@ sol_iio_read_channel_value(struct sol_iio_channel *channel, double *value)
...
     }
 
     len = sol_util_read_file(path, "%" SCNd64, &raw_value);
     if (len < 0) {
         SOL_WRN("Could not read channel [%s] in device%d", channel->name,
             device->device_id);
-        return false;
+        return -EBADMSG;
```
edersondisouza: I think that `-EIO` would be better.

src/lib/io/sol-iio.c
```diff
@@ -1085,64 +1085,64 @@ calc_channel_offset_in_buffer(const struct sol_iio_channel *channel)
...
     }
 
     device->buffer_size = calc_buffer_size(device);
     i = sol_buffer_ensure(&device->buffer, device->buffer_size);
     if (i < 0) {
         SOL_WRN("Could not alloc buffer for device. No readings will be performed");
-        return false;
+        return -EBADMSG;
```
edersondisouza: This should be `-ENOMEM`, don't you think?

amandawrcoelho: Yeap! I'm getting better with errnos :D

amandawrcoelho: I will this errnos! Thank you!

src/modules/flow/iio/nodes.c
amandawrcoelho: Yeap, its really better. 
Tks :)

PR #2140, opened 2016-06-06
API Changes: IIO

amandawrcoelho:
Rename sol_iio_device_trigger_now() to sol_iio_device_trigger()

Change the type of iio_read_buffer_channel_value(), sol_iio_read_channel_value()
sol_iio_device_trigger() and sol_iio_device_start_buffer. Now they return an int,
in case of fail, they return a negative error code.

issue #1742


edersondisouza:
I guess that checkstyle fails...

PR #2142, opened 2016-06-07
i2c & aio: set errno to a positive error value

nagineni:
As mentioned here https://github.com/solettaproject/soletta/pull/2129#discussion_r65670041 global `errno` should always be positive, so this patch sets `errno` to a positive error value.


nagineni:
@gabrielschulhof @cabelitos, please have a look.

barbieri:
+1

Issue #1663, opened 2016-03-18
API Changes: Sol-str-slice


cabelitos:
#### Task Description

- [ ] Replace the `caseeq` keyword with`case_eq`. Example: `sol_str_slice_str_caseeq()` should be `sol_str_slice_str_case_eq()`
- [ ] `sol_str_slice_to_int()` should receive a pointer to `long int` instead of `int` as argument.

#### Dependencies
None

ceolin:
Fixed !

PR #2143, opened 2016-06-07
API Changes: Sol-str-slice

amandawrcoelho:
Replace the caseeq keyword withcase_eq

sol_str_slice_to_int() now receive a pointer to long int.

issue #1663


src/lib/comms/sol-mavlink.c
```diff
@@ -633,6 +633,7 @@ sol_mavlink_parse_addr_protocol(const char *str, struct sol_str_slice *addr, int
     struct sol_vector tokens;
     struct sol_str_slice slice = sol_str_slice_from_str(str);
     const void *init;
+    long int *r = (long int*)port;
```
otaviobp: Casting pointers from different sizes can lead to invalid memory access or undefined behaviors in system with different endianness.
The ideal is to use a local variable and then attribute the value to port.

PR #2146, opened 2016-06-08
check if sol_i2c_set_slave_address returns -ERRNO

brunobottazzini:
Since last change from patch 026c6f6945998ee1b365e8bb7f20e556f42efba4 the
function sol_i2c_set_slave_address does not return bool and yet it returns
a -ERRNO informing the probable cause of the error.

Thus the module should use in the same way instead of checking if false.


otaviobp:
+1

bdilly:
:+1:

Issue #1759, opened 2016-03-28
API Changes: EFIVARS


bdilly:
#### Task Description

Some changes are required to match [Soletta's API design conventions](https://github.com/solettaproject/soletta/wiki/API-design-conventions)

 - [ ] Macros and inline functions missing docs. Review CREATE_BLOB and CREATE_BUFFER
 - [ ] `read_raw()` requires better documentation

ceolin:
https://github.com/solettaproject/soletta/pull/2150/commits/ebe1f8f3d3b7b69749846595542aeb540683e4e2

Issue #1718, opened 2016-03-23
API Changes: MQTT


cabelitos:
#### Task Description
- [ ] Rename `keepalive` to `keep_alive` in `struct sol_mqtt_config`

#### Dependencies
None

ceolin:
https://github.com/solettaproject/soletta/commit/959ecb6e5f3a5eab550501c2412c50352aeef2d9

Issue #1662, opened 2016-03-18
API Changes: Sol_vector and Sol_ptr_vector


cabelitos:
#### Task Description

- [ ] Rename `sol_ptr_vector_get_nocheck()`/`sol_vector_get_nocheck()` to `sol_ptr_vector_get_no_check()`/`sol_vector_get_no_check()`
- [ ] In `sol_ptr_vector_match_first()` rename the `elem` parameter to `template`
- [ ] Rename all functions that containts `take` keyword to `steal`. Example: `sol_vector_take_data()` should be `sol_vector_steal_data()`
- [ ] `sol_ptr_vector_find_last()`/`sol_ptr_vector_find_first()` should return an `int32_t` type. 

#### Dependencies

None.

ceolin:
fixed

PR #2150, opened 2016-06-09
General fixes

ceolin:
It fix the follow issues:

https://github.com/solettaproject/soletta/issues/1718
https://github.com/solettaproject/soletta/issues/1662
https://github.com/solettaproject/soletta/issues/1759

### DIFF
--- a/src/lib/io/include/sol-efivarfs-storage.h
+++ b/src/lib/io/include/sol-efivarfs-storage.h
@@ -65,6 +65,10 @@ int sol_efivars_write_raw(const char *name, struct sol_blob *blob,
 /**
  * @brief Read stored contents and set to buffer.
  *
+ * This function will read the contents of the EFI variable @a name
+ * into the buffer @a buffer. It will try to resize the buffer if
+ * needed.
+ *
  * @param name name of property. It will look for an EFI variable with
  * this name.
  * @param buffer buffer that will be set with read contents.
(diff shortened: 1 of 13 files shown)


src/lib/io/include/sol-efivarfs-storage.h
```diff
@@ -252,6 +430,21 @@ sol_efivars_read_string(const char *name, char **value)
...
+ *
+ * @param name name of property. It will create a new EFI variable with
+ * this name.
+ * @param value The value to be written.
+ * @param cb callback to be called when writing finishes. It contains status
+ * of writing: if failed, is lesser than zero.
+ * @param data user data to be sent to callback @a cb
+ *
+ * return @c 0 on success, a negative number on failure
+ */
```
bdilly: It seems good.
I'm wondering if there isn't any way to enforce inline functions documentation on doxygen / doxy-coverage,
otherwise we may let some functions behind.

bdilly:
PR seems good to be merged

src/lib/io/include/sol-efivarfs-storage.h
ceolin: @amandawrcoelho is looking for it in doxygen since she is documenting those missing functions.
Otherwise will be really hard to keep it tracked.

ceolin:
thanks

Issue #1753, opened 2016-03-28
API Changes: PWM


bdilly:
#### Task Description

Some changes are required to match [Soletta's API design conventions](https://github.com/solettaproject/soletta/wiki/API-design-conventions)

 - [ ] Some structs are missing a "@brief" command
 - [ ] Change return type of functions that may fail to return error code instead of boolean `false`. Example: `sol_pwm_get_enabled()`
 - [ ] Rename `sol_pwm_get_enabled()` to `sol_pwm_is_enabled()`

ceolin:
https://github.com/solettaproject/soletta/commit/fc2ac5dcd6afa05137b5d4f6486379e34d87dc62

PR #2152, opened 2016-06-09
PWM fixes

ceolin:
https://github.com/solettaproject/soletta/issues/1753


bdilly:
Changing the return type seems good, I'm ok with merging this commit.
I'm not so sure about the other one, s/get_enabled/is_enabled/, since we are using get/set for properties.
But it's a minor change anyway. So please feel free to merge it.

ceolin:
merged, thanks

PR #2155, opened 2016-06-10
Flow apichanges v3

kaspersky:
#1677 

#2063 

The "function read_file() on struct sol_flow_parser_client should use sol_buffer" and "sol_flow_parse_buffer_metatype() should use sol_buffer too" changes are more complex and I will do them in a separate PR at a later point.


kaspersky:
#2159

PR #2156, opened 2016-06-10
build: Avoid warning about BOARD_NAME redefined

ibriano:
BOARD_NAME is now always available in the config, even as an empty
string, so it comes as a #define through the generated sol_config.h.
Having it also in the command line, in some cases, generates a
redefinition warning.


ceolin:
:+1:

ibriano:
Merged

PR #2158, opened 2016-06-10
oic: Don't stop processing on unsupported cbor types and doxygen fixes

otaviobp:


bdilly:
Looks ok
I'm changing semaphoreci commands a bit to avoid these timeouts and I'll trigger builds again

otaviobp:
merged

PR #2166, opened 2016-06-10
Add Soletta trademark for every first usage in each file - V2

brunobottazzini:
Differences from V1: 

- Adding C, H and JS files as suggested by @ceolin


bdilly:
@brunobottazzini it seems good.

I would suggest you to do the same in our wiki repository https://github.com/solettaproject/soletta.wiki.git

Thanks

bdilly:
merged, thanks

PR #2153, opened 2016-06-10
Node.js bindings: Adapt JS bindings to the latest API changes

nagineni:
Adapt AIO,SPI and I2C JS bindings to the latest API changes.

Replaces #2129 
 
Signed-off-by: Sudarsana Nagineni sudarsana.nagineni@intel.com

PR #2168, opened 2016-06-13
Node.js bindings: Adapt JS bindings to the latest API changes

nagineni:
Adapt AIO,SPI and I2C JS bindings to the latest API changes. Also, to make the bindings consistent, always require the low-level bindings.

Replaces #2153

### DIFF
--- a/bindings/nodejs/src/functions/i2c.cc
+++ b/bindings/nodejs/src/functions/i2c.cc
@@ -16,6 +16,7 @@
  * limitations under the License.
  */
 
+#include <errno.h>
 #include <v8.h>
 #include <node.h>
 #include <nan.h>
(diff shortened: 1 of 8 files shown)


nagineni:
@gabrielschulhof

bindings/nodejs/src/functions/i2c.cc
```diff
@@ -75,7 +77,11 @@ NAN_METHOD(bind_sol_i2c_set_slave_address)
     sol_i2c *i2c = (sol_i2c *)SolI2c::Resolve(jsI2c);
 
     int returnValue = sol_i2c_set_slave_address(i2c, info[1]->Uint32Value());
-    info.GetReturnValue().Set(Nan::New(returnValue));
+    if (returnValue < 0) {
+        info.GetReturnValue().Set(ReverseLookupConstant("E", abs(returnValue)));
```
gabrielschulhof: It is sufficient to use ```-returnValue``` here. No need for abs, since you already know that it's negative.

gabrielschulhof:
Just one minor nitpick. Feel free to land this with that one change in place. :+1:

PR #2056, opened 2016-05-18
Add IO Convenience API, Platform API, Update API. Synchronize APIs.

zolkis:
Add specification for Platform and Update APIs.

Add an experimental IO API that uses the same style for all IO types (construct, configure, read, write, abort, onread, onerror). 

Other small changes coming from implementation feedback are synchronized here.

PR #2171, opened 2016-06-13
Properly organize an open pull request

bdilly:
PR #2056

Instead of using many merges, still valid commits
were cherry-picked and applied on top of current HEAD.

@ceolin ptal


ceolin:
it is ok to me

bdilly:
merged, thanks

Issue #1713, opened 2016-03-23
API Changes: Flow-Metatypes


cabelitos:
#### Task Description
- [ ] Remove typdefs
- [x] Replace all occurences of metatype to meta_type. Example: `sol_flow_metatype_context` to `sol_flow_meta_type_context`
- [ ] Function `read_file` in `sol_flow_metatype_context` should receive a `struct sol_buffer` as argument instead of `const char **buf, size_t *size`

#### Dependencies

None.

barbieri:
I'm not sure i like meta_type more than metatype... :-D

cabelitos:
Me too :[

bdilly:
ok, so let it as metatype

ceolin:
well, use typedef there will turn the illegible, i.e
`int (*(sol_flow_metatype_get_generate_code_start_func)(const struct sol_str_slice name))(const struct sol_flow_metatype_context *ctx, struct sol_buffer *out);`

i think is better let as it is.

ceolin:
done

PR #2172, opened 2016-06-14
Metatype fixes

ceolin:
https://github.com/solettaproject/soletta/issues/1713
https://github.com/solettaproject/soletta/issues/1677

### DIFF
--- a/src/lib/flow/include/sol-flow-parser.h
+++ b/src/lib/flow/include/sol-flow-parser.h
@@ -20,6 +20,7 @@
 
 #include <stdbool.h>
 
+#include "sol-buffer.h"
 #include "sol-flow.h"
 #include "sol-flow-resolver.h"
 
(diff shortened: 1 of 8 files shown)


src/lib/flow/include/sol-flow-parser.h
```diff
@@ -107,16 +108,15 @@ int sol_flow_parser_del(
  * @brief Parses a flow description contained in @a buf and returns the resulting node type.
  *
  * @param parser The Flow Parser handle
- * @param buf Flow description to be parsed
- * @param len Buffer's size
+ * @param buf Flow description to be parsed. @note The buffer will be
```
otaviobp: @note will create a note section in the middle of parameters definition. I found this weird. Is this desired? I prefer to add notes at the end of the comments.

otaviobp:
lgtm. Just check the @note comments.

src/lib/flow/include/sol-flow-parser.h
ceolin: i did not noticed that. I'll remove the `@note` from there. I prefer this than a `@note` in the final because it's clear and the this info is related with the parameter.

ceolin:
merged with the fixes.

PR #2173, opened 2016-06-14
API Changes: HTTP Documentation v2

amandawrcoelho:
Describe  some enums

### DIFF
--- a/src/lib/comms/include/sol-http.h
+++ b/src/lib/comms/include/sol-http.h
@@ -45,15 +45,49 @@ extern "C" {
  * sol_http_client_request() or sol_http_client_request_with_interface()
  */
 enum sol_http_method {
+    /**
+     * Requests data from a specified resource
+     */
     SOL_HTTP_METHOD_GET,
+    /**
+     * Submits data to be processed to a specified resource
+     */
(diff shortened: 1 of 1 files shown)


src/lib/comms/include/sol-http.h
```diff
@@ -45,15 +45,49 @@ extern "C" {
...
+     * Performs a message loop-back test along the path to the
+     * target resource
+     */
     SOL_HTTP_METHOD_TRACE,
+    /**
+     * Used to update partial resources
+     */
     SOL_HTTP_METHOD_PATCH,
+    /**
+     * Invalid input method
```
bdilly: not clear imo
I believe we could elaborate a bit more about it

bdilly:
seems good overall

src/lib/comms/include/sol-http.h
amandawrcoelho: What do you about "client makes an HTTP request by using an HTTP method that does not comply with the HTTP specifications"?

bdilly: yeah, it's better.
thanks

amandawrcoelho:
@bdilly tks for the review! :D

PR #2175, opened 2016-06-14
http: Returning better error code

ceolin:


src/lib/comms/sol-http-common.c
```diff
@@ -40,7 +40,7 @@ sol_http_params_add(struct sol_http_params *params,
     ptr = sol_vector_append(&params->params);
     if (!ptr) {
         SOL_WRN("Could not append option to parameter vector");
-        return -EINVAL;
+        return -ENOMEM;
```
bdilly: makes sense
+1

ceolin:
in, thanks

PR #2177, opened 2016-06-15
platform: Remove unreachable code

otaviobp:
When SOL_FEATURE_FILESYSTEM is set, we don't need to use the default
app name, so move this code to #else section.


ceolin:
:+1:

PR #2178, opened 2016-06-15
WIP: lwm2m: adding the Bootstrap Interface

bsmelo:
This patch introduces:

- Bootstrap Interface Operations
 - Request (client -> bs-server)
 - Delete (bs-server -> client)
 - Write (bs-server -> client)
 - Finish (bs-server -> client)
- A LWM2M Bootstrap Server sample app
- Modifications in the LWM2M Client app to receive an argument indicating if it should behave as a factory-bootstrapped device or not

There is a _"cry for help comment"_ in the last commit of this branch, for which I would like some help, mainly to solve the `TODO`s currently in the code - as well as some issues pointed out by `valgrind`, described in the commit comment itself.

_**Edit:** Oh wow, I didn't know the comment from the commit would appear here. Well, I'll just leave it there. Let me know if it's best to get rid of it and send by e-mail or something._

### DIFF
--- a/src/lib/comms/sol-lwm2m.c
+++ b/src/lib/comms/sol-lwm2m.c
@@ -2656,7 +2656,7 @@ handle_write(struct sol_lwm2m_client *client,
 {
     int r;
 
-    //If write_resource is not NULL then write_tlv is guaramteed to ve valid as well.
+    //If write_resource is not NULL then write_tlv is guaranteed to be valid as well.
     if (!obj_ctx->obj->write_resource) {
         SOL_WRN("Object %" PRIu16 " does not support the write method",
             obj_ctx->obj->id);
--- a/src/lib/comms/sol-lwm2m.c
+++ b/src/lib/comms/sol-lwm2m.c
@@ -4733,11 +4733,6 @@ exit:
     sol_coap_header_set_code(resp, header_code);
     r = sol_coap_send_packet(server, resp, cliaddr);
 
-    if (header_code == SOL_COAP_RESPONSE_CODE_DELETED) {
-        obj_instance_clear(client, obj_ctx, obj_instance);
-        (void)sol_vector_del_element(&obj_ctx->instances, obj_instance);
-    }
-
     return r;
 }
 
--- a/src/lib/comms/sol-lwm2m.c
+++ b/src/lib/comms/sol-lwm2m.c
@@ -226,6 +226,20 @@ struct management_ctx {
     const void *data;
 };
 
+enum bootstrap_type {
+    BOOTSTRAP_DELETE,
+    BOOTSTRAP_WRITE
+};
+
+struct bootstrap_ctx {
+    enum bootstrap_type type;
+    struct sol_lwm2m_bootstrap_server *server;
+    struct sol_lwm2m_bootstrap_client_info *cinfo;
+    char *path;
+    void *cb;
+    const void *data;
+};
+
 struct resource_ctx {
     char *str_id;
     struct sol_coap_resource *res;
@@ -286,6 +300,8 @@ static int bootstrap_with_server(struct sol_lwm2m_client *client,
 static int handle_resource(void *data, struct sol_coap_server *server,
     const struct sol_coap_resource *resource, struct sol_coap_packet *req,
     const struct sol_network_link_addr *cliaddr);
+static struct resource_ctx *find_resource_ctx_by_id(struct obj_instance *instance,
+    uint16_t id);
 
 static void
 send_ack_if_needed(struct sol_coap_server *coap, struct sol_coap_packet *msg,
@@ -1786,6 +1802,35 @@ exit:
 }
 
 static int
+instances_to_tlv(struct sol_lwm2m_resource **instances,
+    size_t *instances_len, uint16_t *instances_ids, size_t len, struct sol_buffer *tlvs)
+{
+    int r;
+    size_t i, res_id, instance_data_len, data_len;
+
+    for (i = 0; i < len; i++) {
+        for (res_id = 0, instance_data_len = 0; res_id < instances_len[i]; res_id++) {
+            r = get_resource_len(&instances[i][res_id], res_id, &data_len);
+            SOL_INT_CHECK(r, < 0, r);
+
+            instance_data_len += data_len;
+        }
+
+        r = setup_tlv_header(SOL_LWM2M_TLV_TYPE_OBJECT_INSTANCE,
+            instances_ids[i], tlvs, instance_data_len);
+        SOL_INT_CHECK(r, < 0, r);
+
+        r = resources_to_tlv(instances[i], instances_len[i], tlvs);
+        SOL_INT_CHECK_GOTO(r, < 0, exit);
+    }
+
+    return 0;
+
+exit:
+    return r;
+}
+
+static int
 add_coap_int_option(struct sol_coap_packet *pkt,
     enum sol_coap_option opt, const void *data, uint16_t len)
 {
@@ -1817,7 +1862,8 @@ static int
 setup_coap_packet(enum sol_coap_method method,
     enum sol_coap_message_type type, const char *objects_path, const char *path,
     uint8_t *obs, int64_t *token, struct sol_lwm2m_resource *resources,
-    size_t len,
+    struct sol_lwm2m_resource **instances, size_t *instances_len,
+    uint16_t *instances_ids, size_t len,
     const char *execute_args,
     struct sol_coap_packet **pkt)
 {
@@ -1869,8 +1915,10 @@ setup_coap_packet(enum sol_coap_method method,
     r = sol_buffer_append_slice(&buf, sol_str_slice_from_str(path));
     SOL_INT_CHECK_GOTO(r, < 0, exit);
 
-    r = sol_coap_packet_add_uri_path_option(*pkt, buf.data);
-    SOL_INT_CHECK_GOTO(r, < 0, exit);
+    if (strcmp(path, "/") != 0) {
+        r = sol_coap_packet_add_uri_path_option(*pkt, buf.data);
+        SOL_INT_CHECK_GOTO(r, < 0, exit);
+    }
 
     if (execute_args) {
         size_t str_len;
@@ -1883,6 +1931,14 @@ setup_coap_packet(enum sol_coap_method method,
     } else if (resources) {
         content_type = SOL_LWM2M_CONTENT_TYPE_TLV;
         r = resources_to_tlv(resources, len, &tlvs);
+        SOL_INT_CHECK_GOTO(r, < 0, exit);
+        r = -ENOMEM;
+        SOL_INT_CHECK_GOTO(tlvs.used, >= UINT16_MAX, exit);
+        content_data = tlvs.data;
+        content_len = tlvs.used;
+    } else if (instances) {
+        content_type = SOL_LWM2M_CONTENT_TYPE_TLV;
+        r = instances_to_tlv(instances, instances_len, instances_ids, len, &tlvs);
         SOL_INT_CHECK_GOTO(r, < 0, exit);
         r = -ENOMEM;
         SOL_INT_CHECK_GOTO(tlvs.used, >= UINT16_MAX, exit);
@@ -2010,7 +2066,7 @@ sol_lwm2m_server_add_observer(struct sol_lwm2m_server *server,
         return 0;
 
     r = setup_coap_packet(SOL_COAP_METHOD_GET, SOL_COAP_MESSAGE_TYPE_CON,
-        client->objects_path, path, &obs, &entry->token, NULL, 0, NULL, &pkt);
+        client->objects_path, path, &obs, &entry->token, NULL, NULL, NULL, NULL, 0, NULL, &pkt);
     SOL_INT_CHECK(r, < 0, r);
 
     return sol_coap_send_packet_with_reply(server->coap, pkt, &client->cliaddr,
@@ -2187,8 +2243,8 @@ send_management_packet(struct sol_lwm2m_server *server,
     struct management_ctx *ctx;
 
     r = setup_coap_packet(method, SOL_COAP_MESSAGE_TYPE_CON,
-        client->objects_path, path, NULL, NULL, resources, len,
-        execute_args, &pkt);
+        client->objects_path, path, NULL, NULL, resources, NULL, NULL, NULL,
+        len, execute_args, &pkt);
     SOL_INT_CHECK(r, < 0, r);
 
     if (!cb)
@@ -2592,7 +2648,7 @@ sol_lwm2m_resource_clear(struct sol_lwm2m_resource *resource)
 
 static int
 extract_path(struct sol_lwm2m_client *client, struct sol_coap_packet *req,
-    uint16_t *path_id, uint16_t *path_size)
+    uint16_t *path_id, uint16_t *path_size, bool allow_empty_path)
 {
     struct sol_str_slice path[16] = { };
     int i, j, r, count;
@@ -2601,8 +2657,10 @@ extract_path(struct sol_lwm2m_client *client, struct sol_coap_packet *req,
         sol_util_array_size(path));
     count = r;
     SOL_INT_CHECK_GOTO(r, < 0, err_exit);
-    r = -ENOENT;
-    SOL_INT_CHECK_GOTO(count, == 0, err_exit);
+    if (!allow_empty_path) {
+        r = -ENOENT;
+        SOL_INT_CHECK_GOTO(count, == 0, err_exit);
+    }
 
     for (i = client->splitted_path_len ? client->splitted_path_len : 0, j = 0;
         i < count; i++, j++) {
@@ -3338,7 +3396,7 @@ handle_resource(void *data, struct sol_coap_server *server,
     if (r < 0)
         content_format = SOL_LWM2M_CONTENT_TYPE_TEXT;
 
-    r = extract_path(client, req, path, &path_size);
+    r = extract_path(client, req, path, &path_size, false);
     header_code = SOL_COAP_RESPONSE_CODE_BAD_REQUEST;
     SOL_INT_CHECK_GOTO(r, < 0, exit);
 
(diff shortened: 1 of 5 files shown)


src/lib/comms/sol-lwm2m.c
```diff
@@ -629,6 +689,55 @@ fill_client_info(struct sol_lwm2m_client_info *cinfo,
...
+    for (i = 0; i < count; i++) {
+        struct sol_str_slice key, value;
+        const char *sep;
+
+        SOL_DBG("Query:%.*s", SOL_STR_SLICE_PRINT(query[i]));
+        sep = memchr(query[i].data, '=', query[i].len);
+
+        if (!sep) {
+            SOL_WRN("Could not find the separator '=' at: %.*s",
+                SOL_STR_SLICE_PRINT(query[i]));
```
cabelitos: set the `r` value here. Maybe `-EINVAL`

src/lib/comms/sol-lwm2m.c
```diff
@@ -629,6 +689,55 @@ fill_client_info(struct sol_lwm2m_client_info *cinfo,
...
+        key.data = query[i].data;
+        key.len = sep - query[i].data;
+        value.data = sep + 1;
+        value.len = query[i].len - key.len - 1;
+
+        if (sol_str_slice_str_eq(key, "ep")) {
+            //Required info
+            has_name = true;
+            bs_cinfo->name = sol_str_slice_to_str(value);
+            SOL_NULL_CHECK_GOTO(bs_cinfo->name, err_bs_cinfo_prop);
```
cabelitos: You can put a `break` after this NULL check

src/lib/comms/sol-lwm2m.c
```diff
@@ -629,6 +689,55 @@ fill_client_info(struct sol_lwm2m_client_info *cinfo,
...
+        }
```
cabelitos: btw this code above is wrong, because we're not setting the `r` value in case of a failure.
It should be
```c
bs_cinfo->name = sol_str_slice_to_str(value);
r = -ENOMEM;
SOL_NULL_CHECK_GOTO(bs_cinfo->name, err_bs_cinfo_prop);
....
```

src/lib/comms/sol-lwm2m.c
```diff
@@ -930,6 +1050,75 @@ static const struct sol_coap_resource registration_interface = {
...
+    SOL_INT_CHECK_GOTO(r, < 0, err_exit_del_client);
+
+    // TODO: Should I keep clients like this for some reason?
+/*
+    r = sol_ptr_vector_append(&server->clients, cinfo);
+    SOL_INT_CHECK_GOTO(r, < 0, err_exit_del_client);
+ */
+
+    for (i = 0; server->known_clients[i]; i++) {
+        if (strcmp(server->known_clients[i], bs_cinfo->name) == 0)
```
cabelitos: Be careful. In `sol_lwm2m_bootstrap_server_new()` you're not checking if `known_clients` is `NULL`.
There's also a possibility that `server->known_clients[i]` is `NULL`

cabelitos: Btw I would use `size_t` for the `i` variable, because this is an array.

src/lib/comms/sol-lwm2m.c
```diff
@@ -1127,6 +1316,124 @@ sol_lwm2m_server_del_registration_monitor(struct sol_lwm2m_server *server,
...
+sol_lwm2m_bootstrap_server_new(uint16_t port, const char **known_clients)
+{
+    struct sol_lwm2m_bootstrap_server *server;
+    struct sol_network_link_addr servaddr = { .family = SOL_NETWORK_FAMILY_INET6,
+                                              .port = port };
+    int r;
+
+    SOL_LOG_INTERNAL_INIT_ONCE;
+
+    server = calloc(1, sizeof(struct sol_lwm2m_bootstrap_server));
```
cabelitos: Check if `known_clients` is not `NULL`

src/lib/comms/sol-lwm2m.c
```diff
@@ -930,6 +1050,75 @@ static const struct sol_coap_resource registration_interface = {
...
+    r = sol_ptr_vector_append(&server->clients, cinfo);
+    SOL_INT_CHECK_GOTO(r, < 0, err_exit_del_client);
+ */
+
+    for (i = 0; server->known_clients[i]; i++) {
+        if (strcmp(server->known_clients[i], bs_cinfo->name) == 0)
+            know_client = true;
+    }
+
+    if (!know_client) {
```
cabelitos: I have a better idea.

As you can reject a bootstrap request, because it's not a known client. I recommend you to refactor the code and allocate memory for the `struct sol_lwm2m_bootstrap_client_info` only when we know that it's a known client.

This is show you should proceed:

* After `sol_coap_packet_new()`, you should call `fill_bootstrap_client_info()`. However you should change the function signature. 
** `fill_bootstrap_client_info()` should be renamed `extract_bootstrap_client_info(struct sol_coap_packet *req, struct sol_str_slice *client_name)` (note the new arguments). Note that the client_name will be filled by `extract_bootstrap_client_info()`
* After the `client_name` is filled, check if it's a known client.
* Allocate the bootscrap_client info...

src/lib/comms/sol-lwm2m.c
```diff
@@ -1127,6 +1316,124 @@ sol_lwm2m_server_del_registration_monitor(struct sol_lwm2m_server *server,
...
+    SOL_NULL_CHECK(client, -EINVAL);
+
+    m = sol_monitors_append(&client->bootstrap,
+        (sol_monitors_cb_t)cb, data);
+    SOL_NULL_CHECK(m, -ENOMEM);
+    return 0;
+}
+
+SOL_API int
+sol_lwm2m_client_del_bootstrap_finish_monitor(struct sol_lwm2m_client *client,
```
cabelitos: the functions `sol_lwm2m_client_del_bootstrap_finish_monitor()`, `sol_lwm2m_client_add_bootstrap_finish_monitor()`, `sol_lwm2m_bootstrap_server_del_request_monitor()` and 
`sol_lwm2m_bootstrap_server_add_request_monitor()` are basically the same thing.

You should do something like:
```C
static int
add_to_monitors(struct sol_monitors *monitors, sol_monitors_cb_t cb, const void *data)
{
   struct sol_monitors_entry *m;
   SOL_NULL_CHECK(cb, -EINVAL);

   m = sol_monitors_append(monitors, cb, data);
   SOL_NULL_CHECK(m, -ENOMEM);
   return 0;
}
static int
remove_from_monitors(struct sol_monitors *monitors, sol_monitors_cb_t cb, const void *data)
{
   int i;
   SOL_NULL_CHECK(cb, -EINVAL);
   i = sol_monitors_find(monitors, cb, data);
   SOL_INT_CHECK(i, < 0, i);
   return sol_monitors_del(monitors, i);
}

....
sol_lwm2m_client_del_bootstrap_finish_monitor(...)
{
   ...
   return remove_from_monitors(&client->bootstrap, cb, data);
}

//Do the same thing for ADD
//If possible try to use these new functions to the old monitors add/del that are already present in the LWM2M code
```

src/lib/comms/sol-lwm2m.c
```diff
@@ -930,6 +1050,75 @@ static const struct sol_coap_resource registration_interface = {
...
+    r = new_bootstrap_client_info(&bs_cinfo, cliaddr);
+    SOL_INT_CHECK_GOTO(r, < 0, err_exit);
+
+    r = fill_bootstrap_client_info(bs_cinfo, req);
+    SOL_INT_CHECK_GOTO(r, < 0, err_exit_del_client);
+
+    // TODO: Should I keep clients like this for some reason?
+/*
+    r = sol_ptr_vector_append(&server->clients, cinfo);
+    SOL_INT_CHECK_GOTO(r, < 0, err_exit_del_client);
```
cabelitos: Yes you should keep a list of clients, otherwise if one calls `sol_lwm2m_bootstrap_server_del()` you will leak all the clients!

src/samples/coap/lwm2m-client.c
```diff
@@ -17,12 +17,22 @@
  */
 
 /*
-   To run: ./lwm2m-sample-client [client name]
+   To run: ./lwm2m-sample-client <client name> [bootstrap]
```
cabelitos: You are mixing things. This changes should not be in this commit.

src/lib/comms/sol-lwm2m.c
```diff
@@ -3778,25 +4248,59 @@ sol_lwm2m_client_start(struct sol_lwm2m_client *client)
...
+        } else {
+            sol_lwm2m_resource_clear(&res[0]);
+            bootstrap_account_idx = i;
         }
-        clear_resource_array(res, sol_util_array_size(res));
     }
 
+    //If all attempts to register failed, try to bootstrap
     if (!has_server) {
         SOL_WRN("The client did not specify a LWM2M server to connect");
```
cabelitos: Change this to DBG and add a message telling that we'll try to bootstrap

src/lib/comms/sol-lwm2m.c
```diff
@@ -3778,25 +4248,59 @@ sol_lwm2m_client_start(struct sol_lwm2m_client *client)
...
+                res[1].data[0].integer);
             r = -ENOMEM;
-            SOL_NULL_CHECK_GOTO(conn_ctx, err_clear);
+            SOL_NULL_CHECK_GOTO(conn_ctx, err_clear_2);
             has_server = true;
+
+            clear_resource_array(res, 2);
+        } else {
+            sol_lwm2m_resource_clear(&res[0]);
+            bootstrap_account_idx = i;
```
cabelitos: What happens if you have more than one bootstrap server?

src/lib/comms/sol-lwm2m.c
```diff
@@ -3778,25 +4248,59 @@ sol_lwm2m_client_start(struct sol_lwm2m_client *client)
...
+
+        //Create '/bs' CoAP resource to receive Bootstrap Finish notification
+        r = sol_coap_server_register_resource(client->coap_server,
+            &bootstrap_finish_interface, client);
+        SOL_INT_CHECK_GOTO(r, < 0, err_clear_3);
+
+        SOL_INF("Expecting server-initiated Bootstrap for"
+            " %" PRId64 " seconds", res[1].data[0].integer);
+
+        // TODO: block and expect server-initiated bootstrap before client-initiated bootstrap
```
cabelitos: No no, we should never block

src/lib/comms/sol-lwm2m.c
```diff
@@ -3759,9 +4158,80 @@ lifetime_client_timeout(void *data)
...
+    pkt = sol_coap_packet_new_request(SOL_COAP_METHOD_POST, SOL_COAP_MESSAGE_TYPE_CON);
+    r = -ENOMEM;
+    SOL_NULL_CHECK_GOTO(pkt, err_exit);
+
+    r = sol_coap_add_option(pkt, SOL_COAP_OPTION_URI_PATH, "bs", strlen("bs"));
+    SOL_INT_CHECK_GOTO(r, < 0, err_coap);
+
+    SOL_DBG("Sending Bootstrap Finish to LWM2M Client %s", client->name);
+    r = sol_coap_send_packet(server->coap, pkt, &client->cliaddr);
+    return r;
```
cabelitos: You're leaking the client info. You should delete it.
We should also mention in the docs that after `sol_lwm2m_bootstrap_server_send_finish()` returns the client_info handle will be INVALID!

src/lib/comms/sol-lwm2m.c
```diff
@@ -3237,32 +3562,33 @@ obj_ctx_clear(struct sol_lwm2m_client *client, struct obj_ctx *ctx)
...
+server_connection_ctx_clear(struct server_conn_ctx *conn_ctx, bool is_bootstrap)
 {
     if (conn_ctx->pending_pkt)
         sol_coap_packet_unref(conn_ctx->pending_pkt);
     if (conn_ctx->hostname_handle)
         sol_network_hostname_pending_cancel(conn_ctx->hostname_handle);
     sol_vector_clear(&conn_ctx->server_addr_list);
-    free(conn_ctx->location);
+    if (!is_bootstrap)
+        free(conn_ctx->location);
```
cabelitos: Why you did this?
calling `free(NULL)` is totally safe

src/lib/comms/sol-lwm2m.c
```diff
@@ -3658,8 +3978,81 @@ register_with_server(struct sol_lwm2m_client *client,
...
+    }
+    if (!sol_network_link_addr_to_str(server_addr, &addr))
+        SOL_WRN("Could not convert the server address to string");
+
+    sol_coap_header_get_code(pkt, &code);
+    SOL_INT_CHECK_GOTO(code, != SOL_COAP_RESPONSE_CODE_CHANGED, err_exit);
+
+    SOL_DBG("Bootstrap process with server %.*s can start",
+        SOL_STR_SLICE_PRINT(sol_buffer_get_slice(&addr)));
+
```
cabelitos: Save some memory and call sol_vector_clear(conn_ctx->server_addr_list)

src/lib/comms/sol-lwm2m.c
bsmelo: Ok.

If it's NULL though, I should just terminate the execution, since there's no job to be performed by this Bootstrap Server, right? If this is the case, I think it's better if I check it before allocating memory.

src/lib/comms/sol-lwm2m.c
bsmelo: Changed the type of `i`.

Regarding the first comment, it should be enough to check if `known_clients` is `NULL` in `sol_lwm2m_bootstrap_server_new()`, right? `known_clients` is supposed to be `NULL`-terminated itself.

src/lib/comms/sol-lwm2m.c
bsmelo: Wow, I really believe I was getting some compiler warning/error... But I just tried removing it and all is well. oO

Anyway, fixed. Thanks!

src/samples/coap/lwm2m-client.c
```diff
@@ -376,30 +444,45 @@ main(int argc, char *argv[])
     struct sol_lwm2m_client *client;
     static const struct sol_lwm2m_object *objects[] =
     { &security_object, &server_object, &location_object, NULL };
-    bool has_location_instance = false;
+    struct client_data_ctx data_ctx = {
+        .is_bootstrap = false,
+        .has_location_instance = false
+    };
```
cabelitos: You can do `struct client_data_ctx data_ctx = { 0 };`. It's much cleaner

src/lib/comms/sol-lwm2m.c
bsmelo: You shouldn't:

_"The LWM2M Client MUST have at most one LWM2M Bootstrap Server Account."_ [Spec, Pg. 19]

Basically, if you have more than one, you are already breaking the spec. The code will be nice and go on with the first one, though. :)

src/samples/coap/lwm2m-client.c
```diff
@@ -376,30 +444,45 @@ main(int argc, char *argv[])
...
-        fprintf(stderr, "Could not add a server object instance\n");
-        goto exit_del;
+    if (!data_ctx.is_bootstrap) {
+        r = sol_lwm2m_add_object_instance(client, &server_object, NULL);
+        if (r < 0) {
+            fprintf(stderr, "Could not add a server object instance\n");
+            goto exit_del;
+        }
+    } else {
+        r = sol_lwm2m_client_add_bootstrap_finish_monitor(client, bootstrap_cb,
```
cabelitos: call it `bootstrap_finish_cb`

src/samples/coap/lwm2m-client.c
```diff
@@ -337,15 +389,31 @@ del_location_obj(void *instance_data, void *user_data,
...
+    if (event == SOL_LWM2M_BOOTSTRAP_EVENT_FINISHED) {
+        printf("...<Call local Bootstrap clean-up operations>...\n"
+            "...<Now that it should have a Server, try to register again>\n");
+        //sol_lwm2m_client_start(client);
+        return;
+    } else if (event == SOL_LWM2M_BOOTSTRAP_EVENT_ERROR) {
+        fprintf(stderr, "Bootstrap Request or Bootstrap Finish Failed!\n");
+        return;
+    }
+}
```
cabelitos: Remove the `return` from both if branchs

src/samples/coap/lwm2m-client.c
```diff
@@ -337,15 +389,31 @@ del_location_obj(void *instance_data, void *user_data,
...
 
+static void
+bootstrap_cb(void *data,
+    struct sol_lwm2m_client *client,
+    enum sol_lwm2m_bootstrap_event event)
+{
+    if (event == SOL_LWM2M_BOOTSTRAP_EVENT_FINISHED) {
+        printf("...<Call local Bootstrap clean-up operations>...\n"
+            "...<Now that it should have a Server, try to register again>\n");
+        //sol_lwm2m_client_start(client);
```
cabelitos: If is this commented out?

src/lib/comms/sol-lwm2m.c
bsmelo: So, how did you figure this out?  hahahah

Indeed, this call solved the leak I had on the `bs-server` app -- together with a change on the app itself, introducing the correct "callback chain" so send_finish is really sent after every write/delete has finished.

Thanks!

src/lib/comms/sol-lwm2m.c
bsmelo: Done.

src/lib/comms/sol-lwm2m.c
bsmelo: Changed to
```C
            bs_cinfo->name = sol_str_slice_to_str(value);
            r = -ENOMEM;
            SOL_NULL_CHECK_GOTO(bs_cinfo->name, err_bs_cinfo_prop);
            r = 0;
            break;
```

src/lib/comms/sol-lwm2m.c
bsmelo: Done right before the `calloc` with `SOL_NULL_CHECK`.

src/lib/comms/sol-lwm2m.c
bsmelo: Done for functions:
```C
observer_entry_add_monitor(); //Added SOL_NULL_CHECK(entry, -EINVAL); as well
observer_entry_del_monitor(); //Added SOL_NULL_CHECK(entry, -EINVAL); as well
sol_lwm2m_server_add_registration_monitor();
sol_lwm2m_server_del_registration_monitor();
sol_lwm2m_bootstrap_server_add_request_monitor();
sol_lwm2m_bootstrap_server_del_request_monitor();
sol_lwm2m_client_add_bootstrap_finish_monitor();
sol_lwm2m_client_del_bootstrap_finish_monitor();
```

src/lib/comms/sol-lwm2m.c
bsmelo: Done.

I thought I would be able to do the same inside `register_reply()`, but I got a few errors when trying it. :/

src/lib/comms/sol-lwm2m.c
bsmelo: Done.

src/samples/coap/lwm2m-client.c
bsmelo: Hm, I didn't know that. Done.

src/samples/coap/lwm2m-client.c
bsmelo: Done.

src/lib/comms/sol-lwm2m.c
cabelitos: oops

src/lib/comms/sol-lwm2m.c
```diff
@@ -4184,6 +4363,384 @@ sol_lwm2m_bootstrap_server_send_finish(struct sol_lwm2m_bootstrap_server *server
...
+        return SOL_COAP_RESPONSE_CODE_UNSUPPORTED_CONTENT_FORMAT;
+    }
+
+err_exit:
+    obj_instance_clear(client, obj_ctx, obj_instance);
+    return SOL_COAP_RESPONSE_CODE_BAD_REQUEST;
+}
+
+static int
+handle_unknown_bootstrap_resource(void *data, struct sol_coap_server *server,
```
cabelitos: Why did you create this function? Why not use `handle_resource` ?
This is pratically copy & pasta. You must adapt the `handle_resource` function in order to support the bootstrap feature.

cabelitos: The same thing applies to the `handle_bootstrap_write_object` and friends.

src/lib/comms/sol-lwm2m.c
```diff
@@ -4288,6 +4845,11 @@ sol_lwm2m_client_start(struct sol_lwm2m_client *client)
             &bootstrap_finish_interface, client);
         SOL_INT_CHECK_GOTO(r, < 0, err_clear_3);
 
+        //Create unknown CoAP resource to handle Bootstrap Write and Bootstrap Delete
+        r = sol_coap_server_set_unknown_resource_handler(client->coap_server,
```
cabelitos: You know unregister this when you receive a bootstrap finish

src/samples/coap/lwm2m-bs-server.c
```diff
@@ -114,6 +114,9 @@ bootstrap_cb(void *data,
     sol_lwm2m_bootstrap_server_write(server, bs_cinfo, "/0/1",
         sec_server_one, sol_util_array_size(sec_server_one), write_cb, NULL);
 
+    sol_lwm2m_bootstrap_server_write(server, bs_cinfo, "/0/1/10",
```
cabelitos: This should be in the bootstrap server commit, not here

src/lib/comms/sol-lwm2m.c
```diff
@@ -4733,11 +4733,6 @@ handle_unknown_bootstrap_resource(void *data, struct sol_coap_server *server,
     sol_coap_header_set_code(resp, header_code);
     r = sol_coap_send_packet(server, resp, cliaddr);
 
-    if (header_code == SOL_COAP_RESPONSE_CODE_DELETED) {
-        obj_instance_clear(client, obj_ctx, obj_instance);
```
cabelitos: This should not be in here

src/samples/coap/lwm2m-client.c
```diff
@@ -364,7 +350,7 @@ write_security_server_res(void *instance_data, void *user_data,
     switch (res->id) {
     case SECURITY_SERVER_SERVER_URI_RES_ID:
         // TODO: Should I check if exists and free() first?
```
cabelitos: what do you mean?

src/samples/coap/lwm2m-bs-server.c
```diff
@@ -0,0 +1,155 @@
...
+main(int argc, char *argv[])
+{
+    struct sol_lwm2m_bootstrap_server *server;
+    uint16_t port = 5783;
+    int r;
+
+    printf("Using port %" PRIu16 "\n", port);
+    sol_init();
+
+    server = sol_lwm2m_bootstrap_server_new(port, known_clients);
```
cabelitos: What do you think if we add an API to add and remove known clients on the fly?

src/samples/coap/lwm2m-bs-server.c
```diff
@@ -0,0 +1,155 @@
...
+        sec_server_one, sol_util_array_size(sec_server_one), write_cb, NULL);
+
+    for (i = 0; i < sol_util_array_size(sec_server_one); i++)
+        sol_lwm2m_resource_clear(&sec_server_one[i]);
+
+    sol_lwm2m_bootstrap_server_send_finish(server, bs_cinfo);
+}
+
+int
+main(int argc, char *argv[])
```
cabelitos: Try to use the `SOL_MAIN_DEFAULT()` macro, this will avoid the main function

src/lib/comms/sol-lwm2m.c
cabelitos: Awesome.
Well, when I saw that you were not controlling the client life cycle it was clear that it was leaking

cabelitos:
@bsmelo The second valgrind error is happening because your function `handle_bootstrap_delete` only calls the del callback and it DOES NOT REMOVE THE INSTANCE FROM THE LIST!
You should remove it from the list after it's deleted.

src/samples/coap/lwm2m-client.c
bsmelo: It's not working yet, I need to debug further and find out why. I'm probably messing something related to memory management and losing information... also probably related to your comment about the second `valgrind` error.

src/samples/coap/lwm2m-client.c
bsmelo: I mean I'm a little lost with the memory management. :-(
I'm trying to solve this out, but I believe it will be more clear to me after we talk on monday.

src/lib/comms/sol-lwm2m.c
bsmelo: Two things here:

1. I believe I got confused a bit with the commits, as we already talked before. I will be more careful next time. Anyway, if you look at the changes from "all commits", you'll see I've actually removed this from here.
2. I believe, as per your comment down below regarding not removing the instance from the list, I need to put it somewhere else. I'll try to figure it out.

src/samples/coap/lwm2m-bs-server.c
bsmelo: Hmm, I'm trying to think of a precise use-case for this, but it seems like a very good idea. I will put a `TODO` aside here.

src/samples/coap/lwm2m-bs-server.c
bsmelo: I'll look into it. For now, I've just used something similar to the other samples.

src/lib/comms/sol-lwm2m.c
```diff
@@ -4184,6 +4363,384 @@ sol_lwm2m_bootstrap_server_send_finish(struct sol_lwm2m_bootstrap_server *server
...
+    if (resource_ctx) {
+        if (content_format == SOL_LWM2M_CONTENT_TYPE_TEXT ||
+            content_format == SOL_LWM2M_CONTENT_TYPE_OPAQUE) {
+            struct sol_lwm2m_resource res;
+            struct sol_blob *blob;
+
+            blob = sol_blob_new(&SOL_BLOB_TYPE_NO_FREE_DATA, NULL, payload.data, payload.len);
+            SOL_NULL_CHECK(blob, SOL_COAP_RESPONSE_CODE_BAD_REQUEST);
+
+            // TODO: what if I want to write INT, FLOAT, BOOLEAN?
```
cabelitos: then the user should use TLV

cabelitos: Then the message should be in TLV format

bsmelo: Right.

Section 6.3 from the spec was updated. It used to read _"The LWM2M Server MUST support all data formats. **The plain text and opaque formats MUST be supported by the LWM2M Client for singular Resources.** The LWM2M Client MUST support the TLV data format for Object Instance and multiple-instance Resource requests."_
Now it says _"The LWM2M Server MUST support all data formats. **The LWM2M Client MUST support the TLV data format for all requests.** In addition a LWM2M Client MAY choose to support other media type formats in the table below i.e., JSON, plain text and opaque."_

Anyway, I'll take note. Thanks!

src/lib/comms/sol-lwm2m.c
cabelitos: ok

src/samples/coap/lwm2m-client.c
```diff
@@ -297,63 +297,49 @@ read_location_obj(void *instance_data, void *user_data,
...
-        default:
-            if (res_id >= 2 && res_id <= 10)
-                r = -ENOENT;
-            else
-                r = -EINVAL;
-        }
+    switch (res_id) {
+    case SECURITY_SERVER_SERVER_URI_RES_ID:
+        blob = sol_blob_new(&SOL_BLOB_TYPE_DEFAULT, NULL,
+            ctx->server_uri, strlen(ctx->server_uri));
```
cabelitos: this can segfaults, if server_uri is NULL you should return -ENOENT

src/samples/coap/lwm2m-client.c
```diff
@@ -623,22 +633,47 @@ main(int argc, char *argv[])
...
+
         if (r < 0) {
             fprintf(stderr, "Could not add a server object instance\n");
             goto exit_del;
         }
+/*
+        security_data->server_uri = (char*)malloc((strlen("coap://localhost:5683") + 1) * sizeof(char));
+        strcpy(security_data->server_uri, "coap://localhost:5683");
+ */
+        security_data->server_uri = sol_str_slice_to_str(sol_str_slice_from_blob(&server_addr));
```
cabelitos: check for NULL

src/samples/coap/lwm2m-client.c
```diff
@@ -623,22 +633,47 @@ main(int argc, char *argv[])
...
+
         if (r < 0) {
             fprintf(stderr, "Could not add a bootstrap monitor\n");
             goto exit_del;
         }
+/*
+        security_data->server_uri = (char*)malloc((strlen("coap://localhost:5783") + 1) * sizeof(char));
+        strcpy(security_data->server_uri, "coap://localhost:5783");
+ */
+        security_data->server_uri = sol_str_slice_to_str(sol_str_slice_from_blob(&bootstrap_server_addr));
```
cabelitos: check for NULL

bsmelo: ditto. The `bootstrap_server_addr` is statically defined in this same file and should never be NULL.

src/samples/coap/lwm2m-client.c
bsmelo: `server_addr` is statically defined at the beginning of this file.

Anyway, I went with the changes to make server_uri a `sol blob *` and these things work well and are more clear now -- currently there are no leaks.

I'm currently changing the Server Object of this client to be "dynamic" (instead of hard-coded in the `read_server_obj` function) as well, and to have write and del functions, to retest the second call to `sol_lwm2m_client_start`.

bsmelo:
An updated version is at #2183. Closing this. Thanks.

PR #2184, opened 2016-06-20
generator: Looking for description files on install dir and env var

edersondisouza:
Currently, to set description files, one has to use `-j` option. This
patch introduces two changes to this: now sol-fbp-generator looks on
Soletta install dir to find description files. In addition, it's
possible to use env var `SOL_FLOW_DESCRIPTION_PATHS` as well. This last
one can be composed of different colon (:) separated paths to files or
directories.

### DIFF
--- a/src/bin/sol-fbp-generator/main.c
+++ b/src/bin/sol-fbp-generator/main.c
@@ -1886,6 +1886,9 @@ print_usage(const char *program)
         "    -j  When resolving types, use the passed DESC files. If DESC is\n"
         "        a directory then all the .json files in the directory will be used.\n"
         "        Multiple -j can be passed.\n"
+        "        Note that Soletta install dir will also be looked, as well as any\n"
+        "        path on SOL_FLOW_DESCRIPTIONS_PATHS (where multiple paths are\n"
+        "        separated by ':').\n"
         "    -s  Define a function named SYMBOL that will return the type from FBP\n"
         "        and don't generate any main function or entry point.\n"
         "    -I  Define search path for FBP files\n"
@@ -1897,7 +1900,6 @@ static bool
 parse_args(int argc, char *argv[])
 {
     char *filename, *dup_path;
-    bool has_json_file = false;
     int opt;
 
     if (argc < 3) {
@@ -1922,7 +1924,6 @@ parse_args(int argc, char *argv[])
             args.conf_file = optarg;
             break;
         case 'j':
-            has_json_file = true;
             if (!handle_json_path(optarg)) {
                 SOL_ERR("Can't access JSON description path '%s': %s",
                     optarg, sol_util_strerrora(errno));
@@ -1945,12 +1946,6 @@ parse_args(int argc, char *argv[])
     if (optind != argc - 2) {
         fprintf(stderr, "A single FBP input file and output file is required."
             " e.g. './sol-fbp-generator -j builtins.json simple.fbp simple-fbp.c'\n");
-        return false;
-    }
-
-    if (!has_json_file) {
-        fprintf(stderr, "At least one JSON file containing the declaration of the nodes"
-            " (module) used in the FBP is required.\n");
         return false;
     }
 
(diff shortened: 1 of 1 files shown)


edersondisouza:
@cmarcelo Is that what you had on mind?

src/bin/sol-fbp-generator/main.c
```diff
@@ -1886,6 +1886,9 @@ print_usage(const char *program)
         "    -j  When resolving types, use the passed DESC files. If DESC is\n"
         "        a directory then all the .json files in the directory will be used.\n"
         "        Multiple -j can be passed.\n"
+        "        Note that Soletta install dir will also be looked, as well as any\n"
```
cmarcelo: The problem with this (always looking at a predefined set of paths) is that if I don't want to look to Soletta install dir. For example, I'm using sol-fbp-generator in my host machine to generate flows targetting Soletta on a non-Linux system with a different set of nodes available.

src/bin/sol-fbp-generator/main.c
```diff
@@ -1886,6 +1886,9 @@ print_usage(const char *program)
...
+        "        path on SOL_FLOW_DESCRIPTIONS_PATHS (where multiple paths are\n"
```
cmarcelo: Maybe SOL_FLOW_DESCRIPTION_PATH instead of PATH (all the other environment variables I know use singular even when they are a list).

src/bin/sol-fbp-generator/main.c
```diff
@@ -1948,12 +1949,6 @@ parse_args(int argc, char *argv[])
         return false;
     }
 
-    if (!has_json_file) {
-        fprintf(stderr, "At least one JSON file containing the declaration of the nodes"
```
cmarcelo: Would be nice to still have a check for whether any JSON file was found (after taking into account all the paths), after parsing arguments and handling the environment variable.

src/bin/sol-fbp-generator/main.c
edersondisouza: On 21-06-2016 16:19, Caio Marcelo de Oliveira Filho wrote:
> In src/bin/sol-fbp-generator/main.c
> <https://github.com/solettaproject/soletta/pull/2184#discussion_r67934240>:
>
>> @@ -1886,6 +1886,9 @@ print_usage(const char *program)
>>          "    -j  When resolving types, use the passed DESC files. If DESC is\n"
>>          "        a directory then all the .json files in the directory will be used.\n"
>>          "        Multiple -j can be passed.\n"
>> +        "        Note that Soletta install dir will also be looked, as well as any\n"
>
> The problem with this (always looking at a predefined set of paths) is
> that if I don't want to look to Soletta install dir. For example, I'm
> using sol-fbp-generator in my host machine to generate flows targetting
> Soletta on a non-Linux system with a different set of nodes available.
>
> —
> Reply to this email directly, view it on GitHub
> <https://github.com/solettaproject/soletta/pull/2184/files/320bbd683fefb9d55375f839186ebecf0926e5dc#r67934240>,
> or mute the thread
> <https://github.com/notifications/unsubscribe/ADB8T6dv1tso0w-WAWqVBcz0bazliPksks5qODk_gaJpZM4I5-j9>.
>

Yes, but not looking and forcing user always know where to look appears 
bad IMHO. Maybe only looking if no -j or envvar is set?

barbieri: I'd go with:
 - no envvar, look into install dir. This allows simple usage for the common case.
 - envvar is set, do not look into install dir. Even if envvar is set to an invalid directory. This allows easy cross compiling, just set the envvar to the cross-compiled directory, or set to some invalid/empty directory.
 - `-j` adds to/extends the above.

cmarcelo: +1 to Gustavo suggestion. Just make sure that this is clear from the help ("if no env var is set, take the soletta installation").

edersondisouza:
v2 #2197

PR #2186, opened 2016-06-20
cheat-sheet: fix broken links that was not pointing to the nodelist

brunobottazzini:
After opening a nodetype and clicking in the nodetypes list, the navigation was
invalid.

In this patch, the back action is executed before selecting the nodetype in the
list thus the nodetype will be valid to be selected.

### DIFF
--- a/doc/node-types-html/index.html.in
+++ b/doc/node-types-html/index.html.in
@@ -38,6 +38,32 @@
             /* getting the DOM element */
             var el = $('#example');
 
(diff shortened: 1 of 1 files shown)


bdilly:
It doesn't seem to be working fine at the following situation:
  * load the cheat-sheet
  * click on the first node type description (accelerometer/adxl...)
  * press back
  * try another category on the left menu

It scroll to the selected category and scroll back to accelerometer

PR #2187, opened 2016-06-20
check-license: Fix check-license script

bdilly:
It was broken since trademark was added to headers.
Also add it to lots of missing files.

@bottazzini could you please take a look on this?


brunobottazzini:
@bdilly, still failing in CI but looking at the errors it looks like to be another thing.

+1 to this PR

bdilly:
merged, thanks

PR #2188, opened 2016-06-20
cheat-sheet: fix broken links that was not pointing to the nodelist - V2

brunobottazzini:
After opening a nodetype and clicking in the nodetypes list, the navigation was
invalid.

In this patch, the back action is executed before selecting the nodetype in the
list thus the nodetype will be valid to be selected.

Difference from V1:

- Fix Firefox compability as pointed by @bdilly (Thanks!)
- Add commit that warns when the selected node has no examples to be shown. Also, the Get Code Button will get hidden.

### DIFF
--- a/doc/node-types-html/css/main.css
+++ b/doc/node-types-html/css/main.css
@@ -183,6 +183,13 @@ body{
     background-color: rgba(255,255,255,0.05);
     overflow: hidden;
 }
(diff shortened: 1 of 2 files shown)


bdilly:
@brunobottazzini it seems ok now

brunobottazzini:
thanks! merged

PR #2189, opened 2016-06-20
build: use syntax that works fine on bash or dash

bdilly:
On make doc rules


bdilly:
and our build on semaphoreci is good again :smile:

brunobottazzini:
+1 :D

PR #2190, opened 2016-06-20
oic: Use an extra socket for unicast coap communications

otaviobp:
We are using SO_REUSEADDR flag in server sockets so we can have multiple
servers listening for multicast packets in the same interface and port.
But in this scenario, only the last started server will receive unicast
packets. In order to solve this we followed the same approach used by
IoTivitiy implementation. We are creating an extra socket to use for
unicast communication.


cabelitos:
lgtm

otaviobp:
merged

PR #2192, opened 2016-06-21
build: simplify versioning and update version to 1

bdilly:
Instead of adoptiong the semantic version for
Major.Minor.Release just increment version
for each release.


barbieri:
bumping the major of soname for every release will not be that good in practice, if we do another release with just additions and no breaks, we'd be backwards compatible but the soname would not work.

barbieri:
As @lucasdemarchi pointed on IRC: https://www.gnu.org/software/libtool/manual/html_node/Updating-version-info.html

bdilly:
closed since we have a new version #2222

PR #2193, opened 2016-06-22
Add Netctl node type and samples for network control

guchaojie:
Add netctl fbp node and samples include list services, connect or disconnect service,

get service state, turn on or off radios

Change-Id: I7f00a8ce849aed185e49e3b6ac54b66d30384ffd


guchaojie:
@barbieri @bdilly @vcgomes @wzhen12 please have a review.

src/modules/flow/netctl/netctl.c
```diff
@@ -0,0 +1,342 @@
...
+#include "sol-flow-internal.h"
+
+#include <sol-netctl.h>
+#include <sol-str-table.h>
+#include <sol-util-internal.h>
+#include <errno.h>
+
+struct list_service_data {
+    struct sol_ptr_vector service_list;
+    struct sol_flow_node *node;
```
barbieri: no need to keep a node here. Instead, when you need both the private data (list_service_data) and the node, pass the node itself as the `void *data` and then call `sol_flow_node_get_private_data()` (https://github.com/solettaproject/soletta/blob/master/src/lib/flow/include/sol-flow.h#L121)

src/modules/flow/netctl/netctl.c
```diff
@@ -0,0 +1,342 @@
...
+#include "sol-flow/netctl.h"
+#include "sol-flow-internal.h"
+
+#include <sol-netctl.h>
+#include <sol-str-table.h>
+#include <sol-util-internal.h>
+#include <errno.h>
+
+struct list_service_data {
+    struct sol_ptr_vector service_list;
```
barbieri: `list_service_data` and its `service_list` shouldn't be per node, since all nodes will have the exact same content. Actually you do not need to maintain a shadow/copy list, since it will be the same as `sol_netctl_get_services()` (https://github.com/solettaproject/soletta/blob/master/src/lib/comms/include/sol-netctl.h#L472).

Then from `network_service_cb` you simply emit flow packets for that node (you give it as `void *data`).

src/modules/flow/netctl/netctl.c
```diff
@@ -0,0 +1,342 @@
...
+
+    if (sol_netctl_service_get_state(service) ==
+        SOL_NETCTL_SERVICE_STATE_REMOVE) {
+        sol_ptr_vector_del_element(&mdata->service_list, service);
+        return;
+    }
+
+    name = sol_netctl_service_get_name(service);
+    r = sol_flow_send_string_packet(node,
+        SOL_FLOW_NODE_TYPE_NETCTL_LIST_SERVICES__OUT__OUT,
```
barbieri: rename this port to `ADDED`

src/modules/flow/netctl/netctl.c
```diff
@@ -0,0 +1,342 @@
...
+network_service_cb(void *data, const struct sol_netctl_service *service)
+{
+    struct list_service_data *mdata = data;
+    struct sol_flow_node *node = mdata->node;
+    int r;
+    const char *name;
+
+    if (sol_netctl_service_get_state(service) ==
+        SOL_NETCTL_SERVICE_STATE_REMOVE) {
+        sol_ptr_vector_del_element(&mdata->service_list, service);
```
barbieri: add a new port `REMOVED` and emit a packet here, like you did below.

src/modules/flow/netctl/netctl.c
```diff
@@ -0,0 +1,342 @@
...
+void
+error_cb(void *data, const struct sol_netctl_service *service,
+    unsigned int error)
+{
+    const char *name;
+
+    name = sol_netctl_service_get_name(service);
+
+    if (name)
+        SOL_ERR("Service %s error is %d", name, error);
```
barbieri: you should emit an error packet to let the flow know about this error. It may go, for instance, in a UI to inform the user.

src/modules/flow/netctl/netctl.c
```diff
@@ -0,0 +1,342 @@
...
+
+    sol_ptr_vector_init(&mdata->service_list);
+
+    if (sol_netctl_add_service_monitor(network_service_cb, mdata))
+        goto error;
+
+    return 0;
+
+error:
+    sol_flow_send_error_packet(node, EINVAL, "%s", errmsg);
```
barbieri: no need to report this error as a packet. The rule of thumb is:

 - internal errors, like unexpected list additions, lack of memory... that the user can't do anything about it, just `SOL_WRN()` and return the error. This is only useful to help development to solve the problem that shouldn't happen.

 - errors user (here as in user of our node types, the developer) can handle go as packets. Then the flow can link the error packet to some "temporarily disable logic", like stop scanning for services, start a 2s timer and try again. Some the user may present in a UI/dashboard and ask the user for an interaction.

src/modules/flow/netctl/netctl.c
```diff
@@ -0,0 +1,342 @@
...
+open_list_services(struct sol_flow_node *node, void *data, const struct sol_flow_node_options *options)
+{
+    static const char *errmsg = "Could not add service monitor";
+    struct list_service_data *mdata = data;
+
+    mdata->node = node;
+
+    sol_ptr_vector_init(&mdata->service_list);
+
+    if (sol_netctl_add_service_monitor(network_service_cb, mdata))
```
barbieri: for the error case, store the return of `sol_netctl_add_service_monitor()`, it's the error code, then return it instead of just using `-EINVAL`. It will help to propagate the originating error.

You may "translate" errors when needed, but here isn't the case.

src/modules/flow/netctl/netctl.c
```diff
@@ -0,0 +1,342 @@
...
+        [SOL_NETCTL_SERVICE_STATE_READY] = "Ready",
+        [SOL_NETCTL_SERVICE_STATE_ONLINE] = "Online",
+        [SOL_NETCTL_SERVICE_STATE_DISCONNECT] = "Disconnect",
+        [SOL_NETCTL_SERVICE_STATE_FAILURE] = "Failure",
+        [SOL_NETCTL_SERVICE_STATE_REMOVE] = "Remove",
+    };
+
+    service_list = sol_netctl_get_services();
+
+    while (service = sol_ptr_vector_get(service_list,i)) {
```
barbieri: that's not the best way to iterate the vector, use `SOL_PTR_VECTOR_FOREACH_IDX()` that is much nicer :-)

src/modules/flow/netctl/netctl.c
```diff
@@ -0,0 +1,342 @@
...
+        state = sol_netctl_service_get_state(service);
+        name = sol_netctl_service_get_name(service);
+
+        if (name && strcmp(name, mdata->service_name) == 0) {
```
barbieri: to reduce indentation you can do a first block to `continue`:

```c
   if (!name || strcmp(name, mdata->service_name))
      continue;

   if (state == SOL_NETCTL_SERVICE_STATE_IDLE && connect)
     ...
```

src/modules/flow/netctl/netctl.c
```diff
@@ -0,0 +1,342 @@
...
+            if (state == SOL_NETCTL_SERVICE_STATE_IDLE && connect) {
+                sol_netctl_service_connect(service);
+            } else if ((state == SOL_NETCTL_SERVICE_STATE_READY) && (!connect)) {
+                  sol_netctl_service_disconnect(service);
```
barbieri: wrong indentation, check the style with the script `tools/check-style.py`. You'll need to install a tool called uncrustify

src/modules/flow/netctl/netctl.c
```diff
@@ -0,0 +1,342 @@
...
+            }
+            state = sol_netctl_service_get_state(service);
+            r = sol_flow_send_string_packet(node,
+                    SOL_FLOW_NODE_TYPE_NETCTL_SERVICE__OUT__OUT,
+                    state_msgs[state]);
```
barbieri: These mappings to/from string should be in `sol-netctl.h`. I just notice they are missing, but you should add it there. The `to_str()` is what you have here, returns a `const char *` to a static array. The `from_str()` uses `sol_str_table()` to lookup. Please add both in a single commit.

See examples at:
 - https://github.com/solettaproject/soletta/blob/master/src/lib/io/include/sol-gpio.h#L251
 - https://github.com/solettaproject/soletta/blob/master/src/lib/io/sol-gpio-common.c#L82

src/modules/flow/netctl/netctl.c
```diff
@@ -0,0 +1,342 @@
...
+            if (r < 0)
+                SOL_WRN("Failed to send network service connect state");
+            return 0;
+        }
+        i++;
+    }
+
+    sol_flow_send_error_packet(node, ENOENT, "Did not found service name %s", mdata->service_name);
```
barbieri: this is a perfect example on when you should emit an error packet. You did it right, you emit it to let the user know the service doesn't exist, then he can wait some time, or use `list-services` to try again later.

src/modules/flow/netctl/netctl.c
```diff
@@ -0,0 +1,342 @@
...
+    while (service = sol_ptr_vector_get(service_list,i)) {
+        state = sol_netctl_service_get_state(service);
+        name = sol_netctl_service_get_name(service);
+
+        if (name && strcmp(name, mdata->service_name) == 0) {
+            if (state == SOL_NETCTL_SERVICE_STATE_IDLE && connect) {
+                sol_netctl_service_connect(service);
+            } else if ((state == SOL_NETCTL_SERVICE_STATE_READY) && (!connect)) {
+                  sol_netctl_service_disconnect(service);
+            }
```
barbieri: a note for the logic above: I don't think you should need to check the current state before doing the operating. Let the backend handle that. If you ask connect and it's already connected, or disconnect and it's already disconnected, the backend should handle that.

Eventually the backend will return an error "already connected" or "already disconnected", but that's harmless.

The logic you do here is prone to race condition. Suppose you look for IDLE && connect, but the service was in `CONNECTED` and few milliseconds later would go back to `IDLE` because a disconnect request was in progress. You'd miss the connect call. This would happen with a quick disconnect/connect cycle.

IOW: just call `sol_netctl_service_connect()` or `sol_netctl_service_disconnect()`. Then this logic can be moved in the *caller* of this function, instead of using a parameter. This function would be converted into a simple `find_service_by_name()` that does the loop and return the `service` handle. (more on that below, you can't get the state right away)

src/modules/flow/netctl/netctl.c
```diff
@@ -0,0 +1,342 @@
...
+            state = sol_netctl_service_get_state(service);
```
barbieri: it is not correct to query the state here. The call is asynchronous, so it's likely to be the old state since DBus signal did not arrive yet.

The proper way is to remove this `get_state()` from here and listen for service change callback. Then you remember the last-emitted value (ie: the enum value) and if it changed, then you emit the new value and remember the new one.

That is independent from connect/disconnect requests.

src/modules/flow/netctl/netctl.c
```diff
@@ -0,0 +1,342 @@
...
+
+    sol_flow_send_error_packet(node, ENOENT, "Did not found service name %s", mdata->service_name);
+
+    return 0;
+}
+
+static int
+connect_service_name(struct sol_flow_node *node, void *data, uint16_t port, uint16_t conn_id, const struct sol_flow_packet *packet)
+{
+    return connection_handle(node, data, true);
```
barbieri: with my proposal, the above `connection_handle()` would be converted into a `find_service_by_name()` and the logic is:

```c
struct sol_netctl_service *service = find_service_by_name(mdata->service_name);
if (!service) {
    sol_flow_send_error_packet(node, ENOENT, "Did not found service name %s", mdata->service_name);
    return 0; // there was no error in the callback itself, so no need to return error here
}

r = sol_netctl_service_connect(service);
if (r < 0)
    sol_flow_send_error_packet(node, -r, "Could not connect to service: %s: %s, mdata->service_name, sol_util_strerrora(-r));
return 0;
```

barbieri: By the way, the `find_service_by_name()` can be in sol-netctl.h as it's very useful. You can create it as a `static inline` function `sol_netctl_find_service_by_name()`, add this code in the header.

src/modules/flow/netctl/netctl.c
```diff
@@ -0,0 +1,342 @@
...
+static void
+close_service_connection(struct sol_flow_node *node, void *data)
+{
+    static const char *errmsg = "Could not delete service monitor";
+    struct connection_service_data *mdata = data;
+
+    free(mdata->service_name);
+
+    if (sol_netctl_del_service_monitor(NULL, NULL))
+        goto error;
```
barbieri: no, otherwise here you'd keep the error monitor :-)

You always delete both. No need to emit an error packet.

src/modules/flow/netctl/netctl.json
```diff
@@ -0,0 +1,139 @@
...
+      "methods": {
+        "open": "open_list_services",
+        "close": "close_list_services"
+      },
+      "name": "netctl/list-services",
+      "out_ports": [
+        {
+          "data_type": "string",
+          "description": "Name of network services supported by system.",
+          "name": "OUT"
```
barbieri: `OUT` is not enough. You need two ports:
 - `ADDED` services that are new.
 - `REMOVED` services that were removed.

No need to keep an internal shadow list, you do not need a private data for this node type.

src/modules/flow/netctl/netctl.json
```diff
@@ -0,0 +1,139 @@
...
+        }
+      ],
+      "private_data_type": "list_service_data"
+    },
+    {
+      "category": "netctl",
+      "description": "Connect or disconnect to one available network access point.",
```
barbieri: this node is 3 purpose:
 - connect
 - disconnect
 - monitor state

You need to add that last one in here, not a separate node :-)

src/modules/flow/netctl/netctl.json
```diff
@@ -0,0 +1,139 @@
...
+          "name": "CONNECT"
+        },
+        {
+          "data_type": "any",
+          "description": "Trigger to disconnect to network service.",
+          "methods": {
+            "process": "disconnect_service_name"
+          },
+          "name": "DISCONNECT"
+        }
```
barbieri: you need two more input ports:
 - `TRIGGER`: Triggers emission of current service state.
 - `NAME`: set the service name to use (allows changing what is an option in runtime)

See example at https://github.com/solettaproject/soletta/blob/master/src/modules/flow/platform/platform.json#L576 we should make this node similar to that one (except ports START/STOP being called CONNECT/DISCONNECT).

src/modules/flow/netctl/netctl.json
```diff
@@ -0,0 +1,139 @@
...
+            "name": "name"
+          }
+        ],
+        "version": 1
+      },
+      "out_ports": [
+        {
+          "data_type": "string",
+          "description": "Output network services connection status.",
+          "name": "OUT"
```
barbieri: Here I'd have an extra output port `ONLINE`, a boolean that checks if the `OUT` is online. Then it's simple to use and most people will just use that one. Providing a `OUT` will all states will be useful for some specific applications.

src/modules/flow/netctl/netctl.json
```diff
@@ -0,0 +1,139 @@
...
+        }
+      ],
+      "private_data_type": "connection_service_data"
+    },
+    {
+      "category": "netctl",
+      "description": "Get state of a network service. Usually provided by netctl/list-services.",
+      "in_ports": [
```
barbieri: this should be merged in `netctl/service`

src/modules/flow/netctl/netctl.c
```diff
@@ -0,0 +1,342 @@
...
+    SOL_INT_CHECK(r, < 0, r);
+
+    r = sol_flow_packet_get_bool(packet, &value);
+    SOL_INT_CHECK(r, < 0, r);
+
+    SOL_WRN("%d", value);
+    r = sol_netctl_set_radios_offline(value);
+    SOL_INT_CHECK(r, < 0, r);
+
+    value = sol_netctl_get_radios_offline();
```
barbieri: this `enable_radios` must be split into two parts: one for input port, another for output port.

This one will be for input port, rename to `enabled_process` and it will only do `sol_flow_packet_get_bool()` and then `sol_netctl_set_radios_offline(value)`. Without adding a service monitor OR querying the current state. As in another comment I did, you can't immediately query the state, since that value will be updated later, asynchronously, via DBus signal.

The output port will be 2 functions in reality. A node `open` that will call `sol_netctl_add_service_monitor()` with a monitor callback (your second function). From this callback you emit packet on `OUT` port.

src/modules/flow/netctl/netctl.json
```diff
@@ -0,0 +1,139 @@
...
+        {
+          "data_type": "boolean",
+          "description": "turn on or off the radios",
+          "methods": {
+            "process": "enable_radios"
+          },
+          "name": "ENABLED"
+        }
+      ],
+      "name": "netctl/radios",
```
barbieri: maybe `netctl/radios-offline`.

I'm not sure about port name, maybe input port called `IN` to match `OUT`? Or rename port `OUT` to `ENABLED`?

@bdilly @ceolin @cabelitos ?

src/samples/flow/netctl/Makefile
```diff
@@ -0,0 +1,33 @@
+sample-$(FLOW_NETCTL_SAMPLE) += \
+	list-network-services \
+	network-connect \
+	network-disconnect
```
barbieri: missing network-radios?

src/samples/flow/netctl/list-network-services.fbp
```diff
@@ -0,0 +1,27 @@
...
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+# This program will print to console a list of all network services
+#
+# Examples:
+#  * list all network services
+#    $ ./list_network_services.fbp
+
+list(netctl/list-services) OUT -> IN _(console:prefix="Service: ")
```
barbieri: since I've requested 2 ports, you do the following:

```sh
list(netctl/list-services)
list ADDED -> IN _(console:prefix="Added: ")
list REMOVED -> IN _(console:prefix="Removed: ")
```

src/samples/flow/netctl/network-connect.fbp
```diff
@@ -0,0 +1,33 @@
...
+#  * connect or disconnect network services
+#    $ ./network-connect.fbp
+
+conn(netctl/service:name="Guest")
+stat(netctl/state:name="Guest")
+tick(constant/empty)
+
+tick OUT -> TRIGGER stat OUT -> IN _(console:prefix="state: ")
+tick OUT -> CONNECT conn
+conn OUT -> IN _(console)
```
barbieri: you shouldn't need the TRIGGER... but you can do a timer if you wish to force that. The initial trigger shouldn't be needed, do that on node open function.

Also, take the AP name from the cmdline: https://github.com/solettaproject/soletta/blob/master/src/samples/flow/power-supply/power-supply-info.fbp

barbieri: also, to make this sample more useful, you can connect to `OUT` port (to be introduced!) and when service is connected you can quit the application.

src/samples/flow/netctl/network-disconnect.fbp
```diff
@@ -0,0 +1,33 @@
...
+#    $ ./network-disconnect.fbp
+
+
+conn(netctl/service:name="Guest")
+stat(netctl/state:name="Guest")
+tick(constant/empty)
+
+tick OUT -> TRIGGER stat ONLINE -> IN _(console:prefix="online: ")
+tick OUT -> DISCONNECT conn
+conn OUT -> IN _(console)
```
barbieri: like for connect, get the service name from cmdline, wait `OUT` port to change to `IDLE` and then quit.

src/samples/flow/netctl/network-radios.fbp
```diff
@@ -0,0 +1,27 @@
...
+
+# This program will print to console a list of all network services
+#
+# Examples:
+#  * turn on or off network connection
+#    $ ./network-radios.fbp
+
+radio(netctl/radios)
+
+_(constant/boolean:value=false) OUT -> ENABLED radio OUT -> IN _(console)
```
barbieri: here you can do a nice example with the logic:
 - `app/argv:index=1`, if there is an argument, it will convert to boolean and set `ENABLED`. As far as I remember, if no argument is given, it will send an `ERROR` packet on that port. Use that error packet as "no argument given" and if that is the case, keep monitoring. If a packet was given, then quit the application when the radio state change to the desired value.
 - `radio OUT` sent to console to show the current state.

PR #2194, opened 2016-06-22
scripts: add script to generate list of node types

bdilly:
And tests or samples using them.
Useful for reports.

### DIFF
--- a/data/scripts/sol-flow-node-type-examples-list.py
+++ b/data/scripts/sol-flow-node-type-examples-list.py
@@ -1,0 +1,117 @@
+#!/usr/bin/env python3
+# -*- coding: utf-8 -*-
+
+# This file is part of the Soletta™ Project
+#
+# Copyright (C) 2016 Intel Corporation. All rights reserved.
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#     http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+# This script should be used to generate a list of node types
+# and examples of use. May be useful to generate reports.
+#
+# TODO - handle types used by conf files
+#
+# Usage example:
+#
+# ./sol-flow-node-type-examples-list.py file.csv soletta/src/ \
+#   src/lib/flow/builtins.json src/modules/flow/*.json
+
+import json
+import os
+import re
+from os import walk
+
+# TODO - handle types used by conf files
+def get_examples(src_dir, path_index):
+    examples = {}
+    fbp_files = []
+
+    for (dirpath, dirnames, filenames) in walk(src_dir):
+        for filename in filenames:
+            if filename.endswith(".fbp"):
+                fbp_files.append(os.path.join(dirpath, filename))
+
+    for fbp_file in fbp_files:
+        with open(fbp_file, "r") as f:
+            content = f.read()
+            # skip error tests
+            if "TEST-EXPECTS-ERROR" in content:
+                continue
+            for line in content.splitlines():
+                nodes = re.findall("\(.*?\)", line)
+                for node in nodes:
+                    # if it hasn't options, find will return -1, required to
+                    # remove ')'.
+                    type_end = node.find(':')
+                    node = node[1:type_end]
+                    node_files = examples.get(node, [])
+                    fbp_url = fbp_file[path_index:]
+                    if fbp_url not in node_files:
+                        node_files.append(fbp_url)
+                        examples[node] = node_files
+
+    return examples
+
+def print_node_type(outfile, node_type, examples, group_id):
+    examples_list = " "
+    examples = examples.get(node_type["name"])
+    if examples:
+        for example in examples:
+            # drop src/ and .fbp
+            examples_list = examples_list + example[:-4] + ", "
+        examples_list = examples_list[:-2]
+
+    outfile.write("""\
+"%(name)s";%(examples)s
+""" % {
+    "name": node_type["name"],
+    "examples": examples_list
+    })
+
+def print_description(outfile, description, infile, examples):
+    modules = description.keys()
+    if len(modules) != 1:
+        print("Warning: a single module is expected per file. Skiping %s" %
+              infile.name)
+        return
+
+    for module in description.keys():
+        for node_type in description[module]:
+            print_node_type(outfile, node_type, examples, module)
+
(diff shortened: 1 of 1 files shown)


data/scripts/sol-flow-node-type-examples-list.py
```diff
@@ -0,0 +1,117 @@
...
+            if "TEST-EXPECTS-ERROR" in content:
+                continue
+            for line in content.splitlines():
+                nodes = re.findall("\(.*?\)", line)
+                for node in nodes:
+                    # if it hasn't options, find will return -1, required to
+                    # remove ')'.
+                    type_end = node.find(':')
+                    node = node[1:type_end]
+                    node_files = examples.get(node, [])
```
barbieri: instead of `[]` (list) you can use a `set()`. Then you simply add and it will take care to keep a single instance in there. I just recommend to print-out the sorted list of the set so it's stable to the user.

barbieri: also, if you use `examples.setdefault(node, set())` (or the list version), you do not need the `example[node] = node_files` below. It will work as "get if exists, otherwise set that second parameter".

data/scripts/sol-flow-node-type-examples-list.py
```diff
@@ -0,0 +1,117 @@
...
+                        node_files.append(fbp_url)
+                        examples[node] = node_files
+
+    return examples
+
+def print_node_type(outfile, node_type, examples, group_id):
+    examples_list = " "
+    examples = examples.get(node_type["name"])
+    if examples:
+        for example in examples:
```
barbieri: if you use `examples.get(key, [])` (or set() if changed to be that), then you can simply iterate without the need to check.

data/scripts/sol-flow-node-type-examples-list.py
```diff
@@ -0,0 +1,117 @@
...
+            # drop src/ and .fbp
+            examples_list = examples_list + example[:-4] + ", "
+        examples_list = examples_list[:-2]
+
+    outfile.write("""\
+"%(name)s";%(examples)s
+""" % {
+    "name": node_type["name"],
+    "examples": examples_list
+    })
```
barbieri: this is the only bad part in the script as it's subject to problems of escaping.

My recommendation is to use the CSV module and its writer. But given the structure of this file, it would be better to generate a JSON. It would make this file immediately usable in dev-app, just put the file in the samples root and we could offer: nodes -> examples, and when we add the configuration files we can extend that with proper key-value objects.

data/scripts/sol-flow-node-type-examples-list.py
```diff
@@ -0,0 +1,117 @@
...
+
+def print_description(outfile, description, infile, examples):
+    modules = description.keys()
+    if len(modules) != 1:
+        print("Warning: a single module is expected per file. Skiping %s" %
+              infile.name)
+        return
+
+    for module in description.keys():
+        for node_type in description[module]:
```
barbieri: these are usually unstable, particularly the keys of a dict. Use `sorted(description.keys())` and the likes to ensure a stable output.

data/scripts/sol-flow-node-type-examples-list.py
```diff
@@ -0,0 +1,117 @@
...
+
+
+if __name__ == "__main__":
+    import argparse
+    parser = argparse.ArgumentParser()
+    parser.add_argument("csv",
+                        help="Path to write csv file",
+                        type=argparse.FileType('w'))
+    parser.add_argument("src_dir",
+                        help="Path to Soletta src/ directory with example files",
```
barbieri: no need to mention `Soletta src/ directory`, just `the directory with examples`. I think we can install this script in `/usr/share/soletta/SOMETHING` and let people use that as well, useful with extensions, like SML.

data/scripts/sol-flow-node-type-examples-list.py
```diff
@@ -0,0 +1,117 @@
...
+                        type=str)
+    parser.add_argument("inputs_list", nargs='+',
+                        help="List of input description files in JSON format",
+                        type=argparse.FileType('r'))
```
barbieri: why do we need to specify all the description files? Isn't it better to give the folder? Then we walk it to generate the list in Python, it's easier to use and not subject to exceeding the maximum cmdline size.

data/scripts/sol-flow-node-type-examples-list.py
```diff
@@ -0,0 +1,117 @@
...
+                fbp_files.append(os.path.join(dirpath, filename))
+
+    for fbp_file in fbp_files:
+        with open(fbp_file, "r") as f:
+            content = f.read()
+            # skip error tests
+            if "TEST-EXPECTS-ERROR" in content:
+                continue
+            for line in content.splitlines():
+                nodes = re.findall("\(.*?\)", line)
```
barbieri: you can use `re.compile("[(](?P<name>[^):]+)[):]")` to compile a regular expression that gives you a group `name` with a list of matches without the leading and trailing parenthesis, as well as the `:`.

data/scripts/sol-flow-node-type-examples-list.py
```diff
@@ -0,0 +1,117 @@
...
+            if filename.endswith(".fbp"):
+                fbp_files.append(os.path.join(dirpath, filename))
+
+    for fbp_file in fbp_files:
+        with open(fbp_file, "r") as f:
+            content = f.read()
+            # skip error tests
+            if "TEST-EXPECTS-ERROR" in content:
+                continue
+            for line in content.splitlines():
```
barbieri: usually it's better to iterate in the file, instead of reading the whole file, allocating the array with split lines. Something like:
```py
# globals
comment_line_re = re.compile("^ *#")
node_name_re = re.compile("[(](?P<name>[^):]+)[):]")

# locals
ignore_file = False
local_nodes = set()
for line in f:
    if "TEST-EXPECTS-ERROR" in line:
        ignore_file = True
    if comment_line_re.match(line):
       continue
    for match in node_name_re.finditer(line):
        node = match.group("name")
        local_nodes.add(node)

if ignore_file:
   return

global_nodes.update(local_nodes)
```

data/scripts/sol-flow-node-type-examples-list.py
bdilly: I'm just interested on importing this on excel actually.

What escaping issues do you see?

barbieri:
overall looks good, could be bit more pythonic and I'd use JSON instead of CSV due the expected structure of the output, making it directly usable by DevApp later.

If you wish I can take over the work to fix the issues I proposed (and test them, since most were just typed here in github)

data/scripts/sol-flow-node-type-examples-list.py
barbieri: I was hoping this was for DevApp and cheat-sheet :-)

The escaping issues shouldn't appear if we do not use the `;` in file names, those are valid yet unexpected.

bdilly:
merged with many of suggested changes

PR #2195, opened 2016-06-22
build: remove stage/include from HEADERDIRS

brunobottazzini:
This patchset includes 2 commits:

1-  Fixes the warning: ./build/stage/include/: No such file or directory
[-Wmissing-include-dirs] 

2- Fixes the warning: nonnull argument ‘value’ compared to NULL [-Wnonnull-compare]

### DIFF
--- a/src/lib/datatypes/include/sol-memdesc.h
+++ b/src/lib/datatypes/include/sol-memdesc.h
@@ -269,7 +269,7 @@ enum sol_memdesc_type {
  * @param str the string representing a valid type.
  * @return the type or SOL_MEMDESC_TYPE_UNKNOWN if invalid.
  */
-enum sol_memdesc_type sol_memdesc_type_from_str(const char *str) SOL_ATTR_WARN_UNUSED_RESULT SOL_ATTR_NON_NULL(1);
+enum sol_memdesc_type sol_memdesc_type_from_str(const char *str) SOL_ATTR_WARN_UNUSED_RESULT;
 
 /**
  * @brief Converts a sol_memdesc_type to a string.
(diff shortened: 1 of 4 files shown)


barbieri:
I don't like this as a final goal, but we can use it as a starting point to make things uniform.

My dream is to have all non-null methods to be flagged like so, that way any user will get warnings/errors if giving null pointers to us.

On the double-safety side, we can remove our own warnings (-Wnonnull-compare) by doing a trick I've used in EFL before: for an implementation file "sol-XXX.c" of "sol-XXX.h", we include system headers first, then a local build header "redefine SOL_ATTR_NON_NULL to empty", then include "sol-XXX.h", then another local header "redefine SOL_ATTR_NON_NULL to its actual value". This allows the "sol-XXX.c" to check for null parameters on its own files while if it passes null arguments to others, it will get the warnings. Result is something like:

```c

#include <unistd.h> // systems headers

#include "disable-attr-non-null.h"
#include "sol-XXX.h"
#include "enable-attr-non-null.h"

SOL_API void
sol_XXX_do(sol_XXX *handle) {
    SOL_NULL_CHECK(handle); // no warnings!
    sol_YYY_do(NULL); // get warnings!
}
```

With `disable-attr-non-null.h` being something like:
```c
#include "sol-macros.h"
#ifdef SOL_ATTR_NON_NULL
#undef SOL_ATTR_NON_NULL
#define SOL_ATTR_NON_NULL
#endif
```

And `enable-attr-non-null.h` being:
```c
#if __GNUC__ >= 4
#undef SOL_ATTR_NON_NULL
#define SOL_ATTR_NON_NULL(...) __attribute__((nonnull(__VA_ARGS__)))
#endif
```

Maybe call the headers a bit more broadly, since in the future we may have other warnings/flags to change for a given file.

brunobottazzini:
I will make a V2 fixing the error that CI pointed.

Talking with @bdilly, we agreed to proceed with this approach as a V1 and after we can makie the suggestion you made.

Thanks

PR #2198, opened 2016-06-23
Some more tests and samples

bdilly:


src/samples/flow/webservices/thingspeak-talkback-identify.fbp
```diff
@@ -0,0 +1,59 @@
...
+#
+# There you should create a new Talkback. You need the talkback
+# id (not the name) and the API Key. Use these values as
+# options for talkback_add and talkback_execute nodes.
+#
+# You'll be able to see it throught the Thingspeak web interface as well,
+# if you want.
+
+
+# TODO wouldn't be better to accept API_KEY and ID through input ports?
```
cabelitos: Are we leaving this TODO here?

src/samples/flow/webservices/thingspeak-talkback-smart-lock.fbp
```diff
@@ -0,0 +1,61 @@
...
+#
+# Register at https://thingspeak.com or use another server (and
+# update node type options).
+#
+# There you should create a new Talkback. You need the talkback
+# id (not the name) and the API Key. Use these values as
+# options for talkback_add and talkback_execute nodes.
+
+
+# TODO wouldn't be better to accept API_KEY and ID through input ports?
```
cabelitos: Same thing

src/modules/flow/network/network.c
```diff
@@ -190,7 +190,7 @@ network_open(struct sol_flow_node *node, void *data, const struct sol_flow_node_
     SOL_FLOW_NODE_OPTIONS_SUB_API_CHECK(options,
         SOL_FLOW_NODE_TYPE_NETWORK_BOOLEAN_OPTIONS_API_VERSION, -EINVAL);
 
-    if (!sol_network_subscribe_events(_on_network_event, mdata))
+    if (sol_network_subscribe_events(_on_network_event, mdata) < 0)
```
cabelitos: Should we use `SOL_INT_CHECK()` and properly return the error value

src/modules/linux-micro/network-up/network-up.c
```diff
@@ -64,7 +64,7 @@ network_up_init(const struct sol_platform_linux_micro_module *module, const char
 {
     SOL_LOG_INTERNAL_INIT_ONCE;
 
-    if (!sol_network_subscribe_events(_network_event_cb, NULL))
+    if (sol_network_subscribe_events(_network_event_cb, NULL) < 0)
```
cabelitos: Same thing

src/samples/network/network-status.c
```diff
@@ -154,7 +154,7 @@ startup_network(void)
     if (!_compile_regex(regexp))
         goto err;
 
-    if (!sol_network_subscribe_events(_on_network_event, NULL))
+    if (sol_network_subscribe_events(_on_network_event, NULL) < 0)
```
cabelitos: Same thing

src/samples/flow/webservices/thingspeak-talkback-identify.fbp
bdilly: I won't be able to modify the node type for now.
I'll open an issue later regarding accepting these info via input ports.
Tbh I'm still unsure about the best way of dealing with that, after all you don't want to let these values readable on memory. We should specify an approach and use it on all webservices node types - all of them need to deal with this issue.

src/modules/flow/network/network.c
bdilly: good point

PR #2199, opened 2016-06-24
Add Version script v2

cabelitos:
* Added commit 7bc636d
* Using glob to export the SOL_FLOW_PACKET_TYPES
* APIs are sorted alphabetically by module


data/scripts/libsoletta.sym
```diff
@@ -0,0 +1,867 @@
...
+        sol_json_token_get_int64;
+        sol_json_token_get_uint64;
+        sol_json_token_get_unescaped_string;
+        sol_json_token_get_unescaped_string_copy;
+
+        sol_update_cancel;
+        sol_update_check;
+        sol_update_fetch;
+        sol_update_get_progress;
+        sol_update_install;
```
barbieri: sol_update here and up in this file. My suggestion is to sort ALL, case SENSITIVE (so `SOL_` comes separated from `sol_`) and do a blank line when the module change (`sol_XXX` and `sol_YYY`).

This will make things easier to maintain

cabelitos: Hmmm, ok. I'll change that.

cabelitos:
#2203

PR #2200, opened 2016-06-24
Fix warnings in build - V2

brunobottazzini:
Previous patch: https://github.com/solettaproject/soletta/pull/2195

Differences from V1:
- Fix build error caught by CI, Instead of removing the stage/include dir from HEADERDIRS we add it only if OIC is enabled since, it is the only module that needs the folder to be created.

------

This patchset includes 2 commits:

1- Fixes the warning: ./build/stage/include/: No such file or directory
[-Wmissing-include-dirs]

2- Fixes the warning: nonnull argument ‘value’ compared to NULL [-Wnonnull-compare]

### DIFF
--- a/src/lib/datatypes/include/sol-memdesc.h
+++ b/src/lib/datatypes/include/sol-memdesc.h
@@ -269,7 +269,7 @@ enum sol_memdesc_type {
  * @param str the string representing a valid type.
  * @return the type or SOL_MEMDESC_TYPE_UNKNOWN if invalid.
  */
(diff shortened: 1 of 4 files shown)


bdilly:
but in this case, building with OIC enabled, won't we see the warning?

brunobottazzini:
@bdilly no because the build with OIC will create the dir.

The whole problem was when building without OIC and the HEADERDIRS was including a directory that does not exists.

cabelitos:
lgtm

PR #2202, opened 2016-06-27
Add sol-netctl API for string mappings to/from enumeration sol_netctl_service_state 

guchaojie:
Add sol-netctl API for converting a string service state to
enumeration sol_netctl_service_state and otherwise.

Change-Id: I85e1d356822cde16fd7f53416f6769b0b6e2e5e7

### DIFF
--- a/src/lib/comms/include/sol-netctl.h
+++ b/src/lib/comms/include/sol-netctl.h
@@ -471,6 +471,33 @@ int sol_netctl_del_error_monitor(sol_netctl_error_monitor_cb cb,
  */
 const struct sol_ptr_vector *sol_netctl_get_services(void);
 
+/**
+ * @brief Converts a string service state to sol_netctl_service_state.
+ *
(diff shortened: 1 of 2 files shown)


guchaojie:
@barbieri @bdilly @wzhen12 please have a review.

guchaojie:
I would merge this into netctl node type PS as one commit to sumbit together, so close it.

barbieri:
this looks good to me and can be merged directly. A minor detail I'd change is to use the strings as lowercase, not the first letter in uppercase.

what you mean with the node type? I've asked this to be in the C API as you did, then it's not limited to the node/flow, like Python/JS bindings can use that.

guchaojie:
@barbieri  I mean I will submit this commit with netctl node/flow patch together as one PR for reviews.

guchaojie:
modify the string to lowercase now.
Merged it , Thanks.

PR #2209, opened 2016-06-28
modules/thingspeak: fix connections management

bdilly:
Some warnings were raised when closing thingspeak nodes,
since finished connections
never were removed from pending vector.


cabelitos:
+1

Issue #2207, opened 2016-06-28
IIO: add sample for iio/proximity-sensor


bdilly:
The recently added node type iio/proximity-sensor is missing any sample.
It would be great to have one just like all the others under the same category.

bdilly:
Hi @yongli3 ,

please, would you be able to help us on that?

I lack hardware to do it myself. Also I suppose you have written one when you tested it before merging the code.

yongli3:
@bdilly No problem !

bdilly:
Thanks a lot, @yongli3 
PR merged, ticket closed =)

PR #2210, opened 2016-06-29
samples: add iio/proximity-sensor sample

yongli3:

### DIFF
--- a/src/samples/flow/iio/sol-flow-intel-minnow-max-linux_gt_3_17.json
+++ b/src/samples/flow/iio/sol-flow-intel-minnow-max-linux_gt_3_17.json
@@ -64,6 +64,14 @@
            "buffer_size": "-1"
      },
      "type": "iio/thermometer"
+  },
+  {
+     "name": "MyProximity",
+     "options": {
+           "iio_device": "apds9960 create,i2c,platform/80860F41:05,0x39,apds9960 0",
+           "buffer_size": "-1"
+     },
+     "type": "iio/proximity-sensor"
   }
  ]
 }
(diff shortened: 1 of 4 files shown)


yongli3:
@bdilly @neofang @peijiajames this PR is used for issue #2207, please help to review

bdilly:
Hi @yongli3 , it seems good.

edersondisouza:
:+1:

src/samples/flow/iio/sol-flow-intel-minnow-max-linux_gt_3_17.json
```diff
@@ -64,6 +64,14 @@
            "buffer_size": "-1"
      },
      "type": "iio/thermometer"
+  },
+  {
+     "name": "MyProximity",
+     "options": {
+           "iio_device": "apds9960 create,i2c,platform/80860F41:05,0x39,apds9960 0",
+           "buffer_size": "-1"
```
barbieri: any reason why disable buffering?

yongli3: @barbieri I found not all sensors support the buffering/trigger reading. In this case, the apds9960 does not support it

barbieri: @edersondisouza explained to me at IRC, but I think the "0" (auto) logic is missing some pieces and should understand devices that do not support buffering. We'll fix that later, after that we could always go with the default value (0) and it should work, the buffer must be used only explicitly.

yongli3: I got your point. Curently the “buffer_size=0” will use the default buffer size, it does not fallback to “normal reading” if the driver does not support buffering. We need to enhance the logic, to check if it supports buffering when “buffer_size > 0”

PR #2211, opened 2016-06-29
Add Netctl node type and samples for network control V2

guchaojie:
Add netctl fbp node and samples include list services, connect or disconnect service, get service state, turn on or off radios

Change-Id: I60cfaff351f0f65b185cf6d48292a98698618082


guchaojie:
@barbieri @bdilly @vcgomes @wzhen12 
According to previous PR comments to update PR, please have a review.

src/modules/flow/netctl/netctl.c
```diff
@@ -0,0 +1,336 @@
...
+    return 0;
+}
+
+static void
+close_list_services(struct sol_flow_node *node, void *data)
+{
+    static const char *errmsg = "Could not delete service monitor";
+
+    if (sol_netctl_del_service_monitor(service_list_cb, node))
+        SOL_WRN("%s", errmsg);
```
barbieri: any reason why you use "%s" with a string? If the string doesn't contain any "%" you can safely use the string itself:

```c
SOL_WRN("Could not delete service monitor");
```

or you simply use the `SOL_INT_CHECK()` like in open function

src/modules/flow/netctl/netctl.c
```diff
@@ -0,0 +1,336 @@
...
+    const char *name;
+    int r;
+
+    name = sol_netctl_service_get_name(service);
+
+    if (!name || strcmp(name, mdata->service_name)) 
+        return;
+
+    current_state = sol_netctl_service_get_state(service);
+    if(current_state != mdata->state) {
```
barbieri: like above, you can do an early return here:

```c
if (current_state == mdata->state)
    return;
```

Also check the style, it's missing a space. Please run `./tools/check-style.py` or even better, use a GIT hook so you don't forget it: https://github.com/solettaproject/soletta/wiki/Contributing#git-hooks

src/modules/flow/netctl/netctl.c
```diff
@@ -0,0 +1,336 @@
...
+                         false);
+             SOL_INT_CHECK_GOTO(r, < 0, error);
+        }
+
+        if (current_state == SOL_NETCTL_SERVICE_STATE_ONLINE) {
+              r = sol_flow_send_bool_packet(node,
+                         SOL_FLOW_NODE_TYPE_NETCTL_SERVICE__OUT__ONLINE,
+                         true);
+             SOL_INT_CHECK_GOTO(r, < 0, error);
+        }
```
barbieri: This logic should be only:

```c
r = sol_flow_send_bool_packet(node,
                         SOL_FLOW_NODE_TYPE_NETCTL_SERVICE__OUT__ONLINE,
                         current_state == SOL_NETCTL_SERVICE_STATE_ONLINE);
```

No need to use the previous state at this point.

barbieri: hum, I guess you tried to do not emit extra booleans with the same value, but this shouldn't matter much. But it's okay to keep what you did.

src/modules/flow/netctl/netctl.c
```diff
@@ -0,0 +1,336 @@
...
+error_cb(void *data, const struct sol_netctl_service *service,
+    unsigned int error)
+{
+    struct sol_flow_node *node = data;
+    const char *name;
+
+    name = sol_netctl_service_get_name(service);
+
+    if (name)
+        sol_flow_send_error_packet(node, ENOENT, "Service %s error is %d", name, error);
```
barbieri: instead of `ENOENT`, use `error` number. In the message, you use the `%s` and then `sol_util_strerrora(error)` so the message gives some meaning.

src/modules/flow/netctl/netctl.c
```diff
@@ -0,0 +1,336 @@
...
+}
+
+static struct sol_netctl_service *find_service_by_name(char *service_name)
```
barbieri: coding style... and what about to move this function to sol-netctl.h as a static-inline? Call it `sol_netctl_find_service_by_name()` and the `service_name` should be a `const char*` (missing `const` modifier)

src/modules/flow/netctl/netctl.c
```diff
@@ -0,0 +1,336 @@
...
+                     mdata->service_name, sol_util_strerrora(-r));
+    }
+
+    return 0;
+}
+
+static int
+connect_service_name(struct sol_flow_node *node, void *data, uint16_t port, uint16_t conn_id, const struct sol_flow_packet *packet)
+{
+    return connection_handle(node, data, true);
```
barbieri: instead of this function, put the code that matters for `connect` here. The only "duplication" is finding the service, which is a small part of the logic. You can keep that as its own function (ie: `find_or_emit_error()` function), or copy those few lines here.

src/modules/flow/netctl/netctl.c
```diff
@@ -0,0 +1,336 @@
...
+}
+
+static int
+disconnect_service_name(struct sol_flow_node *node, void *data, uint16_t port, uint16_t conn_id, const struct sol_flow_packet *packet)
+{
+    return connection_handle(node, data, false);
```
barbieri: same as for `connect_service_name()`, move the contents of `connection_handle()` relative to disconnect here.

src/modules/flow/netctl/netctl.c
```diff
@@ -0,0 +1,336 @@
...
+}
+
+static int
+get_service_state(struct sol_flow_node *node, void *data, uint16_t port, uint16_t conn_id, const struct sol_flow_packet *packet)
```
barbieri: This whole function body is very similar to `service_status_cb()`. You should move the parts that are similar (ie: emit packets given a `service` into its own function, then call that new function from `service_status_cb()` and `get_service_state()`. This function would only call `find_service_by_name()` and then call that new function.

src/modules/flow/netctl/netctl.c
```diff
@@ -0,0 +1,336 @@
...
+set_service_name(struct sol_flow_node *node, void *data, uint16_t port, uint16_t conn_id, const struct sol_flow_packet *packet)
+{
+    struct network_service_data *mdata = data;
+    const char *in_value;
+    int r;
+
+    r = sol_flow_packet_get_string(packet, &in_value);
+    SOL_INT_CHECK(r, < 0, r);
+
+    mdata->service_name = strdup(in_value);
```
barbieri: memory leak. You should `free()` then `strdup()` or you can use the smarter `sol_util_replace_str_if_changed(&mdata->service_name, in_value)` https://github.com/solettaproject/soletta/blob/master/src/shared/include/sol-util.h#L515

src/modules/flow/netctl/netctl.c
```diff
@@ -0,0 +1,336 @@
...
+}
+
+void
+error_cb(void *data, const struct sol_netctl_service *service,
+    unsigned int error)
+{
+    struct sol_flow_node *node = data;
+    const char *name;
+
+    name = sol_netctl_service_get_name(service);
```
barbieri: if this is a node for a given `mdata->service_name`, then you should ignore other services.

src/modules/flow/netctl/netctl.c
```diff
@@ -0,0 +1,336 @@
...
+    }
+
+    mdata->state = SOL_NETCTL_SERVICE_STATE_UNKNOWN;
+
+    r = sol_netctl_add_service_monitor(service_status_cb, node);
+    SOL_INT_CHECK(r, < 0, r);
+
+    r = sol_netctl_add_error_monitor(error_cb, node);
+    SOL_INT_CHECK(r, < 0, r);
+
```
barbieri: I think you should force a `TICK` here by calling `get_service_state()` if there is a desired service name.

For connman implementation it will happen as magic, since the DBus call will deliver results later, that will trigger the "service_status_cb". But for other implementations or even if the FBP was loaded later in the process lifetime, the services list will be known and thus nothing will be dispatched unless the service changes.

Forcing an initial "TICK" (`get_service_state()`) will produce the same result every time. It's consistent.

barbieri: See that in `platform` it does that by calling the `get_state()` and then the dispatch:

https://github.com/solettaproject/soletta/blob/master/src/modules/flow/platform/platform.c#L90

src/modules/flow/netctl/netctl.c
```diff
@@ -0,0 +1,336 @@
...
+}
+
+static int
+open_network_radio(struct sol_flow_node *node, void *data, const struct sol_flow_node_options *options)
+{
+    int r;
+
+    r = sol_netctl_add_manager_monitor(manager_cb, node);
+    SOL_INT_CHECK(r, < 0, r);
+
```
barbieri: missing `get_radios_offline()` + `send_bool_packet()`. IOW, call `manager_cb()` :-)

src/modules/flow/netctl/netctl.c
```diff
@@ -0,0 +1,336 @@
...
+    return 0;
+}
+
+static void
+close_network_radio(struct sol_flow_node *node, void *data)
+{
+    static const char *errmsg = "Could not delete manager monitor";
+
+    if (sol_netctl_del_manager_monitor(manager_cb, node))
+        SOL_WRN("%s", errmsg);
```
barbieri: you can avoid the `%s`

src/modules/flow/netctl/netctl.c
```diff
@@ -0,0 +1,336 @@
...
+    struct network_service_data *mdata = data;
+    const char *in_value;
+    int r;
+
+    r = sol_flow_packet_get_string(packet, &in_value);
+    SOL_INT_CHECK(r, < 0, r);
+
+    mdata->service_name = strdup(in_value);
+    SOL_NULL_CHECK(mdata->service_name, -ENOMEM);
+
```
barbieri: this should also force the reload by calling `get_service_state()`

src/samples/flow/netctl/Makefile
```diff
@@ -0,0 +1,34 @@
...
+
+sample-network-connect-$(FLOW_NETCTL_SAMPLE)-deps := \
+	flow/netctl.mod
+
+#radios
+
+sample-network-radios-$(FLOW_NETCTL_SAMPLE) := \
+	network-radios.fbp
+
+sample-network-radios-$(FLOW_NETCTL_SAMPLE)-deps := \
```
barbieri: you do not need to repeat the `-deps := ` line for all samples. They will just reset. Move these up after `sample-$(FLOW_NETCTL_SAMPLE) += ...` line (before `# list` comment).

src/samples/flow/netctl/network-connect.fbp
```diff
@@ -0,0 +1,40 @@
...
+tick(constant/empty)
+timer(timer:interval=3000)
+btoi(converter/boolean-to-int)
+
+name ERROR -> IN error-msg OUT -> IN Error
+service ERROR -> ERROR quit
+
+name OUT -> NAME service
+tick OUT -> CONNECT service OUT -> IN _(console:prefix="state: ")
+timer OUT -> TRIGGER service ONLINE -> IN btoi OUT -> CODE quit
```
barbieri: this will always quit (will send 0/1 to quit). I believe you should only quit if it's `ONLINE`, in that case use a filter.

src/samples/flow/netctl/network-disconnect.fbp
```diff
@@ -0,0 +1,40 @@
...
+service(netctl/service)
+tick(constant/empty)
+timer(timer:interval=3000)
+
+name ERROR -> IN error-msg OUT -> IN Error
+service ERROR -> ERROR quit
+
+name OUT -> NAME service
+tick OUT -> DISCONNECT service OUT -> IN _(console:prefix="state: ")
+timer OUT -> TRIGGER service ONLINE -> QUIT quit
```
barbieri: you miss a boolean to integer and also a filter to wait online be false.

src/modules/flow/netctl/netctl.c
guchaojie: It would call API sol_netctl_get_services and sol_netctl_service_get_name in find_service_by_name function , so this function cannot be moved into sol-netctl.h :(

src/samples/flow/netctl/network-connect.fbp
guchaojie: I'm not sure which node type could have filter function?

src/modules/flow/netctl/netctl.c
barbieri: what do you mean? these are public functions, they can be called from the `sol-netctl.h`. Just declare this new function after the other prototype.

src/samples/flow/netctl/network-connect.fbp
barbieri: You have two options, see http://solettaproject.github.io/docs/nodetypes/
 - `service ONLINE -> IN _(boolean/filter) TRUE -> IN btoi OUT -> CODE quit` (not that good since it will exit with error code=1, you should change the `false_value` and `true_value` of `btoi` to fix that);
 - `service ONLINE -> PULSE_IF_TRUE _(converter/boolean-to-empty) OUT -> IN _(constant/int:value=0) OUT -> CODE quit`

src/modules/flow/netctl/netctl.c
guchaojie: get_service_state is not totally similar with service_status_cb. 
In service_status_cb, it wil emit online or offline when the status changed. So I use the current_state compared with previous state.
In get_service_state , it will emit online or state by Trigger node, so it always emit the current_state by given a service.
So this part cannot merged into one new function. In service_status_cb there are some logic to compare current state and previous state.  In get_service_state ,just emit current status packet.

src/modules/flow/netctl/netctl.c
guchaojie: @barbieri That's what I am concern. Because all the services got in FBP are from sol_netctl_add_service_monitor callback network_service_cb. 
So we create netctl/service node and then connect to given service name. Maybe we cannot successfully connect because service_list by calling API sol_netctl_find_service_name is not ready yet for asynchronous delay reason.
I think the best time to connect service is after emitting status packet first time in service_status_cb.
Initial Tick maybe still is not time for user to connect service, just delay some time for user to connect service.
Maybe I didn't totally catch your meaning. I saw the examples, this dispatch just send bool packet in the platform_open which still cannot ensure service has been catched in our network_service_cb case.

src/modules/flow/netctl/netctl.c
guchaojie: That's the same confusion as above

src/modules/flow/netctl/netctl.c
barbieri: ok, leave as 2 functions. But I'd make the `get_service_state()` also remember the last value.

src/modules/flow/netctl/netctl.c
barbieri: what you mean? When I reset the service name, if the service exists, then it should immediately send the values. Otherwise I'd be forced to connect to `TRIGGER` to get it done, or wait for some change in the service so it would be dispatched in `service_status_cb`

src/modules/flow/netctl/netctl.c
barbieri: You need to do both:

 - on `open()`, add monitor with `network_service_cb`, get  the existing services and if the desired service is returned in `sol_netctl_get_services()`, then dispatch the status on `OUT` and `ONLINE` ports. If there were no services (ie: no DBus message arrived at this moment, since it was the first user), `sol_netctl_get_services()` will return an empty list, thus will be walked (zero elements) and nothing is dispatched;
 - on `network_service_cb`, like the service appeared due connman scan, or it was acquired via `GetServices` via DBus on the first time, if the service name matches, then you do your logic.

The second item is already correct. You miss the first item, that is basically to call `get_service_state()` from `open()`.

guchaojie: get_service_state is defined by Trigger in netctl/service from netctl.json file, two question here:
1. How to calling Tick get_service_state in  open function?
2. service(netctl/service)
    name OUT -> NAME service
//after create service node and before TICK to connect, the desire service cannot monitor yet.
 initial Tick in open , this method can ensure the desire service name is ready when CONNECT service?
    tick OUT -> CONNECT service OUT -> IN _(console:prefix="state: ")

src/modules/flow/netctl/netctl.c
guchaojie: Really Sorry, I am still not very clear about this comments.  When you reset service name,  you mean "service exists" is that input service name in service_list , right?, if input service in the service list from sol_netctl_get_services, why we should send the values in set_service_name function? and what value should be emitted? Otherwise, connect to TRIGGER.....  What TRIGGER?  I have lost here in fact.

src/modules/flow/netctl/netctl.c
barbieri: From your `open_network_service()` call `get_service_state()`,  that simple :-)

as I said, you should also call that `get_service_state()` when `set_service_name()` is called. That way you always get the correct behavior and do not need the manual `TICK`.

As for your example with `tick OUT -> CONNECT service` you should think if that's a common case, if so, you can flag "pending connect" or "pending disconnect" with a logic similar to:

```c
struct network_service_data {
    ...
    enum pending_service_operation {
        NONE, CONNECT, DISCONNECT
    } pending;
}; 

open_network_service() {
    mdata->pending = NONE;
}

connect_service_name() {
    struct sol_netctl_service *service = NULL;

    if (mdata->service_name)
        service = sol_netctl_find_service_by_name(mdata->service_name);

    if (!service)
        mdata->pending = CONNECT;
    else {
        int r = sol_netctl_service_connect(service);
       // if (r < 0) send error pkt
    }
}

disconnect_service_name() {
    struct sol_netctl_service *service = NULL;

    if (mdata->service_name)
        service = sol_netctl_find_service_by_name(mdata->service_name);

    if (!service)
        mdata->pending = DISCONNECT;
    else {
        int r = sol_netctl_service_disconnect(service);
       // if (r < 0) send error pkt
    }
}

set_service_name() {
    struct sol_netctl_service *service;
    int r;

    r = sol_util_replace_str_if_changed(&mdata->service_name, in_value);
    if (r <= 0)
        return r;

    if (mdata->pending == NONE)
        return 0;

    service = sol_netctl_find_service_by_name(mdata->service_name);
    if (!service)
        return 0;
    if (mdata->pending == CONNECT)
        r = sol_netctl_service_connect(service);
    else
        r = sol_netctl_service_disconnect(service);

   // if (r < 0) send error pkt
}
```

src/modules/flow/netctl/netctl.c
barbieri: Consider the following conditions and the given FBP:

```sh
# when user changes the service name, update the one we monitor
some_user_input OUT -> NAME service
# show the monitor to user
service STATE -> IN some_user_output
```

## Service did not exist, then appeared

In this scenario, when `some_user_input OUT` generates a string packet `"someAP"`, the connman list did **not** know about any "someAP".

The `"someAP"` packet is processed by `set_service_name` and stored.

**After** the `"someAP"` packet is processed, connman finds a new service (ie: manual Scan DBus method, periodic scan...), and sends a DBus Signal "ServicesChanged", it will call our `service_status_cb()`, which will see `strcmp(mdata->service_name, name)` that matches and will send a new status on `service STATE -> IN some_user_output`. This will print out the output to user.

## Service did exist before

In this scenario, when `some_user_input OUT` generates a string packet `"someAP"`, then connman and `sol-netctl` did know about "someAP".

The `"someAP"` packet is processed by `set_service_name` and stored.

In your code, that stops there. No packets will be ever emitted in `service STATE -> IN some_user_output`, since that is only done by a change in `service_status_cb()` or if the user manually send some packet to `TRIGGER service`.

With my proposal, inside `set_service_name()`, before you return you should call `get_service_state()`. In this case it would call `sol_netctl_get_services_list()` and note that `mdata->service_name` (new) is there, thus will emit the `service STATE -> IN some_user_output` that we need.

PR #2214, opened 2016-06-29
modules/thingspeak: some fixes

bdilly:


src/modules/flow/thingspeak/thingspeak.c
```diff
@@ -535,6 +536,7 @@ thingspeak_channel_update_open(struct sol_flow_node *node, void *data, const str
         return -ENOMEM;
     }
 
+    mdata->node = node;
```
barbieri: ok for now, but the best approach is to  remove `mdata->node` and instead of passing `mdata`, give `node` and then query the private data with `sol_flow_node_get_private_data()`.

bdilly: yeah, I know what you mean.
To be honest I believe the approach "mdata->node" is way more used in our code base than get_private_date.
We could change it at some point.

PR #2215, opened 2016-06-30
iio: Node type code refactor for macros

yongli3:
Combine these similar macros

### DIFF
--- a/src/modules/flow/iio/nodes.c
+++ b/src/modules/flow/iio/nodes.c
@@ -25,6 +25,15 @@
 
 #include <sol-iio.h>
 
+
+#define ADD_CHANNEL(_scale, _offset, _channel, _nodename, _error) \
+    if (!mdata->iio_base.use_device_default_scale) \
+        channel_config.scale = _scale; \
+    if (!mdata->iio_base.use_device_default_offset) \
+        channel_config.offset = _offset; \
+    _channel = sol_iio_add_channel(mdata->iio_base.device, _nodename, &channel_config); \
+    SOL_NULL_CHECK_GOTO(_channel, _error);
+
 struct iio_device_config {
     struct sol_iio_config config;
     struct sol_drange_spec out_range;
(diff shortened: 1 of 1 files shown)


yongli3:
@edersondisouza As discussed in PR #2102, I changed these macros, please help to review

src/modules/flow/iio/nodes.c
```diff
@@ -25,6 +25,23 @@
 
 #include <sol-iio.h>
 
+#undef ADD_CHANNEL_AXIS
```
cabelitos: I guess this is not necessary

src/modules/flow/iio/nodes.c
```diff
@@ -25,6 +25,23 @@
...
+#define ADD_CHANNEL_AXIS(_axis, _nodename) \
+    if (!mdata->iio_base.use_device_default_scale) \
+        channel_config.scale = mdata->scale._axis; \
+    if (!mdata->iio_base.use_device_default_offset) \
+        channel_config.offset = mdata->offset._axis; \
+    mdata->channel_ ## _axis = sol_iio_add_channel(mdata->iio_base.device, _nodename # _axis, &channel_config); \
+    SOL_NULL_CHECK_GOTO(mdata->channel_ ## _axis, error);
+
+#undef ADD_CHANNEL_VAL
```
cabelitos: Same thing

src/modules/flow/iio/nodes.c
yongli3: Done, thanks!

src/modules/flow/iio/nodes.c
```diff
@@ -25,6 +25,21 @@
...
+        channel_config.offset = mdata->offset._axis; \
+    mdata->channel_ ## _axis = sol_iio_add_channel(mdata->iio_base.device, _nodename # _axis, &channel_config); \
+    SOL_NULL_CHECK_GOTO(mdata->channel_ ## _axis, error);
+
+#define ADD_CHANNEL_VAL(_val, _nodename) \
+    if (!mdata->iio_base.use_device_default_scale) \
+        channel_config.scale = mdata->scale; \
+    if (!mdata->iio_base.use_device_default_offset) \
+        channel_config.offset = mdata->offset; \
+    mdata->channel_ ## _val = sol_iio_add_channel(mdata->iio_base.device, _nodename, &channel_config);
```
edersondisouza: Guess this `val` is unnecessary: it's always `val`. You could simply call  `ADD_CHANNEL` and use `mdata->channel_val` directly.

edersondisouza: Why no check of `sol_iio_add_channel` result?

src/modules/flow/iio/nodes.c
```diff
@@ -25,6 +25,21 @@
 
 #include <sol-iio.h>
 
+#define ADD_CHANNEL_AXIS(_axis, _nodename) \
```
edersondisouza: Guess that you could also use this for color node type, couldn't you? It would be good to think of a generic name for `axis` and `colour`, though =D

yongli3:
@edersondisouza I combined all these macros into one, please help to review

src/modules/flow/iio/nodes.c
```diff
@@ -817,17 +756,9 @@ adc_create_channels(struct iio_double_data *mdata, int device_id)
...
-    if (!mdata->iio_base.use_device_default_offset) \
-        channel_config.offset = mdata->offset; \
-    mdata->channel_ ## _val = sol_iio_add_channel(mdata->iio_base.device, "in_voltage0", &channel_config); \
-    SOL_NULL_CHECK_GOTO(mdata->channel_ ## _val, error);
+    ADD_CHANNEL(mdata->scale, mdata->offset, mdata->channel_val, "in_voltage0", error);
 
-    ADD_CHANNEL(val);
-
-#undef ADD_CHANNEL
+    SOL_NULL_CHECK_GOTO(mdata->channel_val, error);
```
edersondisouza: I guess you don't need this check, as `ADD_CHANNEL` macro already does that.

edersondisouza:
I like the way it is now and would merge it once minor comment is fixed. However, I think you could change `ADD_CHANNEL` to an inline function returning the newly added channel. Then, error checking would be left to the caller. Reasons: eases error handling and caller doesn't have to declare a `sol_iio_channel_config` that it's not seem to be used (it is, hidden inside the macro). What do you think?

src/modules/flow/iio/nodes.c
yongli3: Thanks!, done

yongli3:
@edersondisouza I agree with you that we can change the macro to an inline function. Actually I found we can refactor these create_channels functions too.  Is it okay to merge this PR first, then I create another PR for the new change?

edersondisouza:
Sure, it is =D

edersondisouza:
Integrated, tks!

PR #2216, opened 2016-06-30
make sol-oic-gen quiet and fix build

bdilly:
@cabelitos, ptal ^


cabelitos:
The build was breaking because of sol_memdup?

cabelitos:
anyway +1

bdilly:
@cabelitos it was failing to build MQTT node type.

PR #2225, opened 2016-07-07
Add Netctl node type and samples for network control V3

guchaojie:
Add netctl fbp node and samples include list services, connect or disconnect service, get service state, turn on or off radios


guchaojie:
@barbieri @bdilly @vcgomes @wzhen12
According to previous PR V2 comments to update PR, please have a review.

src/lib/comms/include/sol-netctl.h
```diff
@@ -498,6 +498,35 @@ enum sol_netctl_service_state sol_netctl_service_state_from_str(const char *stat
...
+ */
+static inline struct sol_netctl_service *
+sol_netctl_find_service_by_name(const char *service_name)
+{
+    const struct sol_ptr_vector *service_list;
+    struct sol_netctl_service *service;
+    const char *name;
+    uint16_t i;
+
+    service_list = sol_netctl_get_services();
```
barbieri: it's better to `if (!service_list) return NULL;` to be safe.

src/lib/comms/include/sol-netctl.h
```diff
@@ -498,6 +498,35 @@ enum sol_netctl_service_state sol_netctl_service_state_from_str(const char *stat
...
+
+    SOL_PTR_VECTOR_FOREACH_IDX (service_list, service, i) {
+        name = sol_netctl_service_get_name(service);
+        if (name && strcmp(name, service_name) == 0)
```
barbieri: using `strcmp()` implies you need to `#include <string.h>` in this file so it will work in all cases.

src/modules/flow/netctl/netctl.c
```diff
@@ -0,0 +1,327 @@
...
+}
+
+static int
+open_list_services(struct sol_flow_node *node, void *data, const struct sol_flow_node_options *options)
+{
+    int r;
+
+    r = sol_netctl_add_service_monitor(service_list_cb, node);
+    SOL_INT_CHECK(r, < 0, r);
+
```
barbieri: hum... I just noticed in the previous reviews I forgot to say that here you should do an initial batch of `ADDED` for existing services:

```c
service_list = sol_netctl_get_services();
if (!service_list)
    return NULL;

SOL_PTR_VECTOR_FOREACH_IDX (service_list, service, i) {
    enum sol_netctl_service_state state = sol_netctl_service_get_state(service);
    const char *name;

    if (state == SOL_NETCTL_SERVICE_STATE_UNKNOWN ||
        state == SOL_NETCTL_SERVICE_STATE_REMOVE)
        continue;

    name = sol_netctl_service_get_name(service);
    r = sol_flow_send_string_packet(node,
        SOL_FLOW_NODE_TYPE_NETCTL_LIST_SERVICES__OUT__ADDED,
        name);
    if (r < 0)
        SOL_WRN("Failed to send added service name: %s", name);
}
```

src/modules/flow/netctl/netctl.c
```diff
@@ -0,0 +1,327 @@
...
+    struct sol_flow_node *node = data;
+    struct network_service_data *mdata = sol_flow_node_get_private_data(node);
+    const char *name;
+
+    name = sol_netctl_service_get_name(service);
+
+    if (!name || strcmp(name, mdata->service_name))
+        return;
+
+    sol_flow_send_error_packet(node, -error, "Service %s error is: %s",
```
barbieri: `error` is already unsigned int... no need to negate it.

src/modules/flow/netctl/netctl.c
```diff
@@ -0,0 +1,327 @@
...
+        name, sol_util_strerrora(-error));
```
barbieri: same here, error is already positive.

src/modules/flow/netctl/netctl.c
```diff
@@ -0,0 +1,327 @@
...
+    if (!service) {
+        sol_flow_send_error_packet(node, ENOENT, "Did not found service name %s", mdata->service_name);
+        return 0;
+    }
+
+    r = sol_netctl_service_connect(service);
+    if (r < 0)
+        sol_flow_send_error_packet(node, -r, "Could not connect to service: %s: %s",
+            mdata->service_name, sol_util_strerrora(-r));
+
```
barbieri: I added few comments to previous PR, check them there, but here you should `mdata->pending = CONNECT` if no service exists. Then you can avoid sending the error packet and it's more usable to the user.

src/modules/flow/netctl/netctl.c
```diff
@@ -0,0 +1,327 @@
...
+{
+    struct network_service_data *mdata = data;
+    struct sol_netctl_service *service;
+    int r;
+
+    service = sol_netctl_find_service_by_name(mdata->service_name);
+    if (!service) {
+        sol_flow_send_error_packet(node, ENOENT, "Did not found service name %s", mdata->service_name);
+        return 0;
+    }
```
barbieri: I added few comments to previous PR, check them there, but here you should `mdata->pending = DISCONNECT` if no service exists. Then you can avoid sending the error packet and it's more usable to the user.

src/modules/flow/netctl/netctl.c
```diff
@@ -0,0 +1,327 @@
...
+    r = sol_netctl_service_disconnect(service);
+    if (r < 0)
+        sol_flow_send_error_packet(node, -r, "Could not disconnect to service: %s: %s",
+            mdata->service_name, sol_util_strerrora(-r));
+
+    return 0;
+}
+
+static int
+state_dispatch(struct sol_flow_node *node, void *data)
```
barbieri: please change the second parameter to be `struct network_service_data *mdata`.

src/modules/flow/netctl/netctl.c
```diff
@@ -0,0 +1,327 @@
...
+    SOL_INT_CHECK(r, < 0, r);
+
+    if (mdata->state == SOL_NETCTL_SERVICE_STATE_ONLINE)
+        r = sol_flow_send_bool_packet(node,
+            SOL_FLOW_NODE_TYPE_NETCTL_SERVICE__OUT__ONLINE,
+            true);
+    else
+        r = sol_flow_send_bool_packet(node,
+            SOL_FLOW_NODE_TYPE_NETCTL_SERVICE__OUT__ONLINE,
+            false);
```
barbieri: why not a single `send_bool_packet()` with the last argument being: `mdata->state == SOL_NETCTL_SERVICE_STATE_ONLINE`? -

src/modules/flow/netctl/netctl.c
```diff
@@ -0,0 +1,327 @@
...
+{
+    struct network_service_data *mdata = data;
+    const char *in_value;
+    int r;
+
+    r = sol_flow_packet_get_string(packet, &in_value);
+    SOL_INT_CHECK(r, < 0, r);
+
+    sol_util_replace_str_if_changed(&mdata->service_name, in_value);
+    SOL_NULL_CHECK(mdata->service_name, -ENOMEM);
```
barbieri: `sol_util_replace_str_if_changed()` already returns error, you can use that. Also note that if `return == 0`, then it did not change, so no need to process the rest of this function.

barbieri: and as I wrote in the previous PR (check there), you should check `mdata->pending` here and do any pending connection/disconnection request. remember to reset `mdata->pending = NONE` afterwards.

src/modules/flow/netctl/netctl.c
```diff
@@ -0,0 +1,327 @@
...
+    return 0;
+}
+
+static int
+open_network_radio(struct sol_flow_node *node, void *data, const struct sol_flow_node_options *options)
+{
+    int r;
+
+    r = sol_netctl_add_manager_monitor(manager_cb, node);
+    SOL_INT_CHECK(r, < 0, r);
```
barbieri: like I wrote for other node type, here you should call `manager_cb()` so it will report the current status (online/offline).

src/samples/flow/netctl/network-connect.fbp
```diff
@@ -0,0 +1,41 @@
...
+timer(timer:interval=3000)
+btoe(converter/boolean-to-empty)
+etoi(converter/empty-to-int:output_value=0)
+
+name ERROR -> IN error-msg OUT -> IN Error
+service ERROR -> ERROR quit
+
+name OUT -> NAME service
+tick OUT -> CONNECT service OUT -> IN _(console:prefix="state: ")
+timer OUT -> TRIGGER service ONLINE -> PULSE_IF_TRUE btoe OUT -> IN etoi OUT -> CODE quit
```
barbieri: why do you need this `timer OUT -> TRIGGER service`?  You can drop it and `timer(timer:interval=3000)`.

barbieri:
we're almost there! See my latest comments to the previous PR and the ones in this one, I guess we can merge in the next round.

PR #2227, opened 2016-07-07
Build: Avoid race condition when generation OIC node types.

cabelitos:
Create rules different rules for the json and C file and generate them
separadly.

### DIFF
--- a/data/scripts/sol-oic-gen.py
+++ b/data/scripts/sol-oic-gen.py
@@ -1784,16 +1784,14 @@ if __name__ == '__main__':
                         help="Path to store the master JSON with node type information",
                         required=True)
     parser.add_argument("--node-type-impl",
-                        help="Path to store the node type implementation",
-                        required=True)
+                        help="Path to store the node type implementation")
     parser.add_argument("--node-type-gen-c",
                         help="Relative path to source generated with "
-                        "sol-flow-node-type-gen for inclusion purposes.",
-                        required=True)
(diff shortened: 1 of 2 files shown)


barbieri:
-0, I dislike this option and I think it's just slower. The "-0" instead of "-1" is to check if you tried my suggestion and what did not work.

My suggestion is to force some dependency, instead of:
```
file.json file.c file.h: generator inputs
     sol-oic-gen.py ...
```

Which expands to:
```
file.json: generator inputs
     sol-oic-gen.py ...
file.c: generator inputs
     sol-oic-gen.py ...
file.h: generator inputs
     sol-oic-gen.py ...
```

that is prone to parallel execution of each rule, which would override the other files, producing errors (ie: file.json generation would run in parallel with file.c, but both of them would generate all the files.

We could introduce some dependencies:
```
file.json: generator inputs
     sol-oic-gen.py ... file.c file.json file.h

file.c: file.json
file.h: file.json
```

This way, when we need `file.c` it would need to process `file.json`, which would generate `file.c`. You can evaluate this with some smaller `Makefile` and explicitly written rules so you can run make with debug and see the actual rules and dependencies. Maybe you just need the order, in that case you write after `|` (bar):

```
file.json: generator inputs
     sol-oic-gen.py ... file.c file.json file.h

file.c: | file.json
file.h: | file.json
```

cabelitos:
@barbieri Indeed your solution is better. I will send a new PR.

cabelitos:
#2228

PR #2229, opened 2016-07-11
I2C: Use lowercase chars for I2C device numbers.

cabelitos:
Soletta should match the Linux I2C name style. This will prevent
some problems that can occur when trying to use the I2C device via IIO.

### DIFF
--- a/src/lib/io/sol-i2c-impl-linux.c
+++ b/src/lib/io/sol-i2c-impl-linux.c
@@ -1060,7 +1060,7 @@ create_device_iter_cb(void *data, const char *dir_path, struct dirent *ent)
                 sol_util_strerrora(errno));
         }
 
-        r = snprintf(path, PATH_MAX, "%s/%s/%s-00%X",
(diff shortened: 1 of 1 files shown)


edersondisouza:
:+1:

bdilly:
ok

have you checked if it's guaranteed to be lowercase?

edersondisouza:
On 11-07-2016 20:16, Bruno Dilly wrote:
> ok
>
> have you checked if it's guaranteed to be lowercase?
>
> —
> You are receiving this because you commented.
> Reply to this email directly, view it on GitHub
> <https://github.com/solettaproject/soletta/pull/2229#issuecomment-231893635>,
> or mute the thread
> <https://github.com/notifications/unsubscribe/ADB8T_5X7VReHNZqfNINJfab_bTpeX6rks5qUs7ZgaJpZM4JJw3M>.
>

Yes, we checked kernel code and as far as we understood, device is 
created with lower case: 
https://github.com/torvalds/linux/blob/master/drivers/i2c/i2c-core.c#L1002

PR #2230, opened 2016-07-11
Zephyr implementation missing functions

bdilly:
Required to have samples working on this platform.

### DIFF
--- a/src/lib/comms/sol-network-impl-zephyr.c
+++ b/src/lib/comms/sol-network-impl-zephyr.c
@@ -169,6 +169,7 @@ sol_network_subscribe_events(void (*cb)(void *data, const struct sol_network_lin
     enum sol_network_event event),
     const void *data)
 {
+    SOL_WRN("Not implemented");
     return -ENOSYS;
 }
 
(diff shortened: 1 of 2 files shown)


src/lib/comms/sol-network-impl-zephyr.c
```diff
@@ -191,3 +191,33 @@ sol_network_link_get_name(const struct sol_network_link *link)
 {
     return NULL;
 }
+
+SOL_API int
+sol_network_hostname_pending_cancel(
+    struct sol_network_hostname_pending *handle)
+{
+    return -ENOTSUP;
```
cabelitos: Maybe we should return -ENOSYS here as well, just like the other functions

cabelitos: And also add a warn

src/lib/comms/sol-network-impl-zephyr.c
```diff
@@ -191,3 +191,33 @@ sol_network_link_get_name(const struct sol_network_link *link)
...
+    return -ENOSYS;
+}
+
+SOL_API struct sol_network_hostname_pending *
+sol_network_get_hostname_address_info(const struct sol_str_slice hostname,
+    enum sol_network_family family, void (*host_info_cb)(void *data,
+    const struct sol_str_slice host, const struct sol_vector *addrs_list),
+    const void *data)
+{
+    return NULL;
```
cabelitos: We should add a warn in here

bdilly: We could do it, at least a INF.
I was following the RIOT implementation and it's used by some samples. So we're kind of expecting it to fail, so not so sure if it should warn.

bdilly:
merged with proposed changes

PR #2239, opened 2016-07-14
WIP: LWM2M: Access Control

bsmelo:
This patch introduces:
- Access Control Object Management
 - Instantiation
   - Bootstrap case
    - 'Create' Operation case
 - Update
   - Write (modifying access rights)
    - Delete (removing associated Access Control Object)
- Authorization
 - `is_authorized()` is called everytime a 'C'reate, 'R'ead, 'W'rite, 'D'elete, or 'E'xecute operation takes place, to check for the required access right
 - Full Support: Create; Read; Write; Delete; Execute operations
 - Partial Support: Observe and Notify operations (both use the 'R'ead bit)

Also the `sol_lwm2m_resource` API had to be improved to add support to:
- Resource Instance IDs
- Init of a Resource of type `SOL_LWM2M_RESOURCE_TYPE_MULTIPLE` with a single Resource Instance
- A way to init a Resource when we don't know beforehand the amount of Resource Instances desired: `sol_lwm2m_resource_init_vector()`.

TODO:
- [x] Fixing a SEGFAULT when adding the code to delete the connection with the Bootstrap Server after Bootstrap Finish is received (https://github.com/bsmelo/soletta/commit/1bf776c964dbb936c924dbaf867a04a35c4900e9#diff-e4058dc14a343a8aceb7210b323358a2R4816)
- [ ] Figuring out how to deal with Notify operation triggered by the client API (https://github.com/bsmelo/soletta/commit/f94f22ca0f59f25f048ba38ee77669b84fee3a83#diff-e4058dc14a343a8aceb7210b323358a2R5495)
- [ ] Observe operation may be related to the issue above, and may require a fix in `sol-coap`.

### DIFF
--- a/src/lib/comms/include/sol-lwm2m.h
+++ b/src/lib/comms/include/sol-lwm2m.h
@@ -52,11 +52,9 @@ extern "C" {
  * - TLV format.
  *
  * Unsupported features for now:
- * - Bootstrap.
  * - LWM2M JSON.
  * - Queue Mode operation (only 'U' is supported for now).
  * - Data encryption.
- * - Access rights.
  *
  * @{
  */
@@ -374,6 +372,42 @@ struct sol_lwm2m_resource {
             bool b;
         } content;
     } *data;
+};
+
+/**
+ * @brief Enum that represents the Access Control Rights.
+ *
+ * Setting each bit means the LWM2M Server has the access right for that operation.
+ */
+enum sol_lwm2m_acl_rights {
+    /**
+     * No bit is set (No access rights for any operation)
+     */
+    SOL_LWM2M_ACL_NONE = (0),
+    /**
+     * 1st lsb: R (Read, Observe, Discover, Write Attributes)
+     */
+    SOL_LWM2M_ACL_READ = (1),
+    /**
+     * 2nd lsb: W (Write)
+     */
+    SOL_LWM2M_ACL_WRITE = (2),
+    /**
+     * 3rd lsb: E (Execute)
+     */
+    SOL_LWM2M_ACL_EXECUTE = (4),
+    /**
+     * 4th lsb: D (Delete)
+     */
+    SOL_LWM2M_ACL_DELETE = (8),
+    /**
+     * 5th lsb: C (Create)
+     */
+    SOL_LWM2M_ACL_CREATE = (16),
+    /**
+     * All 5 lsbs: Full Access Rights
+     */
+    SOL_LWM2M_ACL_ALL = (31)
 };
 
 /**
--- a/src/lib/comms/include/sol-lwm2m.h
+++ b/src/lib/comms/include/sol-lwm2m.h
@@ -361,15 +361,18 @@ struct sol_lwm2m_resource {
     /** @brief The resource data array len */
     uint16_t data_len;
     /** @brief The resource data array */
-    union sol_lwm2m_resource_data {
-        /** @brief The resource is opaque or an string */
-        struct sol_blob *blob;
-        /** @brief The resource is a integer value */
-        int64_t integer;
-        /** @brief The resource is a float value */
-        double fp;
-        /** @brief The resource is a bool value */
-        bool b;
+    struct sol_lwm2m_resource_data {
+        uint16_t id;
+        union {
+            /** @brief The resource is opaque or an string */
+            struct sol_blob *blob;
+            /** @brief The resource is a integer value */
+            int64_t integer;
+            /** @brief The resource is a float value */
+            double fp;
+            /** @brief The resource is a bool value */
+            bool b;
+        } content;
     } *data;
 };
 
@@ -382,15 +385,33 @@ struct sol_lwm2m_resource {
  * @param ret_value_ The return value of sol_lwm2m_resource_init()
  * @param resource_ The resource to be initialized.
  * @param id_ The resource id.
- * @param data_type_ The resource data type.
+ * @param type_ The resource type.
  * @param resource_len_ The resource data size.
+ * @param data_type_ The resource data type.
  * @param ... The LWM2M resource data, respecting the table according to the resource type.
  * @see sol_lwm2m_resource_init()
  */
-#define SOL_LWM2M_RESOURCE_INIT(ret_value_, resource_, id_, resource_len_, data_type_, ...) \
+#define SOL_LWM2M_RESOURCE_INIT(ret_value_, resource_, id_, type_, resource_len_, data_type_, ...) \
     do { \
         SOL_SET_API_VERSION((resource_)->api_version = SOL_LWM2M_RESOURCE_API_VERSION; ) \
-        (ret_value_) = sol_lwm2m_resource_init((resource_), (id_), (resource_len_), (data_type_), __VA_ARGS__); \
+        (ret_value_) = sol_lwm2m_resource_init((resource_), (id_), (type_), (resource_len_), (data_type_), __VA_ARGS__); \
+    } while (0)
+
+/**
+ * @brief A helper macro to init SINGLE resources.
+ *
+ *
+ * @param ret_value_ The return value of sol_lwm2m_resource_init()
+ * @param resource_ The resource to be initialized.
+ * @param id_ The resource id.
+ * @param data_type_ The resource data type.
+ * @param value_ The LWM2M resource data, respecting the table according to the resource type.
+ * @see sol_lwm2m_resource_init()
+ */
+#define SOL_LWM2M_RESOURCE_SINGLE_INIT(ret_value_, resource_, id_, data_type_, value_) \
+    do { \
+        SOL_SET_API_VERSION((resource_)->api_version = SOL_LWM2M_RESOURCE_API_VERSION; ) \
+        (ret_value_) = sol_lwm2m_resource_init((resource_), (id_), SOL_LWM2M_RESOURCE_TYPE_SINGLE, 1, (data_type_), value_); \
     } while (0)
 
 /**
@@ -420,7 +441,7 @@ struct sol_lwm2m_resource {
 #define SOL_LWM2M_RESOURCE_INT_INIT(ret_value_, resource_, id_, value_) \
     do { \
         SOL_SET_API_VERSION((resource_)->api_version = SOL_LWM2M_RESOURCE_API_VERSION; ) \
-        (ret_value_) = sol_lwm2m_resource_init((resource_), (id_), 1, SOL_LWM2M_RESOURCE_DATA_TYPE_INT, SOL_TYPE_CHECK(int64_t, (value_))); \
+        (ret_value_) = sol_lwm2m_resource_init((resource_), (id_), SOL_LWM2M_RESOURCE_TYPE_SINGLE, 1, SOL_LWM2M_RESOURCE_DATA_TYPE_INT, SOL_TYPE_CHECK(int64_t, (value_))); \
     } while (0)
 
 
@@ -722,8 +743,9 @@ void sol_lwm2m_resource_clear(struct sol_lwm2m_resource *resource);
  *
  * @param resource The resource to be initialized.
  * @param id The resource id.
- * @param data_type The resource data type.
+ * @param type The resource type.
  * @param resource_len The resource data size.
+ * @param data_type The resource data type.
  * @param ... The LWM2M resource data, respecting the table according to the resource data type.
  * @return 0 on success, negative errno on error.
  * @see sol_lwm2m_resource_clear()
@@ -733,8 +755,32 @@ void sol_lwm2m_resource_clear(struct sol_lwm2m_resource *resource);
  * before calling this function.
  */
 int sol_lwm2m_resource_init(struct sol_lwm2m_resource *resource,
-    uint16_t id, uint16_t resource_len,
+    uint16_t id, enum sol_lwm2m_resource_type type, uint16_t resource_len,
     enum sol_lwm2m_resource_data_type data_type, ...);
+
+/**
+ * @brief Initializes a LWM2M resource of type multiple using two @c sol_vector.
+ *
+ * This function makes it easier to init a LWM2M resource of type multiple,
+ * when we don't know beforehand the amount of Resource Instances desired.
+ * The last argument is a @c sol_vector holding elements of the relevant
+ * resource data type as per the table in @see sol_lwm2m_resource_init().
+ *
+ * @param resource The resource to be initialized.
+ * @param id The resource id.
+ * @param data_type The resource data type.
+ * @param res_inst_ids The list of Resource Instance IDs, in the same order as their relevant values in @c res_inst_values
+ * @param res_inst_values The list of Resource Instance values, in the same order as their relevant ids in @c res_inst_ids, respecting the table according to the resource data type.
+ * @return 0 on success, negative errno on error.
+ * @see sol_lwm2m_resource_clear()
+ * @see sol_lwm2m_resource_init()
+ *
+ * @note The LWM2M resource api_version must be set,
+ * before calling this function.
+ */
+int sol_lwm2m_resource_init_vector(struct sol_lwm2m_resource *resource,
+    uint16_t id, enum sol_lwm2m_resource_data_type data_type,
+    struct sol_vector *res_inst_ids, struct sol_vector *res_inst_values);
 
 /**
  * @brief Creates a new LWM2M bootstrap server.
--- a/src/lib/comms/sol-lwm2m.c
+++ b/src/lib/comms/sol-lwm2m.c
@@ -2047,7 +2047,7 @@ observation_request_reply(void *data, struct sol_coap_server *coap_server,
             return false;
         }
         SOL_WRN("Could not complete the observation request on client:%s"
-            " path:%s", entry->path, entry->cinfo->name);
+            " path:%s", entry->cinfo->name, entry->path);
         keep_alive = false;
     } else {
         extract_content(req, &code, &type, &content);
--- a/src/lib/comms/sol-lwm2m.c
+++ b/src/lib/comms/sol-lwm2m.c
@@ -3014,9 +3014,106 @@ clear_bootstrap_ctx(struct sol_lwm2m_client *client)
     }
 }
 
+static bool
+is_authorized(struct sol_lwm2m_client *client,
+    int64_t server_id, uint16_t obj_id,
+    int32_t instance_id, int64_t rights_needed)
+{
+    struct obj_ctx *obj_ctx;
+    struct obj_instance *obj_instance;
+    struct sol_lwm2m_resource res[2];
+    int r;
+    bool ret = false;
+    int64_t default_acl = SOL_LWM2M_ACL_NONE;
+    uint16_t i;
+
+    SOL_WRN("connections.len=%" PRIu16, client->connections.len);
+
+    //If only one server or Bootstrap Server ID, then full access rights
+    if (client->connections.len == 1 || server_id == UINT16_MAX)
+        return true;
+
+    SOL_WRN("server_id=%" PRId64 " obj=/%" PRIu16 "/%" PRId32 " needed=%" PRId64,
+        server_id, obj_id, instance_id, rights_needed);
+
+    obj_ctx = find_object_ctx_by_id(client, ACCESS_CONTROL_OBJECT_ID);
+    SOL_VECTOR_FOREACH_IDX (&obj_ctx->instances, obj_instance, i) {
+        r = read_resources(client, obj_ctx, obj_instance, res, 2,
+            ACCESS_CONTROL_OBJECT_OBJECT_RES_ID,
+            ACCESS_CONTROL_OBJECT_INSTANCE_RES_ID);
+        if (r < 0) {
+            SOL_WRN("Could not read Access Control Object's"
+                " [Object ID] and [Instance ID] resources\n");
+            continue;
+        }
+
+        //Retrieve the associated Access Control Object Instance, by matching Object ID
+        // and Instance ID, or if instance_id == -1 and 'R'ead is needed, this
+        // is an Observe Request on Object level and any Instance with this access right
+        // is enough to reply with the Observe falg on CoAP
+        if ((res[0].data[0].content.integer == obj_id &&
+            res[1].data[0].content.integer == instance_id) ||
+            (res[0].data[0].content.integer == obj_id &&
+            instance_id == -1 && (rights_needed & SOL_LWM2M_ACL_READ))) {
+            clear_resource_array(res, sol_util_array_size(res));
+
+            r = read_resources(client, obj_ctx, obj_instance, res, 2,
+                ACCESS_CONTROL_OBJECT_ACL_RES_ID,
+                ACCESS_CONTROL_OBJECT_OWNER_RES_ID);
+            if (r < 0) {
+                SOL_WRN("Could not read Access Control"
+                    " Object's [ACL] and [Owner ID] resources\n");
+                continue;
+            }
+
+            //Retrive this server's ACL Resource Instance
+            for (i = 0; i < res[0].data_len; i++) {
+                if (res[0].data[i].id == server_id) {
+                    if (res[0].data[i].content.integer & rights_needed) {
+                        ret = true;
+                        goto exit;
+                    } else {
+                        ret = false;
+                        goto exit;
+                    }
+                }
+
+                //Keep the default ACL Resource Instance, if any, to save another loop later
+                if (res[0].data[i].id == DEFAULT_SHORT_SERVER_ID)
+                    default_acl = res[0].data[i].content.integer;
+            }
+
+            //If no ACL for this server, check if it is the owner of the object.
+            // If owner and no specific ACL Resource Instance, then full access rights.
+            if (res[1].data[0].content.integer == server_id) {
+                ret = true;
+                goto exit;
+            }
+
+            //If no ACL and not owner, check if the default ACL Resource Instance applies
+            if (default_acl & rights_needed) {
+                ret = true;
+                goto exit;
+            }
+
+            //If Observe operation on Object level, check next instance
+            if (!(instance_id == -1 && (rights_needed & SOL_LWM2M_ACL_READ)))
+                goto exit;
+        }
+
+        clear_resource_array(res, sol_util_array_size(res));
+    }
+
+exit:
+    clear_resource_array(res, sol_util_array_size(res));
+
+    return ret;
+}
+
 static uint8_t
 handle_delete(struct sol_lwm2m_client *client,
-    struct obj_ctx *obj_ctx, struct obj_instance *obj_instance)
+    struct obj_ctx *obj_ctx, struct obj_instance *obj_instance,
+    int64_t server_id)
 {
     int r, ret = SOL_COAP_RESPONSE_CODE_BAD_REQUEST;
     uint16_t i;
@@ -3025,6 +3122,15 @@ handle_delete(struct sol_lwm2m_client *client,
 
     //Specific Instance?
     if (obj_ctx && obj_instance) {
+        if (client->supports_access_control &&
+            !is_authorized(client, server_id, obj_ctx->obj->id,
+            obj_instance->id, SOL_LWM2M_ACL_DELETE)) {
+            SOL_WRN("Server ID %" PRId64 " is not authorized for D on"
+                " Object Instance /%" PRIu16 "/%" PRIu16,
+                server_id, obj_ctx->obj->id, obj_instance->id);
+            return SOL_COAP_RESPONSE_CODE_UNAUTHORIZED;
+        }
+
         if (!obj_ctx->obj->del) {
             SOL_WRN("The object %" PRIu16 " does not implement the delete method",
                 obj_ctx->obj->id);
@@ -3161,7 +3267,8 @@ is_valid_args(const struct sol_str_slice args)
 static uint8_t
 handle_execute(struct sol_lwm2m_client *client,
     struct obj_ctx *obj_ctx, struct obj_instance *obj_instance,
-    uint16_t resource, struct sol_lwm2m_payload payload)
+    uint16_t resource, struct sol_lwm2m_payload payload,
+    int64_t server_id)
 {
     int r;
 
@@ -3169,6 +3276,15 @@ handle_execute(struct sol_lwm2m_client *client,
         SOL_WRN("Object instance was not provided to execute the path"
             "/%" PRIu16 "/?/%" PRIu16 "", obj_ctx->obj->id, resource);
         return SOL_COAP_RESPONSE_CODE_BAD_REQUEST;
+    }
+
+    if (client->supports_access_control &&
+        !is_authorized(client, server_id, obj_ctx->obj->id,
+        obj_instance->id, SOL_LWM2M_ACL_EXECUTE)) {
+        SOL_WRN("Server ID %" PRId64 " is not authorized for E on"
+            " Object Instance /%" PRIu16 "/%" PRIu16,
+            server_id, obj_ctx->obj->id, obj_instance->id);
+        return SOL_COAP_RESPONSE_CODE_UNAUTHORIZED;
     }
 
     if (!obj_ctx->obj->execute) {
@@ -3205,21 +3321,31 @@ handle_execute(struct sol_lwm2m_client *client,
 static uint8_t
 write_instance_tlv_or_resource(struct sol_lwm2m_client *client,
     struct obj_ctx *obj_ctx, struct obj_instance *obj_instance,
-    int32_t resource, struct sol_lwm2m_payload payload)
+    int32_t resource, struct sol_lwm2m_payload payload,
+    int64_t server_id)
 {
     int r;
+
+    if (!obj_instance) {
+        SOL_WRN("Object instance was not provided."
+            " Can not complete the write operation");
+        return SOL_COAP_RESPONSE_CODE_BAD_REQUEST;
+    }
+
+    if (client->supports_access_control &&
+        !is_authorized(client, server_id, obj_ctx->obj->id,
+        obj_instance->id, SOL_LWM2M_ACL_WRITE)) {
+        SOL_WRN("Server ID %" PRId64 " is not authorized for W on"
+            " Object Instance /%" PRIu16 "/%" PRIu16,
+            server_id, obj_ctx->obj->id, obj_instance->id);
+        return SOL_COAP_RESPONSE_CODE_UNAUTHORIZED;
+    }
 
     //If write_resource is not NULL then write_tlv is guaranteed to be valid as well.
     if (!obj_ctx->obj->write_resource) {
         SOL_WRN("Object %" PRIu16 " does not support the write method",
             obj_ctx->obj->id);
         return SOL_COAP_RESPONSE_CODE_NOT_ALLOWED;
-    }
-
-    if (!obj_instance) {
-        SOL_WRN("Object instance was not provided."
-            " Can not complete the write operation");
-        return SOL_COAP_RESPONSE_CODE_BAD_REQUEST;
     }
 
     if (payload.type == SOL_LWM2M_CONTENT_TYPE_TLV) {
@@ -3268,6 +3394,14 @@ handle_create(struct sol_lwm2m_client *client,
 {
     int r;
     struct obj_instance *obj_instance;
+
+    if (client->supports_access_control &&
+        !is_authorized(client, owner_server_id, obj_ctx->obj->id,
+        UINT16_MAX, SOL_LWM2M_ACL_CREATE)) {
+        SOL_WRN("Server ID %" PRId64 " is not authorized for C on"
+            " Object /%" PRIu16, owner_server_id, obj_ctx->obj->id);
+        return SOL_COAP_RESPONSE_CODE_UNAUTHORIZED;
+    }
 
     if (!obj_ctx->obj->create) {
         SOL_WRN("Object %" PRIu16 " does not support the create method",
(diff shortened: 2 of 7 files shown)


cabelitos:
Segfault is fixed by #2242

src/lib/comms/include/sol-lwm2m.h
```diff
@@ -377,6 +375,42 @@ struct sol_lwm2m_resource {
...
 /**
+ * @brief Enum that represents the Access Control Rights.
+ *
+ * Setting each bit means the LWM2M Server has the access right for that operation.
+ */
+enum sol_lwm2m_acl_rights {
+    /**
+     * No bit is set (No access rights for any operation)
+     */
+    SOL_LWM2M_ACL_NONE = (0),
```
cabelitos: There's no need to use "()" in here.

src/lib/comms/sol-lwm2m.c
```diff
@@ -3051,6 +3072,22 @@ handle_delete(struct sol_lwm2m_client *client,
...
+    }
+
+    return ret;
+
+err_access_control:
+    if (client->supports_access_control) {
+        acc_obj_ctx = find_object_ctx_by_id(client, ACCESS_CONTROL_OBJECT_ID);
+        while ((acc_obj_instance = sol_vector_get(&acc_obj_ctx->instances, 0)) != NULL) {
+            obj_instance_clear(client, acc_obj_ctx, acc_obj_instance);
+            (void)sol_vector_del_element(&acc_obj_ctx->instances, acc_obj_instance);
```
cabelitos: remove (void)

src/lib/comms/sol-lwm2m.c
```diff
@@ -4730,9 +5037,18 @@ sol_lwm2m_client_start(struct sol_lwm2m_client *client)
     struct resource_ctx *res_ctx;
     struct sol_lwm2m_resource res[3];
     int r;
+    static bool setup_access_control = true;
```
cabelitos: what happens if you have two clients? This logic will fail.

src/lib/comms/sol-lwm2m.c
```diff
@@ -3606,6 +3660,27 @@ should_dispatch_notifications(uint8_t code, bool is_execute)
...
+get_server_id_by_link_addr(const struct sol_vector *connections,
+    const struct sol_network_link_addr *cliaddr)
+{
+    struct server_conn_ctx *conn_ctx;
+    struct sol_network_link_addr *server_addr;
+    uint16_t i;
+//    SOL_BUFFER_DECLARE_STATIC(addr, SOL_NETWORK_INET_ADDR_STR_LEN);
+
+//    SOL_WRN("GET_SERVER_ID:\ncliaddr=%s", sol_network_link_addr_to_str(cliaddr, &addr));
+    SOL_VECTOR_FOREACH_IDX (connections, conn_ctx, i)
```
cabelitos: Do not leave commented code in here. Remove it or uncomment it.

src/lib/comms/sol-lwm2m.c
```diff
@@ -3606,6 +3660,27 @@ should_dispatch_notifications(uint8_t code, bool is_execute)
     return false;
 }
 
+static int64_t
+get_server_id_by_link_addr(const struct sol_vector *connections,
```
cabelitos: Change this function. you should return an int here. 0 if you find the server or -ENOENT if the server was not found.
than you change the signature to.
```C
int get_server_id_by_link_addr(const struct sol_vector *connections,  const struct sol_network_link_addr *cliaddr, int64_t *TheServerIdToBeFilled);
```

Then insde the function you do 
```C
*TheServerIdToBeFilled = conn_ctx->server_id;
return 0;
```

Please rename *TheServerIdToBeFilled to something else :]

src/lib/comms/sol-lwm2m.c
```diff
@@ -3305,7 +3358,8 @@ handle_write(struct sol_lwm2m_client *client,
                     r = write_instance_tlv_or_resource(client, obj_ctx, obj_instance, -1,
                         instance_payload);
                 } else {
-                    r = handle_create(client, obj_ctx, instance_tlvs->id, instance_payload);
+                    r = handle_create(client, obj_ctx, instance_tlvs->id,
+                        instance_payload, UINT16_MAX, false);
```
cabelitos: Worth mention that this only happen during bootstrap

src/lib/comms/sol-lwm2m.c
```diff
@@ -3337,7 +3391,7 @@ handle_write(struct sol_lwm2m_client *client,
             if (obj_instance) {
                 r = write_instance_tlv_or_resource(client, obj_ctx, obj_instance, -1, payload);
             } else {
-                r = handle_create(client, obj_ctx, path[1], payload);
+                r = handle_create(client, obj_ctx, path[1], payload, UINT16_MAX, false);
```
cabelitos: same thing

src/lib/comms/sol-lwm2m.c
```diff
@@ -3822,6 +3939,7 @@ sol_lwm2m_client_new(const char *name, const char *path, const char *sms,
 
     sol_vector_init(&client->objects, sizeof(struct obj_ctx));
     sol_vector_init(&client->connections, sizeof(struct server_conn_ctx));
+    client->supports_access_control = false;
```
cabelitos: No need, it's already false. `calloc()` handles that for us

src/lib/comms/sol-lwm2m.c
```diff
@@ -3976,6 +4097,26 @@ sol_lwm2m_client_add_object_instance(struct sol_lwm2m_client *client,
...
+        r = setup_access_control_object_instance_for_instance(client,
+            ctx->obj->id, instance->id, UINT16_MAX, NULL, false);
+
+        if (r == 0) {
+            SOL_DBG("Security Object Instance and Access Control Object Instance "
+                " does not need an Access Control Object Instance nor ACLs");
+        } else if (r != SOL_COAP_RESPONSE_CODE_CHANGED && r != SOL_COAP_RESPONSE_CODE_CREATED) {
+            SOL_WRN("Failed to create Access Control Object Instance for Object /%"
+                PRIu16 "/%" PRIu16, ctx->obj->id, instance->id);
+            return -ENOENT;
```
cabelitos: -ENOENT, we should change the return code in here. ANd we must also remove the instance that was added.

src/lib/comms/sol-lwm2m.c
```diff
@@ -4445,7 +4586,7 @@ server_connection_ctx_new(struct sol_lwm2m_client *client,
     SOL_DBG("Fetching hostname info for:%.*s", SOL_STR_SLICE_PRINT(str_addr));
     conn_ctx->hostname_handle =
         sol_network_get_hostname_address_info(uri.host,
-        SOL_NETWORK_FAMILY_UNSPEC, hostname_ready, conn_ctx);
+        SOL_NETWORK_FAMILY_UNSPEC, hostname_ready, conn_ctx); //TODO: INET6
```
cabelitos: remove this todo

src/lib/comms/sol-lwm2m.c
```diff
@@ -4718,6 +4859,172 @@ client_bootstrap(void *data)
...
+    acc_obj_ctx = find_object_ctx_by_id(client, ACCESS_CONTROL_OBJECT_ID);
+
+    //Only the Client itself and the owner server is able to manage
+    // Access Control Object Instances, and Security Object Instances
+    // can only be managed by Bootstrap Servers, so there's no sense in
+    // creating Access Control Object Instances or ACLs for them
+    if (target_object_id == SECURITY_SERVER_OBJECT_ID ||
+        target_object_id == ACCESS_CONTROL_OBJECT_ID) {
+        if (acl_res)
+            sol_lwm2m_resource_clear(acl_res);
```
cabelitos: this is bad. Let the calling function clear the resource array. Not this function

src/lib/comms/sol-lwm2m.c
```diff
@@ -4718,6 +4859,172 @@ client_bootstrap(void *data)
...
+    struct sol_lwm2m_resource *res;
+    struct sol_lwm2m_payload payload = { .type = SOL_LWM2M_CONTENT_TYPE_TLV };
+
+    acc_obj_ctx = find_object_ctx_by_id(client, ACCESS_CONTROL_OBJECT_ID);
+
+    //Only the Client itself and the owner server is able to manage
+    // Access Control Object Instances, and Security Object Instances
+    // can only be managed by Bootstrap Servers, so there's no sense in
+    // creating Access Control Object Instances or ACLs for them
+    if (target_object_id == SECURITY_SERVER_OBJECT_ID ||
```
cabelitos: Move this if before the `acc_obj_ctx = find_object_ctx_by_id(client, ACCESS_CONTROL_OBJECT_ID);` command

src/lib/comms/sol-lwm2m.c
```diff
@@ -4718,6 +4859,172 @@ client_bootstrap(void *data)
...
+        if (acl_res)
+            sol_lwm2m_resource_clear(acl_res);
+        return 0;
+    }
+
+    if (acl_res) {
+        res = calloc(4, sizeof(struct sol_lwm2m_resource));
+        res[2] = *acl_res;
+    } else
+        res = calloc(3, sizeof(struct sol_lwm2m_resource));
```
cabelitos: Forgot to check if calloc() return value is not NULL

src/lib/comms/sol-lwm2m.c
```diff
@@ -4718,6 +4859,172 @@ client_bootstrap(void *data)
...
+    r = handle_create(client, acc_obj_ctx, -1, payload, UINT16_MAX, register_with_coap);
+    sol_lwm2m_tlv_list_clear(&payload.payload.tlv_content);
+    if (r == SOL_COAP_RESPONSE_CODE_CHANGED || r == SOL_COAP_RESPONSE_CODE_CREATED) {
+        SOL_DBG("Access Control Object Instance for Object Instance /%"
+            PRIu16 "/%" PRIu16 " created successfully!",
+            target_object_id, target_instance_id);
+    } else {
+        SOL_WRN("Failed to create Access Control Object Instance for Object Instance /%"
+            PRIu16 "/%" PRIu16, target_object_id, target_instance_id);
+        return -ENOENT;
```
cabelitos: ENOENT is not a good error code here.

src/lib/comms/sol-lwm2m.c
```diff
@@ -4718,6 +4859,172 @@ client_bootstrap(void *data)
...
+            if (r == SOL_COAP_RESPONSE_CODE_CHANGED || r == SOL_COAP_RESPONSE_CODE_CREATED) {
+                SOL_DBG("Access Control Object Instance for Object /%"
+                    PRIu16 " created successfully!", obj_ctx->obj->id);
+            } else if (r == 0) {
+                SOL_DBG("Security Object and Access Control Object does"
+                    " not need an Access Control Object Instance for 'Create' Operation");
+            } else {
+                SOL_WRN("Failed to create Access Control Object Instance for Object /%"
+                    PRIu16, obj_ctx->obj->id);
+                return -ENOENT;
```
cabelitos: ENOENT again

src/lib/comms/sol-lwm2m.c
```diff
@@ -4718,6 +4859,172 @@ client_bootstrap(void *data)
...
+            SOL_SET_API_VERSION(res.api_version = SOL_LWM2M_RESOURCE_API_VERSION; )
+            r = sol_lwm2m_resource_init_vector(&res, ACCESS_CONTROL_OBJECT_ACL_RES_ID,
+                SOL_LWM2M_RESOURCE_DATA_TYPE_INT, &server_ids, &acl_values);
+            if (r < 0) {
+                SOL_WRN("Could not init Access Control Object's [ACL] resource\n");
+                return r;
+            }
+
+            r = setup_access_control_object_instance_for_instance(client,
+                obj_ctx->obj->id, UINT16_MAX, UINT16_MAX, &res, false);
```
cabelitos: why UINT16_MAX for instance ids?

src/lib/comms/sol-lwm2m.c
```diff
@@ -4718,6 +4859,172 @@ client_bootstrap(void *data)
...
+
+            sol_vector_clear(&server_ids);
+            sol_vector_clear(&acl_values);
```
cabelitos: You're erasing the vectors inside the loop. This means only the first object will have the correct values set.

src/lib/comms/sol-lwm2m.c
```diff
@@ -4718,6 +4859,172 @@ client_bootstrap(void *data)
...
+        res = calloc(4, sizeof(struct sol_lwm2m_resource));
+        res[2] = *acl_res;
+    } else
+        res = calloc(3, sizeof(struct sol_lwm2m_resource));
+
+    SOL_LWM2M_RESOURCE_INT_INIT(r, &res[0],
+        ACCESS_CONTROL_OBJECT_OBJECT_RES_ID, target_object_id);
+    if (r < 0) {
+        SOL_WRN("Could not init Access Control Object's [Object ID] resource\n");
+        return r;
```
cabelitos: leaking res

src/lib/comms/sol-lwm2m.c
```diff
@@ -4718,6 +4859,172 @@ client_bootstrap(void *data)
...
+    }
+    SOL_LWM2M_RESOURCE_INT_INIT(r, &res[1],
+        ACCESS_CONTROL_OBJECT_INSTANCE_RES_ID, target_instance_id);
+    if (r < 0) {
+        SOL_WRN("Could not init Access Control Object's [Instance ID] resource\n");
+        return r;
```
cabelitos: leak

src/lib/comms/sol-lwm2m.c
```diff
@@ -4718,6 +4859,172 @@ client_bootstrap(void *data)
...
+    }
+    SOL_LWM2M_RESOURCE_INT_INIT(r, &res[acl_res ? 3 : 2],
+        ACCESS_CONTROL_OBJECT_OWNER_RES_ID, owner_server_id);
+    if (r < 0) {
+        SOL_WRN("Could not init Access Control Object's [Owner ID] resource\n");
```
cabelitos: leak

src/lib/comms/sol-lwm2m.c
```diff
@@ -4718,6 +4859,172 @@ client_bootstrap(void *data)
...
+        SOL_DBG("Access Control Object Instance for Object Instance /%"
+            PRIu16 "/%" PRIu16 " created successfully!",
+            target_object_id, target_instance_id);
+    } else {
+        SOL_WRN("Failed to create Access Control Object Instance for Object Instance /%"
+            PRIu16 "/%" PRIu16, target_object_id, target_instance_id);
+        return -ENOENT;
+    }
+
+    return r;
```
cabelitos: change to return 0;

src/lib/comms/sol-lwm2m.c
cabelitos: I get that you set UINT16_MAX to owner, since the owner is the bootstrap server.

bsmelo: For the Access Control Object Instances related to the Objects supported by the Client (not the Instances created), which are used to indicate whether or not a given Server has the rights to 'C'reate that Object, the Instance ID is UINT16_MAX.

It's on the first table from Pg. 59 of the spec.

src/lib/comms/sol-lwm2m.c
bsmelo: The outermost `if` already has `&& client->is_bootstrapping` as a condition. Should I still add a comment or something in this line?

src/lib/comms/sol-lwm2m.c
bsmelo: Ditto.

src/lib/comms/sol-lwm2m.c
bsmelo: I need the response code in the caller function. From there I `return 0;` already.

src/lib/comms/sol-lwm2m.c
bsmelo: Fixed:
```C
        clear_resource_array(res, (acl_res ? 3 : 2));
        free(res);
```

src/lib/comms/sol-lwm2m.c
bsmelo: Fixed:
```C
        sol_lwm2m_resource_clear(&res[0]);
        sol_lwm2m_resource_clear(&res[2]);
        free(res);
```

src/lib/comms/sol-lwm2m.c
bsmelo: Fixed:
```C
        sol_lwm2m_resource_clear(&res[2]);
        free(res);
```

src/lib/comms/sol-lwm2m.c
bsmelo: Wow, thanks.

Moved these operations to outside the loop (and added them to the failing `else` as well).

src/lib/comms/sol-lwm2m.c
bsmelo: Changed to
```C
    if (acl_res) {
        res = calloc(4, sizeof(struct sol_lwm2m_resource));
        SOL_NULL_CHECK(res, -ENOMEM);
        res[2] = *acl_res;
    } else {
        res = calloc(3, sizeof(struct sol_lwm2m_resource));
        SOL_NULL_CHECK(res, -ENOMEM);
    }
```

src/lib/comms/sol-lwm2m.c
bsmelo: I went through a list of error codes and couldn't grasp anything better... What do you suggest, -EINVAL?

src/lib/comms/sol-lwm2m.c
bsmelo: Ditto.

src/lib/comms/sol-lwm2m.c
bsmelo: [Why] Is it that bad?

Due to changing this behavior I had to change the snippet down below to
```C
    //From array of sol_lwm2m_resource to a sol_buffer in TLV format
    r = resources_to_tlv(res, acl_res ? 4 : 3, &buf);
    sol_lwm2m_resource_clear(&res[0]);
    sol_lwm2m_resource_clear(&res[1]);
    sol_lwm2m_resource_clear(&res[acl_res ? 3 : 2]);
    free(res);
```
as well. And also, the code for every subsequent `    if (r < 0) {` is a little bit less legible imho.

src/lib/comms/sol-lwm2m.c
bsmelo: Done.

src/lib/comms/sol-lwm2m.c
bsmelo: What do you suggest as error code? Fow now, changed it to:
```C
            r = -EINVAL;
            goto err_exit;
```

Also, thanks for the note about removing the added instance... whoops.

src/lib/comms/include/sol-lwm2m.h
bsmelo: Indeed; removed.

src/lib/comms/sol-lwm2m.c
bsmelo: Done.

src/lib/comms/sol-lwm2m.c
bsmelo: Changed it.

The call inside handle_resource now looks like this:
```C
    if (client->supports_access_control) {
        r = get_server_id_by_link_addr(&client->connections, cliaddr, &server_id);
        header_code = SOL_COAP_RESPONSE_CODE_INTERNAL_ERROR;
    }
```

src/lib/comms/sol-lwm2m.c
bsmelo: Done.

src/lib/comms/sol-lwm2m.c
bsmelo: Two clients? I didn't get it.

`setup_access_control_object_instances();` should be called only once, no matter what. It instantiates one Access Control Object Instance for each Object supported by the Client. This doesn't change.

The only case in which `setup_access_control_object_instances();` needs to be called again is if a Bootstrap Server sends `Delete "/"`, since that will delete these Access Control Instances as well, and then they will need to be recreated.

This function (`setup_access_control_object_instances();`) is responsible for the access rights related to each Object, not Object Instance.

cabelitos: What happens if in the same program you have two sol_lwm2m_clients with different objects?

src/samples/coap/lwm2m-client.c
```diff
@@ -707,6 +726,276 @@ del_location_obj(void *instance_data, void *user_data,
...
+
+    return r;
+
+err_free_acl:
+    sol_vector_clear(&instance_ctx->acl);
+    return r;
+}
+
+static int
+create_access_control_obj(void *user_data, struct sol_lwm2m_client *client,
```
cabelitos: This function and  `write_access_control_tlv` shares common code patterns. You could create auxiliary functions instead of doing copy & pasta.

src/lib/comms/sol-lwm2m.c
```diff
@@ -4586,7 +4585,7 @@ server_connection_ctx_new(struct sol_lwm2m_client *client,
     SOL_DBG("Fetching hostname info for:%.*s", SOL_STR_SLICE_PRINT(str_addr));
     conn_ctx->hostname_handle =
         sol_network_get_hostname_address_info(uri.host,
-        SOL_NETWORK_FAMILY_UNSPEC, hostname_ready, conn_ctx); //TODO: INET6
+        SOL_NETWORK_FAMILY_INET6, hostname_ready, conn_ctx);
```
cabelitos: I'm not quite sure about this one.

src/lib/comms/sol-lwm2m.c
```diff
@@ -4792,7 +4791,12 @@ bootstrap_finish(void *data, struct sol_coap_server *coap,
     struct sol_lwm2m_client *client = data;
     struct sol_coap_packet *response;
     int r;
-
+///*
```
cabelitos: Remove this lists

src/lib/comms/sol-lwm2m.c
```diff
@@ -4792,7 +4791,12 @@ bootstrap_finish(void *data, struct sol_coap_server *coap,
...
+    struct server_conn_ctx *conn_ctx;
+    struct sol_network_link_addr *server_addr;
+    uint16_t i;
+    SOL_BUFFER_DECLARE_STATIC(addr, SOL_NETWORK_INET_ADDR_STR_LEN);
+// */
```
cabelitos: same thing

src/lib/comms/sol-lwm2m.c
```diff
@@ -4809,6 +4813,22 @@ bootstrap_finish(void *data, struct sol_coap_server *coap,
...
+    // code in get_server_id_by_link_addr as well
+///*
+    SOL_WRN("FINISH:\ncliaddr=%s", sol_network_link_addr_to_str(cliaddr, &addr));
+    SOL_VECTOR_FOREACH_IDX (&client->connections, conn_ctx, i) {
+        server_addr = sol_vector_get_no_check(&conn_ctx->server_addr_list, conn_ctx->addr_list_idx);
+        SOL_WRN("server_addr=%s", sol_network_link_addr_to_str(server_addr, &addr));
+        if (sol_network_link_addr_eq(cliaddr, server_addr)) {
+            SOL_WRN("EQUAL!");
+//            server_connection_ctx_clear(conn_ctx);
+            //TODO: Why can't I call the function below to clear and del from the vector?
```
cabelitos: What do you mean?

src/lib/comms/sol-lwm2m.c
```diff
@@ -4809,6 +4813,22 @@ bootstrap_finish(void *data, struct sol_coap_server *coap,
 
     r = sol_coap_send_packet(coap, response, cliaddr);
     dispatch_bootstrap_event_to_client(client, SOL_LWM2M_BOOTSTRAP_EVENT_FINISHED);
+    //TODO: delete connection with bs-server after bs-finish, this is conflicting with the
+    // code in get_server_id_by_link_addr as well
+///*
+    SOL_WRN("FINISH:\ncliaddr=%s", sol_network_link_addr_to_str(cliaddr, &addr));
+    SOL_VECTOR_FOREACH_IDX (&client->connections, conn_ctx, i) {
+        server_addr = sol_vector_get_no_check(&conn_ctx->server_addr_list, conn_ctx->addr_list_idx);
+        SOL_WRN("server_addr=%s", sol_network_link_addr_to_str(server_addr, &addr));
```
cabelitos: You should not use this as WRN. You can change to DBG, however add more information about it

src/lib/comms/sol-lwm2m.c
```diff
@@ -4809,6 +4813,22 @@ bootstrap_finish(void *data, struct sol_coap_server *coap,
 
     r = sol_coap_send_packet(coap, response, cliaddr);
     dispatch_bootstrap_event_to_client(client, SOL_LWM2M_BOOTSTRAP_EVENT_FINISHED);
+    //TODO: delete connection with bs-server after bs-finish, this is conflicting with the
```
cabelitos: Remove this TODO?

src/lib/comms/sol-lwm2m.c
```diff
@@ -2047,7 +2047,7 @@ observation_request_reply(void *data, struct sol_coap_server *coap_server,
             return false;
         }
         SOL_WRN("Could not complete the observation request on client:%s"
-            " path:%s", entry->path, entry->cinfo->name);
+            " path:%s", entry->cinfo->name, entry->path);
```
cabelitos: Ok :]

src/lib/comms/include/sol-lwm2m.h
```diff
@@ -733,10 +755,34 @@ void sol_lwm2m_resource_clear(struct sol_lwm2m_resource *resource);
...
+    uint16_t id, enum sol_lwm2m_resource_type type, uint16_t resource_len,
     enum sol_lwm2m_resource_data_type data_type, ...);
 
 /**
+ * @brief Initializes a LWM2M resource of type multiple using two @c sol_vector.
+ *
+ * This function makes it easier to init a LWM2M resource of type multiple,
+ * when we don't know beforehand the amount of Resource Instances desired.
+ * The last argument is a @c sol_vector holding elements of the relevant
+ * resource data type as per the table in @see sol_lwm2m_resource_init().
```
cabelitos: Not very good explanation. You know how many resources instances you have, since you pass it ass vectors.

src/lib/comms/sol-lwm2m.c
```diff
@@ -2198,13 +2198,91 @@ sol_lwm2m_resource_init(struct sol_lwm2m_resource *resource,
...
+            int *ic = v;
+            resource->data[i].content.integer = *ic;
+
+        } else if (resource->data_type == SOL_LWM2M_RESOURCE_DATA_TYPE_OBJ_LINK) {
+            int *ic = v;
+            resource->data[i].content.integer = (uint16_t)*ic;
+            resource->data[i].content.integer = (resource->data[i].content.integer << 16) |
+                (uint16_t)*ic;
+
+        } else {
```
cabelitos: You forgot to check about the other formats like time and int. I suggest that you create an auxiliary function and use it to init the types in `sol_lwm2m_resource_init` and in this function.

src/lib/comms/include/sol-lwm2m.h
```diff
@@ -733,10 +755,34 @@ void sol_lwm2m_resource_clear(struct sol_lwm2m_resource *resource);
...
+ * @param res_inst_ids The list of Resource Instance IDs, in the same order as their relevant values in @c res_inst_values
+ * @param res_inst_values The list of Resource Instance values, in the same order as their relevant ids in @c res_inst_ids, respecting the table according to the resource data type.
+ * @return 0 on success, negative errno on error.
+ * @see sol_lwm2m_resource_clear()
+ * @see sol_lwm2m_resource_init()
+ *
+ * @note The LWM2M resource api_version must be set,
+ * before calling this function.
+ */
+int sol_lwm2m_resource_init_vector(struct sol_lwm2m_resource *resource,
```
cabelitos: To be honest, I'm not very happy with this approach.

src/lib/comms/sol-lwm2m.c
```diff
@@ -3014,9 +3014,106 @@ clear_bootstrap_ctx(struct sol_lwm2m_client *client)
...
+{
+    struct obj_ctx *obj_ctx;
+    struct obj_instance *obj_instance;
+    struct sol_lwm2m_resource res[2];
+    int r;
+    bool ret = false;
+    int64_t default_acl = SOL_LWM2M_ACL_NONE;
+    uint16_t i;
+
+    SOL_WRN("connections.len=%" PRIu16, client->connections.len);
```
cabelitos: Change to DBG?

src/lib/comms/sol-lwm2m.c
```diff
@@ -3014,9 +3014,106 @@ clear_bootstrap_ctx(struct sol_lwm2m_client *client)
...
+
+    //If only one server or Bootstrap Server ID, then full access rights
+    if (client->connections.len == 1 || server_id == UINT16_MAX)
+        return true;
+
+    SOL_WRN("server_id=%" PRId64 " obj=/%" PRIu16 "/%" PRId32 " needed=%" PRId64,
```
cabelitos: DBG?

src/lib/comms/sol-lwm2m.c
```diff
@@ -3014,9 +3014,106 @@ clear_bootstrap_ctx(struct sol_lwm2m_client *client)
...
+    struct sol_lwm2m_resource res[2];
+    int r;
+    bool ret = false;
+    int64_t default_acl = SOL_LWM2M_ACL_NONE;
+    uint16_t i;
+
+    SOL_WRN("connections.len=%" PRIu16, client->connections.len);
+
+    //If only one server or Bootstrap Server ID, then full access rights
+    if (client->connections.len == 1 || server_id == UINT16_MAX)
```
cabelitos: Before returning, log that this is the bootstrap server

src/lib/comms/sol-lwm2m.c
```diff
@@ -3014,9 +3014,106 @@ clear_bootstrap_ctx(struct sol_lwm2m_client *client)
...
+
+    obj_ctx = find_object_ctx_by_id(client, ACCESS_CONTROL_OBJECT_ID);
+    SOL_VECTOR_FOREACH_IDX (&obj_ctx->instances, obj_instance, i) {
+        r = read_resources(client, obj_ctx, obj_instance, res, 2,
+            ACCESS_CONTROL_OBJECT_OBJECT_RES_ID,
+            ACCESS_CONTROL_OBJECT_INSTANCE_RES_ID);
+        if (r < 0) {
+            SOL_WRN("Could not read Access Control Object's"
+                " [Object ID] and [Instance ID] resources\n");
+            continue;
```
cabelitos: I guess you should break it. This is not supposed to happen!

src/lib/comms/sol-lwm2m.c
```diff
@@ -3014,9 +3014,106 @@ clear_bootstrap_ctx(struct sol_lwm2m_client *client)
...
+            if (default_acl & rights_needed) {
+                ret = true;
+                goto exit;
+            }
+
+            //If Observe operation on Object level, check next instance
+            if (!(instance_id == -1 && (rights_needed & SOL_LWM2M_ACL_READ)))
+                goto exit;
+        }
+
```
cabelitos: replace the goto with break

src/lib/comms/sol-lwm2m.c
```diff
@@ -3014,9 +3014,106 @@ clear_bootstrap_ctx(struct sol_lwm2m_client *client)
...
+            instance_id == -1 && (rights_needed & SOL_LWM2M_ACL_READ))) {
+            clear_resource_array(res, sol_util_array_size(res));
+
+            r = read_resources(client, obj_ctx, obj_instance, res, 2,
+                ACCESS_CONTROL_OBJECT_ACL_RES_ID,
+                ACCESS_CONTROL_OBJECT_OWNER_RES_ID);
+            if (r < 0) {
+                SOL_WRN("Could not read Access Control"
+                    " Object's [ACL] and [Owner ID] resources\n");
+                continue;
```
cabelitos: Again, you should break and return the error code.

src/lib/comms/sol-lwm2m.c
```diff
@@ -3014,9 +3014,106 @@ clear_bootstrap_ctx(struct sol_lwm2m_client *client)
...
+                }
+
+                //Keep the default ACL Resource Instance, if any, to save another loop later
+                if (res[0].data[i].id == DEFAULT_SHORT_SERVER_ID)
+                    default_acl = res[0].data[i].content.integer;
+            }
+
+            //If no ACL for this server, check if it is the owner of the object.
+            // If owner and no specific ACL Resource Instance, then full access rights.
+            if (res[1].data[0].content.integer == server_id) {
```
cabelitos: Why not check this first and then if this is not true, you iterate over the res[0] ?

src/lib/comms/sol-lwm2m.c
bsmelo: They are right there:
```C
        } else if (resource->data_type == SOL_LWM2M_RESOURCE_DATA_TYPE_INT ||
            resource->data_type == SOL_LWM2M_RESOURCE_DATA_TYPE_TIME) {
```

I'm not sure an auxiliary function would help that much, since there are differences between this `if-else` chain and the one in `sol_lwm2m_resource_init` (like the pointer variable declaration depending on the `resource->data_type` and the use [or not] of `va_args`).

src/lib/comms/sol-lwm2m.c
```diff
@@ -3014,9 +3014,106 @@ clear_bootstrap_ctx(struct sol_lwm2m_client *client)
     }
 }
 
+static bool
+is_authorized(struct sol_lwm2m_client *client,
```
cabelitos: I suggest that you change this signature and return an int. By doing so we can know if an error happens during the read process.
So basically you can do something like:
If is_authrorized returns < 0 == Error
If is_authorized returns 0 == Not authorized.
If is_authorized returns > 0 == Authorized

src/lib/comms/include/sol-lwm2m.h
bsmelo: Mixed feelings in here.

An alternative I thought of would be to expose the type `struct sol_lwm2m_resource_data` (already used inside `sol_lwm2m_resource`), and the user could provide a single `sol_vector` with elements from this type. I'm not sure about how better that would be, though...

src/lib/comms/include/sol-lwm2m.h
bsmelo: We know dynamically, not statically (I mean, there's no way I could use the regular `sol_lwm2m_init` function with `va_args` for this use case).

Maybe my wording was not good, though... any suggestion? :/

src/lib/comms/sol-lwm2m.c
```diff
@@ -4146,11 +4312,18 @@ read_resources(struct sol_lwm2m_client *client,
...
     va_start(ap, res_len);
 
-    // The va_list contains the resources IDs that we should be read.
+    // The va_list contains the resources IDs that should be read.
     for (i = 0; i < res_len; i++) {
         r = obj_ctx->obj->read((void *)instance->data,
             (void *)client->user_data, client, instance->id,
             (uint16_t)va_arg(ap, int), &res[i]);
+
+        if (r == -ENOENT) {
```
cabelitos: Explain to me, why this if?

src/lib/comms/include/sol-lwm2m.h
```diff
@@ -420,7 +441,7 @@ struct sol_lwm2m_resource {
 #define SOL_LWM2M_RESOURCE_INT_INIT(ret_value_, resource_, id_, value_) \
```
bsmelo: Sure, why not. I will change it.

src/lib/comms/include/sol-lwm2m.h
cabelitos: You should try this one. Let's see

src/lib/comms/sol-lwm2m.c
bsmelo: Keeping state in the client now.

src/samples/coap/lwm2m-client.c
bsmelo: Created a function to `write_or_create_acl()`.

src/lib/comms/sol-lwm2m.c
cabelitos: God, I'm blind. You're right. Leave as is

src/lib/comms/sol-lwm2m.c
cabelitos: Yeah, the resource may not be mandatory. :]

cabelitos: However I suggest you do this check after LWM2M_RESOURCE_CHECK_API_GOTO

src/lib/comms/sol-lwm2m.c
bsmelo: Because spec [Pg. 60; Section 7.3.2.1 Obtaining Access Right].

I guess there could be a scenario in which I'm the owner, but my ACL exists and doesn't allow me to 'D'elete, for instance (maybe I changed it before, as the owner, or the bootstrap server changed it). If I check for owner first, I would be granted the access rights... but if I check for the ACL first (as the spec says), I would not. :)

cabelitos: Ok

src/lib/comms/sol-lwm2m.c
cabelitos: Ignore, you're right. YOu need the r value.

src/lib/comms/sol-lwm2m.c
bsmelo: Changed to -ECANCELED.

bsmelo:
An updated version is at #2255. Closing this. Thanks.

Issue #2072, opened 2016-05-20
Intermittently failed to run fbp at the 1st trial on IIO sensors with "create" command in json file


fulong82:
#### Issue Description

Intermittently observed a failed running error at the 1st try but error gone while doing a 2nd try running with IIO Sensor and using "create" command to create device and load the sensor driver. 

The failed warning/error may happen in a random location for example:
case 1: WRN: ./src/lib/io/sol-iio.c:916 sol_iio_add_channel() Could not find channel [in_magn_x] for device0
case 2: WRN: ./src/lib/io/sol-iio.c:252 set_current_trigger() Could not write to device current_trigger file

Suspect this is a timing issue where the sol flow node is trying to access the iio sysfs interface while iio framework is still creating the sensor interfaces.

#### Sample json & fbp code
====sol-flow.json====
{
  "$schema": "http://solettaproject.github.io/soletta/schemas/config.schema",
  "nodetypes": [
    {
      "name": "MyGyro",
      "type": "iio/gyroscope",
      "options": {
        "iio_device": "l3gd20 create,i2c,pci0000:00/0000:00:17.2/i2c_designware.6,0x6B,l3gd20 0",
        "buffer_size": 0
      }
    },
    {
      "name": "MyMag",
      "type": "iio/magnet",
      "options": {
        "iio_device": "ak09911 create,i2c,pci0000:00/0000:00:17.2/i2c_designware.6,0x0C,ak09911 0",
        "buffer_size": -1
      }
    }
  ]
}

====testGyro.fbp====
#!/usr/bin/env sol-fbp-runner

gyro(MyGyro)
timer(timer:interval=1000)

timer OUT -> TICK gyro OUT -> IN _(console)

====testMagneto.fbp====
#!/usr/bin/env sol-fbp-runner

magneto(MyMag)
timer(timer:interval=1000)

timer OUT -> TICK magneto OUT -> IN _(console)
#### Steps to reproduce

Take the log 1 below as an reproduce steps example:
 1. Cold boot the system with no iio sensor device & driver loaded
 2. Run the fbp with the "create" command in json config file to create sensor device and load the sensor driver (take gyro sensor as an example)
 3. Run another fbp with the "create" command in json config file to create sensor device and load the sensor driver (take magneto sensor as an example)
 4. Error observed: Failed to run flow
 5. Run the same fbp again (take magneto sensor as an example) and no failing observed

#### Logs

 * Soletta version (top commit hash): tag v1_beta19
[...]

PR #2240, opened 2016-07-15
I2C: Match the kernel's I2C address path.

cabelitos:
The device address must have 4 characters. This was causing
some problems in IIO, because it was not possible to find the
correct path in the sysfs.


cabelitos:
Should fix https://github.com/solettaproject/soletta/issues/2072

cabelitos:
@edersondisouza Take a look

bdilly:
:+1:

edersondisouza:
lgtm

Issue #2247, opened 2016-07-18
Warning about invalid sub_api version in calamari/7seg node type


cmarcelo:
#### Issue Description

Using a FBP containing

```
display(calamari/7seg:clear_pin=473,clock_pin=340,data_pin=474,latch_pin=475)
```

either with sol-fbp-runner or sol-fbp-generator, will yield in the end the following error

WRN:sol-flow-calamari ./src/modules/flow/calamari/calamari.c:221 calamari_7seg_child_opts_set() child_opts(0xd33100)->sub_api(0) != SOL_FLOW_NODE_TYPE_GPIO_WRITER_OPTIONS_API_VERSION(1)

#### Board information

Host OS and architecture: Fedora 23

Target OS and platform: MinnowBoard

Soletta version: beta19

bdilly:
@cmarcelo I believe it's fixed now.
If not please reopen this ticket
Thanks

PR #2251, opened 2016-07-18
modules/calamari: only check API version of used options

bdilly:
Fix issue #2247


barbieri:
+1

PR #2252, opened 2016-07-18
all: add typedefs for public facing structs

### LINKED ISSUE #1982: proposal: use typedef for public opaque structs
The page https://github.com/solettaproject/soletta/wiki/API-design-conventions documents Soletta approach for dealing with typedefs:

> We don't typedef structs, enums, function pointers. There are some exceptions around the code, but we prefer not to open space for more.

I propose to add an exception for pointers to opaque structures, at least when used by consumers of the library. For the consumer of the library, there is literally no difference whether `sol_gpio *` is a struct or not, because they can't access its contents.

Note the proposal is to typedef the structs not pointers to the structs. This doesn't necessarily affects how code inside Soletta is written.

The exception won't apply for structs that are not used only via opaque pointers: I do think we would benefit in this case as well, but it's a harder argument to have, and IMHO there is some value in this limited scope proposal.

cmarcelo:
Soletta already use "sol_" namespace and usually some other
module-specific namespace making its identifiers large. By removing the
need of explicitly use "struct " we save some bytes and reduce the noise
in client code.

This patch only adds typedef for public facing structs, per discussion
with the team. The typedefs were added after the declaration/definition
of the structs, to avoid messing with Doxygen documentation.

Fixes #1982.

### DIFF
--- a/src/lib/common/include/sol-certificate.h
+++ b/src/lib/common/include/sol-certificate.h
@@ -45,6 +45,7 @@ extern "C" {
  * This object is the abstraction of certificate.
  */
 struct sol_cert;
+typedef struct sol_cert sol_cert;
 
 /**
  * @brief Load a certificate from an id
--- a/src/lib/common/include/sol-file-reader.h
+++ b/src/lib/common/include/sol-file-reader.h
@@ -35,6 +35,7 @@ extern "C" {
  * @brief Opaque handler for a file reader.
  */
 struct sol_file_reader;
+typedef struct sol_file_reader sol_file_reader;
 
 /**
  * @brief Open a file using its filename.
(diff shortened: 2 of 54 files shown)


cmarcelo:
@bdilly @barbieri

bdilly:
Looks like doxygen only understand the documentation is about the struct X, not typedef X.
Not sure if just adding a command @ typedef below @ struct would make the trick

bdilly:
@cmarcelo actually I believe just documenting the typedef instead of the struct should be good enough, after all the idea is that users would use typedef (direct struct usage would be only for internal usage).

cmarcelo:
Replaced by #2257.

PR #2255, opened 2016-07-18
LWM2M: Access Control [v2]

bsmelo:
- Main changes from v1 (#2239):
 - Changes suggested in the review by @cabelitos 

- TODO/Under discussion:
- [x] Figuring out how to deal with Notify operation triggered by the client API (https://github.com/bsmelo/soletta/commit/6271c96344fa0f775259e0210bbc89ccf58bf067#diff-e4058dc14a343a8aceb7210b323358a2R5577)
- [x] Observe operation may be related to the issue above, and may require a fix in `sol-coap`.
- [x] Restriction to IPv6 (https://github.com/bsmelo/soletta/commit/29836c12703ce6423746099295f339781b420e2c#diff-e4058dc14a343a8aceb7210b323358a2R4585)

- Future Improvements
 - Add tests to `test-lwm2m.c`

### DIFF
--- a/src/lib/comms/sol-lwm2m.c
+++ b/src/lib/comms/sol-lwm2m.c
@@ -2048,7 +2048,7 @@ observation_request_reply(void *data, struct sol_coap_server *coap_server,
             return false;
         }
         SOL_WRN("Could not complete the observation request on client:%s"
-            " path:%s", entry->path, entry->cinfo->name);
+            " path:%s", entry->cinfo->name, entry->path);
         keep_alive = false;
     } else {
         extract_content(req, &code, &type, &content);
--- a/src/lib/comms/sol-lwm2m.c
+++ b/src/lib/comms/sol-lwm2m.c
@@ -73,6 +73,12 @@ SOL_LOG_INTERNAL_DECLARE_STATIC(_lwm2m_domain, "lwm2m");
 #define SERVER_OBJECT_LIFETIME (1)
 #define SERVER_OBJECT_BINDING (7)
 
+#define ACCESS_CONTROL_OBJECT_ID (2)
+#define ACCESS_CONTROL_OBJECT_OBJECT_RES_ID (0)
+#define ACCESS_CONTROL_OBJECT_INSTANCE_RES_ID (1)
+#define ACCESS_CONTROL_OBJECT_ACL_RES_ID (2)
+#define ACCESS_CONTROL_OBJECT_OWNER_RES_ID (3)
+
 #ifndef SOL_NO_API_VERSION
 #define LWM2M_TLV_CHECK_API(_tlv, ...) \
     do { \
@@ -283,6 +289,8 @@ struct sol_lwm2m_client {
     bool running;
     bool removed;
     bool is_bootstrapping;
+    bool supports_access_control;
+    bool need_to_setup_access_control;
 };
 
 struct server_conn_ctx {
@@ -307,6 +315,18 @@ static int bootstrap_with_server(struct sol_lwm2m_client *client,
 static int handle_resource(void *data, struct sol_coap_server *server,
     const struct sol_coap_resource *resource, struct sol_coap_packet *req,
     const struct sol_network_link_addr *cliaddr);
+static int setup_access_control_object_instance_for_instance(
+    struct sol_lwm2m_client *client, uint16_t object_id,
+    uint16_t instance_id, int64_t server_id,
+    struct sol_lwm2m_resource *acl_res, bool register_with_coap);
+static int setup_access_control_object_instances(
+    struct sol_lwm2m_client *client);
+static int
+read_resources(struct sol_lwm2m_client *client,
+    struct obj_ctx *obj_ctx, struct obj_instance *instance,
+    struct sol_lwm2m_resource *res, size_t res_len, ...);
+static void
+clear_resource_array(struct sol_lwm2m_resource *array, uint16_t len);
 
 static void
 send_ack_if_needed(struct sol_coap_server *coap, struct sol_coap_packet *msg,
@@ -2994,6 +3014,8 @@ handle_delete(struct sol_lwm2m_client *client,
 {
     int r, ret = SOL_COAP_RESPONSE_CODE_BAD_REQUEST;
     uint16_t i;
+    struct obj_ctx *acc_obj_ctx;
+    struct obj_instance *acc_obj_instance;
 
     //Specific Instance?
     if (obj_ctx && obj_instance) {
@@ -3043,6 +3065,22 @@ handle_delete(struct sol_lwm2m_client *client,
                     ret = SOL_COAP_RESPONSE_CODE_DELETED;
                 }
             }
+        }
+
+        if (client->supports_access_control) {
+            r = setup_access_control_object_instances(client);
+            SOL_INT_CHECK_GOTO(r, < 0, err_access_control);
+        }
+    }
+
+    return ret;
+
+err_access_control:
+    if (client->supports_access_control) {
+        acc_obj_ctx = find_object_ctx_by_id(client, ACCESS_CONTROL_OBJECT_ID);
+        while ((acc_obj_instance = sol_vector_get(&acc_obj_ctx->instances, 0)) != NULL) {
+            obj_instance_clear(client, acc_obj_ctx, acc_obj_instance);
+            sol_vector_del_element(&acc_obj_ctx->instances, acc_obj_instance);
         }
     }
 
@@ -3219,7 +3257,8 @@ write_instance_tlv_or_resource(struct sol_lwm2m_client *client,
 
 static uint8_t
 handle_create(struct sol_lwm2m_client *client,
-    struct obj_ctx *obj_ctx, int32_t instance_id, struct sol_lwm2m_payload payload)
+    struct obj_ctx *obj_ctx, int32_t instance_id, struct sol_lwm2m_payload payload,
+    uint64_t owner_server_id, bool register_with_coap)
 {
     int r;
     struct obj_instance *obj_instance;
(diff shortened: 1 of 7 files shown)


src/lib/comms/sol-lwm2m.c
```diff
@@ -3244,9 +3283,24 @@ handle_create(struct sol_lwm2m_client *client,
...
+        r = setup_access_control_object_instance_for_instance(client,
+            obj_ctx->obj->id, obj_instance->id, owner_server_id, NULL, register_with_coap);
+
+        if (r == 0) {
+            SOL_DBG("Access Control Object Instance and Security Object Instance does"
+                " not need an Access Control Object Instance nor ACLs");
+        } else if (r != SOL_COAP_RESPONSE_CODE_CHANGED && r != SOL_COAP_RESPONSE_CODE_CREATED) {
+            SOL_WRN("Failed to create Access Control Object Instance for Object /%"
+                PRIu16 "/%" PRIu16, obj_ctx->obj->id, obj_instance->id);
+            return -ENOENT;
```
cabelitos: ECANCELED or something like that

src/lib/comms/sol-lwm2m.c
```diff
@@ -4718,6 +4858,178 @@ client_bootstrap(void *data)
...
+    //Only the Client itself and the owner server is able to manage
+    // Access Control Object Instances, and Security Object Instances
+    // can only be managed by Bootstrap Servers, so there's no sense in
+    // creating Access Control Object Instances or ACLs for them
+    if (target_object_id == SECURITY_SERVER_OBJECT_ID ||
+        target_object_id == ACCESS_CONTROL_OBJECT_ID) {
+        return 0;
+    }
+
+    acc_obj_ctx = find_object_ctx_by_id(client, ACCESS_CONTROL_OBJECT_ID);
```
cabelitos: Check if this is not NULL ?

src/lib/comms/sol-lwm2m.c
```diff
@@ -4718,6 +4858,178 @@ client_bootstrap(void *data)
...
+
+    if (acl_res) {
+        res = calloc(4, sizeof(struct sol_lwm2m_resource));
```
cabelitos: I have a better idea. There's no need to alloc memory for `res`. You can do something like this.

```C
struct sol_lwm2m_resource res[4] = { 0 };
int res_size = 4;

if (!acl_res)
    res_size = 3;
//Do that you need to do.....
```

src/lib/comms/sol-lwm2m.c
```diff
@@ -4830,6 +5150,14 @@ sol_lwm2m_client_start(struct sol_lwm2m_client *client)
 
     return 0;
 
+err_access_control:
+    if (client->supports_access_control) {
+        ctx = find_object_ctx_by_id(client, ACCESS_CONTROL_OBJECT_ID);
+        while ((instance = sol_vector_get(&ctx->instances, 0)) != NULL) {
```
cabelitos: Why not SOL_VECTOR_FOREACH_IDX ?

src/lib/comms/sol-lwm2m.c
```diff
@@ -3008,9 +3008,103 @@ clear_bootstrap_ctx(struct sol_lwm2m_client *client)
...
+
+            //If no ACL and not owner, check if the default ACL Resource Instance applies
+            if (default_acl & rights_needed) {
+                r = 1;
+                goto exit;
+            }
+
+            //If Observe operation on Object level, check next instance
+            if (!(instance_id == -1 && (rights_needed & SOL_LWM2M_ACL_READ)))
+                goto exit;
```
cabelitos: You decided to leave the gotos here?

src/lib/comms/sol-lwm2m.c
```diff
@@ -3008,9 +3008,103 @@ clear_bootstrap_ctx(struct sol_lwm2m_client *client)
     }
 }
 
+static int
+is_authorized(struct sol_lwm2m_client *client,
```
cabelitos: I guess we should document here that 1 means authorized, 0 means not authorized and < 0 mean error.

src/lib/comms/sol-lwm2m.c
```diff
@@ -3008,9 +3008,103 @@ clear_bootstrap_ctx(struct sol_lwm2m_client *client)
...
+            }
+
+            //If no ACL and not owner, check if the default ACL Resource Instance applies
+            if (default_acl & rights_needed) {
+                r = 1;
+                goto exit;
+            }
+
+            //If Observe operation on Object level, check next instance
+            if (!(instance_id == -1 && (rights_needed & SOL_LWM2M_ACL_READ)))
```
cabelitos: You comment does not describe what's really hapenning. You will not check the next instance, you will break the loop and return.

src/lib/comms/sol-lwm2m.c
```diff
@@ -3008,9 +3008,103 @@ clear_bootstrap_ctx(struct sol_lwm2m_client *client)
...
+
+    obj_ctx = find_object_ctx_by_id(client, ACCESS_CONTROL_OBJECT_ID);
+    SOL_VECTOR_FOREACH_IDX (&obj_ctx->instances, obj_instance, i) {
+        r = read_resources(client, obj_ctx, obj_instance, res, 2,
+            ACCESS_CONTROL_OBJECT_OBJECT_RES_ID,
+            ACCESS_CONTROL_OBJECT_INSTANCE_RES_ID);
+        if (r < 0) {
+            SOL_WRN("Could not read Access Control Object's"
+                " [Object ID] and [Instance ID] resources\n");
+            goto exit;
```
cabelitos: So you decided to leave the gotos?

src/lib/comms/sol-lwm2m.c
edersondisouza: On 20-07-2016 15:05, Guilherme Íscaro wrote:
> In src/lib/comms/sol-lwm2m.c
> <https://github.com/solettaproject/soletta/pull/2255#discussion_r71576311>:
>
>> @@ -3008,9 +3008,103 @@ clear_bootstrap_ctx(struct sol_lwm2m_client *client)
>>      }
>>  }
>>
>> +static int
>> +is_authorized(struct sol_lwm2m_client *client,
>
> I guess we should document here that 1 means authorized, 0 means not
> authorized and < 0 mean error.
>

Then, the `is_` prefix won't make much sense, right?

src/lib/comms/sol-lwm2m.c
```diff
@@ -3008,9 +3008,103 @@ clear_bootstrap_ctx(struct sol_lwm2m_client *client)
...
+        }
+
+        //Retrieve the associated Access Control Object Instance, by matching Object ID
+        // and Instance ID, or if instance_id == -1 and 'R'ead is needed, this
+        // is an Observe Request on Object level and any Instance with this access right
+        // is enough to reply with the Observe falg on CoAP
```
cabelitos: s/falg/flag/

cabelitos:
@bsmelo,
How's the progress with the CoAP changes?

src/lib/comms/sol-lwm2m.c
bsmelo: We're sure it's not NULL, since this function (and any other access control-related stuff) is only called if `client->supports_access_control` is true (which is setup during `sol_lwm2m_client_new` if there's an object with `ACCESS_CONTROL_OBJECT_ID` in the list the client passes to that function.

src/lib/comms/sol-lwm2m.c
bsmelo: Oh, you're right, nice one. I have something like this now:
```C
    struct sol_lwm2m_resource res[4] = { };
    int res_last_id = 4;

    if (acl_res)
        res[2] = *acl_res;
    else
        res_last_id = 3;

//Then when I have to clear stuff:
sol_lwm2m_resource_clear(&res[res_last_id - 1]);

//Or to get the resources in TLV format:
r = resources_to_tlv(res, res_last_id, &buf);
```

src/lib/comms/sol-lwm2m.c
bsmelo: Because `sol_vector_del_element(&ctx->instances, instance);`.

Since I'm deleting stuff, it messes up the counter in SOL_VECTOR_FOREACH_IDX... same thing as https://github.com/solettaproject/soletta/pull/2255/files#diff-e4058dc14a343a8aceb7210b323358a2R3165, which is where I first faced this.

src/lib/comms/sol-lwm2m.c
bsmelo: It does, but the condition checking is a negative (mind the `!`), since it's more clear to think about the only case in which it should check next instance (Observe on Object level) then all other cases. I can change it to:
//If **not** Observe operation on Object level, **do not** check next instance**; only break and return**

Really better?

src/lib/comms/sol-lwm2m.c
bsmelo: Yes... Since these two would have to be `goto`s anyway:
https://github.com/solettaproject/soletta/pull/2255/files#diff-e4058dc14a343a8aceb7210b323358a2R3065
https://github.com/solettaproject/soletta/pull/2255/files#diff-e4058dc14a343a8aceb7210b323358a2R3068

I went with the approach of keeping them all syntactically consistent.

bsmelo:
I'm on it, sorry for the delay.

src/lib/comms/sol-lwm2m.c
cabelitos: k

src/lib/comms/sol-lwm2m.c
cabelitos: awesome

src/lib/comms/sol-lwm2m.c
cabelitos: Indeed. We should change the name as well

src/lib/comms/sol-lwm2m.c
cabelitos: As far as I know, you are able to delete and iterate from the sol_vector using SOL_VECTOR_FOREACH_IDX

cabelitos: ignore me

src/lib/comms/sol-lwm2m.c
cabelitos: ok

src/lib/comms/sol-lwm2m.c
cabelitos: I prefer this new version of the comment, otherwise others can be become *dazed* and confused like I did

src/lib/comms/sol-lwm2m.c
edersondisouza: You can always do a reverse loop and delete, using `SOL_VECTOR_FOREACH_REVERSE_IDX`.

edersondisouza: Anyway, as you are actually deleting all elements, it's better to just iterate over all of them calling `obj_instance_clear`, and after that deleting the whole vector by calling `sol_vector_clear`. It avoids unnecessary reallocs while deleting elements.

bsmelo:
An updated version is at #2264. Closing this. Thanks.

PR #2257, opened 2016-07-19
all: add typedefs to public API structs [v2]

### LINKED ISSUE #1982: proposal: use typedef for public opaque structs

cmarcelo:
Soletta already uses "sol_" namespace and most of the time a secondary
namespace that is module-specific. This makes the identifiers very
large. By removing the need of explicitly use "struct" the client code
has less noise.

The convention for Doxygen documentation was:

- Opaque structs appear in typedefs section of the
  documentation. @typedef doxygen marker is used.

- Public structs appear both in data structures and typedefs
  sections. The documentation is not marked so it applies to both (some
  markers were removed).

This patch includes a change in documentation of sol_http_url struct to
escape "#fragment" properly, so Doxygen stops warning. This was
preventing "make doc" to work -- due to a check done after doxygen call.

Fixes #1982.

### DIFF
--- a/src/lib/common/include/sol-certificate.h
+++ b/src/lib/common/include/sol-certificate.h
@@ -38,13 +38,14 @@ extern "C" {
  */
 
 /**
- * @struct sol_cert
+ * @typedef sol_cert
  *
  * @brief Certificate handler
  *
  * This object is the abstraction of certificate.
  */
 struct sol_cert;
+typedef struct sol_cert sol_cert;
 
 /**
  * @brief Load a certificate from an id
--- a/src/lib/common/include/sol-file-reader.h
+++ b/src/lib/common/include/sol-file-reader.h
@@ -31,10 +31,11 @@ extern "C" {
 #endif
 
 /**
- * @struct sol_file_reader
+ * @typedef sol_file_reader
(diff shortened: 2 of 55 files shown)


cmarcelo:
This adds tweaks to documentation after @bdilly suggestion.

cmarcelo:
@bdilly the error from semaphoreci seems unrelated to my patch :-(

bdilly:
yeah, you're right. not related.
commits is good
:+1:

bdilly:
merged

PR #2259, opened 2016-07-19
build: improve make doxygen rule

### LINKED ISSUE #2258: build: doxygen rule is fragile
I can run "make doxygen" and it fails, then run "make doxygen" again and it passes. The issue is we can fail after the product is generated, so a second run will think the product is already there. The three last lines might fail. In my particular case I was failing the test of the logfile (something wrong generated warning in the log file).

In tools/build/Makefile.targets

```
$(DOXYGEN_GENERATED): $(DOXYGEN_RESOURCES) $(shell find $(top_srcdir)src/ -name '*.h' -type f)
	$(Q)$(MKDIR) -p $(build_doxygen_targetdir)
	$(Q)$(RM) $(build_doxygen_logfile)
	$(Q)$(RM) -rf $(build_doxygen_targetdir)/*
	$(Q) cat $(top_srcdir)doc/doxygen/examples_head.dox >> $(build_doxygen_targetdir)/examples.dox
	$(Q) find $(top_srcdir)src/samples -type f -name *.c -printf "@example %h/%f\n" | sort >> $(build_doxygen_targetdir)/examples.dox
	$(Q) cat $(top_srcdir)doc/doxygen/examples_foot.dox >> $(build_doxygen_targetdir)/examples.dox
	$(Q)$(DOXYGEN) doc/doxygen/Doxyfile
	$(Q) test `du $(build_doxygen_logfile) | cut -f1` -eq 0 || (echo "Failed to build doc" && cat $(build_doxygen_logfile) && exit 1)
	$(Q) cd $(build_doxygen_targetdir)/html/; for i in `find . -name "_2src*.html"`; do mv $$i $${i#./_2}; done
	$(Q) cd $(build_doxygen_targetdir)/html/; find . -name "*.html" -exec sed -i -e 's/_2src/src/g' {} \;
```

bdilly:
Make sure it won't consider make doxygen is done
when it fails (it has warnings).

For that create a success file when everything is
checked.

Fix #2258

### DIFF
--- a/tools/build/Makefile.targets
+++ b/tools/build/Makefile.targets
@@ -163,6 +163,7 @@ ifeq (yyy,$(HAVE_DOXYGEN)$(HAVE_BZIP2)$(HAVE_TAR))
 $(DOXYGEN_GENERATED): $(DOXYGEN_RESOURCES) $(shell find $(top_srcdir)src/ -name '*.h' -type f)
 	$(Q)$(MKDIR) -p $(build_doxygen_targetdir)
 	$(Q)$(RM) $(build_doxygen_logfile)
+	$(Q)$(RM) $(build_doxygen_success)
 	$(Q)$(RM) -rf $(build_doxygen_targetdir)/*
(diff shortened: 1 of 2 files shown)


cabelitos:
+1

PR #2262, opened 2016-07-19
build: increment version for V1 release

bdilly:
V1 is a stable C API release, done after almost
3k public commits and about 40 contributors.

Really nice =)


barbieri:
+1, remember to write some NEWS or release notes in the tag message :-)

I may be wrong, but I guess we're not declaring all the API stable, then we should state which headers are stable and which are still subject to changes.

bdilly:
merged

PR #2264, opened 2016-07-21
LWM2M: Access Control [v3]

bsmelo:
- Main changes from v2 (#2255):
 - Minor changes suggested in the review
 - New CoAP API `sol_coap_notify_by_callback`, to support LWM2M Authorization for the Notify operation.

- Future Improvements
 - Add tests to `test-lwm2m.c`

### DIFF
--- a/src/lib/comms/include/sol-coap.h
+++ b/src/lib/comms/include/sol-coap.h
@@ -760,6 +760,33 @@ int sol_coap_notify(struct sol_coap_server *server,
     struct sol_coap_resource *resource, struct sol_coap_packet *pkt);
 
 /**
+ * @brief Sends the notification packet returned by a callback
+ * to all registered observers.
+ *
+ * Sends the notification @c pkt to all the registered observers for the
+ * resource @a resource, setting the correct token for each instance. The @c pkt
+ * will be filled by a callback @c cb given to this function, probably based on the
+ * observer's address and/or any other used data.
+ *
+ * @param server The server through which the notifications will be sent.
+ * @param resource The resource the notification is about.
+ * @param cb A callback that is used to create the @c pkt - @c data User data; @c server The server through which the notifications will be sent; @c resource The resource the notification is about; @c addr The address of the observer; @c pkt The pkt to be filled.
+ * @param cb_data The user data to @c cb.
+ *
+ * @return 0 on success, -errno on failure.
+ *
+ * @see sol_coap_notify()
+ * @see sol_coap_send_packet()
+ * @see sol_coap_send_packet_with_reply()
+ */
+int sol_coap_notify_by_callback(struct sol_coap_server *server,
+    struct sol_coap_resource *resource,
+    int (*cb)(void *cb_data, struct sol_coap_server *server,
+    struct sol_coap_resource *resource, struct sol_network_link_addr *addr,
+    struct sol_coap_packet **pkt),
+    const void *cb_data);
+
+/**
  * @brief Register a #sol_coap_resource with the server.
  *
  * Registers the given resource with the server, in order to be able to
--- a/src/lib/comms/sol-coap.c
+++ b/src/lib/comms/sol-coap.c
@@ -969,9 +969,13 @@ find_context(struct sol_coap_server *server, const struct sol_coap_resource *res
     return NULL;
 }
 
-SOL_API int
-sol_coap_notify(struct sol_coap_server *server,
-    struct sol_coap_resource *resource, struct sol_coap_packet *pkt)
+static int
+sol_coap_notify_full(struct sol_coap_server *server,
+    struct sol_coap_resource *resource, struct sol_coap_packet *pkt,
+    int (*cb)(void *cb_data, struct sol_coap_server *server,
+    struct sol_coap_resource *resource, struct sol_network_link_addr *addr,
+    struct sol_coap_packet **pkt),
+    const void *cb_data)
 {
     struct resource_observer *o;
     struct resource_context *c;
@@ -979,37 +983,52 @@ sol_coap_notify(struct sol_coap_server *server,
     uint16_t i;
     int r = -EINVAL;
 
-    SOL_NULL_CHECK(pkt, -EINVAL);
-    SOL_NULL_CHECK_GOTO(server, error_pkt);
-    SOL_NULL_CHECK_GOTO(resource, error_pkt);
+    SOL_NULL_CHECK_GOTO(server, err_pkt);
+    SOL_NULL_CHECK_GOTO(resource, err_pkt);
 
-    COAP_RESOURCE_CHECK_API_GOTO(error_pkt);
+    COAP_RESOURCE_CHECK_API_GOTO(err_pkt);
 
     c = find_context(server, resource);
     if (!c) {
-        SOL_WRN("Context not found for expecified resource");
+        SOL_WRN("Context not found for specified resource");
         r = -ENOENT;
-        goto error_pkt;
+        goto err_pkt;
     }
 
     r = 0;
     SOL_PTR_VECTOR_FOREACH_IDX (&c->observers, o, i) {
         uint8_t type, code;
 
+        if (cb) {
+            r = cb((void *)cb_data, server, resource, &o->cliaddr, &pkt);
+            if (r < 0 && r != -EPERM) {
+                SOL_WRN("Error creating notification packet. Reason: %d", r);
+
+                goto err_pkt;
+            } else if (r == -EPERM) {
+                SOL_BUFFER_DECLARE_STATIC(addr, SOL_NETWORK_INET_ADDR_STR_LEN);
+
+                sol_network_link_addr_to_str(&o->cliaddr, &addr);
+                SOL_WRN("Observer at %.*s is not authorized for CoAP Resource %p",
+                    SOL_STR_SLICE_PRINT(sol_buffer_get_slice(&addr)), resource);
+                continue;
+            }
+        }
+
         header = sol_coap_packet_new(NULL);
         if (!header) {
             r = -ENOMEM;
-            goto error_pkt;
+            goto err_pkt;
         }
 
         sol_coap_header_get_code(pkt, &code);
         r = sol_coap_header_set_code(header, code);
-        SOL_INT_CHECK_GOTO(r, < 0, err);
+        SOL_INT_CHECK_GOTO(r, < 0, err_header);
         sol_coap_header_get_type(pkt, &type);
         r = sol_coap_header_set_type(header, type);
-        SOL_INT_CHECK_GOTO(r, < 0, err);
+        SOL_INT_CHECK_GOTO(r, < 0, err_header);
         r = sol_coap_header_set_token(header, o->token, o->tkl);
-        SOL_INT_CHECK_GOTO(r, < 0, err);
+        SOL_INT_CHECK_GOTO(r, < 0, err_header);
 
         r = enqueue_packet(server, pkt, header, &o->cliaddr);
         if (r < 0) {
@@ -1018,20 +1037,48 @@ sol_coap_notify(struct sol_coap_server *server,
             sol_network_link_addr_to_str(&o->cliaddr, &addr);
             SOL_WRN("Failed to enqueue packet %p to %.*s", header,
                 SOL_STR_SLICE_PRINT(sol_buffer_get_slice(&addr)));
-            goto error_pkt;
+            goto err_header;
         }
 
+        if (cb)
+            sol_coap_packet_unref(pkt);
         sol_coap_packet_unref(header);
     }
 
-    sol_coap_packet_unref(pkt);
+    if (!cb)
+        sol_coap_packet_unref(pkt);
+
     return r;
 
-err:
+err_header:
     sol_coap_packet_unref(header);
-error_pkt:
+err_pkt:
     sol_coap_packet_unref(pkt);
     return r;
+}
+
+SOL_API int
+sol_coap_notify(struct sol_coap_server *server,
+    struct sol_coap_resource *resource, struct sol_coap_packet *pkt)
+{
+    SOL_NULL_CHECK(pkt, -EINVAL);
+
+    return sol_coap_notify_full(server, resource, pkt, NULL, NULL);
+}
+
+SOL_API int
+sol_coap_notify_by_callback(struct sol_coap_server *server,
+    struct sol_coap_resource *resource,
+    int (*cb)(void *cb_data, struct sol_coap_server *server,
+    struct sol_coap_resource *resource, struct sol_network_link_addr *addr,
+    struct sol_coap_packet **pkt),
+    const void *cb_data)
+{
+    struct sol_coap_packet *pkt = NULL;
+
+    SOL_NULL_CHECK(cb, -EINVAL);
+
+    return sol_coap_notify_full(server, resource, pkt, cb, cb_data);
 }
 
 SOL_API struct sol_coap_packet *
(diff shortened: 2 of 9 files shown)


src/lib/comms/sol-lwm2m.c
```diff
@@ -4759,6 +5451,14 @@ sol_lwm2m_client_start(struct sol_lwm2m_client *client)
...
     return 0;
 
+err_access_control:
+    if (client->supports_access_control) {
+        ctx = find_object_ctx_by_id(client, ACCESS_CONTROL_OBJECT_ID);
+        while ((instance = sol_vector_get(&ctx->instances, 0)) != NULL) {
+            obj_instance_clear(client, ctx, instance);
+            sol_vector_del_element(&ctx->instances, instance);
+        }
+    }
```
edersondisouza: As you are actually deleting all elements, it's better to just iterate over all of them calling `obj_instance_clear`, and after that delete the whole vector by calling `sol_vector_clear`. It avoids unnecessary reallocs while deleting elements.

src/lib/comms/sol-lwm2m.c
```diff
@@ -2973,6 +3190,22 @@ handle_delete(struct sol_lwm2m_client *client,
...
+
+    return ret;
+
+err_access_control:
+    if (client->supports_access_control) {
+        acc_obj_ctx = find_object_ctx_by_id(client, ACCESS_CONTROL_OBJECT_ID);
+        while ((acc_obj_instance = sol_vector_get(&acc_obj_ctx->instances, 0)) != NULL) {
+            obj_instance_clear(client, acc_obj_ctx, acc_obj_instance);
+            sol_vector_del_element(&acc_obj_ctx->instances, acc_obj_instance);
+        }
```
edersondisouza: Ditto.

src/lib/comms/sol-lwm2m.c
```diff
@@ -2198,13 +2225,84 @@ sol_lwm2m_resource_init(struct sol_lwm2m_resource *resource,
...
 }
 
+SOL_API int
+sol_lwm2m_resource_init_vector(struct sol_lwm2m_resource *resource,
+    uint16_t id, enum sol_lwm2m_resource_data_type data_type,
+    struct sol_vector *res_instances)
+{
+    uint16_t i;
+    int r = -EINVAL;
+    uint16_t resource_len = res_instances->len;
```
edersondisouza: This will segfault it `res_instances` is NULL.

src/lib/comms/sol-lwm2m.c
```diff
@@ -2198,13 +2225,84 @@ sol_lwm2m_resource_init(struct sol_lwm2m_resource *resource,
...
+
+    if (!resource || data_type == SOL_LWM2M_RESOURCE_DATA_TYPE_NONE ||
+        !resource_len)
+        return -EINVAL;
```
edersondisouza: Any reason why not use SOL_CHECKS here?

src/lib/comms/sol-coap.c
```diff
@@ -969,47 +969,66 @@ find_context(struct sol_coap_server *server, const struct sol_coap_resource *res
...
         r = sol_coap_header_set_code(header, code);
-        SOL_INT_CHECK_GOTO(r, < 0, err);
+        SOL_INT_CHECK_GOTO(r, < 0, err_header);
         sol_coap_header_get_type(pkt, &type);
         r = sol_coap_header_set_type(header, type);
-        SOL_INT_CHECK_GOTO(r, < 0, err);
+        SOL_INT_CHECK_GOTO(r, < 0, err_header);
         r = sol_coap_header_set_token(header, o->token, o->tkl);
-        SOL_INT_CHECK_GOTO(r, < 0, err);
+        SOL_INT_CHECK_GOTO(r, < 0, err_header);
```
cabelitos: This will be only necessary is `!cb`, otherwise the packet returned by `cb` will already contains the correct token and stuff. 
The user is supposed to use `sol_coap_notification_packet_new()`

src/lib/comms/include/sol-coap.h
```diff
@@ -760,6 +760,33 @@ int sol_coap_notify(struct sol_coap_server *server,
     struct sol_coap_resource *resource, struct sol_coap_packet *pkt);
 
 /**
+ * @brief Sends the notification packet returned by a callback
```
cabelitos: @vcgomes Could you please take a look at this patch?
thanks

src/lib/comms/sol-lwm2m.c
cabelitos: Agreed

bsmelo: What about a case like this one: https://github.com/solettaproject/soletta/pull/2264/files#diff-e4058dc14a343a8aceb7210b323358a2R3174 ?

If I successfully delete 2 instances and fail when deleting the 3rd, it would still call `sol_vector_clear` outside the loop, trying to delete a `sol_vector` with elements still in it (at least the 3rd instance, in this case), and cause a bigger failure, wouldn't it? I mean, if I call `sol_vector_del_element` right after `obj_instance_clear`, it seems to me that I'm assuring more consistency.

src/lib/comms/sol-lwm2m.c
bsmelo: I will add a `SOL_NULL_CHECK` before the attribution, thanks!

src/lib/comms/sol-lwm2m.c
bsmelo: Not really, I just based myself on existing code: https://github.com/solettaproject/soletta/pull/2264/files#diff-e4058dc14a343a8aceb7210b323358a2R2169

I will change it in both places.

src/lib/comms/sol-lwm2m.c
edersondisouza: In there, there are checks, not all elements are deleted. Here you are deleting all elements. It doesn't compare.

bsmelo: All elements should be deleted in there too (it's LWM2M Bootstrap Delete on '/'). But I see your point, I will change these 2 you suggested.

edersondisouza: =D
But now I'm curious: leaving 'dangling' elements on vector there is a good thing? 
@cabelitos ^

cabelitos: @bsmelo In that case I think we should still remove from the vector even if the del callback returns < 0

src/lib/comms/sol-lwm2m.c
```diff
@@ -4733,6 +5039,14 @@ sol_lwm2m_client_start(struct sol_lwm2m_client *client)
 
     SOL_NULL_CHECK(client, -EINVAL);
 
+    if (client->supports_access_control &&
+        client->need_to_setup_access_control) {
+        r = setup_access_control_object_instances(client);
+        SOL_INT_CHECK_GOTO(r, < 0, err_access_control);
```
cabelitos: Be carefull, becuase if you jump to this goto, you will call  sol_lwm2m_resource_clear(&res[0]);, sol_lwm2m_resource_clear(&res[1]); and sol_lwm2m_resource_clear(&res[2]); Which are uninitialized. Thus can give as a headache

src/lib/comms/sol-lwm2m.c
```diff
@@ -3605,8 +3854,14 @@ dispatch_notifications(struct sol_lwm2m_client *client,
             resource->path[path_idx]))
             continue;
 
-        r = send_notification_pkt(client, obj_ctx, NULL, -1, obj_ctx->obj_res);
-        SOL_EXP_CHECK(!r, false);
+        ctx.obj_ctx = obj_ctx;
+        ctx.resource_id = -1;
+        r = sol_coap_notify_by_callback(client->coap_server, obj_ctx->obj_res,
```
cabelitos: By carefull `r` is a bool and not an int!

src/lib/comms/sol-lwm2m.c
```diff
@@ -4718,6 +4859,171 @@ client_bootstrap(void *data)
...
+        // one ACL will be created per existing [Server ID]
+        SOL_VECTOR_FOREACH_IDX (&obj_ctx->instances, instance, i) {
+            r = read_resources(client, obj_ctx, instance, &res, 1,
+                SERVER_OBJECT_SERVER_ID);
+            if (r < 0) {
+                SOL_WRN("Could not read Server Object's [Server ID] resource\n");
+                return r;
+            }
+            res_data = sol_vector_append(&acl_instances);
+            res_data->id = res.data->content.integer;
```
cabelitos: missing null check

src/lib/comms/sol-lwm2m.c
```diff
@@ -4718,6 +4859,171 @@ client_bootstrap(void *data)
...
+        //Else, there's no way to know which Server is authorized to
+        // instantiate which Object, and thus <no Server> should be authorized
+        // to instantiate <any> Object. However, since there's no way for the Bootstrap
+        // Server to Read /2, in order to get the Access Control Instance relevant for
+        // 'C'reate on a given Object, and change the ACLs granting access rights to a given
+        // Server, <every Server> will be authorized to instantiate <every> Object
+        SOL_DBG("LWM2M Server object not provided! No factory bootstrap.");
+
+        res_data = sol_vector_append(&acl_instances);
+        res_data->id = DEFAULT_SHORT_SERVER_ID;
```
cabelitos: missing null check

src/lib/comms/sol-lwm2m.c
bsmelo: Ok, changed for all cases here. :)

src/lib/comms/sol-coap.c
```diff
@@ -1018,22 +1037,50 @@ sol_coap_notify(struct sol_coap_server *server,
...
+
+SOL_API int
+sol_coap_notify_by_callback(struct sol_coap_server *server,
+    struct sol_coap_resource *resource,
+    int (*cb)(void *cb_data, struct sol_coap_server *server,
+    struct sol_coap_resource *resource, struct sol_network_link_addr *addr,
+    struct sol_coap_packet **pkt),
+    const void *cb_data)
+{
+    struct sol_coap_packet *pkt = NULL;
```
cabelitos: You don't need to do this here. Create this variable localy inside sol_coap_notify_full

src/lib/comms/sol-coap.c
```diff
@@ -969,47 +969,66 @@ find_context(struct sol_coap_server *server, const struct sol_coap_resource *res
...
-        goto error_pkt;
+        goto err_pkt;
     }
 
     r = 0;
     SOL_PTR_VECTOR_FOREACH_IDX (&c->observers, o, i) {
         uint8_t type, code;
 
+        if (cb) {
+            r = cb((void *)cb_data, server, resource, &o->cliaddr, &pkt);
```
vcgomes: What's the reason of creating the packet inside the callback? I.e. what does change from one observer to another?

What I had in mind is that this callback would be basically a filter, I wouldn't send a notification to that observer. Honest question: what more does it need to do?

src/lib/comms/sol-coap.c
```diff
@@ -1018,22 +1037,50 @@ sol_coap_notify(struct sol_coap_server *server,
...
         }
 
+        if (cb)
+            sol_coap_packet_unref(pkt);
         sol_coap_packet_unref(header);
     }
 
-    sol_coap_packet_unref(pkt);
+    if (!cb)
+        sol_coap_packet_unref(pkt);
```
cabelitos: You messing the variable pkt around, using for different purposes. I suggest that you stop trying to unref it here. unref it in sol_coap_notify.

src/lib/comms/sol-coap.c
```diff
@@ -1018,22 +1037,50 @@ sol_coap_notify(struct sol_coap_server *server,
...
+
     return r;
 
-err:
+err_header:
     sol_coap_packet_unref(header);
-error_pkt:
+err_pkt:
     sol_coap_packet_unref(pkt);
```
cabelitos: Same thing, do not unref it. Let the caller function unref it

src/lib/comms/sol-coap.c
bsmelo: Done.

src/lib/comms/sol-lwm2m.c
bsmelo: Nice catch... Changed above to `    struct sol_lwm2m_resource res[3] = { };`

src/lib/comms/sol-lwm2m.c
bsmelo: Oops, changed to int.
Now I know why the compiler was complaining about my `SOL_INT_CHECK(r, < 0, false);` and I had to use `if`s. Changed that too, thanks.

src/lib/comms/sol-coap.c
cabelitos: Different servers may receive different information

src/lib/comms/sol-lwm2m.c
```diff
@@ -4647,6 +5166,171 @@ client_bootstrap(void *data)
...
+    obj_ctx = find_object_ctx_by_id(client, SERVER_OBJECT_ID);
+    if (obj_ctx && obj_ctx->instances.len > 0) {
+        //If any Server Object Instance exists, this is already Bootstrapped and
+        // one ACL will be created per existing [Server ID]
+        SOL_VECTOR_FOREACH_IDX (&obj_ctx->instances, instance, i) {
+            r = read_resources(client, obj_ctx, instance, &res, 1,
+                SERVER_OBJECT_SERVER_ID);
+            if (r < 0) {
+                SOL_WRN("Could not read Server Object's [Server ID] resource\n");
+                return r;
```
cabelitos: leaking acl_instances

src/lib/comms/sol-lwm2m.c
```diff
@@ -4647,6 +5166,171 @@ client_bootstrap(void *data)
...
+        // and Access Control Objects
+        if (obj_ctx->obj->id != SECURITY_SERVER_OBJECT_ID &&
+            obj_ctx->obj->id != SERVER_OBJECT_ID &&
+            obj_ctx->obj->id != ACCESS_CONTROL_OBJECT_ID) {
+            SOL_SET_API_VERSION(res.api_version = SOL_LWM2M_RESOURCE_API_VERSION; )
+            r = sol_lwm2m_resource_init_vector(&res, ACCESS_CONTROL_OBJECT_ACL_RES_ID,
+                SOL_LWM2M_RESOURCE_DATA_TYPE_INT, &acl_instances);
+            if (r < 0) {
+                SOL_WRN("Could not init Access Control Object's [ACL] resource\n");
+                return r;
```
cabelitos: leaking acl_instances

src/lib/comms/sol-lwm2m.c
bsmelo: Added
```C
        r = -ENOMEM;
        SOL_NULL_CHECK_GOTO(res_data, err_vector);
...
err_vector:
    sol_vector_clear(&acl_instances);
    return r;
```

src/lib/comms/sol-coap.c
bsmelo: I think it's more clear this way, since `sol_coap_notify` and `sol_coap_notify_full` already has a local with this name (third parameter).

src/lib/comms/sol-lwm2m.c
bsmelo: Fixed.

src/lib/comms/sol-lwm2m.c
bsmelo: Fixed.

src/lib/comms/sol-lwm2m.c
bsmelo: Fixed above.

src/lib/comms/sol-lwm2m.c
bsmelo: Fixed.

src/lib/comms/sol-coap.c
bsmelo: Changed the memory management for `pkt`.

bsmelo:
An updated version is at #2265. Closing this. Thanks.

PR #2266, opened 2016-07-22
Network: Properly compare IPv6 Addresses mapped into IPv4

cabelitos:

### DIFF
--- a/src/lib/comms/include/sol-network.h
+++ b/src/lib/comms/include/sol-network.h
@@ -24,6 +24,7 @@
 #include <sol-vector.h>
 #include <sol-str-slice.h>
 #include <sol-buffer.h>
+#include <sol-util.h>
 
 #ifdef __cplusplus
 extern "C" {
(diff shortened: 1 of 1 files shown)


cabelitos:
@vcgomes @bsmelo

src/lib/comms/include/sol-network.h
```diff
@@ -234,31 +235,66 @@ static inline bool
...
+
+        struct ipv6_map_prefix {
+            const uint8_t zeroes[10];
+            const uint16_t ones;
+        } __attribute__ ((packed)) prefix = {
+            { 0 }, sol_util_be16_to_cpu(0xffff)
+        };
+        const uint8_t *addr_ipv6, *addr_ipv4;
+
+
```
bsmelo: Extra blank line.

src/lib/comms/include/sol-network.h
```diff
@@ -234,31 +235,66 @@ static inline bool
...
+    }
+
+    if ((a->family == SOL_NETWORK_FAMILY_INET &&
+         b->family == SOL_NETWORK_FAMILY_INET6) ||
+        (a->family == SOL_NETWORK_FAMILY_INET6 &&
+         b->family == SOL_NETWORK_FAMILY_INET)) {
+
+        struct ipv6_map_prefix {
+            const uint8_t zeroes[10];
+            const uint16_t ones;
```
vcgomes: From what I understand, this would be an `uint32_t`, please confirm.

bsmelo: Why is that?

Also, looked at the code and tested here: lgtm.

vcgomes: From `netinet/in.h`:

```C
# define IN6_IS_ADDR_V4MAPPED(a) \
  (__extension__							      \
   ({ const struct in6_addr *__a = (const struct in6_addr *) (a);	      \
      __a->s6_addr32[0] == 0						      \
      && __a->s6_addr32[1] == 0						      \
      && __a->s6_addr32[2] == htonl (0xffff); }))
```

The use of `htonl()` and the `s6_addr32` name makes me think that this should be 32-bit wide.

bsmelo: Hmm... Maybe I'm missing something, but I guess `s6_addr32[0]` through `s6_addr32[2]` just captures the first 96 bits of the IPv6 address, the same as `uint8_t zeroes[10]` through `uint16_t ones`. The difference is that in `netinet/in.h`, since it's using "32-bit chunks", `htonl()` just pads the last 32-bit chunk with 16 `0` bits before those 16 `1` bits from `0xffff`. `ipv6_map_prefix` just uses an 80-bit chunk followed by a 16-bit chunk instead.

If that's true, both cases seems to be correctly checking for a 96-bit portion with 80 `0`s followed by 16 `1`s. But I guess @cabelitos can clear it out a little better. :)

src/lib/comms/include/sol-network.h
cabelitos: indeed

src/lib/comms/include/sol-network.h
cabelitos: IPV6 are 128 bits.
So the first 80 bits should be zero.
The next 16 bits should contain ffff
the next 32bits is the ipv4 itself.

cabelitos: http://www.tcpipguide.com/free/t_IPv6IPv4AddressEmbedding-2.htm

vcgomes: Nice. LGTM.

cabelitos:
merged

PR #2267, opened 2016-07-27
LWM2M: Split the module between different files

bsmelo:
Since the LWM2M module has been increasing a lot latelly, this patch splits the functionality of the module between different, specific files, in order to make maintenance easier.

Old file structure organization:
```
src/lib/comms/include/sol-lwm2m.h
src/lib/comms/sol-lwm2m.c
```

New file structure organization:
```
src/lib/comms/include/sol-lwm2m.h
src/lib/comms/include/sol-lwm2m-client.h
src/lib/comms/include/sol-lwm2m-server.h
src/lib/comms/include/sol-lwm2m-bs-server.h
src/lib/comms/sol-lwm2m-client.c
src/lib/comms/sol-lwm2m-server.c
src/lib/comms/sol-lwm2m-bs-server.c
src/lib/comms/sol-lwm2m-common.h
src/lib/comms/sol-lwm2m-common.c
```

Besides a few `-Wunused-function` warnings I couldn't get rid of, I belive I'm missing something related to the logging infrastructure and the `"SOL_LOG_DOMAIN" redefined` warnings as well.

When running the samples after `export SOL_LOG_LEVEL="DBG"` I can see logging related to the infrastructure for that specific sample (`sol-lwm2m-{client, server, bs-server}.c`) but I can't see stuff from `sol-lwm2m-common.{h,c}` (at least not all of it, from some tests I made here).


src/lib/comms/sol-lwm2m-common.h
```diff
@@ -0,0 +1,679 @@
...
+#pragma once
+
+#include <errno.h>
+#include <float.h>
+#include <stdlib.h>
+#include <time.h>
+#include <math.h>
+#include <ctype.h>
+
+#define SOL_LOG_DOMAIN &_lwm2m_domain
```
barbieri: you shouldn't export the `SOL_LOG_DOMAIN` here since this file is included by others, which will lead them to use this instead of theirs.

src/lib/comms/sol-lwm2m-common.h
```diff
@@ -0,0 +1,679 @@
...
+#include "sol-macros.h"
+#include "sol-mainloop.h"
+#include "sol-monitors.h"
+#include "sol-random.h"
+#include "sol-str-slice.h"
+#include "sol-str-table.h"
+#include "sol-util.h"
+#include "sol-http.h"
+
+SOL_LOG_INTERNAL_DECLARE_STATIC(_lwm2m_domain, "lwm2m");
```
barbieri: likewise, this should go in the `.c`

src/lib/comms/sol-lwm2m-common.h
```diff
@@ -0,0 +1,679 @@
...
+    uint32_t lifetime;
+};
+
+struct sol_lwm2m_client_object {
+    struct sol_ptr_vector instances;
+    uint16_t id;
+};
+
+static void
+send_ack_if_needed(struct sol_coap_server *coap, struct sol_coap_packet *msg,
```
cabelitos: I don't think we should have all this functions in the header file. Move them to sol-lwm2m-common.c

cabelitos: If we will leave them here, you should add static inline. Otherwise the compiler will complain

bsmelo:
An updated version is at #2269. Closing this. Thanks.

PR #2269, opened 2016-07-28
LWM2M: Split the module between different files [v2]

bsmelo:
Main changes from v1 (#2267):
- Moved the functions from `sol_lwm2m_common.h` to `sol_lwm2m_common.c`, removing the `static` keyword.
- Added 2 functions `sol_lwm2m_common_init()` and `sol_lwm2m_common_shutdown()` responsible for initializing the logging for the common functions, as suggested by @barbieri and discussed with @cabelitos. These 2 functions are called from `sol-mainloop.c::sol_init()`.


src/lib/comms/sol-lwm2m-common.c
```diff
@@ -0,0 +1,976 @@
...
+#include "sol-random.h"
+#include "sol-str-slice.h"
+#include "sol-str-table.h"
+#include "sol-util.h"
+#include "sol-http.h"
+
+SOL_LOG_INTERNAL_DECLARE_STATIC(_lwm2m_common_domain, "lwm2m-common");
+
+int sol_lwm2m_common_init(void);
+void sol_lwm2m_common_shutdown(void);
```
cabelitos: Why are you declaring them here?

bsmelo: I went directly with the function definition here, instead of declaration... but the compiler threw a warning that there was no declaration :/
So I declared them here, and defined in the end of this file. Better options?

barbieri: no, that's the way since we require forward definition for non-static functions (so people remember to add them to headers or to make them static). These are the exceptions and we're handling like you did.

cabelitos: Indeed. Sorry

Issue #2235, opened 2016-07-13
Proposal: IIO: Activate device feature before read raw data


laykuanloon:
Task Description

Hi @barbieri , @edersondisouza , @yongli3 

There is a need to activate device feature before read raw data for some sensor. 

The apds9930 ambient light and proximity sensor which have following entry that need to be enable before read raw data.
/sys/.../iio:deviceX/in_proximity_en
/sys/.../iio:deviceX/in_illuminance_en

Following is the description and sysfs entry example for enable iio device feature:
What:		/sys/.../iio:deviceX/in_energy_en
What:		/sys/.../iio:deviceX/in_distance_en
What:		/sys/.../iio:deviceX/in_velocity_sqrt(x^2+y^2+z^2)_en
What:		/sys/.../iio:deviceX/in_steps_en
KernelVersion:	3.19
Contact:	linux-iio@vger.kernel.org
Description:
		Activates a device feature that runs in firmware/hardware.
		E.g. for steps: the pedometer saves power while not used;
		when activated, it will count the steps taken by the user in
		firmware and export them through in_steps_input.

Thanks.

Dependencies

N/A

barbieri:
For ease of use, we want to include this in our side in some way (to be thought), but these can be done at the "system level" with some script that setups the modules. Something analogous to "modprobe.conf" and "modules-autoload.conf" that would setup the sensors. I've discussed with @yongli3 team in the past as we already setup i2c devices. The outcome of that discussion was that it should be done with some devicetree or ACPI overlay that the board/system provides to state which extra devices were attached, doing the setup at a single place. The steps counter could be also configured at that level.

However, data such as steps counter is more like an application or runtime configuration, maybe you enable or disable that feature based on some need. In this case it may result in some extra IIO device ports in addition to node type options, so we can dynamically change them.

yongli3:
@barbieri After discussed with @laykuanloon , we think we can add a logic in soletta, when open a channel, to check if there is any in_???_en sysfs entities, if exits, then write 1 to enable it, before close the channel, write 0 to the entities to save power. How do you think about it ?

barbieri:
I think it's a good approach, we should test to see if any problems will appear, but overall looks good.

laykuanloon:
@barbieri and @yongli3 

I already submitted the PR. Please take a look
#2270

laykuanloon:
Close this since already merge.

PR #2270, opened 2016-08-01
IIO: activate firmware/hardware feature before read out sensor data

laykuanloon:
Some device firmware/hardware feature default state is off, thus need to
activate it before read out sensor data. Do activate in sol_iio_add_channel()
and deactivate in sol_iio_close().

Activate `'in_<channel>_en'` before read `'in_<channel>_input'` or `'in_<channel>_raw'`.
For example: activate 'in_steps_en' before read 'in_steps_input'.

### DIFF
--- a/src/lib/io/sol-iio.c
+++ b/src/lib/io/sol-iio.c
@@ -104,6 +104,7 @@ struct resolve_absolute_path_data {
 #define SYSFS_TRIGGER_SYSFS_ADD_TRIGGER SYSFS_DEVICES_PATH "/iio_sysfs_trigger/add_trigger"
 #define SYSFS_TRIGGER_SYSFS_PATH SYSFS_DEVICES_PATH "/iio_sysfs_trigger"
 
+#define CHANNEL_ENABLE_PATH SYSFS_DEVICES_PATH "/iio:device%d/%s_en"
 #define CHANNEL_RAW_PATH SYSFS_DEVICES_PATH "/iio:device%d/%s_raw"
 #define CHANNEL_PROCESSED_PATH SYSFS_DEVICES_PATH "/iio:device%d/%s_input"
 #define CHANNEL_OFFSET_PATH SYSFS_DEVICES_PATH "/iio:device%d/%s_offset"
@@ -362,6 +363,24 @@ set_buffer_enabled(struct sol_iio_device *device, bool enabled)
 
     if (sol_util_write_file(path, "%d", enabled) < 0)
         return false;
+
+    return true;
+}
+
+static bool
+set_channel_enabled(struct sol_iio_device *device, const char *channel_name, bool enabled)
+{
+    char path[PATH_MAX];
+    bool r;
+
+    r = craft_filename_path(path, sizeof(path), CHANNEL_ENABLE_PATH, device->device_id, channel_name);
+    if (!r)
+        return false;
+
+    if (check_file_existence(path)) {
+        if (sol_util_write_file(path, "%d", enabled) < 0)
+            return false;
+    }
 
     return true;
 }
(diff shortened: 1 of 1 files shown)


src/lib/io/sol-iio.c
```diff
@@ -973,6 +994,9 @@ sol_iio_add_channel(struct sol_iio_device *device, const char *name, const struc
     r = sol_ptr_vector_append(&channel->device->channels, channel);
     SOL_INT_CHECK_GOTO(r, < 0, error);
 
+    /* Some device firmware/hardware feature default state is off, thus need to activate it before read */
+    set_channel_enabled(device, name, true);
```
bdilly: Hi @laykuanloon ,

don't you think it could be useful to log case set_channel_enabled fail (return false)?

laykuanloon: @bdilly 
Agree, I already push in the changes. Thanks.

src/lib/io/sol-iio.c
```diff
@@ -870,6 +889,8 @@ sol_iio_close(struct sol_iio_device *device)
     SOL_NULL_CHECK(device);
 
     SOL_PTR_VECTOR_FOREACH_REVERSE_IDX (&device->channels, channel, i) {
+        /* Deactivate device firmware/hardware after use */
+        set_channel_enabled(device, channel->name, false);
```
edersondisouza: Maybe put this call into `iio_del_channel`?

laykuanloon: @edersondisouza, @bdilly, @yongli3, @barbieri 

Currently the `iio_del_channel(struct sol_iio_channel *channel) `only take one parameter. 

If want to` set_channel_enabled(false)` in `iio_del_channel()`, then need change to `iio_del_channel(sol_iio_device *device, struct sol_iio_channel *channel)` 

Because in set_channel_enabled() need the device to craft the path to deactivate 

All agree with the above changes? If yes then I will push the changes.

edersondisouza: Sounds fair: it's better to centralise this operation. So if code flow changes in future, any call to `iio_del_channel` is sure to do a proper clean up.

laykuanloon:
@edersondisouza,

Push in the changes already.

src/lib/io/sol-iio.c
```diff
@@ -366,6 +367,24 @@ set_buffer_enabled(struct sol_iio_device *device, bool enabled)
...
+set_channel_enabled(struct sol_iio_device *device, const char *channel_name, bool enabled)
+{
+    char path[PATH_MAX];
+    bool r;
+
+    r = craft_filename_path(path, sizeof(path), CHANNEL_ENABLE_PATH, device->device_id, channel_name);
+    if (!r)
+        return false;
+
+    if (check_file_existence(path)) {
```
bdilly: @laykuanloon I would like to suggest you to verify if you're using this call ^ (check_file_existence()) correctly.
It doesn't seem the case to me.
I'm almost sure you just want to call sol_util_write_file() directly after crafting the file name.

edersondisouza: @laykuanloon @bdilly 
I think that it's important to verify file existence, as it's not always there - only for some devices, as per #2235

laykuanloon: @bdilly 
@edersondisouza  is right. The _en sysfs entry not always there. If is not there then `set_channel_enabled`  just return true

edersondisouza: @laykuanloon Do not forget to update use of `check_file_existence` - it already does craft path, no need to call `craft_filename_path` before. Check other usage, please.

laykuanloon: @edersondisouza 
I need to craft path for sol_util_write_file(). Should I change check_file_existence to !stat(path, &st)  ?

edersondisouza: Oh, it's true, we overlooked it :sweat_smile:

bdilly:
@laykuanloon so I'm merging it.

Also I'll grant you write access to this repository.
I just would ask you to read our Contribution, wiki page, specially the section regarding contribution flow: https://github.com/solettaproject/soletta/wiki/Contributing#contributions-flow

PR #2271, opened 2016-08-02
Some fixes for Soletta

cabelitos:
@ceolin Could you check eabb5eb ?
@bsmelo

### DIFF
--- a/src/lib/datatypes/include/sol-str-slice.h
+++ b/src/lib/datatypes/include/sol-str-slice.h
@@ -190,7 +190,6 @@ static inline void
 sol_str_slice_copy(char *dst, const struct sol_str_slice src)
 {
     memcpy(dst, src.data, src.len);
-    dst[src.len] = 0;
 }
 
 /**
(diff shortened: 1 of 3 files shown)


bsmelo:
`make check` has an error:
```
# Running ./build/stage/test/test-buffer
- fini_null_is_fine
- test_resize
- test_ensure
- test_set_slice
./src/test/test-buffer.c:120: test_set_slice: Assertion string_eq("Hello+", "Hello") failed.
```

Besides that, lgtm.

src/lib/datatypes/include/sol-str-slice.h
```diff
@@ -190,7 +190,6 @@ static inline void
 sol_str_slice_copy(char *dst, const struct sol_str_slice src)
 {
     memcpy(dst, src.data, src.len);
-    dst[src.len] = 0;
```
barbieri: this is not invalid, when you copy from a slice to a `char*` (string), it is expected to have space for the contents AND the trailing nul byte. Maybe this needs documentation to be enhanced?

eventually using `sol_buffer` is a better approach as it can choose to use nul-byte or not.

cabelitos:
#2283

PR #2272, opened 2016-08-03
LWM2M: Support for DTLS with Pre-Shared Key security mode

bsmelo:
This patch adds support for CoAP Data Encryption using DTLS with Pre-Shared Key mode (using `TLS_PSK_WITH_AES_128_CCM_8` Cipher Suite).

It includes modifications on the LWM2M Samples to demonstrate the support as well.

This patch depends on #2271.

### DIFF
--- a/src/lib/comms/include/sol-lwm2m-bs-server.h
+++ b/src/lib/comms/include/sol-lwm2m-bs-server.h
@@ -60,10 +60,13 @@ typedef struct sol_lwm2m_bootstrap_client_info sol_lwm2m_bootstrap_client_info;
  *
  * @param port The UDP port to be used.
  * @param known_clients An array with the name of all clients this server has Bootstrap Information for.
+ * @param sec_mode The DTLS Security Mode to be used.
+ * @param known_psks The Clients' Pre-Shared Keys this Bootstrap Server has previous knowledge of.
  * @return The LWM2M bootstrap server or @c NULL on error.
  */
 struct sol_lwm2m_bootstrap_server *sol_lwm2m_bootstrap_server_new(uint16_t port,
-    const char **known_clients);
+    const char **known_clients, enum sol_lwm2m_security_mode sec_mode,
+    struct sol_vector *known_psks);
 
 /**
  * @brief Adds a bootstrap request monitor to the server.
--- a/src/lib/comms/sol-lwm2m-client.c
+++ b/src/lib/comms/sol-lwm2m-client.c
@@ -64,57 +64,6 @@ struct resource_ctx {
     uint16_t id;
 };
 
-struct obj_instance {
-    uint16_t id;
-    bool should_delete;
-    char *str_id;
-    const void *data;
-    struct sol_vector resources_ctx;
-    struct sol_coap_resource *instance_res;
-};
-
-struct obj_ctx {
-    const struct sol_lwm2m_object *obj;
-    char *str_id;
-    struct sol_vector instances;
-    struct sol_coap_resource *obj_res;
-};
-
-struct sol_lwm2m_client {
-    struct sol_coap_server *coap_server;
-    struct lifetime_ctx lifetime_ctx;
-    struct sol_ptr_vector connections;
-    struct sol_vector objects;
-    struct sol_monitors bootstrap;
-    struct {
-        struct sol_timeout *timeout;
-        struct sol_blob *server_uri;
-    } bootstrap_ctx;
-    const void *user_data;
-    uint16_t splitted_path_len;
-    char *name;
-    char **splitted_path;
-    char *sms;
-    bool running;
-    bool removed;
-    bool is_bootstrapping;
-    bool supports_access_control;
-    bool need_to_setup_access_control;
-};
-
-struct server_conn_ctx {
-    struct sol_network_hostname_pending *hostname_handle;
-    struct sol_lwm2m_client *client;
-    struct sol_vector server_addr_list;
-    struct sol_coap_packet *pending_pkt; //Pending registration or bootstrap reply
-    int64_t server_id;
-    int64_t lifetime;
-    uint16_t port;
-    uint16_t addr_list_idx;
-    time_t registration_time;
-    char *location;
-};
-
 static bool lifetime_client_timeout(void *data);
 static int register_with_server(struct sol_lwm2m_client *client,
     struct server_conn_ctx *conn_ctx, bool is_update);
@@ -129,10 +78,6 @@ static int setup_access_control_object_instance_for_instance(
     struct sol_lwm2m_resource *acl_res, bool register_with_coap);
 static int setup_access_control_object_instances(
     struct sol_lwm2m_client *client);
-static int
-read_resources(struct sol_lwm2m_client *client,
-    struct obj_ctx *obj_ctx, struct obj_instance *instance,
-    struct sol_lwm2m_resource *res, size_t res_len, ...);
 
 static void
 dispatch_bootstrap_event_to_client(struct sol_lwm2m_client *client,
@@ -197,20 +142,6 @@ extract_path(struct sol_lwm2m_client *client, struct sol_coap_packet *req,
 
     *path_size = j;
     return 0;
-}
-
-static struct obj_ctx *
-find_object_ctx_by_id(struct sol_lwm2m_client *client, uint16_t id)
-{
-    uint16_t i;
-    struct obj_ctx *ctx;
-
-    SOL_VECTOR_FOREACH_IDX (&client->objects, ctx, i) {
-        if (ctx->obj->id == id)
-            return ctx;
-    }
-
-    return NULL;
 }
 
 static struct obj_instance *
@@ -420,15 +351,6 @@ clear_bootstrap_ctx(struct sol_lwm2m_client *client)
         client->bootstrap_ctx.timeout = NULL;
         client->bootstrap_ctx.server_uri = NULL;
     }
-}
-
-static void
-clear_resource_array(struct sol_lwm2m_resource *array, uint16_t len)
-{
-    uint16_t i;
-
-    for (i = 0; i < len; i++)
-        sol_lwm2m_resource_clear(&array[i]);
 }
 
 /* Returns 1 if authorized; 0 if unauthorized and < 0 if error */
@@ -1192,25 +1114,6 @@ err_exit:
 }
 
 static int
-get_server_id_by_link_addr(const struct sol_ptr_vector *connections,
-    const struct sol_network_link_addr *cliaddr, int64_t *server_id)
-{
-    struct server_conn_ctx *conn_ctx;
-    struct sol_network_link_addr *server_addr;
-    uint16_t i;
-
-    SOL_PTR_VECTOR_FOREACH_IDX (connections, conn_ctx, i) {
-        server_addr = sol_vector_get_no_check(&conn_ctx->server_addr_list, conn_ctx->addr_list_idx);
-        if (sol_network_link_addr_eq(cliaddr, server_addr)) {
-            *server_id = conn_ctx->server_id;
-            return 0;
-        }
-    }
-
-    return -ENOENT;
-}
-
-static int
 notification_cb(void *data, struct sol_coap_server *server,
     struct sol_coap_resource *resource, struct sol_network_link_addr *addr,
     struct sol_coap_packet **pkt)
@@ -1815,48 +1718,6 @@ sol_lwm2m_client_add_object_instance(struct sol_lwm2m_client *client,
 
 err_exit:
     sol_vector_del_element(&ctx->instances, instance);
-    return r;
-}
-
-static int
-read_resources(struct sol_lwm2m_client *client,
-    struct obj_ctx *obj_ctx, struct obj_instance *instance,
-    struct sol_lwm2m_resource *res, size_t res_len, ...)
-{
-    size_t i;
-    int r = 0;
-    va_list ap;
-
-    SOL_NULL_CHECK(obj_ctx->obj->read, -ENOTSUP);
-
-    va_start(ap, res_len);
-
-    // The va_list contains the resources IDs that should be read.
-    for (i = 0; i < res_len; i++) {
-        r = obj_ctx->obj->read((void *)instance->data,
-            (void *)client->user_data, client, instance->id,
-            (uint16_t)va_arg(ap, int), &res[i]);
-
-        if (r == -ENOENT) {
-            res[i].data_len = 0;
-            res[i].data = NULL;
-            continue;
-        }
-
-        LWM2M_RESOURCE_CHECK_API_GOTO(res[i], err_exit_api);
-        SOL_INT_CHECK_GOTO(r, < 0, err_exit);
-    }
-
-    va_end(ap);
-    return 0;
-
-#ifndef SOL_NO_API_VERSION
-err_exit_api:
-    r = -EINVAL;
-#endif
-err_exit:
-    clear_resource_array(res, i);
-    va_end(ap);
     return r;
 }
 
--- a/src/lib/comms/sol-lwm2m-client.c
+++ b/src/lib/comms/sol-lwm2m-client.c
@@ -1040,6 +1040,13 @@ handle_read(struct sol_lwm2m_client *client,
     uint16_t i;
     int r;
 
+    if ((obj_ctx->obj->id == SECURITY_OBJECT_ID) &&
+        (server_id != UINT16_MAX)) {
+        SOL_WRN("Only the Bootstrap Server is allowed to access the Security Object."
+            " Server ID %" PRId64 " trying to access it", server_id);
+        return SOL_COAP_RESPONSE_CODE_UNAUTHORIZED;
+    }
+
     if (client->supports_access_control && obj_instance) {
         r = check_authorization(client, server_id, obj_ctx->obj->id,
             obj_instance->id, SOL_LWM2M_ACL_READ);
@@ -1362,10 +1369,17 @@ handle_resource(void *data, struct sol_coap_server *server,
         goto exit;
     }
 
-    if (client->supports_access_control) {
-        r = get_server_id_by_link_addr(&client->connections, cliaddr, &server_id);
-        header_code = SOL_COAP_RESPONSE_CODE_INTERNAL_ERROR;
-        SOL_INT_CHECK_GOTO(r, < 0, exit);
+    r = get_server_id_by_link_addr(&client->connections, cliaddr, &server_id);
+    header_code = SOL_COAP_RESPONSE_CODE_INTERNAL_ERROR;
+    SOL_INT_CHECK_GOTO(r, < 0, exit);
+
+    if (path_size >= 1 &&
+        ((obj_ctx->obj->id == SECURITY_OBJECT_ID) &&
+        (server_id != UINT16_MAX))) {
+        SOL_WRN("Only the Bootstrap Server is allowed to access the Security Object."
+            " Server ID %" PRId64 " trying to access it", server_id);
+        header_code = SOL_COAP_RESPONSE_CODE_UNAUTHORIZED;
+        goto exit;
     }
 
     switch (method) {
--- a/src/lib/comms/sol-lwm2m-client.c
+++ b/src/lib/comms/sol-lwm2m-client.c
@@ -170,6 +170,10 @@ obj_instance_clear(struct sol_lwm2m_client *client, struct obj_ctx *obj_ctx,
         if (!client->removed && res_ctx->res) {
             sol_coap_server_unregister_resource(client->coap_server,
                 res_ctx->res);
+
+            if (client->security)
+                sol_coap_server_unregister_resource(client->dtls_server,
+                    res_ctx->res);
         }
         free(res_ctx->res);
         free(res_ctx->str_id);
@@ -178,6 +182,10 @@ obj_instance_clear(struct sol_lwm2m_client *client, struct obj_ctx *obj_ctx,
     if (!client->removed && obj_instance->instance_res) {
         sol_coap_server_unregister_resource(client->coap_server,
             obj_instance->instance_res);
+
+        if (client->security)
+            sol_coap_server_unregister_resource(client->dtls_server,
+                obj_instance->instance_res);
     }
     free(obj_instance->instance_res);
     free(obj_instance->str_id);
@@ -265,6 +273,12 @@ setup_resources_ctx(struct sol_lwm2m_client *client, struct obj_ctx *obj_ctx,
             r = sol_coap_server_register_resource(client->coap_server,
                 res_ctx->res, client);
             SOL_INT_CHECK_GOTO(r, < 0, err_exit);
+
+            if (client->security) {
+                r = sol_coap_server_register_resource(client->dtls_server,
+                    res_ctx->res, client);
+                SOL_INT_CHECK(r, < 0, r);
+            }
         }
     }
 
@@ -275,6 +289,11 @@ err_exit:
         if (res_ctx->res) {
             sol_coap_server_unregister_resource(client->coap_server,
                 res_ctx->res);
+
+            if (client->security)
+                sol_coap_server_unregister_resource(client->dtls_server,
+                    res_ctx->res);
+
             free(res_ctx->res);
         }
         free(res_ctx->str_id);
@@ -324,6 +343,12 @@ setup_instance_resource(struct sol_lwm2m_client *client,
         r = sol_coap_server_register_resource(client->coap_server,
             obj_instance->instance_res, client);
         SOL_INT_CHECK_GOTO(r, < 0, err_register);
+
+        if (client->security) {
+            r = sol_coap_server_register_resource(client->dtls_server,
+                obj_instance->instance_res, client);
+            SOL_INT_CHECK(r, < 0, r);
+        }
     }
 
     r = setup_resources_ctx(client, obj_ctx, obj_instance, register_with_coap);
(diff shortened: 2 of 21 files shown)


src/lib/comms/sol-lwm2m-security.c
```diff
@@ -0,0 +1,470 @@
...
+#include "sol-socket.h"
+#include "sol-str-slice.h"
+#include "sol-util-internal.h"
+#include "sol-vector.h"
+
+SOL_LOG_INTERNAL_DECLARE(_lwm2m_security_domain, "lwm2m-security");
+
+#ifdef DTLS
+
+struct sol_socket *sol_coap_server_get_socket(const struct sol_coap_server *server);
```
cabelitos: Why this is defined here?

src/lib/comms/sol-lwm2m-security.c
```diff
@@ -0,0 +1,470 @@
...
+    SOL_VECTOR_FOREACH_IDX (known_psks, stored_psk, i) {
+        if (sol_str_slice_eq(sol_str_slice_from_blob(stored_psk->id), id)) {
+            if (stored_psk->key->size != SOL_DTLS_PSK_KEY_LEN) {
+                SOL_WRN("The PSK '%.*s' is %zu-bytes long; expecting a %d-bytes long PSK",
+                    SOL_STR_SLICE_PRINT(sol_str_slice_from_blob(stored_psk->key)),
+                    stored_psk->key->size, SOL_DTLS_PSK_KEY_LEN);
+                return -EINVAL;
+            }
+
+            memcpy(psk, stored_psk->key->mem, stored_psk->key->size);
```
cabelitos: Are we 100% sure that `psk_len` is at least `stored_psk->key->size` ?

src/lib/comms/sol-lwm2m-security.c
```diff
@@ -0,0 +1,470 @@
...
+            return (ssize_t)SOL_DTLS_PSK_KEY_LEN;
```
cabelitos: Should we return `stored_psk->key->size` ?

src/lib/comms/sol-lwm2m-security.c
```diff
@@ -0,0 +1,470 @@
...
+    }
+
+    SOL_VECTOR_FOREACH_IDX (&obj_ctx->instances, instance, i) {
+        r = read_resources(lwm2m_client, obj_ctx, instance, res, 3,
+            SECURITY_SECURITY_MODE,
+            SECURITY_PUBLIC_KEY_OR_IDENTITY,
+            SECURITY_SECRET_KEY);
+        SOL_INT_CHECK_GOTO(r, < 0, err_clear);
+
+        //If -ENOENT when reading SECURITY_SECRET_KEY or
```
cabelitos: watch it out, because if `-ENOENT` is returned the `SOL_INT_CHECK_GOTO` above will jump to `err_clear`

src/lib/comms/sol-lwm2m-security.c
```diff
@@ -0,0 +1,470 @@
...
+
+    //If the caller expects a PSK_ID_LEN less than 16 bytes
+    if (id_len < SOL_DTLS_PSK_ID_LEN)
+        return -ENOBUFS;
+
+    if (ctx->type == LWM2M_SERVER) {
+        SOL_DBG("ctx->type = LWM2M_SERVER");
+        lwm2m_server = ctx->entity;
+        known_psks = lwm2m_server->known_psks;
+        cli_name = sol_str_slice_from_str("cli1");
```
cabelitos: OUch, what happens when you have multiple clients?

src/lib/comms/sol-lwm2m-security.c
```diff
@@ -0,0 +1,470 @@
...
+    } else {
+        SOL_DBG("ctx->type = LWM2M_BOOTSTRAP_SERVER");
+        lwm2m_bs_server = ctx->entity;
+        known_psks = lwm2m_bs_server->known_psks;
+        cli_name = sol_str_slice_from_str("cli1-bs");
```
cabelitos: Same thing

src/lib/comms/sol-lwm2m-security.c
cabelitos: Maybe we should provide a callback so lwm2m-server/bs-server can fetch the proper id for the client.

src/lib/comms/sol-lwm2m-security.c
```diff
@@ -0,0 +1,470 @@
...
+
+    SOL_VECTOR_FOREACH_IDX (&obj_ctx->instances, instance, i) {
+        r = read_resources(lwm2m_client, obj_ctx, instance, res, 4,
+            SECURITY_SECURITY_MODE,
+            SECURITY_SERVER_ID,
+            SECURITY_PUBLIC_KEY_OR_IDENTITY,
+            SECURITY_IS_BOOTSTRAP);
+        SOL_INT_CHECK_GOTO(r, < 0, err_clear);
+
+        //If -ENOENT when reading SECURITY_PUBLIC_KEY_OR_IDENTITY or
```
cabelitos: Some comment about the `-ENOENT`

src/lib/comms/sol-lwm2m-security.c
```diff
@@ -0,0 +1,470 @@
...
+
+            memcpy(id, res[2].data[0].content.blob->mem, psk_id_len);
+            clear_resource_array(res, sol_util_array_size(res));
+
+            return (ssize_t)psk_id_len;
+        }
+
+        clear_resource_array(res, sol_util_array_size(res));
+    }
+
```
cabelitos: We could add an WRN here saying that we could not find the server.

src/lib/comms/sol-lwm2m-security.c
```diff
@@ -0,0 +1,470 @@
...
+        break;
+    case LWM2M_SERVER:
+        ((struct sol_lwm2m_server *)(security->entity))->security = NULL;
+        break;
+    case LWM2M_BOOTSTRAP_SERVER:
+        ((struct sol_lwm2m_bootstrap_server *)(security->entity))->security = NULL;
+        break;
+    }
+
+    sol_util_clear_memory_secure(security, sizeof(*security));
```
cabelitos: This makes me thinks in something... All blobs that holds the PSKs must be freed using sol_util_clear_memory.
I guess we need a new blob type!

src/lib/comms/sol-lwm2m-security.c
```diff
@@ -0,0 +1,470 @@
...
+    enum sol_lwm2m_security_mode sec_mode)
+{
+    struct sol_lwm2m_security *security;
+    struct sol_socket *socket_dtls = NULL;
+    int r = 0;
+
+    ssize_t (*get_psk_cb)(const void *creds, struct sol_str_slice id,
+        char *psk, size_t psk_len) = NULL;
+    ssize_t (*get_id_cb)(const void *creds, struct sol_network_link_addr *addr,
+        char *id, size_t id_len) = NULL;
```
cabelitos: I wish we used typedefs...

src/lib/comms/sol-lwm2m-security.c
```diff
@@ -0,0 +1,470 @@
...
+    case LWM2M_BOOTSTRAP_SERVER:
+        get_psk_cb = get_psk_from_server_or_bs_server;
+        get_id_cb = get_id_from_server_or_bs_server;
+        socket_dtls = sol_coap_server_get_socket(
+            ((struct sol_lwm2m_bootstrap_server *)entity)->coap);
+        if (((struct sol_lwm2m_bootstrap_server *)entity)->security) {
+            security = ((struct sol_lwm2m_bootstrap_server *)entity)->security;
+            has_security = true;
+        }
+        break;
```
cabelitos: Add a `default:` entry and return error if it happens..

src/lib/comms/sol-lwm2m-client.c
```diff
@@ -1567,12 +1594,28 @@ sol_lwm2m_client_new(const char *name, const char *path, const char *sms,
     client->coap_server = sol_coap_server_new(&servaddr, false);
     SOL_NULL_CHECK_GOTO(client->coap_server, err_coap);
 
+    client->dtls_server = sol_coap_server_new(&servaddr, true);
+    if (!client->dtls_server) {
+        if (errno == ENOSYS) {
```
cabelitos: This should be negative, no?

src/lib/comms/sol-lwm2m-client.c
```diff
@@ -2487,10 +2541,48 @@ sol_lwm2m_client_start(struct sol_lwm2m_client *client)
...
-            SECURITY_SERVER_IS_BOOTSTRAP);
+            SECURITY_SECURITY_MODE);
         SOL_INT_CHECK_GOTO(r, < 0, err_exit);
 
+        switch (res[0].data[0].content.integer) {
+        case SOL_LWM2M_SECURITY_MODE_PRE_SHARED_KEY:
+            client->security = sol_lwm2m_client_security_add(client,
+                SOL_LWM2M_SECURITY_MODE_PRE_SHARED_KEY);
+            if (!client->security) {
+                r = -errno;
```
cabelitos: Be carefull I guess you're already setting -errno inside lwm2m sec.

src/lib/comms/sol-lwm2m-client.c
```diff
@@ -2131,11 +2181,12 @@ server_connection_ctx_new(struct sol_lwm2m_client *client,
     SOL_INT_CHECK_GOTO(r, < 0, err_append);
     conn_ctx->client = client;
     conn_ctx->server_id = server_id;
+    conn_ctx->secure = secure;
     sol_vector_init(&conn_ctx->server_addr_list,
         sizeof(struct sol_network_link_addr));
 
     if (!uri.port)
-        conn_ctx->port = SOL_LWM2M_DEFAULT_SERVER_PORT;
```
cabelitos: I guess we should check `conn_ctx->secure` in here. If it `true` we should use the default DTLS port.

src/samples/coap/lwm2m-client.c
```diff
@@ -1058,20 +1217,55 @@ main(int argc, char *argv[])
...
+
+    while ((r = getopt(argc, argv, "bs:")) != -1) {
+        switch (r) {
+        case 'b':
+            data_ctx.is_bootstrap = true;
+            break;
+        case 's':
+            sec_mode = atoi(optarg);
+            if (sec_mode < 0 || sec_mode > 3) {
+                fprintf(stderr, "%s", usage);
```
cabelitos: Missing \n

src/samples/coap/lwm2m-client.c
```diff
@@ -1058,20 +1217,55 @@ main(int argc, char *argv[])
...
+                return -1;
+            }
+            break;
+        default:
+            fprintf(stderr, "%s", usage);
```
cabelitos: Ditto

src/samples/coap/lwm2m-client.c
```diff
@@ -1058,20 +1217,55 @@ main(int argc, char *argv[])
...
+            return -1;
+        }
+    }
+
+    cli_name = argv[optind];
+    if (!cli_name) {
+        fprintf(stderr, "%s", usage);
```
cabelitos: Ditto

src/lib/comms/sol-lwm2m-client.c
```diff
@@ -1362,10 +1369,17 @@ handle_resource(void *data, struct sol_coap_server *server,
...
 
-    if (client->supports_access_control) {
-        r = get_server_id_by_link_addr(&client->connections, cliaddr, &server_id);
-        header_code = SOL_COAP_RESPONSE_CODE_INTERNAL_ERROR;
-        SOL_INT_CHECK_GOTO(r, < 0, exit);
+    r = get_server_id_by_link_addr(&client->connections, cliaddr, &server_id);
+    header_code = SOL_COAP_RESPONSE_CODE_INTERNAL_ERROR;
+    SOL_INT_CHECK_GOTO(r, < 0, exit);
+
+    if (path_size >= 1 &&
```
cabelitos: why not: `if (!client->is_bootstrapping && obj_ctx->obj->id == SECURITY_OBJECT_ID) {

}

src/lib/comms/sol-lwm2m-client.c
```diff
@@ -1040,6 +1040,13 @@ handle_read(struct sol_lwm2m_client *client,
     uint16_t i;
     int r;
 
+    if ((obj_ctx->obj->id == SECURITY_OBJECT_ID) &&
```
cabelitos: Why do you need this here if you already check that on `handle_resource` ?

src/lib/comms/include/sol-lwm2m-bs-server.h
```diff
@@ -60,10 +60,13 @@ typedef struct sol_lwm2m_bootstrap_client_info sol_lwm2m_bootstrap_client_info;
...
  * @param port The UDP port to be used.
  * @param known_clients An array with the name of all clients this server has Bootstrap Information for.
+ * @param sec_mode The DTLS Security Mode to be used.
+ * @param known_psks The Clients' Pre-Shared Keys this Bootstrap Server has previous knowledge of.
  * @return The LWM2M bootstrap server or @c NULL on error.
  */
 struct sol_lwm2m_bootstrap_server *sol_lwm2m_bootstrap_server_new(uint16_t port,
-    const char **known_clients);
+    const char **known_clients, enum sol_lwm2m_security_mode sec_mode,
+    struct sol_vector *known_psks);
```
cabelitos: This signature is very strange now, we're receiving two types of vector... A plain C vector and a sol_vector :/

src/lib/comms/sol-lwm2m-server.c
```diff
@@ -802,33 +812,87 @@ observer_entry_del_monitor(struct observer_entry *entry,
...
     server = calloc(1, sizeof(struct sol_lwm2m_server));
     SOL_NULL_CHECK(server, NULL);
 
-    server->coap = sol_coap_server_new(&servaddr, false);
+    server->coap = sol_coap_server_new(&servaddr_coap, false);
     SOL_NULL_CHECK_GOTO(server->coap, err_coap);
 
+    if (sec_mode != SOL_LWM2M_SECURITY_MODE_NO_SEC) {
+        if (sec_mode == SOL_LWM2M_SECURITY_MODE_PRE_SHARED_KEY)
+            server->known_psks = known_psks;
```
cabelitos: Watch out if this kind of attribution. The user can call sol_vector_clear() right after the functions returns.

src/lib/comms/sol-lwm2m-security.c
bsmelo: Because it's an internal API, as stated here: https://github.com/solettaproject/soletta/blob/master/src/lib/comms/sol-coap.c#L138.

`sol-oic-security.c` does the same, so I just followed the pattern... if I remove it, I get compiler warnings.

src/lib/comms/sol-lwm2m-security.c
bsmelo: Yes. At the beginning of this function we check:
```C
    //If the caller expects a PSK_LEN less than 16 bytes
    if (psk_len < SOL_DTLS_PSK_KEY_LEN)
        return -ENOBUFS;
```

And then on line 67 we make sure `stored_psk->key->size` is equal to `SOL_DTLS_PSK_KEY_LEN`.

src/lib/comms/sol-lwm2m-security.c
bsmelo: `stored_psk->key->size` MUST be equal to `SOL_DTLS_PSK_KEY_LEN` (see answer above an line 67), since the PSK Key has a fixed size. So, basically, there's no difference... I just think it's a little better to read it this way.

src/lib/comms/sol-lwm2m-security.c
bsmelo: `-ENOENT` is never actually returned by `read_resources`, since it's a valid situation to have an optional non-supported resource in there (in which case it will just be filled with empty values by `read_resources` itself, see https://github.com/solettaproject/soletta/blob/master/src/lib/comms/sol-lwm2m-client.c#L1840).

So, the `SOL_INT_CHECK_GOTO` would not jump to `err_clear`, and that's why this check is needed.

src/lib/comms/sol-lwm2m-security.c
bsmelo: This is exactly what we were discussing the other day and I thought it was ruled "out of scope" since the LWM2M spec doesn't really have a say on how this is done.

Maybe we can discuss this idea of a callback offline on IRC... even if we do provide a callback, what should be its parameters? I mean, the server will retrieve a PSK ID based on what information, since at this point it doesn't even have the client name itself, since registration didn't occur yet. IP address, which is the thing we used at the client side, feels really weird to be used here, since Clients' IP addresses should be expected to change and so on.

src/lib/comms/sol-lwm2m-security.c
bsmelo: Ditto.

src/lib/comms/sol-lwm2m-security.c
bsmelo: Ditto.

src/lib/comms/sol-lwm2m-security.c
bsmelo: Nice one, done.

src/lib/comms/sol-lwm2m-security.c
```diff
@@ -0,0 +1,470 @@
...
+    case LWM2M_CLIENT:
+        ((struct sol_lwm2m_client *)(security->entity))->security = NULL;
+        break;
+    case LWM2M_SERVER:
+        ((struct sol_lwm2m_server *)(security->entity))->security = NULL;
+        break;
+    case LWM2M_BOOTSTRAP_SERVER:
+        ((struct sol_lwm2m_bootstrap_server *)(security->entity))->security = NULL;
+        break;
+    }
```
bsmelo: So I can rely on the following check, mostly needed after Client- or Server-initiated Bootstrap: https://github.com/bsmelo/soletta/blob/lwm2m-dtls_psk-v1/src/lib/comms/sol-lwm2m-client.c#L2534.

Of course I would need to set only the `case LWM2M_CLIENT` to `NULL`, but then I thought it was better to keep them consistent.

src/lib/comms/sol-lwm2m-security.c
bsmelo: Yup.

I kept the whole thing just as pointers, so basically the only place in memory to actually store the PSK (Both key and ID) is the application code itself (such as `lwm2m-sample-client`, for instance), but still, I'm using static blobs in the samples and there's no [secure] cleaning there; see https://github.com/bsmelo/soletta/blob/lwm2m-dtls_psk-v1/src/samples/coap/lwm2m-client.c#L148.

src/lib/comms/sol-lwm2m-security.c
bsmelo: [2], and not for the first time, but during the whole development of LWM2M code :(

src/lib/comms/sol-lwm2m-security.c
bsmelo: Added
```C
    default:
        r = -EINVAL;
        SOL_WRN("Invalid Entity Type");
        goto err_socket_dtls;
```

src/lib/comms/sol-lwm2m-client.c
bsmelo: No... If DTLS is not supported, that function will return with `errno` set to ENOSYS, in which case we just print a message using `SOL_INF`. Otherwise, DTLS is indeed supported and the error was due to something else; in that case, we use `SOL_WRN` and `goto err_dtls`.

This is similar to https://github.com/bsmelo/soletta/blob/lwm2m-dtls_psk-v1/src/lib/comms/sol-oic-server.c#L615.

src/lib/comms/sol-lwm2m-client.c
bsmelo: `errno` will be positive (as it should, from what I understand) upon return from `sol_lwm2m_client_security_add()` [1], so in here we just need a negative `r` obtained from errno.

[1] https://github.com/bsmelo/soletta/blob/lwm2m-dtls_psk-v1/src/lib/comms/sol-lwm2m-security.c#L410

src/lib/comms/sol-lwm2m-client.c
bsmelo: Nice catch, I overlooked this function.
Fixed:
```C
    if (!uri.port)
        if (secure)
            conn_ctx->port = SOL_LWM2M_DEFAULT_SERVER_PORT_DTLS;
        else
            conn_ctx->port = SOL_LWM2M_DEFAULT_SERVER_PORT_COAP;
    else
        conn_ctx->port = uri.port;
```

Also, I've added an URI Scheme enforcement, right before the `calloc` call:
```C
    SOL_NULL_CHECK(&uri.scheme, NULL);
    if (sol_str_slice_str_case_eq(uri.scheme, "coaps"))
        SOL_EXP_CHECK(!secure, NULL);
    else if (sol_str_slice_str_case_eq(uri.scheme, "coap"))
        SOL_EXP_CHECK(secure, NULL);
    else
        return NULL;
```

src/samples/coap/lwm2m-client.c
bsmelo: It's up there on the `usage` string itself: https://github.com/solettaproject/soletta/pull/2272/files#diff-31f229684ffba7b5bad2c6f184b0afd2R1231.

src/samples/coap/lwm2m-client.c
bsmelo: Ditto. :)

src/samples/coap/lwm2m-client.c
bsmelo: Ditto. :)

src/lib/comms/sol-lwm2m-client.c
bsmelo: I would still need the first check (`path_size >= 1`) to enforce `obj_ctx != NULL`, but I guess your version is shorter/more clear.
Changed to:
```C
    if (path_size >= 1 && !client->is_bootstrapping &&
        obj_ctx->obj->id == SECURITY_OBJECT_ID) {
        SOL_WRN("Only the Bootstrap Server is allowed to access the Security Object."
            " Server ID %" PRId64 " trying to access it", server_id);
        header_code = SOL_COAP_RESPONSE_CODE_UNAUTHORIZED;
        goto exit;
    }
```

src/lib/comms/sol-lwm2m-client.c
bsmelo: Because `handle_read()` can have other entry points, such as `notification_cb()`, which will not pass through `handle_resource()` in case it's called due to `sol_lwm2m_client_notify()`.

src/lib/comms/include/sol-lwm2m-bs-server.h
bsmelo: Hmm... Should I change the plain C vector to a `sol_vector`? :)

src/lib/comms/sol-lwm2m-server.c
bsmelo: Hmm, that makes sense.

`server->known_psks` would need to become `sol_vector` instead of `sol_vector *` then, and store elements of type `sol_lwm2m_security_psk` (so I can do
```C
uint16_t i;
struct sol_lwm2m_security_psk *psk, *given_psk;

sol_vector_init(&server->known_psks, sizeof(sol_lwm2m_security_psk));

SOL_VECTOR_FOREACH_IDX (known_psks, given_psk, i) {
    psk = sol_vector_append(server->known_psks);
    SOL_NULL_CHECK_GOTO(psk, err_psk); //TODO: add new label at the end
    psk->id = sol_blob_ref(given_psk->id);
    psk->key = sol_blob_ref(given_psk->key);
}
```
, right?

I'm not sure about adding references to the sensitive data (PSK ID and KEY) like this, though... Let's discuss this one offline as well.

src/lib/comms/sol-lwm2m-security.c
cabelitos: I don't like this. If you need this kind of thing. Make the client himself set it to NULL

src/lib/comms/sol-lwm2m-client.c
cabelitos: indeed

cabelitos: But the user is not supposed to observe a security object anyway...

src/lib/comms/include/sol-lwm2m-bs-server.h
cabelitos: dont think so

bsmelo: Change the `sol_vector` to a plain C vector? Keep it like this? :)

src/lib/comms/sol-lwm2m-client.c
bsmelo: That's true, but the check inside

src/lib/comms/sol-lwm2m-security.c
bsmelo: Ok, changed.

src/lib/comms/sol-lwm2m-client.c
cabelitos: Damn coap......We need to fix that

src/lib/comms/sol-lwm2m-server.c
cabelitos: Right

src/lib/comms/sol-lwm2m-security.c
bsmelo: As discussed offline, changed to use `machine_id` like OIC does.

PR #2274, opened 2016-08-04
libmicrohttp: bump version requirement.

glima:
This slipped in wrong for v1 and nobody noticed. With version before
that one make check-fbp fail big time.

### DIFF
--- a/data/jsons/dependencies.json
+++ b/data/jsons/dependencies.json
@@ -642,7 +642,7 @@
       "dependency": "libmicrohttpd",
       "type": "pkg-config",
       "pkgname": "libmicrohttpd",
-      "atleast-version": "0.9.43"
+      "atleast-version": "0.9.47"
     },
     {
       "dependency": "tinycbor_src",


glima:
Gotta check which libmicrohttpd is on semaphoreci's server. I wonder how the tests passed with a version < 47, since they don't for us here.

ceolin:
The version on semaphoreci is probably less than 47 and with this patch the server was not built and then it failed. You need to check wheter in semaphoreci we are doing some hack using LD_LIBRARY_PATH to enforce soletta using a different/custom version of libmicrohttpd

glima:
From a quick glance, it seems there's no hack for that. @bdilly, is there any?

bdilly:
Semaphoreci seems to be 43. No hacks, it's the version of the library on the ppa of soletta deps we're using:

https://launchpad.net/~solettaproject/+archive/ubuntu/deps

bdilly:
@anselmolsm is the only person with access to that ppa I suppose, but shouldn't be hard to update it to 47.

Anyway, since it was using version 43 and check-fbp was passing I'm wondering if your patch is correct. Why do we need version 47?

glima:
Because with 46 (current version) here it fails big time. Maybe with both 43 and 47 (the latter was tested) they work? LOL

bdilly:
hahaha
amazing =)

so I'm ok with the patch, but it would be great to poke @anselmolsm and see if he could help us updating this package version before merging it

glima:
@anselmolsm, would you please do the bump OR give me access to the ppa cited above? :)

anselmolsm:
@glima, ack. I'll take a look a this.

anselmolsm:
libmicrohttpd package updated to 0.47

anselmolsm:
Regarding access to PPA: I wrote info about it a "few" months ago: Oct 20  2015 README-ppa ;)
Updating the package is easy. The pain is to [re]create the env to do so.

"All checks have passed" \o/

glima:
Thanks a lot, @anselmolsm ! Now merging! :dancers:

PR #2275, opened 2016-08-04
lwm2m: Handle error condition correctly

ibriano:
The function returns an uint8_t, so returning negative values is wrong.
It's also leaking the handles in the case of an error, as it's not
following the clean-up path.


bdilly:
seems correct
:+1: 
thanks, @ibriano

ibriano:
Merged

Issue #2205, opened 2016-06-28
Proposal: Add support for iio mount matrix


laykuanloon:
#### Task Description
Please support iio mount matrix in soletta, I need this feature for the soletta sensor application(fbp or c api) can read out the data that following android orientation.

iio mount matrix is merge into mainline kernel recently (https://git.kernel.org/cgit/linux/kernel/git/torvalds/linux.git/commit/?id=dfc57732ad38f93ae6232a3b4e64fd077383a0f1).  

The mount matrix is store in iio mount matrix entry, it will give "0, 1, 0; 1, 0, 0; 0, 0, -1". Store this into a float array and use the following method to process.
https://github.com/01org/android-iio-sensors-hal/blob/master/transform.c#L200 is the example of handling mount matrix.

Thanks.
#### Dependencies
N/A

barbieri:
I'm thinking about this and would likely solve it outside of the IIO:
 - `sol-util.h` to provide a matrix transformation code, for use in C code;
 - introduce a node type to convert matrix, not specific about IIO, but any direction vector

Is this enough? For sure we could add the matrix transformation to the IIO node using that `sol-util.h` function to be introduced, I'm just wondering if that is going to help if we provide a more generic `direction-vector/transform` node that can be used for all inputs, like other kinds of sensors, data from the internet, etc.

laykuanloon:
But this is iio mount matrix, is in iio driver. I think is better to put under iio layer.

laykuanloon:
@barbieri
Can include this feature in V1?

@fulong82 @lblim

barbieri:
ok, then it makes sense to have both the option and the node type to transform. The option will write directly to the kernel options.

As for v1, I don't want to commit to that in order to avoid delaying the v1 even further. However, if for bureaucracy purposes the v1 is delayed and we have the code ready, then it could go in v1.

laykuanloon:
FYI. The iio mount matrix sysfs entry is READ only.

lblim:
@laykuanloon please close the ticket

Issue #2101, opened 2016-05-26
Proposal: Add support of IIO sampling frequency to channel level besides device level


fulong82:
#### Task Description

Hi @barbieri, @edersondisouza & @yongli3,

Base on the current design, IIO sampling frequency supported in soletta only cater to IIO sensor driver that its sampling frequency is shared by all (Refer to `src/lib/io/sol-iio.c` function `iio_set_sampling_frequency()` implemetation).

Whereas, there are a lot IIO sensor drivers exposing its sampling frequency that is shared by channel type. Let's take the sensor list [Link](https://github.com/ostroproject/ostro-os/blob/master/doc/hardware/sensors.rst) that is claimed in Ostro as examples to which are using channel type sampling frequency.

Accelerometer:
- BMC150 series -- http://lxr.free-electrons.com/source/drivers/iio/accel/bmc150-accel-core.c#L983
- MMA7455 series -- http://lxr.free-electrons.com/source/drivers/iio/accel/mma7455_core.c#L210

Gyro:
- BMG160 -- http://lxr.free-electrons.com/source/drivers/iio/gyro/bmg160_core.c#L729

Magneto:
- HMC5843 series -- http://lxr.free-electrons.com/source/drivers/iio/magnetometer/hmc5843_core.c#L468
- MAG3110 series -- http://lxr.free-electrons.com/source/drivers/iio/magnetometer/mag3110.c#L277

So a proposal here is to add support for the sampling frequency that are shared by channel type in IIO sensor.

Thanks.

#### Dependencies

N/A

laykuanloon:
Hi @barbieri, @edersondisouza & @yongli3,

I have 2 sensor need this enhancement also.

Following is the sensors:
Accelerometer - kxcjk1013 
/sys/bus/iio/devices/iio:device0/in_accel_sampling_frequency

Magnetometer - mmc35240
/sys/bus/iio/devices/iio:device5/in_magn_sampling_frequency

lblim:
@barbieri @bdilly Can this be pulled in for v1? The override sampling frequency option for IIO node will not work for certain sensors if the issue is not resolved.

barbieri:
not for v1, it's frozen. For v2 yes :-)

I'll talk to @bdilly so we can make v2 provide all the IIO enhancements you've asked, they seem to be quite interesting and useful.

lblim:
[...]

PR #2278, opened 2016-08-05
WIP iio: Add support for iio mount matrix

yongli3:
Fix issue #2205

### DIFF
--- a/src/lib/io/include/sol-iio.h
+++ b/src/lib/io/include/sol-iio.h
@@ -230,6 +230,20 @@ struct sol_str_slice
 sol_iio_read_channel_raw_buffer(struct sol_iio_channel *channel);
 
 /**
+ * @brief Perform the mount calibration.
+ *
+ * This function is meaningful only when mount_matrix is supported.
+ *
+ * @param device device to get mount matrix
+ * @param value data to be calibrated
+ *
+ * @return 0 if mount_matrix is exits and perform the calibration, -1 means does not support mount_matrix.
+ *
+ */
+int
+sol_iio_mount_calibration(struct sol_iio_device *device, sol_direction_vector *value);
+
+/**
  * @}
  */
 
--- a/src/lib/io/sol-iio.c
+++ b/src/lib/io/sol-iio.c
@@ -43,6 +43,7 @@
 #endif
 
 struct sol_iio_device {
+    double *mount_matrix;
     char *trigger_name;
     void (*reader_cb)(void *data, struct sol_iio_device *device);
     const void *reader_cb_data;
@@ -118,8 +119,13 @@ struct resolve_absolute_path_data {
 #define SAMPLING_FREQUENCY_BUFFER_PATH SYSFS_DEVICES_PATH "/iio:device%d/buffer/sampling_frequency"
 #define SAMPLING_FREQUENCY_TRIGGER_PATH SYSFS_DEVICES_PATH "/trigger%d/sampling_frequency"
 
+#define SYSFS_MOUNT_MATRIX SYSFS_DEVICES_PATH "/iio:device%d/mount_matrix"
+#define SYSFS_OUT_MOUNT_MATRIX SYSFS_DEVICES_PATH "/iio:device%d/out_mount_matrix"
+#define SYSFS_IN_MOUNT_MATRIX SYSFS_DEVICES_PATH "/iio:device%d/in_mount_matrix"
+
 #define I2C_DEVICES_PATH "/sys/bus/i2c/devices/%u-%04u/"
 
+#define MOUNT_MATRIX_LEN 9
 #define REL_PATH_IDX 2
 #define DEV_NUMBER_IDX 3
 #define DEV_NAME_IDX 4
@@ -618,6 +624,56 @@ iio_set_channel_offset(struct sol_iio_channel *channel, int offset)
     return result;
 }
 
+static bool
+get_mount_matrix(struct sol_iio_device *device, double *mount_matrix)
+{
+    char buf[100];
+    char *tmp1, *tmp2;
+    char path[PATH_MAX];
+    bool result = false;
+    int len, i;
+    double d;
+    const char *base = NULL;
+
+    /* Find the mount matrix sysfs node */
+    if (check_file_existence(SYSFS_MOUNT_MATRIX, device->device_id))
+        base = SYSFS_MOUNT_MATRIX;
+    else if (check_file_existence(SYSFS_OUT_MOUNT_MATRIX, device->device_id))
+        base = SYSFS_OUT_MOUNT_MATRIX;
+    else if (check_file_existence(SYSFS_IN_MOUNT_MATRIX, device->device_id))
+        base = SYSFS_IN_MOUNT_MATRIX;
+    else {
+        SOL_WRN("Could not find mount_matrix for device%d", device->device_id);
+        return false;
+    }
+
+    result = craft_filename_path(path, sizeof(path), base, device->device_id);
+    if (!result) {
+        SOL_WRN("Could not open device mount_matrix file");
+        return false;
+    }
+
+    len = sol_util_read_file(path, "%99[^\n]", buf);
+    if (len < 0) {
+        SOL_WRN("Coult not read mount matrix %s on sysfs.", path);
+        return false;
+    }
+
+    SOL_DBG("in_mount_matrix=%s", buf);
+
+    tmp1 = buf;
+    for (i = 0; i < MOUNT_MATRIX_LEN; i++) {
+        d = strtod(tmp1, &tmp2);
+        if (tmp1 == tmp2)
+            return false;
+        mount_matrix[i] = d;
+        tmp1 = tmp2 + 1;
+        SOL_DBG("matrix[%d]=%f", i, mount_matrix[i]);
+    }
+
+    return result;
+}
+
 SOL_API struct sol_iio_device *
 sol_iio_open(int device_id, const struct sol_iio_config *config)
 {
@@ -720,10 +776,19 @@ sol_iio_open(int device_id, const struct sol_iio_config *config)
             SOL_WRN("Could not set device%d sampling frequency", device->device_id);
     }
 
+    device->mount_matrix = calloc(MOUNT_MATRIX_LEN, sizeof(double));
+    SOL_NULL_CHECK_GOTO(device->mount_matrix, error);
+
+    r = get_mount_matrix(device, device->mount_matrix);
+    if (!r) {
+        free(device->mount_matrix);
+        device->mount_matrix = NULL;
+    }
+
     SOL_DBG("iio device created. device%d - buffer_enabled: %d - manual_trigger: %d"
-        " - trigger_name: %s - trigger_id: %d", device->device_id,
+        " - trigger_name: %s - trigger_id: %d - mount_matrix: %p", device->device_id,
         device->buffer_enabled, device->manual_triggering, device->trigger_name,
-        device->trigger_id);
+        device->trigger_id, device->mount_matrix);
 
     return device;
 
@@ -902,6 +967,7 @@ sol_iio_close(struct sol_iio_device *device)
     if (device->fd_handler) sol_fd_del(device->fd_handler);
     if (device->fd > -1) close(device->fd);
     if (device->name_fd > -1) close(device->name_fd);
+    free(device->mount_matrix);
 
     sol_buffer_fini(&device->buffer);
     free(device->trigger_name);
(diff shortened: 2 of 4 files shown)


yongli3:
@laykuanloon please help to review

laykuanloon:
@yongli3 

For the mount matrix sysfs entry. Below is entries that possible mount matrix sysfs entry for the sensor. It's related to #2101, For e.g: accelerometer may have in_accel_sampling_frequency and in_accel_mount_matrix. Any suggestion on how to handle this?


What:           /sys/bus/iio/devices/iio:deviceX/mount_matrix
What:           /sys/bus/iio/devices/iio:deviceX/in_mount_matrix
What:           /sys/bus/iio/devices/iio:deviceX/out_mount_matrix
What:           /sys/bus/iio/devices/iio:deviceX/in_anglvel_mount_matrix
What:           /sys/bus/iio/devices/iio:deviceX/in_accel_mount_matrix
KernelVersion:  4.6
Contact:        linux-iio@vger.kernel.org
Description:

yongli3:
@laykuanloon On my boards, I only get one read-only sysfs node: in_mount_matrix. What is the difference about the in_mount_matrix and out_mount_matrix? All of them should be read-only?

laykuanloon:
@yongli3 
I only see in_mount_matrix also. I suggest we can implement out_mount_matrix later since we not sure the usage. What I want to discuss is the in_accel_, and in_anglevel_.

laykuanloon:
@yongli3 

I tested the patch and it's working for fbp. I suggest:

1) Do sol_iio_get_mount_matrix() one time only in sol_iio_open() and store in a double array to save the processing time on keep on open the same file and read the same content

2) If want to support for both fbp and c. Can we do the mount correction in 
https://github.com/solettaproject/soletta/blob/master/src/lib/io/sol-iio.c#L1062 and
https://github.com/solettaproject/soletta/blob/master/src/lib/io/sol-iio.c#L1084

yongli3:
Thanks @laykuanloon, I just finished the test on my Pi3 boards. Let me update the PR to support both fbp and c, and other different mount matrix sysfs nodes

src/lib/io/include/sol-iio.h
```diff
@@ -230,6 +230,20 @@ struct sol_str_slice
 sol_iio_read_channel_raw_buffer(struct sol_iio_channel *channel);
 
 /**
+ * @brief Returns mount_atrix data array.
```
edersondisouza: s/atrix/matrix

src/lib/io/sol-iio.c
```diff
@@ -117,6 +117,8 @@ struct resolve_absolute_path_data {
 #define SAMPLING_FREQUENCY_BUFFER_PATH SYSFS_DEVICES_PATH "/iio:device%d/buffer/sampling_frequency"
 #define SAMPLING_FREQUENCY_TRIGGER_PATH SYSFS_DEVICES_PATH "/trigger%d/sampling_frequency"
 
+#define SYSFS_MOUNT_MATRIX SYSFS_DEVICES_PATH "/iio:device%d/in_mount_matrix"
```
edersondisouza: From kernel doc, it appears that are other names for this file. Are you going with only `in_mount_matrix` because of accessible devices (for test)?

src/lib/io/sol-iio.c
```diff
@@ -1427,3 +1429,44 @@ sol_iio_read_channel_raw_buffer(struct sol_iio_channel *channel)
...
+    if (len < 0) {
+        SOL_INF("Coult not read mount matrix %s on sysfs.", path);
+        return false;
+    }
+
+    SOL_DBG("in_mount_matrix=%s", buf);
+
+    tmp1 = buf;
+    for (i = 0; i < 9; i++) {
+        f = strtof(tmp1, &tmp2);
```
edersondisouza: Why `strtof` if `f` is `double`?

src/lib/io/sol-iio.c
```diff
@@ -1427,3 +1429,44 @@ sol_iio_read_channel_raw_buffer(struct sol_iio_channel *channel)
...
+
+SOL_API bool
+sol_iio_get_mount_matrix(struct sol_iio_device *device, double mount_matrix[9])
+{
+    char buf[100];
+    char *tmp1, *tmp2;
+    char path[PATH_MAX];
+    bool result = false;
+    int len, i;
+    double f;
```
edersondisouza: `double f`?

src/lib/io/sol-iio.c
yongli3: Thanks @edersondisouza. FYI, the mount matrix data processing logic is github.com:01org/android-iio-sensors-hal.git.  I will add the "out_mount_matrix" checking later

yongli3:
@laykuanloon @edersondisouza I updated the PR, to encapsulate the mount calibration logic into sol-iio.c, and add different sysfs node names support, please help to review

src/modules/flow/iio/nodes.c
```diff
@@ -143,6 +143,11 @@ iio_direction_vector_reader_cb(void *data, struct sol_iio_device *device)
     r = sol_iio_read_channel_value(mdata->channel_z, &out.z);
     SOL_INT_CHECK_GOTO(r, < 0, error);
 
+    SOL_DBG("Before mount_caliration: %f-%f-%f", out.x, out.y, out.z);
```
edersondisouza: s/caliration/calibration

src/lib/io/include/sol-iio.h
```diff
@@ -230,6 +230,22 @@ struct sol_str_slice
...
+ *
+ * @param device device to get mount matrix
+ * @param x X coordinate data to be calibrated
+ * @param y Y coordinate data to be calibrated
+ * @param z Z coordinate data to be calibrated
+ *
+ * @return true if mount_matrix is exits, false means does not support mount_matrix.
+ *
+ */
+bool
```
edersondisouza: Soletta APIs usually return an `int` value, where 0 is success and a negative number, failure.

src/modules/flow/iio/nodes.c
```diff
@@ -143,6 +143,11 @@ iio_direction_vector_reader_cb(void *data, struct sol_iio_device *device)
...
+
+    // mount correction
+    sol_iio_mount_calibration(device, &out.x, &out.y, &out.z);
```
edersondisouza: Curiosity: why is applying mount matrix optional? Wouldn't be better to always apply it if available? Or are there use cases to not apply it?
If we could always apply it, we could do it internally, and not create a new API.

src/lib/io/include/sol-iio.h
```diff
@@ -230,6 +230,22 @@ struct sol_str_slice
...
+sol_iio_mount_calibration(struct sol_iio_device *device, double *x, double *y, double *z);
```
cabelitos: Why not a sol_direction_vector instead of double x, double y, double z...

src/lib/io/sol-iio.c
```diff
@@ -618,6 +624,59 @@ iio_set_channel_offset(struct sol_iio_channel *channel, int offset)
...
+    else if (check_file_existence(SYSFS_IN_MOUNT_MATRIX, device->device_id))
+        base = SYSFS_IN_MOUNT_MATRIX;
+    else {
+        SOL_WRN("Could not find mount_matrix for device%d", device->device_id);
+        return false;
+    }
+
+    result = craft_filename_path(path, sizeof(path), base, device->device_id);
+    if (!result) {
+        SOL_DBG("Could not open device mount_matrix file");
```
cabelitos: This is an error. Change to SOL_WRN

src/lib/io/sol-iio.c
```diff
@@ -618,6 +624,59 @@ iio_set_channel_offset(struct sol_iio_channel *channel, int offset)
...
+        return false;
+    }
+
+    len = sol_util_read_file(path, "%99[^\n]", buf);
+    if (len < 0) {
+        SOL_INF("Coult not read mount matrix %s on sysfs.", path);
```
cabelitos: Same thing change to SOL_WRN

src/lib/io/sol-iio.c
```diff
@@ -902,6 +970,7 @@ sol_iio_close(struct sol_iio_device *device)
     if (device->fd_handler) sol_fd_del(device->fd_handler);
     if (device->fd > -1) close(device->fd);
     if (device->name_fd > -1) close(device->name_fd);
+    if (NULL != device->mount_matrix) free(device->mount_matrix);
```
cabelitos: Two things here. First we're not using yoda style in Soletta and you don't need to check if mount_matrix is not null to call free. Remove the check, calling free(NULL) is safe.

src/lib/io/sol-iio.c
```diff
@@ -720,10 +779,19 @@ sol_iio_open(int device_id, const struct sol_iio_config *config)
             SOL_WRN("Could not set device%d sampling frequency", device->device_id);
     }
 
+    device->mount_matrix = calloc(1, sizeof(double) * MOUNT_MATRIX_LEN);
```
cabelitos: change to
`device->mount_matrix = calloc(MOUNT_MATRIX_LEN, sizeof(double));`

src/lib/io/sol-iio.c
```diff
@@ -1454,3 +1523,27 @@ sol_iio_read_channel_raw_buffer(struct sol_iio_channel *channel)
...
     return slice;
 }
+
+SOL_API bool
+sol_iio_mount_calibration(struct sol_iio_device *device, double *x, double *y, double *z)
+{
+    int i;
+    double tmp[3];
+
+    if (NULL == device->mount_matrix)
```
cabelitos: We do not use yoda style. Change it to if (!device->mount_matrix)

src/lib/io/sol-iio.c
```diff
@@ -618,6 +624,59 @@ iio_set_channel_offset(struct sol_iio_channel *channel, int offset)
...
+    for (i = 0; i < MOUNT_MATRIX_LEN; i++) {
+        d = strtod(tmp1, &tmp2);
+        if (tmp1 == tmp2)
+            return false;
+        mount_matrix[i] = d;
+        tmp1 = tmp2 + 1;
+    }
+
+    for (i = 0; i < MOUNT_MATRIX_LEN; i++) {
+        SOL_DBG("matrix[%d]=%f", i, mount_matrix[i]);
```
cabelitos: This for is useless, why you do not print it this inside the for above?

src/lib/io/sol-iio.c
```diff
@@ -43,6 +43,7 @@
 #endif
 
 struct sol_iio_device {
+    double *mount_matrix;
```
cabelitos: I see that the size is constant. Why not change it to: double mount_matrix[MOUNT_MATRIX_LEN]; ?

yongli3: Thanks for your comments. Not all sensors support such mount_matrix sysfs node, By using point instead of array, we can malloc memory for these supported sensors only, to save some memory

src/modules/flow/iio/nodes.c
yongli3: @edersondisouza Based on my understanding, the mount_matrix calibraton needs all these X/Y/Z raw data, only nodes.c can capture all these X/Y/Z data together.Is there any a proper location internally for the X/Y/Z data calibration?

src/lib/io/sol-iio.c
cabelitos: I don't get it. Even as a pointer you will always alloc an array of size 9.

Quoting your code:
```C
device->mount_matrix = calloc(1, sizeof(double) * MOUNT_MATRIX_LEN);
```

yongli3: Please check on the line 787, if there is no such mount_matrix(get_mount_marix return false), the pointer will be free

src/modules/flow/iio/nodes.c
edersondisouza: Yeah, you're right, calculation needs all three elements so it can't be done on a channel alone.

src/lib/io/sol-iio.c
cabelitos: Ok then...

yongli3: Thanks @cabelitos. Let me update the PR to follow other comments

cabelitos: Thank you.

yongli3:
@laykuanloon  @edersondisouza @cabelitos I updated the PR, please help to review

laykuanloon:
@yongli3, i tested is working fine.

cabelitos:
lgtm. However I did not test it. I think we're ok to merge it. @edersondisouza is it ok for you?

edersondisouza:
Merged, tks!

PR #2281, opened 2016-08-09
LWM2M: Fix check_authorization() for Access Control Objects

bsmelo:
I'm improving the `src/test/test-lwm2m.c` program to cover all new functionalities
added during GSoC, and found this issue.

This patch fixes a bug when a given LWM2M Server who owns a given
Access Control Object Instance tries to modify it (for instance,
to add or remove an ACL for a given third-party LWM2M Server).

There was no way to modify an Access Control Object Instance, since
the check_authorization() would fail, not finding an Access Control
Object Instance related to the target instance - which should not
exist anyway. To fix it, a new condition was added in case the target
`obj_id` == `ACCESS_CONTROL_OBJECT_ID`, in which we should only look
for the target `instance_id` itself and then inside the OWNER Resource,
not looking for ACLs or anything else.

### DIFF
--- a/src/lib/comms/sol-lwm2m-client.c
+++ b/src/lib/comms/sol-lwm2m-client.c
@@ -452,6 +452,28 @@ check_authorization(struct sol_lwm2m_client *client,
     }
 
     obj_ctx = find_object_ctx_by_id(client, ACCESS_CONTROL_OBJECT_ID);
+    //If the target Object is an Access Control Object itself,
+    // the server is authorized iff it is the owner of the object instance.
+    if (obj_id == ACCESS_CONTROL_OBJECT_ID) {
+        SOL_VECTOR_FOREACH_IDX (&obj_ctx->instances, obj_instance, i) {
+            if (obj_instance->id == instance_id) {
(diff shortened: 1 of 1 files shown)


cabelitos:
lgtm

src/lib/comms/sol-lwm2m-client.c
```diff
@@ -452,14 +452,36 @@ check_authorization(struct sol_lwm2m_client *client,
...
+                }
+
+                if (res[0].data[0].content.integer == server_id)
+                    r = 1;
+                else
+                    r = 0;
+                goto exit_clear_1;
+            }
+        }
+    }
```
cabelitos: I guess if the instance was not found we can return an error as well.

bsmelo: Fixed this (and added `return -EINVAL` at the end of the "normal" case as well).

Should I make a new PR or update this one?

cabelitos: Send a new one

bsmelo:
An updated version is at #2282. Thanks!

PR #2282, opened 2016-08-11
LWM2M: Fix check_authorization() for Access Control Objects [v2]

bsmelo:
Changes from v1 (#2281):
- Added error returning.

This patch fixes a bug when a given LWM2M Server who owns a given
Access Control Object Instance tries to modify it (for instance,
to add or remove an ACL for a given third-party LWM2M Server).

There was no way to modify an Access Control Object Instance, since
the check_authorization() would fail, not finding an Access Control
Object Instance related to the target instance - which should not
exist anyway. To fix it, a new condition was added in case the target
'obj_id' == 'ACCESS_CONTROL_OBJECT_ID', in which we should only look
for the target 'instance_id' itself and then inside the OWNER Resource,
not looking for ACLs or anything else.


cabelitos:
+1

PR #2284, opened 2016-08-11
Fix for sol_platform_get_machine_id()

bsmelo:
In case SOL_PLATFORM_LINUX is defined, this function expects an UUID
from the environment variable SOL_MACHINE_ID. It was only considering
UUIDs formed by 32 chars, and not 36.

This patch fixes that, leaving the length to be check by
sol_util_uuid_str_is_valid(), which already does that, and if the length
is 36, we remove the hyphens, since the returned value is expected to
contain only hex digits. Valid formats (32 chars and 36 chars + hyphens)
for UUIDs were obtained from [1].

[1] https://en.wikipedia.org/wiki/Universally_unique_identifier

### DIFF
--- a/src/lib/common/sol-platform.c
+++ b/src/lib/common/sol-platform.c
@@ -423,13 +423,28 @@ sol_platform_get_machine_id(void)
     char *env_id = getenv("SOL_MACHINE_ID");
 
     if (env_id) {
-        if (strlen(env_id) > 32 ||
-            !sol_util_uuid_str_is_valid(sol_str_slice_from_str(env_id))) {
+        if (!sol_util_uuid_str_is_valid(sol_str_slice_from_str(env_id))) {
             SOL_WRN("Malformed UUID passed on environment variable "
                 "SOL_MACHINE_ID: %s", env_id);
             return NULL;
         }
-        return env_id;
+
+        if (strlen(env_id) == 36) {
+            int i, j;
+
+            /* remove hyphens on positions 8, 13, 18, 23 (from 0) */
+            for (i = 0, j = 0; i < 36; i++) {
+                if (i != 8 && i != 13 && i != 18 && i !=  23) {
+                    id[j] = env_id[i];
+                    j++;
+                }
+            }
+            id[32] = '\0';
+
+            return id;
+        } else {
+            return env_id;
+        }
     }
 #endif
     if (!id[0]) {


cabelitos:
The code looks good, however it's failing a fbp test.

src/lib/common/sol-platform.c
```diff
@@ -423,13 +423,26 @@ sol_platform_get_machine_id(void)
...
+            /* remove hyphens on positions 8, 13, 18, 23 (from 0) */
+            for (i = 0, j = 0; i < 36; i++) {
+                if (i != 8 && i != 13 && i != 18 && i !=  23) {
+                    id[j] = env_id[i];
+                    j++;
+                }
+            }
+            id[32] = '\0';
+        }
+
```
cabelitos: Hey, if strlen(env_id) is != 36 you still need to copy it to the id variable otherwise you will return a wrong value!

bsmelo: Wow, how did I miss that? Can I change it and update this same PR?
I have no way of retesting the failing test locally, because I couldn't get the installation of pcre correctly, so I was expecting the CI could do that for me.

cabelitos: ok, send it again

cabelitos:
+1

glima:
Fair for me as well

src/lib/common/sol-platform.c
```diff
@@ -423,13 +423,28 @@ sol_platform_get_machine_id(void)
...
+            }
+            id[32] = '\0';
+
+            return id;
+        } else {
+            return env_id;
+        }
     }
 #endif
     if (!id[0]) {
```
bdilly: I know it's not touched by this patch, but I don't get this check ^
you only touch id if you're going to return it.

bdilly:
I've granted you write access to the repository since you have lots of contributions already.
Just make sure to read  https://github.com/solettaproject/soletta/wiki/Contributing#contributions-flow please.
Thanks

src/lib/common/sol-platform.c
cabelitos: `id` is a static variable, it is basically checking if it was initialized, if not it will call `spl_platform_impl_get_machine_id(id);`

cabelitos:
Merged

PR #2287, opened 2016-08-12
Network: Add parameter to sol_network_link_addr_eq()

bsmelo:
This patch adds a boolean parameter to the function
`sol_network_link_addr_eq()`, to indicate whether or not the `port`
field inside the `struct sol_network_link_addr` should be considered
for address comparison as well.

### DIFF
--- a/data/scripts/sol-oic-gen.py
+++ b/data/scripts/sol-oic-gen.py
@@ -1167,7 +1167,7 @@ state_changed(void *data, enum sol_coap_response_code response_code, struct sol_
     if (!cliaddr || !repr_vec)
         return;
 
-    if (!sol_network_link_addr_eq(cliaddr, &resource->resource->addr)) {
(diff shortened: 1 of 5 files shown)


cabelitos:
Hey bruno, the patch looks good. However I think we should create a new version of sol_network_link_addr_eq(). And leave sol_network_link_addr_eq() as is. By doing so we will not break the API. So basically what you need to do is:

* Create a new function called sol_network_link_addr_full_eq(addr1, addr2, bool COMPARE_PORTS) (Or a better name) - (Note the bool argument)
* Move the current implementation of the sol_network_link_addr_eq() to sol_network_link_addr_full_eq()
* sol_network_link_addr_eq() should call sol_network_link_addr_full_eq() - passing false to COMPARE_PORTS

bsmelo:
New version at #2288.

PR #2288, opened 2016-08-15
Network: New API sol_network_link_addr_eq_full()

bsmelo:
This patch adds a boolean parameter to the function sol_network_link_addr_eq(),
renaming it to sol_network_link_addr_eq_full(). The additional
parameter indicates whether or not the port field inside the
struct sol_network_link_addr should be considered for the comparison.

The old function sol_network_link_addr_eq() is still present, and now
only calls sol_network_link_addr_eq_full() passing false to the new parameter.


cabelitos:
+1

cabelitos:
merged

PR #2285, opened 2016-08-11
LWM2M: Support for DTLS with Pre-Shared Key security mode [v2]

bsmelo:
Main changes since v1 (#2272):
https://github.com/solettaproject/soletta/pull/2272#discussion-diff-74065997
https://github.com/solettaproject/soletta/pull/2272#discussion-diff-73922416

This patch depends on #2283.
_**Edit**: Depends on #2288 as well._

Signed-off-by: Bruno Melo bsilva.melo@gmail.com

PR #2289, opened 2016-08-15
LWM2M: Support for DTLS with Pre-Shared Key security mode [v3]

bsmelo:
Main changes since v2 (#2285):
https://github.com/solettaproject/soletta/pull/2285#discussion-diff-74800812


cabelitos:
Merged

PR #2290, opened 2016-08-16
Fixes lwm2m

cabelitos:


cabelitos:
@otaviobp Could you please take a look at  fe62a4e

src/lib/comms/sol-lwm2m-server.c
```diff
@@ -1295,26 +1295,41 @@ send_management_packet(struct sol_lwm2m_server *server,
...
- * This checks if the path has the following form: /2/0/1
- */
-static bool
-is_resource_set(const char *path)
+enum lwm2m_path_props {
+    PATH_IS_INVALID = (1 << 0),
+    PATH_HAS_OBJECT =  (1 << 1),
+    PATH_HAS_INSTANCE = (1 << 2),
+    PATH_HAS_RESOURCE = (1 << 3)
+};
```
bsmelo: Just out of curiosity: is there any advantage in using left-shifts instead of `(1)`, `(2)`, `(4)` and `(8)`?

src/lib/comms/sol-lwm2m-server.c
```diff
@@ -1295,26 +1295,41 @@ send_management_packet(struct sol_lwm2m_server *server,
...
+            SOL_WRN("The path '%s' contains a nondigit character: '%c'",
+                path, path[i]);
+            return PATH_IS_INVALID;
         }
     }
 
-    if (slashes < 3 || *(last_slash + 1) == '\0')
-        return false;
-    return true;
+    if (i - slashes == 0) {
```
bsmelo: Since this means "The `path` is only formed by `slashes` and has no digits", wouldn't it be easier to read an `if (i == slashes)`?

src/lib/comms/sol-lwm2m-server.c
cabelitos: 0 advantages.

src/lib/comms/sol-lwm2m-server.c
cabelitos: Indeed. I will change that.

src/lib/comms/sol-lwm2m-server.c
```diff
@@ -1295,26 +1295,41 @@ send_management_packet(struct sol_lwm2m_server *server,
...
+enum lwm2m_path_props {
+    PATH_IS_INVALID = (1 << 0),
+    PATH_HAS_OBJECT =  (1 << 1),
+    PATH_HAS_INSTANCE = (1 << 2),
+    PATH_HAS_RESOURCE = (1 << 3)
+};
+
+static enum lwm2m_path_props
+get_path_props(const char *path)
 {
```
bsmelo: I would suggest moving this to sol-lwm2m-common, and adding these checks to operations in other interfaces as well (Bootstrap Write, Bootstrap Delete, Observe and Notify).

bsmelo:
+1 in general - I really like the one for `DTLS_EXTRAS`.  It would be nice to add the checks in the other suggested operations/interfaces though.

cabelitos:
#2292

PR #2291, opened 2016-08-17
LWM2M: Tests covering new functionality from GSoC

bsmelo:
This patch adds lwm2m tests covering all new functionality
added during GSoC, including: Bootstrap Interface, Access
Control Objects and CoAP Encryption (DTLS Security Modes).

Depends on #2290.

### DIFF
--- a/src/test/test-lwm2m.c
+++ b/src/test/test-lwm2m.c
@@ -16,6 +16,31 @@
  * limitations under the License.
  */
 
+/*
+ * Testing Scenario:
+ *
+ * +----------------+                +---------------+
+ * |                |                |               |
+ * |            [5693\CoAP]  (4)     |  sec_client   |
+ * | sec_server [5684\DTLS]-----[*\DTLS] w/ Access   |
+ * |                |        +--[*\DTLS]  Control [*\CoAP]-+
+ * +----------------+        |       |               |  (3)|   +--------------+
+ *                           |(2)    +---------------+     |   |              |
+ *                           |                             |   | nosec_server |
+ * +----------------+        |       +----------------+  [5683\CoAP]          |
+ * |                |        |       |                |    |   |              |
+ * |            [5784\DTLS]--+       |  nosec_client  |    |   +--------------+
+ * |  bs_server     |                |   w/o Access   | (1)|
+ * |                |                |    Control [*\CoAP]-+
+ * +----------------+                |                |
+ *                                   +----------------+
+ *
+ * (1)        1:1 Client-Server communication through CoAP
+ * (2)        1:1 Client-BootstrapServer communication through DTLS (w/ PSK)
+ * (3)[+(1)]  2:1 Client-Server communication through CoAP
+ * (4)[+(3)]  1:2 Client-Server communication through CoAP and DTLS (w/ PSK)
+ */
+
 #include <time.h>
 #include <stdint.h>
 #include <string.h>
@@ -26,6 +51,7 @@
 #include "sol-lwm2m.h"
 #include "sol-lwm2m-client.h"
 #include "sol-lwm2m-server.h"
+#include "sol-lwm2m-bs-server.h"
 #include "sol-mainloop.h"
 #include "sol-coap.h"
 #include "sol-util.h"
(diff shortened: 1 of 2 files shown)


src/test/test-lwm2m.c
```diff
@@ -460,7 +1169,7 @@ observe_res_cb(void *data,
     enum sol_lwm2m_content_type content_type,
     struct sol_str_slice content)
 {
-    static int i = 0;
+    static int nosec_i = 0, sec_i = 0;
```
cabelitos: I would create an ENUM to describe which state we're right know.

src/test/test-lwm2m.c
```diff
@@ -562,94 +1322,522 @@ delete_cb(void *data,
...
+#ifdef DTLS
+    struct sol_lwm2m_bootstrap_server *bs_server;
+    struct sol_lwm2m_server *sec_server;
+    struct sol_lwm2m_client *sec_client;
+    static const struct sol_lwm2m_object *sec_objects[] =
+    { &security_object, &server_object,
+      &access_control_object, &dummy_object, NULL };
+    struct security_obj_instance_ctx *sec_security_data;
+    struct sol_lwm2m_security_psk *bs_psk, *sec_psk;
+    struct sol_vector bs_known_keys = { }, sec_known_keys = { };
```
cabelitos: you can use
`struct sol_vector bs_known_keys = SOL_VECTOR_INIT(struct sol_lwm2m_security_psk);`

src/test/test-lwm2m.c
```diff
@@ -562,94 +1322,522 @@ delete_cb(void *data,
...
+    struct sol_lwm2m_bootstrap_server *server,
+    struct sol_lwm2m_bootstrap_client_info *bs_cinfo)
+{
+    int r;
+
+    ASSERT(!strcmp("cli1", sol_lwm2m_bootstrap_client_info_get_name(bs_cinfo)));
+
+    printf("DBG: [Bootstrap Delete /]---> '%s'\n",
+        sol_lwm2m_bootstrap_client_info_get_name(bs_cinfo));
+    r = sol_lwm2m_bootstrap_server_delete_object_instance(server, bs_cinfo, "/",
```
cabelitos: I guess we should add a test that a server will send a delete on '/' while NOT in bootstrap.

cabelitos: To make sure that we implemented correctly

src/test/test-lwm2m.c
```diff
@@ -562,94 +1322,522 @@ delete_cb(void *data,
...
+static void
+read_unauthorized_cb(void *data,
+    struct sol_lwm2m_server *server,
+    struct sol_lwm2m_client_info *client,
+    const char *path,
+    enum sol_coap_response_code response_code,
+    enum sol_lwm2m_content_type content_type,
+    struct sol_str_slice content)
+{
+    ASSERT(response_code == SOL_COAP_RESPONSE_CODE_UNAUTHORIZED);
```
cabelitos: I think we should chain the events and not leave them hanging around like this.

src/test/test-lwm2m.c
```diff
@@ -562,94 +1322,522 @@ delete_cb(void *data,
...
+            SOL_LWM2M_RESOURCE_DATA_TYPE_BOOL, false);
+        ASSERT(r == 0);
+        SOL_LWM2M_RESOURCE_SINGLE_INIT(r, &res[1], DUMMY_OBJECT_FLOAT_ID,
+            SOL_LWM2M_RESOURCE_DATA_TYPE_FLOAT, FLOAT_VALUE);
+        ASSERT(r == 0);
+
+        printf("DBG: '%s' ---[Write /999]---> '%s'\n",
+            server_type, sol_lwm2m_client_info_get_name(cinfo));
+        r = sol_lwm2m_server_write(server, cinfo, "/999",
+            res, sol_util_array_size(res), NULL, data);
```
cabelitos: Why the callback here is NULL ?

src/test/test-lwm2m.c
cabelitos: Nevermind. You're already doing this.

cabelitos:
It would be nice to add comments about what kind of objects each client has and what the servers are trying to do. Could you add this documentation? You should also add to this docs operations that the servers will try to do, but will fail since they are not allowed.

src/test/test-lwm2m.c
bsmelo: Which states would those be? I must confess I didn't "deep dive" in this part - the sequence of tests performed originally, which apparently includes a step of modifying one integer resource and so on. I just changed the `static i` to `static "i(s)"` to support two different clients independently.

src/test/test-lwm2m.c
bsmelo: Why? I mean, I'm only doing this "hang around" thing for the negative- and corner-cases. Since they should fail anyway there's no dependency between the operations (like a Create needing to be executed before a Read can happen and so on).

src/test/test-lwm2m.c
bsmelo: Because it must fail without a CoAP Request being even sent to the Client. Since the callback would never be called anyway (there's no Request -> there's no Response), we save code.

cabelitos: God. You're right. It's funny, because I fixed this problem. lol

src/test/test-lwm2m.c
cabelitos: States:
INT_VALUE_NOT_SET and INT_VALUE_REPLACE

cabelitos: I know, I just want to keep it more readable.

bsmelo:
An updated version is at #2294.

PR #2292, opened 2016-08-17
Fixes lwm2m v2

cabelitos:
Changes since V1:
*  5c6edd8 now enforces path restrictions on the observation/bootstrap interface. (I'm not touching the notification interface, since it already does the necessary checks)

### DIFF
--- a/src/lib/comms/sol-lwm2m-bs-server.c
+++ b/src/lib/comms/sol-lwm2m-bs-server.c
@@ -446,9 +446,14 @@ sol_lwm2m_bootstrap_server_write_object(struct sol_lwm2m_bootstrap_server *serve
     enum sol_coap_response_code response_code),
     const void *data)
 {
+    enum sol_lwm2m_path_props props;
+
     SOL_NULL_CHECK(server, -EINVAL);
     SOL_NULL_CHECK(client, -EINVAL);
     SOL_NULL_CHECK(path, -EINVAL);
+
+    props = sol_lwm2m_common_get_path_props(path);
+    SOL_EXP_CHECK(props != PATH_HAS_OBJECT, -EINVAL);
 
     return send_bootstrap_packet(server, client, path,
         BOOTSTRAP_WRITE, cb, data, SOL_COAP_METHOD_PUT, NULL, instances,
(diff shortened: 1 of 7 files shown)


src/lib/comms/sol-lwm2m-bs-server.c
```diff
@@ -465,10 +470,15 @@ sol_lwm2m_bootstrap_server_write(struct sol_lwm2m_bootstrap_server *server,
...
     const void *data)
 {
+    enum sol_lwm2m_path_props props;
+
     SOL_NULL_CHECK(server, -EINVAL);
     SOL_NULL_CHECK(client, -EINVAL);
     SOL_NULL_CHECK(path, -EINVAL);
 
+    props = sol_lwm2m_common_get_path_props(path);
+    SOL_EXP_CHECK(props == PATH_IS_INVALID, -EINVAL);
```
bsmelo: This check should be `path < PATH_HAS_INSTANCE`.
The separate API `sol_lwm2m_bootstrap_server_write_object()` must be used if `path == PATH_HAS_OBJECT`.

src/lib/comms/sol-lwm2m-bs-server.c
```diff
@@ -482,10 +492,15 @@ sol_lwm2m_bootstrap_server_delete_object_instance(struct sol_lwm2m_bootstrap_ser
...
     const void *data)
 {
+    enum sol_lwm2m_path_props props;
+
     SOL_NULL_CHECK(server, -EINVAL);
     SOL_NULL_CHECK(client, -EINVAL);
     SOL_NULL_CHECK(path, -EINVAL);
 
+    props = sol_lwm2m_common_get_path_props(path);
+    SOL_EXP_CHECK(props != PATH_HAS_INSTANCE, -EINVAL);
```
bsmelo: An empty path `/` or a full object `/999` is accepted for Bootstrap Delete. So, this should be `SOL_EXP_CHECK(props == PATH_HAS_RESOURCE, -EINVAL);`.

For readability only - so it's not weird that we're accepting a property called `...INVALID` -, maybe we could have an additional property `PATH_IS_EMPTY` between `PATH_IS_INVALID` and `PATH_HAS_OBJECT`?

cabelitos: Indeed I forget that during bootstrap I can send a delete '/'. I will rename the enum value.

cabelitos:
@2293

PR #2296, opened 2016-08-21
LWM2M: Support for DTLS with Raw Public Key security mode

bsmelo:
This patch adds support for CoAP Data Encryption using DTLS with Raw Public Key mode (using `TLS_ECDHE_ECDSA_WITH_AES_128_CCM_8` Cipher Suite).

It includes modifications on the LWM2M Samples and LWM2M Tests to demonstrate the support as well.

### DIFF
--- a/src/lib/comms/include/sol-coap.h
+++ b/src/lib/comms/include/sol-coap.h
@@ -18,6 +18,7 @@
 
 #pragma once
 
+#include <sol-socket.h>
 #include <sol-network.h>
 #include <sol-str-slice.h>
 
@@ -502,6 +503,25 @@ bool sol_coap_server_is_secure(const struct sol_coap_server *server);
  *
  */
 struct sol_coap_server *sol_coap_server_new(const struct sol_network_link_addr *addr, bool secure);
+
+/**
+ * @brief Creates a new DTLS-Secured CoAP server instance.
+ *
+ * Creates a new, DTLS-Secured CoAP server instance listening on address
+ * @a addr. If the server cannot be created, NULL will be returned and
+ * @c errno will be set to indicate the reason.
+ *
+ * @param addr The address where the server will listen on.
+ * @param cipher_suites Indicates the desired cipher suites to use.
+ * @param cipher_suites_len Indicates the length of the @c cipher_suites array.
+ *
+ * @return A new server instance, or NULL in case of failure.
+ *
+ */
+struct sol_coap_server *sol_coap_server_new_by_cipher_suites(
+    const struct sol_network_link_addr *addr,
+    enum sol_socket_dtls_cipher *cipher_suites,
+    uint16_t cipher_suites_len);
 
 /**
  * @brief Take a reference of the given server.
--- a/src/lib/comms/include/sol-lwm2m-bs-server.h
+++ b/src/lib/comms/include/sol-lwm2m-bs-server.h
@@ -60,13 +60,20 @@ typedef struct sol_lwm2m_bootstrap_client_info sol_lwm2m_bootstrap_client_info;
  *
  * @param port The UDP port to be used.
  * @param known_clients An array with the name of all clients this server has Bootstrap Information for.
- * @param sec_mode The DTLS Security Mode to be used.
- * @param known_psks The Clients' Pre-Shared Keys this Bootstrap Server has previous knowledge of.
+ * @param num_sec_modes The number of DTLS Security Modes this Bootstrap Server will support.
+ * @param ... At least one @c sol_lwm2m_security_mode followed by its relevant parameters, as per the table below:
+ *
+ * Security Mode | Follow-up arguments | Description
+ * ------------- | ------------------- | ------------------
+ * SOL_LWM2M_SECURITY_MODE_PRE_SHARED_KEY | struct sol_lwm2m_security_psk **known_psks | @c known_psks The Clients' Pre-Shared Keys this Bootstrap Server has previous knowledge of.
+ * SOL_LWM2M_SECURITY_MODE_RAW_PUBLIC_KEY | struct sol_lwm2m_security_rpk *rpk, struct sol_blob **known_pub_keys | @c rpk This Bootstrap Server's Key Pair - @c known_pub_keys The Clients' Public Keys this Bootstrap Server has previous knowledge of.
+ *
+ * @note: SOL_LWM2M_SECURITY_MODE_CERTIFICATE is not supported yet.
+ *
  * @return The LWM2M bootstrap server or @c NULL on error.
  */
 struct sol_lwm2m_bootstrap_server *sol_lwm2m_bootstrap_server_new(uint16_t port,
-    const char **known_clients, enum sol_lwm2m_security_mode sec_mode,
-    struct sol_vector *known_psks);
+    const char **known_clients, uint16_t num_sec_modes, ...);
 
 /**
  * @brief Adds a bootstrap request monitor to the server.
--- a/src/lib/comms/include/sol-socket.h
+++ b/src/lib/comms/include/sol-socket.h
@@ -42,6 +42,15 @@ struct sol_socket;
 typedef struct sol_socket sol_socket;
 
 /**
+ * @brief Represents supported Cipher Suites for use with DTLS.
+ */
+enum sol_socket_dtls_cipher {
+    SOL_SOCKET_DTLS_CIPHER_ECDH_ANON_AES128_CBC_SHA256,
+    SOL_SOCKET_DTLS_CIPHER_PSK_AES128_CCM8,
+    SOL_SOCKET_DTLS_CIPHER_ECDHE_ECDSA_AES128_CCM8,
+};
+
+/**
  * @brief Defines the behaviour of a socket instance
  */
 typedef struct sol_socket_options {
@@ -116,6 +125,19 @@ typedef struct sol_socket_ip_options {
      * sol_socket_bind()
      */
     bool reuse_addr;
+
+    /**
+     * @brief If @c secure is true, this should be considered.
+     * It indicates which DTLS cipher suites are supported and could
+     * be used for communication.
+     */
+    enum sol_socket_dtls_cipher *cipher_suites;
+
+    /**
+     * @brief If @c secure is true, this should be considered.
+     * It indicates the length of the @c cipher_suites array.
+     */
+    uint16_t cipher_suites_len;
 } sol_socket_ip_options;
 
 /**
--- a/src/lib/comms/sol-coap.c
+++ b/src/lib/comms/sol-coap.c
@@ -1885,6 +1885,9 @@ sol_coap_server_new_full(struct sol_socket_ip_options *options, const struct sol
         goto err_bind;
     }
 
+    SOL_DBG("server=%p, socket=%p, addr=%p, port=%" PRIu16 ", reuse_addr=%s binded!",
+        server, s, servaddr, servaddr->port, options->reuse_addr ? "True" : "False");
+
     server->refcnt = 1;
 
     sol_vector_init(&server->contexts, sizeof(struct resource_context));
@@ -1944,19 +1947,43 @@ err:
 }
 
 SOL_API struct sol_coap_server *
-sol_coap_server_new(const struct sol_network_link_addr *addr, bool secure)
+sol_coap_server_new(const struct sol_network_link_addr *addr,
+    bool secure)
 {
     return sol_coap_server_new_full(&((struct sol_socket_ip_options) {
-        .base = {
-            SOL_SET_API_VERSION(.api_version = SOL_SOCKET_OPTIONS_API_VERSION, )
-            SOL_SET_API_VERSION(.sub_api = SOL_SOCKET_IP_OPTIONS_SUB_API_VERSION, )
-            .on_can_read = on_can_read,
-            .on_can_write = on_can_write,
-        },
-        .family = addr->family,
-        .secure = secure,
-        .reuse_addr = addr->port ? true : false,
-    }), addr);
+            .base = {
+                SOL_SET_API_VERSION(.api_version = SOL_SOCKET_OPTIONS_API_VERSION, )
+                SOL_SET_API_VERSION(.sub_api = SOL_SOCKET_IP_OPTIONS_SUB_API_VERSION, )
+                .on_can_read = on_can_read,
+                .on_can_write = on_can_write,
+            },
+            .family = addr->family,
+            .secure = secure,
+            .cipher_suites = secure ?
+            (enum sol_socket_dtls_cipher []){ SOL_SOCKET_DTLS_CIPHER_PSK_AES128_CCM8 } : NULL,
+            .cipher_suites_len = secure ? 1 : 0,
+            .reuse_addr = addr->port ? true : false,
+        }), addr);
+}
+
+SOL_API struct sol_coap_server *
+sol_coap_server_new_by_cipher_suites(
+    const struct sol_network_link_addr *addr,
+    enum sol_socket_dtls_cipher *cipher_suites, uint16_t cipher_suites_len)
+{
+    return sol_coap_server_new_full(&((struct sol_socket_ip_options) {
+            .base = {
+                SOL_SET_API_VERSION(.api_version = SOL_SOCKET_OPTIONS_API_VERSION, )
+                SOL_SET_API_VERSION(.sub_api = SOL_SOCKET_IP_OPTIONS_SUB_API_VERSION, )
+                .on_can_read = on_can_read,
+                .on_can_write = on_can_write,
+            },
+            .family = addr->family,
+            .secure = true,
+            .cipher_suites = cipher_suites,
+            .cipher_suites_len = cipher_suites_len,
+            .reuse_addr = addr->port ? true : false,
+        }), addr);
 }
 
 SOL_API bool
--- a/src/lib/comms/sol-coap.c
+++ b/src/lib/comms/sol-coap.c
@@ -48,7 +48,7 @@ SOL_LOG_INTERNAL_DECLARE(_sol_coap_log_domain, "coap");
  * FIXME: use a random number between ACK_TIMEOUT (2000ms)
  * and ACK_TIMEOUT * ACK_RANDOM_FACTOR (3000ms)
  */
-#define ACK_TIMEOUT_MS 2345
+#define ACK_TIMEOUT_MS 12345
 #define MAX_RETRANSMIT 4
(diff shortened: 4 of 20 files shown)


src/lib/comms/include/sol-socket.h
```diff
@@ -116,6 +125,19 @@ typedef struct sol_socket_ip_options {
      * sol_socket_bind()
      */
     bool reuse_addr;
+
+    /**
+     * @brief If @c secure is true, this should be considered.
```
cabelitos: This comment makes no sense. "If secure is true" ??

src/lib/comms/include/sol-socket.h
```diff
@@ -116,6 +125,19 @@ typedef struct sol_socket_ip_options {
...
+     * It indicates which DTLS cipher suites are supported and could
+     * be used for communication.
+     */
+    enum sol_socket_dtls_cipher *cipher_suites;
+
+    /**
+     * @brief If @c secure is true, this should be considered.
```
cabelitos: Same thing

src/lib/comms/sol-socket-dtls-impl-tinydtls.c
```diff
@@ -708,14 +796,50 @@ sol_socket_default_dtls_new(const struct sol_socket_options *options)
...
+            s->handler.get_psk_info = get_psk_info;
+            break;
+        case SOL_SOCKET_DTLS_CIPHER_ECDHE_ECDSA_AES128_CCM8:
+            SOL_DBG("Adding get_ecdsa_* callbacks to %p handler", &s->handler);
+            s->handler.get_ecdsa_key = get_ecdsa_key;
+            s->handler.verify_ecdsa_key = verify_ecdsa_key;
+            break;
+        case SOL_SOCKET_DTLS_CIPHER_ECDH_ANON_AES128_CBC_SHA256:
+#ifdef DTLS_EXTRAS
+            //FIXME: What should be done here?
```
cabelitos: *universe explodes* - Say it's not supported and return

src/lib/comms/sol-socket-dtls-impl-tinydtls.c
```diff
@@ -655,6 +656,98 @@ get_psk_info(struct dtls_context_t *ctx, const session_t *session,
     return r;
 }
 
+static int
+get_ecdsa_key(struct dtls_context_t *ctx, const session_t *session,
+    const dtls_ecdsa_key_t **result)
+{
+    struct sol_socket_dtls *socket = dtls_get_app_data(ctx);
+    struct sol_network_link_addr addr;
+    static unsigned char ecdsa_priv_key[SOL_DTLS_ECDSA_PRIV_KEY_LEN],
```
cabelitos: What out for this static variable. Image what happens if you have two sockets on the same process using RPK....

src/lib/comms/sol-socket-dtls-impl-tinydtls.c
```diff
@@ -655,6 +656,98 @@ get_psk_info(struct dtls_context_t *ctx, const session_t *session,
...
+        ecdsa_pub_key_x[SOL_DTLS_ECDSA_PUB_KEY_X_LEN],
+        ecdsa_pub_key_y[SOL_DTLS_ECDSA_PUB_KEY_Y_LEN];
+    static dtls_ecdsa_key_t ecdsa_key;
```
cabelitos: Same thing

src/lib/comms/include/sol-coap.h
```diff
@@ -504,6 +505,25 @@ bool sol_coap_server_is_secure(const struct sol_coap_server *server);
 struct sol_coap_server *sol_coap_server_new(const struct sol_network_link_addr *addr, bool secure);
```
cabelitos: Add to the docs that this will create a server using PSK. And the recommend function is sol_coap_server_new_by_cipher_suites from now on.

cabelitos: Also add a @see sol_coap_server_new_by_cipher_suites

src/lib/comms/include/sol-lwm2m-bs-server.h
```diff
@@ -60,13 +60,20 @@ typedef struct sol_lwm2m_bootstrap_client_info sol_lwm2m_bootstrap_client_info;
...
+ * SOL_LWM2M_SECURITY_MODE_RAW_PUBLIC_KEY | struct sol_lwm2m_security_rpk *rpk, struct sol_blob **known_pub_keys | @c rpk This Bootstrap Server's Key Pair - @c known_pub_keys The Clients' Public Keys this Bootstrap Server has previous knowledge of.
+ *
+ * @note: SOL_LWM2M_SECURITY_MODE_CERTIFICATE is not supported yet.
+ *
  * @return The LWM2M bootstrap server or @c NULL on error.
  */
 struct sol_lwm2m_bootstrap_server *sol_lwm2m_bootstrap_server_new(uint16_t port,
-    const char **known_clients, enum sol_lwm2m_security_mode sec_mode,
-    struct sol_vector *known_psks);
+    const char **known_clients, uint16_t num_sec_modes, ...);
```
cabelitos: Use the enum

cabelitos: Forget it.

src/lib/comms/sol-lwm2m-bs-server.c
```diff
@@ -212,67 +213,95 @@ static const struct sol_coap_resource bootstrap_request_interface = {
...
-        SOL_WRN("Certificate security mode is not supported yet.");
-        return NULL;
-    case SOL_LWM2M_SECURITY_MODE_NO_SEC:
-        SOL_WRN("Bootstrap Server MUST use DTLS.");
-        return NULL;
-    default:
-        SOL_WRN("Unknown DTLS Security Mode: %d", sec_mode);
-        return NULL;
+
+    va_start(ap, num_sec_modes);
```
cabelitos: If num_sec_modes is 0 return an error

src/lib/comms/sol-lwm2m-bs-server.c
```diff
@@ -212,67 +213,95 @@ static const struct sol_coap_resource bootstrap_request_interface = {
...
-            sec_mode_str);
-        goto err_security;
-    } else {
-        SOL_DBG("Using %s security mode", sec_mode_str);
+    server->coap = sol_coap_server_new_by_cipher_suites(&servaddr,
+        cipher_suites, num_sec_modes);
+    SOL_NULL_CHECK_GOTO(server->coap, err_rpk_pair);
+
+    for (i = 0; i < num_sec_modes; i++) {
+        server->security = sol_lwm2m_bootstrap_server_security_add(server, sec_modes[i]);
```
cabelitos: This is useless. You're overwriting the server->security.

src/lib/comms/sol-lwm2m-bs-server.c
```diff
@@ -212,67 +213,95 @@ static const struct sol_coap_resource bootstrap_request_interface = {
...
+
+    sec_modes = calloc(num_sec_modes, sizeof(enum sol_lwm2m_security_mode));
+    SOL_NULL_CHECK_GOTO(sec_modes, err_cipher_suites);
+
+    for (i = 0; i < num_sec_modes; i++) {
+        sec_modes[i] = va_arg(ap, enum sol_lwm2m_security_mode);
+
+        switch (sec_modes[i]) {
+        case SOL_LWM2M_SECURITY_MODE_PRE_SHARED_KEY:
+            known_psks = va_arg(ap, struct sol_lwm2m_security_psk **);
```
cabelitos: You should sol_blob_ref them

src/lib/comms/sol-lwm2m-bs-server.c
```diff
@@ -212,67 +213,95 @@ static const struct sol_coap_resource bootstrap_request_interface = {
...
+            SOL_NULL_CHECK_GOTO(known_psks, err_sec_modes);
+
+            cipher_suites[i] = SOL_SOCKET_DTLS_CIPHER_PSK_AES128_CCM8;
+            break;
+        case SOL_LWM2M_SECURITY_MODE_RAW_PUBLIC_KEY:
+            my_rpk = va_arg(ap, struct sol_lwm2m_security_rpk *);
+            SOL_NULL_CHECK_GOTO(my_rpk, err_sec_modes);
+            known_pub_keys = va_arg(ap, struct sol_blob **);
```
cabelitos: You need to sol_blob_ref

src/lib/comms/sol-lwm2m-bs-server.c
```diff
@@ -212,67 +213,95 @@ static const struct sol_coap_resource bootstrap_request_interface = {
...
+
+        switch (sec_modes[i]) {
+        case SOL_LWM2M_SECURITY_MODE_PRE_SHARED_KEY:
+            known_psks = va_arg(ap, struct sol_lwm2m_security_psk **);
+            SOL_NULL_CHECK_GOTO(known_psks, err_sec_modes);
+
+            cipher_suites[i] = SOL_SOCKET_DTLS_CIPHER_PSK_AES128_CCM8;
+            break;
+        case SOL_LWM2M_SECURITY_MODE_RAW_PUBLIC_KEY:
+            my_rpk = va_arg(ap, struct sol_lwm2m_security_rpk *);
```
cabelitos: You need to sol_blob_ref them

src/lib/comms/sol-lwm2m-bs-server.c
```diff
@@ -212,67 +213,95 @@ static const struct sol_coap_resource bootstrap_request_interface = {
...
-        sol_vector_init(&server->known_psks, sizeof(sol_lwm2m_security_psk));
-
-        SOL_VECTOR_FOREACH_IDX (known_psks, given_psk, i) {
-            psk = sol_vector_append(&server->known_psks);
-            SOL_NULL_CHECK_GOTO(psk, err_psk);
-            psk->id = sol_blob_ref(given_psk->id);
-            psk->key = sol_blob_ref(given_psk->key);
+    for (i = 0; i < num_sec_modes; i++) {
+        if (sec_modes[i] == SOL_LWM2M_SECURITY_MODE_PRE_SHARED_KEY) {
+            server->known_psks = known_psks;
```
cabelitos: Don't do this. If the user frees the known_psks array we're dommed.

src/lib/comms/sol-lwm2m-bs-server.c
```diff
@@ -212,67 +213,95 @@ static const struct sol_coap_resource bootstrap_request_interface = {
...
+        } else if (sec_modes[i] == SOL_LWM2M_SECURITY_MODE_RAW_PUBLIC_KEY) {
+            server->rpk_pair.private_key = sol_blob_ref(my_rpk->private_key);
+            server->rpk_pair.public_key = sol_blob_ref(my_rpk->public_key);
+
+            server->known_pub_keys = known_pub_keys;
```
cabelitos: Same thing

src/lib/comms/sol-lwm2m-bs-server.c
```diff
@@ -212,67 +213,95 @@ static const struct sol_coap_resource bootstrap_request_interface = {
...
     //LWM2M Bootstrap Server MUST always use DTLS
-    if (sec_mode == SOL_LWM2M_SECURITY_MODE_PRE_SHARED_KEY) {
-        sol_vector_init(&server->known_psks, sizeof(sol_lwm2m_security_psk));
-
-        SOL_VECTOR_FOREACH_IDX (known_psks, given_psk, i) {
-            psk = sol_vector_append(&server->known_psks);
-            SOL_NULL_CHECK_GOTO(psk, err_psk);
-            psk->id = sol_blob_ref(given_psk->id);
-            psk->key = sol_blob_ref(given_psk->key);
+    for (i = 0; i < num_sec_modes; i++) {
```
cabelitos: I don't link this. You can do all the stuff below in the for/switch above.

src/lib/comms/sol-lwm2m-bs-server.c
```diff
@@ -212,67 +213,95 @@ static const struct sol_coap_resource bootstrap_request_interface = {
...
+    SOL_NULL_CHECK_GOTO(cipher_suites, err_va_list);
+
+    sec_modes = calloc(num_sec_modes, sizeof(enum sol_lwm2m_security_mode));
+    SOL_NULL_CHECK_GOTO(sec_modes, err_cipher_suites);
+
+    for (i = 0; i < num_sec_modes; i++) {
+        sec_modes[i] = va_arg(ap, enum sol_lwm2m_security_mode);
+
+        switch (sec_modes[i]) {
+        case SOL_LWM2M_SECURITY_MODE_PRE_SHARED_KEY:
```
cabelitos: You should add a check here. If you find the same security mode more than once you should return an error.

src/lib/comms/sol-lwm2m-server.c
```diff
@@ -812,67 +813,104 @@ observer_entry_del_monitor(struct observer_entry *entry,
...
-    case SOL_LWM2M_SECURITY_MODE_NO_SEC:
-        sec_mode_str = "NoSec";
-        SOL_DBG("Using NoSec Security Mode (No DTLS).");
-        break;
-    default:
-        SOL_WRN("Unknown DTLS Security Mode: %d", sec_mode);
-        return NULL;
+    va_start(ap, num_sec_modes);
+
+    if (num_sec_modes > 0) {
```
cabelitos: Same comments as found in the bootstrap.
* Do it in one pass.
* Do not allow the user to pass repeated security modes.
* remember to ref all the blobs (psks and rpks)
* Do not server->known_psks = known_psks.

src/lib/comms/sol-coap.c
```diff
@@ -48,7 +48,7 @@ SOL_LOG_INTERNAL_DECLARE(_sol_coap_log_domain, "coap");
  * FIXME: use a random number between ACK_TIMEOUT (2000ms)
  * and ACK_TIMEOUT * ACK_RANDOM_FACTOR (3000ms)
  */
-#define ACK_TIMEOUT_MS 2345
+#define ACK_TIMEOUT_MS 12345
```
cabelitos: This is OK for me. @vcgomes What do you think?

src/samples/coap/lwm2m-client.c
```diff
@@ -1263,6 +1250,46 @@ main(int argc, char *argv[])
...
+            r = sol_util_base16_decode(buf_aux, sizeof(buf_aux),
+                sol_str_slice_from_str(BS_SERVER_PUBLIC_KEY), SOL_DECODE_BOTH);
+            server_public_key = sol_blob_new_dup(buf_aux, RPK_PUBLIC_KEY_LEN);
+        } else {
+            r = sol_util_base16_decode(buf_aux, sizeof(buf_aux),
+                sol_str_slice_from_str(SERVER_PUBLIC_KEY), SOL_DECODE_BOTH);
+            server_public_key = sol_blob_new_dup(buf_aux, RPK_PUBLIC_KEY_LEN);
+        }
+        break;
+    case SOL_LWM2M_SECURITY_MODE_CERTIFICATE:
```
cabelitos: This is case, print that it's not supported and return

bsmelo: I left it like this so we can go through the sample code itself and only fail on a "Certificate Mode Not Supported" at the actual APIs; kind of a test. What do you think?

src/lib/comms/include/sol-socket.h
bsmelo: Why?

There's a `bool secure` inside this same `struct`, and these two fields only apply to the case in which that `bool` is `true`.

src/lib/comms/include/sol-socket.h
bsmelo: Ditto.

src/lib/comms/sol-socket-dtls-impl-tinydtls.c
bsmelo: Well, I will remove the `#ifdef` then, since I will just say it's unsupported and return in both `#ifdef` branches.

src/lib/comms/sol-socket-dtls-impl-tinydtls.c
bsmelo: If the two sockets try to handshake at the same time, a race-condition can happen on this variable indeed; otherwise, no problem.

What do you suggest? The only place I see where this could go into to solve that is the `struct sol_socket_dtls`; is it ok to move it there?

src/lib/comms/sol-socket-dtls-impl-tinydtls.c
bsmelo: Ditto. (actually only this one would be present inside the `sol_socket_dtls` and the auxiliary vars above would not be needed.)

src/lib/comms/include/sol-coap.h
bsmelo: Done.

src/lib/comms/sol-lwm2m-bs-server.c
bsmelo: Added `SOL_INT_CHECK(num_sec_modes, == 0, NULL);` before the `va_start()` call.

src/lib/comms/sol-lwm2m-bs-server.c
bsmelo: No I'm not. Look at the code for `_security_add`: it works like an append operation. I add the sec_modes[i] to the same `server->security` object everytime - I only create the object if it doesn't exist already.

src/lib/comms/sol-lwm2m-bs-server.c
bsmelo: I don't get it. I don't have the `server` allocated in the `switch` there, and I don't have the `sol_coap_server server->coap` allocated in the `for` there. I need both those objects set to use `_security_add()`.

src/lib/comms/sol-lwm2m-bs-server.c
bsmelo: Done.

src/lib/comms/include/sol-socket.h
cabelitos: nevermind

src/lib/comms/sol-socket-dtls-impl-tinydtls.c
cabelitos: k

src/lib/comms/sol-lwm2m-bs-server.c
cabelitos: My bad

src/samples/coap/lwm2m-client.c
cabelitos: I think you should do this here as well.

bsmelo:
An updated version is at #2298

PR #2298, opened 2016-08-23
LWM2M: Support for DTLS with Raw Public Key security mode [v2]

bsmelo:
Main changes since v1 (#2296)
- Change of a `static` variable in tinydtls https://github.com/solettaproject/soletta/pull/2296#discussion-diff-75605632
- `sol_blob_ref()`s to solve https://github.com/solettaproject/soletta/pull/2296#discussion-diff-75606155

Signed-off-by: Bruno Melo bsilva.melo@gmail.com

PR #2299, opened 2016-08-23
LWM2M: Support for DTLS with Raw Public Key security mode [v3]

bsmelo:
Main changes since v2 (#2298)
- Minor memory leaking fixes
- Removed the "Ugly" patch for CoAP timeout. The timeout issue (which generates duplicated packets) seems to happen only in Client-Server/BSServer communication between the samples, and not in the same communication when it's done inside the `test-lwm2m.c` program. I guess it has something to do with the fact that the samples are running in different processes (since that's the only difference between samples and tests), but I still don't know what/where exactly is the issue.

### DIFF
--- a/src/lib/comms/sol-socket-dtls-impl-tinydtls.c
+++ b/src/lib/comms/sol-socket-dtls-impl-tinydtls.c
@@ -47,6 +47,8 @@ struct sol_socket_dtls {
     struct sol_socket *wrapped;
     struct sol_timeout *retransmit_timeout;
     dtls_context_t *context;
+    dtls_handler_t handler;
+    dtls_ecdsa_key_t ecdsa_key;
     struct {
         bool (*cb)(void *data, struct sol_socket *s);
         struct sol_vector queue;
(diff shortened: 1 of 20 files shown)


cabelitos:
@bsmelo In my machine the RPK is working perfectly without the necesseity to increase the timeout. Maybe is something related to your machine?
Well, I'll merge this series. Looks good as is.

src/lib/comms/sol-socket-dtls-impl-tinydtls.c
```diff
@@ -655,6 +661,109 @@ get_psk_info(struct dtls_context_t *ctx, const session_t *session,
...
+        socket->credentials->clear(socket->credentials->data);
+
+    return 0;
+
+err_getters:
+    free(socket->ecdsa_key.pub_key_y);
+err_pub_key_y:
+    free(socket->ecdsa_key.pub_key_x);
+err_pub_key_x:
+    free(socket->ecdsa_key.priv_key);
```
cabelitos: you forgot the return r in here!

bsmelo: updated @_@

cabelitos:
Merged!

PR #2302, opened 2016-08-30
LWM2M: Change type of bs_server->known_clients

bsmelo:
This patch changes the type of the `known_clients` frield from
`struct sol_lwm2m_bootstrap_server`, to avoid possible issues in
case the user of the API deallocates the strings with the clients'
names.

### DIFF
--- a/src/lib/comms/sol-lwm2m-bs-server.c
+++ b/src/lib/comms/sol-lwm2m-bs-server.c
@@ -148,7 +148,7 @@ bootstrap_request(void *data, struct sol_coap_server *coap,
     struct sol_lwm2m_bootstrap_client_info *bs_cinfo;
     struct sol_lwm2m_bootstrap_server *server = data;
     struct sol_coap_packet *response;
-    struct sol_str_slice client_name = SOL_STR_SLICE_EMPTY;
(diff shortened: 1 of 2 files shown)


src/lib/comms/sol-lwm2m-bs-server.c
```diff
@@ -314,23 +315,31 @@ sol_lwm2m_bootstrap_server_new(uint16_t port, const char **known_clients,
...
 
-    free(sec_modes);
-    free(cipher_suites);
+    sol_vector_init(&server->known_clients, sizeof(struct sol_str_slice));
 
-    server->known_clients = known_clients;
+    for (i = 0; known_clients[i]; i++) {
+        known_client = sol_vector_append(&server->known_clients);
+        SOL_NULL_CHECK_GOTO(known_client, err_security);
+        *known_client = sol_str_slice_from_str(known_clients[i]);
```
cabelitos: We still have the problem, the user can free(known_clients[i]) from outside. Maybe you should create a sol_ptr_vector and strdup() the known_clients.

cabelitos: By doing so, we will be completly safe

bsmelo:
You're right.
An updated version is at #2306.

PR #2297, opened 2016-08-23
comms: Add sol-netctl agent API

wzhen12:
The agent is necessary in the process of network connection,
that will receive input information from the user and send it
to the network(such as AP and so on).

The API is based on connman D-Bus APIs, it provides register agent,
input login information and so on netctl agent API.

Signed-off-by: Wu Zheng wu.zheng@intel.com

PR #2303, opened 2016-09-01
comms: Add sol-netctl agent API V3

wzhen12:

### DIFF
--- a/src/lib/comms/include/sol-netctl.h
+++ b/src/lib/comms/include/sol-netctl.h
@@ -75,6 +75,13 @@ typedef struct sol_netctl_network_params {
     struct sol_network_link_addr gateway;
 } sol_netctl_network_params;
 
+#define SOL_NETCTL_AGENT_NAME         "Name"    /**< @brief the agent input name type string */
+#define SOL_NETCTL_AGENT_IDENTITY     "Identity"    /**< @brief the agent input identity type string */
+#define SOL_NETCTL_AGENT_PASSPHRASE   "Passphrase"    /**< @brief the agent input passphrase type string */
+#define SOL_NETCTL_AGENT_WPS          "WPS"    /**< @brief the agent input WPS type string */
+#define SOL_NETCTL_AGENT_USERNAME     "Username"    /**< @brief the agent input username type string */
+#define SOL_NETCTL_AGENT_PASSWORD     "Password"    /**< @brief the agent input password type string */
+
 /**
  * @enum sol_netctl_service_state
  *
@@ -207,6 +214,60 @@ enum sol_netctl_state {
      */
     SOL_NETCTL_STATE_OFFLINE,
 };
+
+/**
+ * @brief agent input struct
+ *
+ * This struct contains the information of agent input.
+ */
+typedef struct sol_netctl_agent_input {
+    /**
+     * @brief The agent prompt type
+     */
+    char *type;
+    /**
+     * @brief The agent input value
+     */
+    char *input;
+} sol_netctl_agent_input;
+
+/**
+ * @brief agent callback functions
+ *
+ * This struct contains the callback functions of agent.
+ */
+typedef struct sol_netctl_agent {
+    /**
+     * @brief connection error callback used to inform connection failure
+     *
+     * @param data the user data
+     * @param service the connection failure service
+     * @param error the error information
+     */
+    void (*report_error)(void *data, const struct sol_netctl_service *service,
+        const char *error);
+    /**
+     * @brief connection input callback used to inform connection login input
+     *
+     * @param data the user data
+     * @param service the connection login input service
+     * @param inputs the vector of login input type
+     */
+    void (*request_input)(void *data, const struct sol_netctl_service *service,
+        const struct sol_vector *inputs);
+    /**
+     * @brief connection cancel callback used to inform connection cancel
+     *
+     * @param data the user data
+     */
+    void (*cancel)(void *data);
+    /**
+     * @brief agent release callback used to inform agent release
+     *
+     * @param data the user data
+     */
+    void (*release)(void *data);
+} sol_netctl_agent;
 
 /**
  * @brief Service monitor callback used to inform a service changed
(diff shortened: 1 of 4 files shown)


wzhen12:
The Version 3 of Add sol-netctl agent API.

1. Modify the version 2 errors. (#2297)

2. At current, scan feature only exists in wifi plugin in connman.
The other technology plugins don't exist scan feature.
Therefore, at current, wifi scan is the all technology scan. 
And the scan feature isn't modified in the patch.

3. In the patches, the vector isn't released in sol_netctl_request_input API interfaces. 
The suggestion of Gustavo has been implemented,
the system vector only provides the types of inputs.
The user will create the new vector and add the all items to the vector. (including input, type and so on)
At the same time, the user will release the new vector too.
You can check the agent sample for how to use it.

3. ptr_vector has been replaced by vector.

4. WPS, Passphrase, Identity, username, password.
a. It has been checked in _agent_input_properties.
b. In the function of _agent_input_properties,
wether the input item exists in vector or not can't be checked.
(connman has done for it)
c. wether the wps item has been got or not can be checked.
if wps has been got, we will ignore the other items (such as Identity, passphrase and so).
d. username and password will not be got with passphrase, Identity at the same time.
Therefore, we don't care about  username and password/passphrase and Identity.
e. passphrase and Identity can be got at the same time.
Therefore, we don't care about passphrase and Identity.
More cases? Any suggestions? Thanks.

@cabelitos @barbieri
Please check it.

wzhen12:
The compiling is failure.
It is the format issue.
It will be fixed in other patch.

src/lib/comms/include/sol-netctl.h
```diff
@@ -553,6 +614,67 @@ sol_netctl_find_service_by_name(const char *service_name)
...
+ * @see struct sol_netctl_agent
+ *
+ * @param agent the agent Callback struct to be called when the related information is updated.
+ * NULL is used to unregister the agent.
+ * @param data Add a user data per callback.
+ *
+ * @return 0 on success, -errno on failure.
+ */
+int
+sol_netctl_register_agent(const struct sol_netctl_agent *agent, const void *data);
```
ceolin: soletta does not break line for the function type in the headers

ceolin: I'm wondering whether this function is too complicated or not. I mean, this function has 3 different behaviours

1- Register a new agent when there is no previous agent registered - Success
2- Register a new agent when there is an agent - Fail
3- Unregister an agent passing NULL - What will return if there is no agent registered ?

Maybe it's worth add a separated function to unregister the agent.

src/lib/comms/include/sol-netctl.h
```diff
@@ -553,6 +614,67 @@ sol_netctl_find_service_by_name(const char *service_name)
...
+ * If retry connection is selected, the failure network connection will be tried to
+ * connect. If not retry connection is selected, the failure network connection will
+ * not be tried to connect.
+ * The connection failure information is informed via the agent callback.
+ * The agent must be registered before using sol_netctl_report_error.
+ *
+ * @see sol_netctl_register_agent
+ *
+ * @param service The service structure which the network connection is desired.
+ * @param is_type Ture is retry the failure selected network connection,
```
ceolin: s/Ture/True/

ceolin: wouldn't `retry` a better name ?

src/lib/comms/include/sol-netctl.h
```diff
@@ -553,6 +614,67 @@ sol_netctl_find_service_by_name(const char *service_name)
...
+ * @param service The service structure which the network connection is desired.
+ * @param inputs the vector of input types and inputs from the user.
+ *
+ * @return 0 on success, -errno on failure.
+ */
+int
+sol_netctl_request_input(struct sol_netctl_service *service,
+    const struct sol_vector *inputs);
+
+/**
```
ceolin: I think is better you merge this commit with the one that implements these functions.

src/lib/comms/sol-netctl-impl-connman.c
```diff
@@ -58,8 +61,15 @@ struct ctx {
...
     sd_bus_slot *state_slot;
+    sd_bus_slot *agent_slot;
+    sd_bus_slot *vtable_slot;
+    sd_bus_message *agent_msg;
+    struct sol_netctl_service *auth_service;
+    const struct sol_netctl_agent *agent;
+    const void *agent_data;
     enum sol_netctl_state connman_state;
     int32_t refcount;
+    struct sol_vector agent_vector;
```
ceolin: put it just after `agent_data` to avoid holes in the struct

src/lib/comms/sol-netctl-impl-connman.c
```diff
@@ -1212,3 +1242,255 @@ sol_netctl_service_state_to_str(enum sol_netctl_service_state state)
...
+    pending->agent_data = NULL;
+
+    _ctx.agent_msg = sd_bus_message_unref(_ctx.agent_msg);
+    _ctx.auth_service = NULL;
+}
+
+static int
+agent_cancel(sd_bus_message *m, void *data, sd_bus_error *ret_error)
+{
+    struct ctx *context = (struct ctx *)data;
```
ceolin: not necessary cast

src/lib/comms/sol-netctl-impl-connman.c
```diff
@@ -1212,3 +1242,255 @@ sol_netctl_service_state_to_str(enum sol_netctl_service_state state)
...
+    const struct sol_netctl_agent *agent = context->agent;
+    void *userdata = (void *)context->agent_data;
```
ceolin: you don't need this variable, you can cast `context->agent_data` when you use.

src/lib/comms/sol-netctl-impl-connman.c
```diff
@@ -1212,3 +1242,255 @@ sol_netctl_service_state_to_str(enum sol_netctl_service_state state)
...
+
+    return 0;
+}
+
+static int
+agent_release(sd_bus_message *m, void *data, sd_bus_error *ret_error)
+{
+    struct ctx *context = (struct ctx *)data;
+    const struct sol_netctl_agent *agent = context->agent;
+    void *userdata = (void *)context->agent_data;
```
ceolin: ditto

src/lib/comms/sol-netctl-impl-connman.c
```diff
@@ -1212,3 +1242,255 @@ sol_netctl_service_state_to_str(enum sol_netctl_service_state state)
...
+    if (agent->cancel)
+        agent->cancel(userdata);
+
+    return 0;
+}
+
+static int
+agent_release(sd_bus_message *m, void *data, sd_bus_error *ret_error)
+{
+    struct ctx *context = (struct ctx *)data;
```
ceolin: ditto

src/lib/comms/sol-netctl-impl-connman.c
```diff
@@ -1212,3 +1242,255 @@ sol_netctl_service_state_to_str(enum sol_netctl_service_state state)
...
+        goto error;
+
+    r = sd_bus_message_read(m, "os", &path, &err);
+    SOL_INT_CHECK(r, < 0, r);
+
+    service = find_service_by_path(path);
+    SOL_NULL_CHECK_GOTO(service, error);
+
+    context->agent_msg = sd_bus_message_unref(context->agent_msg);
+    context->agent_msg = sd_bus_message_ref(m);
```
ceolin: check the return of `sd_bus_message_ref`

src/lib/comms/sol-netctl-impl-connman.c
```diff
@@ -1212,3 +1242,255 @@ sol_netctl_service_state_to_str(enum sol_netctl_service_state state)
...
+    int r;
+    sd_bus *bus = sol_bus_client_get_bus(_ctx.connman);
+    const char *path = CONNMAN_AGENT_PATH;
+
+    SOL_NULL_CHECK(bus, -EINVAL);
+
+    if (_ctx.agent && agent)
+        return -EEXIST;
+
+    if (!agent) {
```
ceolin: i guess it should be `if (!agent && _ctx.agent)`

src/lib/comms/include/sol-netctl.h
wzhen12: ok, the unregister the agent will be added.

src/lib/comms/include/sol-netctl.h
wzhen12: ok

src/lib/comms/include/sol-netctl.h
wzhen12: ok

src/lib/comms/sol-netctl-impl-connman.c

src/lib/comms/sol-netctl-impl-connman.c
wzhen12: It will be changed.

src/lib/comms/sol-netctl-impl-connman.c

src/lib/comms/sol-netctl-impl-connman.c

src/lib/comms/sol-netctl-impl-connman.c

src/lib/comms/sol-netctl-impl-connman.c

src/lib/comms/sol-netctl-impl-connman.c
wzhen12: It will be changed.

PR #2305, opened 2016-09-03
linux:mainloop:log: Fix multiple initializations

ceolin:
There is problem when sol_log_impl_init()/sol_log_impl_shutdown() is
called more than once. The global mutex _mutex is statically initialized
but soletta is destroying it on shutdown(). So, all further functions
that use this mutex (for lock() and unlock())will fail because the init
is not initializing it.

This commit remove the destroy() from the shutdown(). Because there is
no reason to init/destroy the mutex instead of let it available.

### DIFF
--- a/src/lib/common/sol-log-impl-linux.c
+++ b/src/lib/common/sol-log-impl-linux.c
@@ -266,7 +266,6 @@ sol_log_impl_shutdown(void)
     _main_pid = 0;
 #ifdef PTHREAD
     _main_thread = 0;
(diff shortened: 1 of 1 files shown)


cabelitos:
lgtm

barbieri:
+0

maybe another option is to always init it... but what you did is okay, does valgrind bitch?

ceolin:
@barbieri i was in doubt about always init or never destroy, as the second one is simpler i proceed with that. I didn't run valgrind, I'll do it now and if no problem appears I'll commit this.

PR #2307, opened 2016-09-08
Add sol-netctl API to get extra service properties V2

guchaojie:
This PR is for to improve getting extra service properties for user
All the properties provided by Connman D-Bus Hierarchy

### DIFF
--- a/src/lib/comms/include/sol-netctl.h
+++ b/src/lib/comms/include/sol-netctl.h
@@ -153,6 +153,86 @@ enum sol_netctl_service_state {
     SOL_NETCTL_SERVICE_STATE_REMOVE,
 };
 
+/**
+ * @brief method of proxy generated for a network link.
+ *
+ */
+enum sol_proxy_method {
+    SOL_PROXY_METHOD_DIRECT,
+    SOL_PROXY_METHOD_AUTO,
+    SOL_PROXY_METHOD_MANUAL,
+};
+
+/**
+ * @brief Structure to represent a network proxy.
+ *
+ * This struct contains the necessary information to present a
+ * network proxy for user.
+ */
+typedef struct sol_proxy {
+#ifndef SOL_NO_API_VERSION
+#define SOL_PROXY_API_VERSION (1)
+    /** @brief API version */
+    uint16_t api_version;
+#endif
+    /** @brief proxy method provided */
+    enum sol_proxy_method method;
+    /** @brief automatic proxy configuration URL */
+    char *url;
+    /** @brief List of proxy URIs when manual method is set */
+    sol_ptr_vector servers;
+    /** @brief List of host can be accessed directly */
+    sol_ptr_vector excludes;
+} sol_proxy;
+
+/**
+ * @brief Structure to represent a network provider.
+ *
+ * This struct contains the necessary information to present a
+ * network provider for user.
+ */
+typedef struct sol_provider {
+#ifndef SOL_NO_API_VERSION
+#define SOL_PROVIDER_API_VERSION (1)
+    /** @brief API version */
+    uint16_t api_version;
+#endif
+    /** @brief VPN host IP */
+    char *host;
+    /** @brief VPN Domain */
+    char *domain;
+    /** @brief provider name */
+    char *name;
+    /** @brief provider type */
+    char *type;
+} sol_provider;
+
+/**
+ * @brief Structure to represent a network ethernet.
+ *
+ * This struct contains the necessary information to present a
+ * network ethernet for user.
+ */
+typedef struct sol_ethernet {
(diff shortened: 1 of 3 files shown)


src/lib/comms/include/sol-netctl.h
```diff
@@ -153,6 +153,86 @@ enum sol_netctl_service_state {
...
+    SOL_PROXY_METHOD_MANUAL,
+};
+
+/**
+ * @brief Structure to represent a network proxy.
+ *
+ * This struct contains the necessary information to present a
+ * network proxy for user.
+ */
+typedef struct sol_proxy {
```
ceolin: I think it would be better use the complete namespace `sol_netctl_*` as it's done in all other types in this file.

src/lib/comms/include/sol-netctl.h
```diff
@@ -153,6 +153,86 @@ enum sol_netctl_service_state {
...
+    sol_ptr_vector excludes;
+} sol_proxy;
+
+/**
+ * @brief Structure to represent a network provider.
+ *
+ * This struct contains the necessary information to present a
+ * network provider for user.
+ */
+typedef struct sol_provider {
```
ceolin: ditto

src/lib/comms/include/sol-netctl.h
```diff
@@ -268,6 +348,20 @@ enum sol_netctl_service_state sol_netctl_service_get_state(
     const struct sol_netctl_service *service);
 
 /**
+ * @brief Gets the service error
+ *
+ * This function gets the error for the netctl service,
+ * after the service is to give by service monitor.
+ *
+ * @param service The service structure which the name is desired
```
ceolin: s/structure/struct/

src/lib/comms/include/sol-netctl.h
```diff
@@ -268,6 +348,20 @@ enum sol_netctl_service_state sol_netctl_service_get_state(
...
+ *
+ * @return The error during connection or disconnection, NULL on no
```
ceolin: NULL if no error occur

src/lib/comms/include/sol-netctl.h
```diff
@@ -303,12 +397,152 @@ int sol_netctl_service_get_network_address(
...
 /**
+ * @brief Gets the service favorite
+ *
+ * This function checks if the service is favorite or not for the netctl
+ * service, after the service is to give by service monitor.
+ *
+ * @param service The service structure which the service favorite is
+ * desired
+ *
+ * @return a boolean value if favorite is true or false.
```
ceolin: `@c true or @c false`

ceolin: review this in all commit

src/lib/comms/sol-netctl-impl-connman.c
```diff
@@ -328,8 +484,317 @@ get_service_ip(sd_bus_message *m,
...
+                { }
+            };
+
+            r = sd_bus_message_enter_container(m, SD_BUS_TYPE_VARIANT, "s");
+            SOL_INT_CHECK(r, < 0, r);
+
+            r = sd_bus_message_read_basic(m, SD_BUS_TYPE_STRING, &method);
+            SOL_INT_CHECK_GOTO(r, < 0, end);
+
+            if (method)
```
ceolin: is it possible method be NULL here ? if yes shouldn't it be treated as an error ?

src/lib/comms/include/sol-netctl.h
```diff
@@ -153,6 +153,86 @@ enum sol_netctl_service_state {
...
+    /** @brief Interface name */
+    char *interface;
+    /** @brief Ethernet device address */
+    char *address;
+    /** @brief The ethernet MTU */
+    uint16_t mtu;
+    /** @brief Selected speed of line */
+    uint16_t speed;
+    /** @brief Selected duplex settings of line */
+    char *duplex;
```
ceolin: put this above `mtu` to avoid possible holes in the struct

ceolin:
the commit looks good to me, just minor things to review.

cabelitos:
@ceolin I guess @guchaojie forgot to set the api_versions!

src/lib/comms/sol-netctl-impl-connman.c
guchaojie: it is possible ,  but according to API spec , there is no reference to treat as an error.
please check: https://www.freedesktop.org/software/systemd/man/sd_bus_message_read_basic.html

src/lib/comms/include/sol-netctl.h
guchaojie: sorry, I don't get your meaning . Put what above mtu?

PR #2308, opened 2016-09-12
comms: Add sol-netctl agent API V4

wzhen12:

### DIFF
--- a/src/lib/comms/include/sol-netctl.h
+++ b/src/lib/comms/include/sol-netctl.h
@@ -75,6 +75,13 @@ typedef struct sol_netctl_network_params {
     struct sol_network_link_addr gateway;
 } sol_netctl_network_params;
 
+#define SOL_NETCTL_AGENT_NAME         "Name"    /**< @brief the agent input name type string */
+#define SOL_NETCTL_AGENT_IDENTITY     "Identity"    /**< @brief the agent input identity type string */
+#define SOL_NETCTL_AGENT_PASSPHRASE   "Passphrase"    /**< @brief the agent input passphrase type string */
+#define SOL_NETCTL_AGENT_WPS          "WPS"    /**< @brief the agent input WPS type string */
+#define SOL_NETCTL_AGENT_USERNAME     "Username"    /**< @brief the agent input username type string */
+#define SOL_NETCTL_AGENT_PASSWORD     "Password"    /**< @brief the agent input password type string */
+
 /**
  * @enum sol_netctl_service_state
  *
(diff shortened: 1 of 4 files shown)


wzhen12:
The Version 4 of Add sol-netctl agent API.
1. Modify the version 3 errors. (#2303)
2. Split register_agent to register_agent and unregister_agent
3. Rename report_error to request_retry

@ceolin  @barbieri  Please check it. Thanks.

src/lib/comms/sol-netctl-impl-connman.c
```diff
@@ -1212,3 +1242,382 @@ sol_netctl_service_state_to_str(enum sol_netctl_service_state state)
...
+        agent->release((void *)context->agent_data);
+
+    return 0;
+}
+
+static int
+agent_report_error(sd_bus_message *m, void *data, sd_bus_error *ret_error)
+{
+    struct sol_netctl_service *service;
+    struct ctx *context = (struct ctx *)data;
```
ceolin: C does not require this cast.

src/lib/comms/sol-netctl-impl-connman.c
```diff
@@ -1212,3 +1242,382 @@ sol_netctl_service_state_to_str(enum sol_netctl_service_state state)
...
+
+    return 0;
+}
+
+SOL_API int
+sol_netctl_unregister_agent(void)
+{
+    int r;
+    sd_bus *bus = sol_bus_client_get_bus(_ctx.connman);
+    const char *path = CONNMAN_AGENT_PATH;
```
ceolin: why don't use `CONNMAN_AGENT_PATH` directly ?

src/lib/comms/sol-netctl-impl-connman.c
```diff
@@ -1212,3 +1242,382 @@ sol_netctl_service_state_to_str(enum sol_netctl_service_state state)
...
+
+    return -ENOMEM;
+}
+
+static int
+agent_request_input(sd_bus_message *m, void *data, sd_bus_error *ret_error)
+{
+    struct sol_netctl_service *service;
+    const char *path;
+    struct ctx *context = (struct ctx *)data;
```
ceolin: same here

src/lib/comms/include/sol-netctl.h
```diff
@@ -209,6 +216,60 @@ enum sol_netctl_state {
...
+        const char *error);
+    /**
+     * @brief connection input callback used to inform connection login input
+     *
+     * @param data the user data
+     * @param service the connection login input service
+     * @param inputs the vector of login input type
+     */
+    void (*request_input)(void *data, const struct sol_netctl_service *service,
+        const struct sol_vector *inputs);
```
ceolin: seing the sample this method is not clear to me. This method is called when connman needs info for a specific input right ? In this implementation the user (that have registered the agent) will be called with a list of inputs, then they will need to create another list of inputs with the answers and then call `sol_netctl_request_input` ? Why we simply don't allow the user fill the answer and we use it ?

What I told is right ?

wzhen12: @ceolin 
It is gustavo's suggestion.
If any not clear, @barbieri please figure it out. Thanks.
The reason is that the user can keep the malloc/free the vector data.
If  fill the answer, it is confused to malloc/free the vector data.

src/lib/comms/sol-netctl-impl-connman.c
wzhen12: ok, remove path.

src/lib/comms/sol-netctl-impl-connman.c

src/lib/comms/sol-netctl-impl-connman.c

PR #2312, opened 2016-09-18
comms: Add sol-netctl agent API V5

wzhen12:

### DIFF
--- a/data/scripts/libsoletta.sym
+++ b/data/scripts/libsoletta.sym
@@ -892,4 +892,5 @@ global:
         sol_netctl_unregister_agent;
         sol_netctl_request_retry;
         sol_netctl_request_input;
+        sol_netctl_scan;
 } LIBSOLETTA_1;
--- a/data/scripts/libsoletta.sym
+++ b/data/scripts/libsoletta.sym
@@ -885,3 +885,11 @@ global:
 local:
         *;
 };
+
+LIBSOLETTA_2 {
+global:
+        sol_netctl_register_agent;
+        sol_netctl_unregister_agent;
+        sol_netctl_request_retry;
+        sol_netctl_request_input;
+} LIBSOLETTA_1;
--- a/src/lib/comms/include/sol-netctl.h
+++ b/src/lib/comms/include/sol-netctl.h
@@ -679,6 +679,18 @@ int sol_netctl_request_input(struct sol_netctl_service *service,
     const struct sol_ptr_vector *inputs);
 
 /**
+ * @brief request scan the surrounding devices
+ *
+ * Request scan for the surrounding devices.
+ * This must be inovked AFTER sol_netctl_add_service_monitor.
+ *
+ * @see sol_netctl_add_service_monitor
+ *
+ * @return 0 on success, -errno on failure.
+ */
+int sol_netctl_scan(void);
+
+/**
  * @}
  */
 
--- a/src/lib/comms/include/sol-netctl.h
+++ b/src/lib/comms/include/sol-netctl.h
@@ -75,6 +75,13 @@ typedef struct sol_netctl_network_params {
     struct sol_network_link_addr gateway;
 } sol_netctl_network_params;
 
+#define SOL_NETCTL_AGENT_NAME         "Name"    /**< @brief the agent input name type string */
+#define SOL_NETCTL_AGENT_IDENTITY     "Identity"    /**< @brief the agent input identity type string */
+#define SOL_NETCTL_AGENT_PASSPHRASE   "Passphrase"    /**< @brief the agent input passphrase type string */
+#define SOL_NETCTL_AGENT_WPS          "WPS"    /**< @brief the agent input WPS type string */
+#define SOL_NETCTL_AGENT_USERNAME     "Username"    /**< @brief the agent input username type string */
+#define SOL_NETCTL_AGENT_PASSWORD     "Password"    /**< @brief the agent input password type string */
+
 /**
  * @enum sol_netctl_service_state
  *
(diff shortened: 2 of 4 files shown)


wzhen12:
The Version 5 of Add sol-netctl agent API.
1. Modify the version 3 errors. (#2308)
2.This method is called when connman needs info for a specific input right ? In this implementation the user (that have registered the agent) will be called with a list of inputs, then they will need to create another list of inputs with the answers and then call sol_netctl_request_input ? Why we simply don't allow the user fill the answer and we use it ?
Answer: It is gustavo's suggestion.
If any not clear, @barbieri please figure it out. Thanks.
The reason is that the user can keep the malloc/free the vector data.
If fill the answer, it is confused to malloc/free the vector data for the user.

@ceolin and @barbieri Please check it.

src/lib/comms/include/sol-netctl.h
```diff
@@ -553,6 +614,73 @@ sol_netctl_find_service_by_name(const char *service_name)
...
+ * @param data Add a user data per callback.
+ *
+ * @return 0 on success, -errno on failure.
+ */
+int sol_netctl_register_agent(const struct sol_netctl_agent *agent, const void *data);
+
+/**
+ * @brief unregister a agent for network connection
+ *
+ * unregister the previous and to unregister the last agent you should use it.
```
ceolin: it's very confusing this phrase. To my mind you don't need this, the brief is good enough.

src/lib/comms/include/sol-netctl.h
```diff
@@ -553,6 +614,73 @@ sol_netctl_find_service_by_name(const char *service_name)
...
+ * @see sol_netctl_register_agent
+ *
+ * @param service The service structure which the network connection is desired.
+ * @param is_type True is retry the failure selected network connection,
+ * false is not retry the failure selected network connection.
+ *
+ * @return 0 on success, -errno on failure.
+ */
+int sol_netctl_request_retry(struct sol_netctl_service *service,
+    bool is_type);
```
ceolin: the name of the variable `is_type` is unclear, shouldn't be better `retry` ?

src/lib/comms/sol-netctl-impl-connman.c
```diff
@@ -797,6 +807,23 @@ sol_netctl_service_disconnect(struct sol_netctl_service *service)
         "net.connman.Service", "Disconnect", _service_disconnect, service, NULL);
 }
 
+static void
+_destory_agent_vector(struct sol_vector *vector)
```
ceolin: typo, s/_destory/_destroy/g

src/lib/comms/sol-netctl-impl-connman.c
```diff
@@ -797,6 +807,23 @@ sol_netctl_service_disconnect(struct sol_netctl_service *service)
...
+{
+    uint16_t i;
+    sol_netctl_agent_input *input;
+
+    if (!vector)
+        return;
+
+    SOL_VECTOR_FOREACH_IDX (vector, input, i) {
+        if (input->type)
+            free(input->type);
```
ceolin: if you never use `input->input` why we are keeping a vector of `sol_netctl_agent_input` ? we can keep only the input types (char *).

src/lib/comms/include/sol-netctl.h
```diff
@@ -209,6 +216,60 @@ enum sol_netctl_state {
...
+        const char *error);
+    /**
+     * @brief connection input callback used to inform connection login input
+     *
+     * @param data the user data
+     * @param service the connection login input service
+     * @param inputs the vector of login input type
+     */
+    void (*request_input)(void *data, const struct sol_netctl_service *service,
+        const struct sol_vector *inputs);
```
ceolin: well, if the user needs to call `sol_netctl_request_input` and can not simply fill the field input value, why we are using `struct sol_nectl_agent_input`  in inputs ? I think it will be better change inputs to a `sol_ptr_vector` of ` char *` with the types, don't you think ?

src/lib/comms/sol-netctl-impl-connman.c
```diff
@@ -1212,3 +1242,380 @@ sol_netctl_service_state_to_str(enum sol_netctl_service_state state)
...
+    _release_agent(&_ctx);
+
+    return 0;
+}
+
+SOL_API int
+sol_netctl_unregister_agent(void)
+{
+    int r;
+    sd_bus *bus = sol_bus_client_get_bus(_ctx.connman);
```
ceolin: shouldn't check if `bus != NULL` too ?

data/scripts/libsoletta.sym
```diff
@@ -885,3 +885,12 @@ global:
...
 };
+
+LIBSOLETTA_2 {
+global:
+        sol_netctl_register_agent;
+        sol_netctl_unregister_agent;
+        sol_netctl_request_retry;
+        sol_netctl_request_input;
+        sol_netctl_scan;
+} LIBSOLETTA_1;
```
ceolin: these entries should be added with the commit that adds the function, otherwise we are broken between some commits.

ceolin:
Please review the comments, all the things are quite simple to fix.

src/lib/comms/include/sol-netctl.h
wzhen12: Okay

src/lib/comms/include/sol-netctl.h
wzhen12: It can be changed.

src/lib/comms/sol-netctl-impl-connman.c

src/lib/comms/sol-netctl-impl-connman.c
wzhen12: let me try

src/lib/comms/include/sol-netctl.h
wzhen12: Ok, Let me check it.

src/lib/comms/sol-netctl-impl-connman.c

data/scripts/libsoletta.sym
wzhen12: It will be modified.

Issue #1998, opened 2016-05-09
IIO node type: add support for multiple channels


edersondisouza:
#### IIO node type: add support for multiple channels

Currently, IIO node types support only one logic channel. Some devices, however, have more than one channel, whose readings are independent. They usually have channels `in_readingX`, like `in_voltage0` or `in_proximity2`.
To address that, IIO node types would need to have a new option, `channel`, default -1. If -1, it means that are no specifier, so channel name would be `in_proximity`. If > 0, then this should be used as channel specifier.
### Implementation notes

Some thoughts on implementation, of course different ideas are welcome.
Even though the channels are independent, they do belong to the same device. So, it's necessary that different nodes share devices. This can be done by creating a list of open devices on nodes.c.
After opening a device, it shall be sent to this list. I believe that this list should store device id, sol_iio_device handle and device/channel config. Last one is to prevent that some channel is opened with different device config - so it's there to compare.
So, when creating a node, it should first call `sol_iio_address_device`, and check if there's an sol_iio_device open with same id. If so, compare configs and if OK, add new channel. If not, fail on mismatch configuration.
This is also an opportunity to make iio/nodes.c code share more common code and avoid it's current state of duplicating code.

edersondisouza:
@yongli3 See this issue, please.

lblim:
This will allow multiple access, will the implementation only allow multiple access to the same IIO device from a single application or it will allow multiple access from multiple applications? Where is the access controller here.

yongli3:
@edersondisouza I am thinking we need to add another option: count. Default = 1, it means we only output one channel. > 1 means output multi-channel data. Using the ADC sensor on Galileo as an example. Below is the sysfs entities:
ADC sensor on Galileo boards:
root@intel-quark:/sys/bus/iio/devices/iio:device0# ls -l
drwxr-xr-x    2 root     root             0 May  2 06:44 buffer
-r--r--r--    1 root     root          4096 May  2 06:44 dev
-rw-r--r--    1 root     root          4096 May  2 06:44 in_voltage0_raw
-rw-r--r--    1 root     root          4096 May  2 06:44 in_voltage0_scale
-rw-r--r--    1 root     root          4096 May  2 06:44 in_voltage1_raw
-rw-r--r--    1 root     root          4096 May  2 06:44 in_voltage1_scale
-rw-r--r--    1 root     root          4096 May  2 06:44 in_voltage2_raw
[...]

PR #2314, opened 2016-09-22
WIP: iio: add support for multiple channels

yongli3:
Fix issue #1998

### DIFF
--- a/src/lib/io/sol-iio.c
+++ b/src/lib/io/sol-iio.c
@@ -116,6 +116,8 @@ struct resolve_absolute_path_data {
 #define CHANNEL_SCAN_TYPE_PATH SYSFS_DEVICES_PATH "/iio:device%d/scan_elements/%s_type"
 
 #define SAMPLING_FREQUENCY_DEVICE_PATH SYSFS_DEVICES_PATH "/iio:device%d/sampling_frequency"
+#define CHANNEL_SAMPLING_FREQUENCY_DEVICE_PATH SYSFS_DEVICES_PATH "/iio:device%d/%ssampling_frequency"
+
 #define SAMPLING_FREQUENCY_BUFFER_PATH SYSFS_DEVICES_PATH "/iio:device%d/buffer/sampling_frequency"
 #define SAMPLING_FREQUENCY_TRIGGER_PATH SYSFS_DEVICES_PATH "/trigger%d/sampling_frequency"
 
@@ -521,14 +523,22 @@ channel_get_pure_name(struct sol_iio_channel *channel)
 }
 
 static bool
-iio_set_sampling_frequency(struct sol_iio_device *device, int frequency)
+iio_set_sampling_frequency(struct sol_iio_device *device, const struct sol_iio_config *config)
 {
     char path[PATH_MAX];
+    int frequency = config->sampling_frequency;
 
     SOL_NULL_CHECK(device, false);
 
     if (craft_filename_path(path, sizeof(path), SAMPLING_FREQUENCY_DEVICE_PATH,
         device->device_id)) {
+
+        if (sol_util_write_file(path, "%d", frequency) > 0)
+            return true;
+    }
+
+    if (craft_filename_path(path, sizeof(path), CHANNEL_SAMPLING_FREQUENCY_DEVICE_PATH,
+        device->device_id, config->sampling_frequency_name)) {
 
         if (sol_util_write_file(path, "%d", frequency) > 0)
             return true;
@@ -772,7 +782,7 @@ sol_iio_open(int device_id, const struct sol_iio_config *config)
     }
 
     if (config->sampling_frequency > -1) {
-        if (!iio_set_sampling_frequency(device, config->sampling_frequency))
+        if (!iio_set_sampling_frequency(device, config))
             SOL_WRN("Could not set device%d sampling frequency", device->device_id);
     }
 
--- a/src/modules/flow/iio/nodes.c
+++ b/src/modules/flow/iio/nodes.c
@@ -275,6 +275,8 @@ gyroscope_open(struct sol_flow_node *node, void *data, const struct sol_flow_nod
 
     mdata->iio_base.config.buffer_size = opts->buffer_size;
     mdata->iio_base.config.sampling_frequency = opts->sampling_frequency;
+    snprintf(mdata->iio_base.config.sampling_frequency_name,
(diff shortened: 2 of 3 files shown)


yongli3:
@fulong82 @laykuanloon @edersondisouza please help to check on this PR. I am using an ADC sensor for test

laykuanloon:
@yongli3 

Previously you add this for multichannel https://github.com/solettaproject/soletta/blob/master/src/modules/flow/iio/nodes.c#L937

I have [http://lxr.free-electrons.com/source/drivers/iio/proximity/sx9500.c](sx9500) which have in_proximity0, in_proximity1, in_proximity2, in_proximity3. Is a proximity sensor but have 4 channel. 

Any solution can cater for N-channel entry? Can it be specify an the channel as a array (proximity0,proximity1,proximity2,proximity3) ? So that it's flexible for user to insert what channel they want that no need require to code change.

src/modules/flow/iio/nodes.c
```diff
@@ -776,12 +776,16 @@ humidity_open(struct sol_flow_node *node, void *data, const struct sol_flow_node
 }
 
 static bool
-adc_create_channels(struct iio_double_data *mdata, int device_id)
+adc_create_channels(struct iio_double_data *mdata, int device_id, int channel_id)
 {
+    char channel_name[NAME_MAX];
+
+    snprintf(channel_name, sizeof(channel_name), "in_voltage%d", channel_id);
```
edersondisouza: Maybe it's different for ADC devices, but couldn't we have a channel named only `in_voltage`, with no index? This is so for proximity sensors, for instance. So, maybe '-1' should mean 'no index' instead of 'first index'. If common case is to have index, default could be '0', then.

yongli3:
@laykuanloon Thanks for your comment. I would like to know which channels do you want at one time reading? Currently the proximity node only return on double data from the first available channel(0 or 2). This PR adds a new option parameter, you can set the channel number for fbp files. If you want to get all these channel data at the same time, we need to change the reader callback functions

src/modules/flow/iio/iio.json
```diff
@@ -822,6 +822,12 @@
              "name": "sampling_frequency"
          },
          {
+             "data_type": "int",
+             "default": -1,
+             "description": "Channel number. If -1, use the first channel",
+             "name": "option"
```
yongli3: Thanks @edersondisouza  for your comments. As the initial PR for the "multi-channel" support, I only tested the simple use-case(no buffer, no trigger, single node reading). Let me add the "two nodes on an app" in the next change

laykuanloon:
@yongli3. I am using channel 2. 

I would suggest:
In FBP:
I suggest to make the N-channel name can be specific in json file so does not require code change if the channel name is not handle in nodes.c. 
In C:
I think user can use sol_iio_add_channel(device, channel-1, &channel_config);, sol_iio_add_channel(device, channel-2, &channel_config);

src/modules/flow/iio/nodes.c
ceolin: please always check the return of `snprintf`

src/modules/flow/iio/iio.json
yongli3: @edersondisouza @laykuanloon I changed the sol_iio_open/close API, Using a list to trace these opened devices, Please help to review. The test sensor I am using does support two different channels, but it does not support buffer reading. I will find another one...

src/modules/flow/iio/nodes.c
yongli3: Thanks @ceolin for the snprintf, I found there are several different places in this nodes.c files, let me add the error checking in another PR

edersondisouza:
HI @yongli3 ,

I'm afraid that your patch is a bit messed up: it contains changes from an already integrated PR. Please, rebase it on top of master. Also, commit message is wrong, from this other PR.

src/lib/io/sol-iio.c
```diff
@@ -690,6 +711,19 @@ sol_iio_open(int device_id, const struct sol_iio_config *config)
...
 #endif
+    // find the open list 
+    SOL_LIST_FOREACH (&iio_open_list, it) {
+        entry = SOL_LIST_GET_CONTAINER(it, struct open_list, list);
+        SOL_INF("");
+        if (device_id == entry->device->device_id) {
+            if (0 == memcmp(config, entry->config, sizeof(struct sol_iio_config))) {// the same
+                return entry->device;
+            } else { // same device_id but different config
+            SOL_WRN("Could not open IIO device%d with different config", device_id);
```
edersondisouza: Please, make this message more informative. Say that device is already open, but couldn't reuse it because of different config.

src/lib/io/sol-iio.c
```diff
@@ -690,6 +711,19 @@ sol_iio_open(int device_id, const struct sol_iio_config *config)
         return NULL;
     }
 #endif
+    // find the open list 
+    SOL_LIST_FOREACH (&iio_open_list, it) {
+        entry = SOL_LIST_GET_CONTAINER(it, struct open_list, list);
+        SOL_INF("");
+        if (device_id == entry->device->device_id) {
+            if (0 == memcmp(config, entry->config, sizeof(struct sol_iio_config))) {// the same
```
edersondisouza: I don't think that we use this inverted comparison on Soletta.

src/lib/io/sol-iio.c
```diff
@@ -790,6 +824,10 @@ sol_iio_open(int device_id, const struct sol_iio_config *config)
         device->buffer_enabled, device->manual_triggering, device->trigger_name,
         device->trigger_id, device->mount_matrix);
 
+    entry = malloc(sizeof(struct open_list));
```
edersondisouza: Always check if malloc was ok.

src/lib/io/sol-iio.c
```diff
@@ -88,6 +89,14 @@ struct resolve_absolute_path_data {
     int id;
 };
 
+struct open_list {
+    struct sol_list list;
+    sol_iio_device *device;
+    sol_iio_config *config;
+};
```
edersondisouza: Besides keeping a list of open devices, you need to refcount uses. Otherwise, first one that closes a device will make it invalid for others.

yongli3:
Thanks @edersondisouza for the comments, it seems that I amended an incorrect commit id, let me submit another PR

PR #2317, opened 2016-09-26
mqtt: Check libmosquitto version

ceolin:
Soletta requires libmosquitto >= 1.4.2
Since libmosquitto does not have a pkg-config to check the library
version, it's checking for the version exported in mosquitto.h


cabelitos:
lgtm

ceolin:
thx iscaro !!!

Issue #2244, opened 2016-07-18
Proposal: Add support for IIO Ambient Light which use in_intensity_both & in_intensity_ir as its channel data


fulong82:
@barbieri, @edersondisouza & @yongli3,

#### Task Description
Current Soletta only support IIO Ambient Light with `in_illuminance` processed channel type or `in_intensity_<COLORS>` raw channel type. I would like to make a request here to add support for `in_intensity_both` & `in_intensity_ir` raw channel type where the below ambient light sensor are using it:
1) [RPR0521](http://lxr.free-electrons.com/source/drivers/iio/light/rpr0521.c#L165)
2) [LTR501](http://lxr.free-electrons.com/source/drivers/iio/light/ltr501.c#L585)
3) [APDS9300](http://lxr.free-electrons.com/source/drivers/iio/light/apds9300.c#L373)
4) [TSL2563](http://lxr.free-electrons.com/source/drivers/iio/light/tsl2563.c#L553)
5) [HID ALS](http://lxr.free-electrons.com/source/drivers/iio/light/hid-sensor-als.c#L48)

#### Dependencies
N/A

yongli3:
@fulong82  Below is the sysfs nodes for TSL2563. Currently Soletta will read raw data from in_illuminance0_input, do you want to support both the _input and _raw data reading?

 cat name
tsl2563
root@raspberrypi3:/sys/bus/i2c/devices/1-0039/iio:device1# ls -l
-r--r--r--    1 root     root          4096 Sep 18 08:19 dev
drwxr-xr-x    2 root     root             0 Sep 18 08:19 events
-rw-r--r--    1 root     root          4096 Sep 18 08:19 in_illuminance0_input
-rw-r--r--    1 root     root          4096 Sep 18 08:19 in_intensity_both_calibscale
-rw-r--r--    1 root     root          4096 Sep 18 08:19 in_intensity_both_raw
-rw-r--r--    1 root     root          4096 Sep 18 08:19 in_intensity_ir_calibscale
-rw-r--r--    1 root     root          4096 Sep 18 08:19 in_intensity_ir_raw
-r--r--r--    1 root     root          4096 Sep 18 08:19 name
drwxr-xr-x    2 root     root             0 Sep 18 08:19 power
lrwxrwxrwx    1 root     root             0 Sep 18 08:19 subsystem -> >./../../../n./../../bus/iio
-rw-r--r--    1 root     root        ! 4096 Sep 18 08:18 uevent

[...]

PR #2318, opened 2016-09-27
iio: add support for multiple channels

yongli3:
Fix issue #1998

### DIFF
--- a/src/lib/io/sol-iio.c
+++ b/src/lib/io/sol-iio.c
@@ -42,11 +42,15 @@
 #include <sol-i2c.h>
 #endif
 
+struct reader_cb_data {
+    void (*reader_cb)(void *data, struct sol_iio_device *device);
+    const void *data;
+};
+
 struct sol_iio_device {
     double *mount_matrix;
     char *trigger_name;
-    void (*reader_cb)(void *data, struct sol_iio_device *device);
-    const void *reader_cb_data;
+    struct sol_vector reader_cb_list;
     struct sol_fd *fd_handler;
     struct sol_buffer buffer;
     size_t buffer_size;
@@ -87,6 +91,14 @@ struct resolve_absolute_path_data {
     const char *path;
     int id;
 };
+
+struct iio_opened_device {
+    struct sol_iio_device *device;
+    struct sol_iio_config *config;
+    int refcount;
+};
+
+static struct sol_vector iio_opened_devices = SOL_VECTOR_INIT(struct iio_opened_device);
 
 #define DEVICE_PATH "/dev/iio:device%d"
 #define SYSFS_DEVICES_PATH "/sys/bus/iio/devices"
@@ -437,6 +449,22 @@ set_buffer_enabled(struct sol_iio_device *device, bool enabled)
 }
 
 static bool
+get_buffer_value(struct sol_iio_device *device, int *value)
+{
+    char path[PATH_MAX];
+    bool r;
+
+    r = craft_filename_path(path, sizeof(path), BUFFER_ENABLE_DEVICE_PATH, device->device_id);
+    if (!r)
+        return false;
+
+    if (sol_util_read_file(path, "%d", value) < 0)
+        return false;
+
+    return true;
+}
+
+static bool
 set_channel_enabled(struct sol_iio_device *device, const char *channel_name, bool enabled)
 {
     char path[PATH_MAX];
@@ -472,8 +500,10 @@ static bool
 device_reader_cb(void *data, int fd, uint32_t active_flags)
 {
     struct sol_iio_device *device = data;
+    struct reader_cb_data *reader_cb_data_elem = NULL;
     bool result = true;
     ssize_t ret;
+    int i;
 
     if (active_flags & (SOL_FD_FLAGS_ERR | SOL_FD_FLAGS_HUP | SOL_FD_FLAGS_NVAL)) {
         SOL_WRN("Unexpected reading");
@@ -484,8 +514,8 @@ device_reader_cb(void *data, int fd, uint32_t active_flags)
     if (ret <= 0) {
         result = false;
     } else if (device->buffer.used == device->buffer_size) {
-        if (device->reader_cb) {
-            device->reader_cb((void *)device->reader_cb_data, device);
+        SOL_VECTOR_FOREACH_IDX (&device->reader_cb_list, reader_cb_data_elem, i) {
+            reader_cb_data_elem->reader_cb((void *)reader_cb_data_elem->data, device);
         }
         sol_buffer_reset(&device->buffer);
     }
@@ -748,11 +778,14 @@ get_mount_matrix(struct sol_iio_device *device, double *mount_matrix)
 SOL_API struct sol_iio_device *
 sol_iio_open(int device_id, const struct sol_iio_config *config)
 {
+    int i;
     int prefix_len = strlen(HRTIMER_TRIGGER_PREFIX);
     bool r;
     bool buffer_existence;
     char path[PATH_MAX];
     struct sol_iio_device *device = NULL;
+    struct iio_opened_device *entry = NULL;
+    struct reader_cb_data *reader_cb_data_elem = NULL;
 
     SOL_NULL_CHECK(config, NULL);
 #ifndef SOL_NO_API_VERSION
@@ -762,12 +795,31 @@ sol_iio_open(int device_id, const struct sol_iio_config *config)
         return NULL;
     }
 #endif
+    SOL_VECTOR_FOREACH_IDX (&iio_opened_devices, entry, i) {
+        if (device_id == entry->device->device_id) {
+            if ((config->buffer_size == entry->config->buffer_size)
+                && (config->sampling_frequency == entry->config->sampling_frequency)) {
+                device = entry->device;
+                reader_cb_data_elem = sol_vector_append(&device->reader_cb_list);
+                SOL_NULL_CHECK(reader_cb_data_elem, NULL);
+                reader_cb_data_elem->data = config->data;
+                reader_cb_data_elem->reader_cb = config->sol_iio_reader_cb;
+                entry->refcount++;
+                return device;
+            } else {
+                SOL_ERR("device%d is already open, but could not reuse it with different config",
+                    device_id);
+                return NULL;
+            }
+        }
+    }
 
     device = calloc(1, sizeof(struct sol_iio_device));
     SOL_NULL_CHECK(device, NULL);
     sol_ptr_vector_init(&device->channels);
     device->buffer = SOL_BUFFER_INIT_EMPTY;
     device->fd = -1;
+    sol_vector_init(&device->reader_cb_list, sizeof(struct reader_cb_data));
 
     device->device_id = device_id;
 
@@ -804,8 +856,11 @@ sol_iio_open(int device_id, const struct sol_iio_config *config)
                 " was defined.'", device->device_id);
             goto error;
         }
-        device->reader_cb = config->sol_iio_reader_cb;
-        device->reader_cb_data = config->data;
+
+        reader_cb_data_elem = sol_vector_append(&device->reader_cb_list);
+        SOL_NULL_CHECK_GOTO(reader_cb_data_elem, error);
+        reader_cb_data_elem->reader_cb = config->sol_iio_reader_cb;
+        reader_cb_data_elem->data = config->data;
 
         if (config->trigger_name && config->trigger_name[0] != '\0') {
             /* if name start with HRTIMER_TRIGGER_PREFIX, it is a hrtimer trigger */
@@ -868,8 +923,19 @@ sol_iio_open(int device_id, const struct sol_iio_config *config)
         device->buffer_enabled, device->manual_triggering, device->trigger_name,
         device->trigger_id, device->mount_matrix);
 
+    entry = sol_vector_append(&iio_opened_devices);
+    SOL_NULL_CHECK_GOTO(entry, error);
+    entry->device = device;
+
+    entry->config = calloc(1, sizeof(struct sol_iio_config));
+    SOL_NULL_CHECK_GOTO(entry->config, error_config);
+    memcpy(entry->config, config, sizeof(struct sol_iio_config));
+
+    entry->refcount = 1;
     return device;
 
+error_config:
+    sol_vector_del_last(&iio_opened_devices);
 error:
     sol_iio_close(device);
 
@@ -881,7 +947,7 @@ enable_channel_scan(struct sol_iio_channel *channel)
 {
     struct sol_iio_device *device = channel->device;
     char path[PATH_MAX];
-    int current_value, len;
+    int current_value, len, ret, buffer_enable;
 
     if (!craft_filename_path(path, sizeof(path), CHANNEL_SCAN_ENABLE_PATH,
         device->device_id, channel->name)) {
@@ -896,9 +962,32 @@ enable_channel_scan(struct sol_iio_channel *channel)
         return false;
     }
 
-    if (current_value != 1)
-        return sol_util_write_file(path, "%d", 1) > 0;
+    if (current_value != 1) {
+        if (!get_buffer_value(device, &buffer_enable)) {
+            SOL_WRN("Coult not get buffer status for device%d", device->device_id);
+            return false;
+        }
 
+        if (buffer_enable) {
+            if (!set_buffer_enabled(device, false)) {
+                SOL_WRN("Could not disable buffer for device%d", device->device_id);
+                return false;
+            }
+        }
+
+        ret = sol_util_write_file(path, "%d", 1);
+        if (ret < 0) {
+            SOL_WRN("Could not enable scan %s for device%d", path, device->device_id);
+        }
+
+        if (buffer_enable) {
+            if (!set_buffer_enabled(device, true)) {
+                SOL_WRN("Could not enable buffer for device%d", device->device_id);
+                return false;
+            }
+        }
+        return ret > 0;
+    }
     return true;
 }
 
@@ -1031,8 +1120,31 @@ sol_iio_close(struct sol_iio_device *device)
 {
     int i;
     struct sol_iio_channel *channel;
+    struct iio_opened_device *entry = NULL;
+    struct iio_opened_device *remove_entry = NULL;
 
     SOL_NULL_CHECK(device);
+
+    SOL_VECTOR_FOREACH_IDX (&iio_opened_devices, entry, i) {
+        if (device->device_id == entry->device->device_id) {
+            entry->refcount--;
+            if (entry->refcount == 0) {
+                SOL_DBG("Close device%d", device->device_id);
+                remove_entry = entry;
+                free(entry->config);
+            } else {
+                SOL_DBG("Won't close device%d refcount=%d > 0", device->device_id, entry->refcount);
+                return;
+            }
+            break;
+        }
+    }
+
+    if (remove_entry)
+        sol_vector_del_element(&iio_opened_devices, remove_entry);
+
+    if (iio_opened_devices.len == 0)
+        sol_vector_clear(&iio_opened_devices);
 
     SOL_PTR_VECTOR_FOREACH_REVERSE_IDX (&device->channels, channel, i) {
         iio_del_channel(device, channel);
@@ -1046,6 +1158,8 @@ sol_iio_close(struct sol_iio_device *device)
     if (device->fd > -1) close(device->fd);
     if (device->name_fd > -1) close(device->name_fd);
     free(device->mount_matrix);
+
+    sol_vector_clear(&device->reader_cb_list);
 
     sol_buffer_fini(&device->buffer);
     free(device->trigger_name);
--- a/src/modules/flow/iio/iio.json
+++ b/src/modules/flow/iio/iio.json
@@ -822,6 +822,12 @@
              "name": "sampling_frequency"
          },
          {
+             "data_type": "int",
+             "default": -1,
+             "description": "Channel id. If -1, use the default channel",
+             "name": "channel_id"
+         },
+         {
              "data_type": "drange-spec",
              "default": {
                "max": "DBL_MAX",
@@ -924,6 +930,12 @@
              "name": "sampling_frequency"
          },
          {
+             "data_type": "int",
+             "default": -1,
+             "description": "Channel id. If -1, use the default channel",
+             "name": "channel_id"
+         },
+         {
              "data_type": "drange-spec",
              "default": {
                "max": "DBL_MAX",
@@ -1021,6 +1033,12 @@
              "default": -1,
              "description": "Sampling frequency of the sensor. If -1, use device default",
              "name": "sampling_frequency"
+         },
+         {
+             "data_type": "int",
+             "default": -1,
+             "description": "Channel id. If -1, use the default channel",
+             "name": "channel_id"
          },
          {
              "data_type": "drange-spec",
--- a/src/modules/flow/iio/nodes.c
+++ b/src/modules/flow/iio/nodes.c
@@ -25,6 +25,17 @@
 
 #include <sol-iio.h>
 
+#define GEN_CHANNEL_NAME(_channel_name, _name, _error, _id) \
+    if (_id >= 0) { \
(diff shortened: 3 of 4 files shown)


yongli3:
@edersondisouza @ceolin @laykuanloon please help to review

src/modules/flow/iio/nodes.c
```diff
@@ -275,8 +275,10 @@ gyroscope_open(struct sol_flow_node *node, void *data, const struct sol_flow_nod
 
     mdata->iio_base.config.buffer_size = opts->buffer_size;
     mdata->iio_base.config.sampling_frequency = opts->sampling_frequency;
-    snprintf(mdata->iio_base.config.sampling_frequency_name,
+    ret = snprintf(mdata->iio_base.config.sampling_frequency_name,
```
laykuanloon: After check snprintf manual. ret is  Upon successful return, these functions return the number of characters printed (excluding the null byte used to end output to strings).

For e.g 
"in_anglvel_sampling_frequency" is 29 bytes.
sizeof(mdata->iio_base.config.sampling_frequency_name) is NAME_MAX. I think is 256 bytes
29 >= NAME_MAX which is wrong will go to err (Line 280)

yongli3: @laykuanloon  It seems to me that only the expr is ture, it will goto err?

laykuanloon: @yongli3 yes, you are right. my mistake. 

So if you write 29 bytes. but ret return less than 29 bytes will not go to err.

yongli3:
@laykuanloon @ceolin @edersondisouza please help to review. I add a new option channel_name in the fbp files. user can set the channel name

edersondisouza:
Commit about checking snprintf return merged.

src/lib/io/sol-iio.c
```diff
@@ -88,6 +89,15 @@ struct resolve_absolute_path_data {
     int id;
 };
 
+struct open_list {
```
edersondisouza: This struct is not actually the list, but an entry/item of the list. Maybe a more explicit name, to avoid confusion?

src/lib/io/sol-iio.c
```diff
@@ -963,9 +1035,26 @@ sol_iio_close(struct sol_iio_device *device)
...
+    SOL_LIST_FOREACH_SAFE (&iio_open_list, it, it_next) {
+        entry = SOL_LIST_GET_CONTAINER(it, struct open_list, list);
+        if (device->device_id == entry->device->device_id) {
+            entry->refcount--;
+            if (entry->refcount == 0) {
+                SOL_DBG("Close %d", device->device_id);
+                sol_list_remove(it);
+                free(entry);
+            } else {
+                SOL_DBG("cannot close %d", device->device_id);
```
edersondisouza: I don't think this message is useful. Either delete it or write something like 'device X still have N references'.

src/modules/flow/iio/iio.json
```diff
@@ -822,6 +822,12 @@
              "name": "sampling_frequency"
          },
          {
+             "data_type": "string",
+             "default": null,
+             "description": "Channel name. If none, try to auto detect channel",
+             "name": "channel_name"
```
edersondisouza: I don't like this. What if user uses a channel name that isn't compatible with channel output data? User could subvert node type use drastically, leading to a complete mess - using a gyro node to get acceleration info, for instance.
Soletta approach on node type is to provide node for each information that can be extracted from device. I guess that this is motivated by #2244. But I believe that we need to understand what information is being exposed there. Maybe it's a case for new node types.
The indices approach used before was really useful: same kind of information being exposed via different channels, because they were independent.

yongli3: @edersondisouza I encountered a problem about the "multi-channel"  with buffer reading. Below is the test fbp file. It should output both adc1 and adc2 values, but it just print adc1 two times, and there is no adc2 output, how to narrow down this issue?
#!/usr/bin/env sol-fbp-runner
adc1(iio/adc:iio_device="INT3495 0",buffer_size=64)
adc2(iio/adc:iio_device="INT3495 0",buffer_size=64,channel_name="in_voltage5")
timer(timer:interval=1000)

timer OUT -> TICK adc1 OUT -> IN padc1(console:prefix="adc1:")
timer OUT -> TICK adc2 OUT -> IN padc2(console:prefix="adc2:")

edersondisouza: Yeah, it will call device reader callback twice. That's one more thing that needs to be coordinated. 
A simple approach would use a list of reader callbacks per device, instead of one. But then, a single tick will trigger all nodes, is this a problem? (In your test case, it would trigger 4 reads each time.) So, each device added would add to a list of callbacks, and the whole list is called by sol-iio.c.
Other ideas?

yongli3: Thanks @edersondisouza. I got your point. Just change the reader_cb_data to a list, and store all these config->data. Please check the new PR. YES, one single tick will output 4 RAW data:
adc1:698.242188 (float range)
adc2:3696.289062 (float range)
adc1:805.664062 (float range)
adc2:3608.398438 (float range)

laykuanloon: @edersondisouza 
I suggest to @yongli3  this method is for easy for user to use the channel that is not supported yet. Can we add custom iio node (not tie to any sensor type) that can let user specify this option.

I got another scenario to discuss on how to handle in iio flow node.
sx9500 sensor have more channel(in_proximity0_raw, in_proximity1_raw, in_proximity2_raw, in_proximity3_raw), how can user select 0 and 2 only in json?

yongli3: @laykuanloon @edersondisouza I updated the PR to use the channel_id(as int). For the "string" method @laykuanloon mentioned, let use use another PR to implement it

src/modules/flow/iio/nodes.c
```diff
@@ -776,12 +782,24 @@ humidity_open(struct sol_flow_node *node, void *data, const struct sol_flow_node
...
 {
+    char channel_name[NAME_MAX];
+    int ret;
+
     mdata->iio_base.device = sol_iio_open(device_id, &mdata->iio_base.config);
     SOL_NULL_CHECK(mdata->iio_base.device, false);
 
-    mdata->channel_val = iio_add_channel(mdata->scale, mdata->offset, "in_voltage0", &mdata->iio_base);
+    if (channelname) {
+        ret = snprintf(channel_name, sizeof(channel_name), "%s", channelname);
```
fulong82: I believe the `channelname` here is directly obtained from json/fbp setting. But I do not see any checking to it. Would it break the sensor node consistency?
For example: a iio combo sensor [imu MPU6050](http://lxr.free-electrons.com/source/drivers/iio/imu/inv_mpu6050/inv_mpu_core.c#L694), this implementation seems to allow user using a gyro sensor node but when filling in `in_accel_x`, user can obtain accelerometer data from gyro sensor node. So the consistency to a sensor node type is no longer there.

Actually iio sysfs interface format are:

```
<in | out>_<channel type name>[index number]_[modifier name]_<data type name>_[extended name]
|            c*              |      o      |       o       |       c        |       o       |

c = compulsory
o = optional
* most of the time compulsory, unless it is used to be shared by all
```
- channel type name could be referred to http://lxr.free-electrons.com/source/drivers/iio/industrialio-core.c#L54
- modifier name could be referred to http://lxr.free-electrons.com/source/drivers/iio/industrialio-core.c#L85
- data type name could be referred to http://lxr.free-electrons.com/source/drivers/iio/industrialio-core.c#L124
- extended name can be anything

My suggestion is to allow user input index number instead of the whole channel name so that sensor node consistency can still be under controlled across whole Soletta.

What do you guys think?

yongli3: @fulong82 "allow user input index number instead of the whole channel name" is the current solution in this PR. please update the webpage to get the latest PR #2318

fulong82: Oops... Okay :)

src/lib/io/sol-iio.c
```diff
@@ -36,17 +36,24 @@
 #include <sol-util-file.h>
 #include <sol-vector.h>
 
+#include "sol-list.h"
```
edersondisouza: May I ask you why do you favor `sol-list` over `sol-vector` or `sol-ptr-vector`?

src/lib/io/sol-iio.c
```diff
@@ -800,6 +866,12 @@ sol_iio_open(int device_id, const struct sol_iio_config *config)
         device->buffer_enabled, device->manual_triggering, device->trigger_name,
         device->trigger_id, device->mount_matrix);
 
+    entry = calloc(1, sizeof(struct iio_opened_device));
+    SOL_NULL_CHECK_GOTO(entry, error);
+    entry->device = device;
+    entry->config = config;
```
edersondisouza: `sol_iio_open` shouldn't take ownership of config struct. I guess that use on `iio/nodes.c` is ok (as it saves config struct), but this 'config to open' structs are supposed to be short lived. If we need to keep them here, we need to copy it.

src/lib/io/sol-iio.c
```diff
@@ -700,12 +741,33 @@ sol_iio_open(int device_id, const struct sol_iio_config *config)
...
+        }
+    }
 
     device = calloc(1, sizeof(struct sol_iio_device));
     SOL_NULL_CHECK(device, NULL);
     sol_ptr_vector_init(&device->channels);
     device->buffer = SOL_BUFFER_INIT_EMPTY;
-    device->fd = -1;
+    device->reader_cb_data_list.next = &device->reader_cb_data_list;
+    device->reader_cb_data_list.prev = &device->reader_cb_data_list;
```
edersondisouza: I didn't get this change. Why remove `device->fd` initialisation? Why are you making list point to list?

src/lib/io/sol-iio.c
```diff
@@ -700,12 +741,33 @@ sol_iio_open(int device_id, const struct sol_iio_config *config)
         return NULL;
     }
 #endif
+    SOL_LIST_FOREACH (&iio_open_list, it) {
+        entry = SOL_LIST_GET_CONTAINER(it, struct iio_opened_device, list);
+        if (device_id == entry->device->device_id) {
+            if ((config->buffer_size == entry->config->buffer_size)
+                && (config->sol_iio_reader_cb == entry->config->sol_iio_reader_cb)
```
edersondisouza: Different 'readers callbacks' should be allowed, don't you think? If a device has both gyro and temperature, why limiting to the very same callback?

src/lib/io/sol-iio.c
```diff
@@ -963,9 +1058,27 @@ sol_iio_close(struct sol_iio_device *device)
...
+    SOL_LIST_FOREACH_SAFE (&iio_open_list, it, it_next) {
+        entry = SOL_LIST_GET_CONTAINER(it, struct iio_opened_device, list);
+        if (device->device_id == entry->device->device_id) {
+            entry->refcount--;
+            if (entry->refcount == 0) {
+                SOL_DBG("Close device%d", device->device_id);
+                sol_list_remove(it);
+                free(entry);
+            } else {
+                SOL_DBG("Cannot close device%d refcount=%d > 0", device->device_id, entry->refcount);
```
edersondisouza: I still think that 'Cannot' is too negative, appearing to be a problem. Maybe 'Won't'?

src/modules/flow/iio/nodes.c
```diff
@@ -782,12 +782,24 @@ humidity_open(struct sol_flow_node *node, void *data, const struct sol_flow_node
...
+    int ret;
+
     mdata->iio_base.device = sol_iio_open(device_id, &mdata->iio_base.config);
     SOL_NULL_CHECK(mdata->iio_base.device, false);
 
-    mdata->channel_val = iio_add_channel(mdata->scale, mdata->offset, "in_voltage0", &mdata->iio_base);
+    if (channel_id > 0) {
+        ret = snprintf(channel_name, sizeof(channel_name), "in_voltage%d", channel_id);
+    } else {
+        ret = snprintf(channel_name, sizeof(channel_name), "%s", "in_voltage0");
```
edersondisouza: What if channel name is `in_voltage`, with no index? I'd say that this should be our '-1' case.

src/lib/io/sol-iio.c
```diff
@@ -36,17 +36,24 @@
...
+    const void *data;
+};
+
 struct sol_iio_device {
     double *mount_matrix;
     char *trigger_name;
     void (*reader_cb)(void *data, struct sol_iio_device *device);
-    const void *reader_cb_data;
+    struct sol_list reader_cb_data_list;
+    int list_len;
```
edersondisouza: It appears to me that this variable isn't used.

src/lib/io/sol-iio.c
yongli3: Done, using sol-ptr-vector now

src/lib/io/sol-iio.c
yongli3: Done

src/lib/io/sol-iio.c
yongli3: Done

src/lib/io/sol-iio.c
yongli3: Done, using ptr-vector now

src/lib/io/sol-iio.c
yongli3: Changed to memcpy now

src/lib/io/sol-iio.c
yongli3: Done

src/modules/flow/iio/nodes.c
yongli3: Changed the logic, if < 0(without change_id option in .fbp files), will use "in_voltage"

src/lib/io/sol-iio.c
```diff
@@ -42,11 +42,15 @@
...
+struct reader_cb_data {
+    const void *data;
+};
+
 struct sol_iio_device {
     double *mount_matrix;
     char *trigger_name;
     void (*reader_cb)(void *data, struct sol_iio_device *device);
-    const void *reader_cb_data;
+    struct sol_ptr_vector reader_cb_data_list;
```
edersondisouza: You also need to change `reader_cb` to a `reader_cb_list`. This list would be composed of a 
```
struct {
   void (*reader_cb)(void *data, struct sol_iio_device *device);
   const void *data;
}
```
So different callbacks (with their respective user data) can be defined. The way you've done work properly because `iio/nodes.c` has a single callback function, but on `sol-iio.c` we need to cover a more generic usage.

Minor: check if a `sol_vector` isn't enough, as you won't (probably) share list items elsewhere.

src/lib/io/sol-iio.c
```diff
@@ -700,12 +735,31 @@ sol_iio_open(int device_id, const struct sol_iio_config *config)
...
+    SOL_PTR_VECTOR_FOREACH_IDX (&iio_opened_devices, entry, i) {
+        if (device_id == entry->device->device_id) {
+            if ((config->buffer_size == entry->config->buffer_size)
+                && (config->sampling_frequency == entry->config->sampling_frequency)) {
+                entry->refcount++;
+                device = entry->device;
+                reader_cb_data_elem = calloc(1, sizeof(struct reader_cb_data));
+                SOL_NULL_CHECK(reader_cb_data_elem, NULL);
+                reader_cb_data_elem->data = config->data;
+                sol_ptr_vector_append(&device->reader_cb_data_list, reader_cb_data_elem);
```
edersondisouza: `sol_ptr_vector_append` may fail, check its result.

src/lib/io/sol-iio.c
```diff
@@ -800,9 +858,20 @@ sol_iio_open(int device_id, const struct sol_iio_config *config)
...
+    entry = calloc(1, sizeof(struct iio_opened_device));
+    SOL_NULL_CHECK_GOTO(entry, error);
+    entry->device = device;
+
+    entry->config = calloc(1, sizeof(struct sol_iio_config));
+    SOL_NULL_CHECK_GOTO(entry->config, error);
+    memcpy(entry->config, config, sizeof(struct sol_iio_config));
+
+    entry->refcount = 1;
+    sol_ptr_vector_append(&iio_opened_devices, entry);
```
edersondisouza: Ditto.

src/lib/io/sol-iio.c
```diff
@@ -963,9 +1055,33 @@ sol_iio_close(struct sol_iio_device *device)
...
+                SOL_DBG("Won't close device%d refcount=%d > 0", device->device_id, entry->refcount);
+                return;
+            }
+        }
+    }
+
+    if (remove_entry)
+        sol_ptr_vector_remove(&iio_opened_devices, remove_entry);
+
+    if (0 == iio_opened_devices.base.len)
```
edersondisouza: Soletta usually don't use this reverse comparison.

src/lib/io/sol-iio.c
```diff
@@ -88,6 +92,14 @@ struct resolve_absolute_path_data {
     int id;
 };
 
+struct iio_opened_device {
+    struct sol_iio_device *device;
+    struct sol_iio_config *config;
+    int refcount;
+};
+
+static struct sol_ptr_vector iio_opened_devices = SOL_PTR_VECTOR_INIT;
```
edersondisouza: Maybe a `sol_vector` is easier to use. A good way to know if you need a `sol_ptr_vector` or a `sol_vector` is to check if you need to share pointers to list items. If so, use `sol_ptr_vector`. If not, all you need is an internal list that will always be iterated when needed, a `sol_vector` should suffice.

For instance, the `channels` of a device are stored on a `sol_ptr_vector`, as we give pointers to `channel` to users. As `sol_vector` internal pointers may change, they are not fit for this task.

On the other hand, the instructions to 'address a device' are stored on a `sol_vector`, as pointer to instructions are not shared - we always iterate the `sol_vector` to use them.

Let me know if you had any doubts.

src/modules/flow/iio/iio.json
yongli3: @laykuanloon Please check the new added channel_id option, we can use it to read in_proximity2_raw/in_proximity3_raw

src/lib/io/sol-iio.c
```diff
@@ -700,12 +733,31 @@ sol_iio_open(int device_id, const struct sol_iio_config *config)
...
     }
 #endif
+    SOL_VECTOR_FOREACH_IDX (&iio_opened_devices, entry, i) {
+        if (device_id == entry->device->device_id) {
+            if ((config->buffer_size == entry->config->buffer_size)
+                && (config->sampling_frequency == entry->config->sampling_frequency)) {
+                entry->refcount++;
+                device = entry->device;
+                reader_cb_data_elem = sol_vector_append(&device->reader_cb_list);
+                SOL_NULL_CHECK(reader_cb_data_elem, NULL);
```
edersondisouza: If this check fails, you'd need to decrease `refcount`. Increase `refcount` just before `return device` to keep things simpler.

src/modules/flow/iio/nodes.c
```diff
@@ -861,17 +873,26 @@ adc_open(struct sol_flow_node *node, void *data, const struct sol_flow_node_opti
...
     SOL_NULL_CHECK(mdata->iio_base.device, false);
 
-    mdata->channel_val = iio_add_channel(mdata->scale, mdata->offset, "in_illuminance0", &mdata->iio_base);
-    if (!mdata->channel_val) {
-        mdata->channel_val = iio_add_channel(mdata->scale, mdata->offset, "in_illuminance", &mdata->iio_base);
-        SOL_NULL_CHECK_GOTO(mdata->channel_val, error);
+    if (channel_id >= 0) {
+        ret = snprintf(channel_name, sizeof(channel_name), "in_illuminance%d", channel_id);
+    } else {
+        ret = snprintf(channel_name, sizeof(channel_name), "%s", "in_illuminance");
```
edersondisouza: Very minor: as this is an `snprintf` of `%s` only, a `strncpy(channel_name, "in_illumincance")` should suffice.

src/lib/io/sol-iio.c
yongli3: Done, thanks @edersondisouza

src/modules/flow/iio/nodes.c

yongli3:
@edersondisouza I changed the commit message and add a note about sol_iio_open in sol-iio.h. 
I will submit a new PR about the "new nodes for intensity_both/intensity_ir after this PR checked in
Please help to review

src/lib/io/sol-iio.c
```diff
@@ -963,9 +1052,31 @@ sol_iio_close(struct sol_iio_device *device)
...
+        if (device->device_id == entry->device->device_id) {
+            entry->refcount--;
+            if (entry->refcount == 0) {
+                SOL_DBG("Close device%d", device->device_id);
+                remove_entry = entry;
+                free(entry->config);
+            } else {
+                SOL_DBG("Won't close device%d refcount=%d > 0", device->device_id, entry->refcount);
+                return;
+            }
```
edersondisouza: Minor: you can `break` at the end of the `if (device->...` to save some iterations.

src/modules/flow/iio/nodes.c
```diff
@@ -861,17 +873,26 @@ adc_open(struct sol_flow_node *node, void *data, const struct sol_flow_node_opti
...
-    if (!mdata->channel_val) {
-        mdata->channel_val = iio_add_channel(mdata->scale, mdata->offset, "in_illuminance", &mdata->iio_base);
-        SOL_NULL_CHECK_GOTO(mdata->channel_val, error);
+    if (channel_id >= 0) {
+        ret = snprintf(channel_name, sizeof(channel_name), "in_illuminance%d", channel_id);
+        SOL_INT_CHECK_GOTO(ret, >= (int)sizeof(channel_name), error);
+        SOL_INT_CHECK_GOTO(ret, < 0, error);
+    } else {
+        strncpy(channel_name, "in_illuminance", sizeof(channel_name) - 1);
+        channel_name[sizeof(channel_name) - 1] = '\0';
```
edersondisouza: Minor: This block `if (channel_id >= 0) {...} else {...}` could be made a function or a macro, don't you think?

src/lib/io/sol-iio.c
yongli3: Done, thanks!

src/modules/flow/iio/nodes.c
```diff
@@ -25,6 +25,16 @@
 
 #include <sol-iio.h>
 
+#define GEN_CHANNEL_NAME(_name) \
```
edersondisouza: I'd also ask `error` label and `channel_name` as parameters. What do you think?

yongli3: Agree with you, done!

src/modules/flow/iio/nodes.c
```diff
@@ -25,6 +25,16 @@
 
 #include <sol-iio.h>
 
+#define GEN_CHANNEL_NAME(_channel_name, _name, _error) \
+    if (channel_id >= 0) { \
```
edersondisouza: Oh, my bad. I forgot to mention that `channel_id` is also a good candidate to be a parameter. Idea is not depend on local variables from functions where this macro is called.

src/modules/flow/iio/nodes.c
```diff
@@ -25,6 +25,16 @@
...
+        ret = snprintf(_channel_name, sizeof(_channel_name), _name "%d", channel_id); \
```
edersondisouza: Same for this `ret` variable. Better having it local here.

src/modules/flow/iio/nodes.c
yongli3: Got it! Please check again

src/modules/flow/iio/nodes.c
yongli3: Done, thanks!

src/modules/flow/iio/nodes.c
```diff
@@ -25,6 +25,17 @@
 
 #include <sol-iio.h>
 
+#define GEN_CHANNEL_NAME(_channel_name, _name, _error, _id) \
+    if (_id >= 0) { \
+        int ret; \
```
yongli3: Good suggestion, thanks @edersondisouza !

PR #2321, opened 2016-09-29
memort fixes

ceolin:
Fix user-after-free and invalid memory access

### DIFF
--- a/src/lib/comms/sol-mqtt-impl-mosquitto.c
+++ b/src/lib/comms/sol-mqtt-impl-mosquitto.c
@@ -274,7 +274,8 @@ end:
 
     sol_mqtt_message_free(message);
 
-    mqtt->message_timeout = NULL;
+    if (mqtt)
+        mqtt->message_timeout = NULL;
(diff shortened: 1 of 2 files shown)


src/lib/comms/sol-mqtt-impl-mosquitto.c
```diff
@@ -274,7 +274,8 @@ sol_mqtt_on_message_wrapper(void *data)
 
     sol_mqtt_message_free(message);
 
-    mqtt->message_timeout = NULL;
+    if (mqtt)
+        mqtt->message_timeout = NULL;
```
cabelitos: Why not just set to NULL?

ceolin: because `mqtt` can be NULL

cabelitos: ok. +1

ceolin:
merged, thank you for reviewing it.

PR #2324, opened 2016-10-03
test:lwm2m: Fix unused variable

ceolin:
The variable access_ocontrol_object is only used when we build with
DTLS, so it should be under a ifdef instruction to avoid

./src/test/test-lwm2m.c:1130:38: warning: ‘access_control_object’
defined but not used [-Wunused-const-variable=]


cabelitos:
+1

ceolin:
merged, thanks for reviewing it.

Issue #2256, opened 2016-07-19
Proposal: IIO: Add read back to ensure the value written is being taken effect


fulong82:
@barbieri, @edersondisouza, @yongli3, @laykuanloon, @lblim

#### Task Description
The discussion is initiated from #2233. Open up this new ticket for task record as well as continue the discussion.

Here just a summary:
A secure coding practice would use the read back value instead of use the write value. A suggestion here:
Instead of
```
    if (config->scale > -1)
        iio_set_channel_scale(channel, config->scale);
    else
        channel_get_scale(channel);

    if (config->use_custom_offset)
        iio_set_channel_offset(channel, config->offset);
    else
        channel_get_offset(channel);
```
Do
```
    if (config->scale > -1)
        iio_set_channel_scale(channel, config->scale);

    channel_get_scale(channel);

    if (config->use_custom_offset)
        iio_set_channel_offset(channel, config->offset);

    channel_get_offset(channel);
```

Besides the scale and offset, suggest to apply this to sampling frequency as well to provide warning if the written value is not taken in effect.

Any different thought?

#### Dependencies
N/A

bdilly:
It seems this way you could make unnecessary reads.

I agree with you the way it is right now doesn't look secure, since it may fail to set channel scale (or offset).
But it could be done just checking for the return value of iio_set_channel_scale, don't you think?

Something like:
```C
bool r = false;
if (config->scale > -1)
    r = iio_set_channel_scale(channel, config->scale);

if (!r)
    channel_get_scale(channel);
```
But even get_scale() could fail and it would fallback to default value (1.0).

fulong82:
Hi @bdilly,

What i mean even a successful write, it is more secure to have a read back to ensure. Let take a scenario as an example: Some one is reading the raw data of the sensor and Soletta is doing a node initialization to this sensor. So the sensor HW may due to it already start sampling data and may ignore whatever write to some specific registers. In this case, use the read back value could prevent us using a wrong value if the HW does not take in effect.

Agree that most of the time this read back is un-necessary when thing just go perfectly. But it is a secure practice to take care corner case. Just to share some learning from secure coding training and open this discussion to seek better idea for it.

Thanks & Regards,
Wilson

edersondisouza:
Hi @fulong82,

> So the sensor HW may due to it already start sampling data and may ignore whatever write to some specific registers.

[...]

PR #2325, opened 2016-10-07
io: Fix add channel

ceolin:
Ensure the values written for scale and offset are being taken.


ceolin:
@peijiajames @fulong82  could you check if this fixes 
https://github.com/solettaproject/soletta/issues/2256

peijiajames:
Yes, I think that fixes #2256.  One minor question is if we need a SOL_WARN message when set failure happens, to indicate default value "xxx" is used clearly.

ceolin:
@peijiajames i think it's worth the warn message. I merged with this suggestion. Thanks for reviewing it.

PR #2319, opened 2016-09-27
comms: Add sol-netctl agent API V6

wzhen12:
The agent is necessary in the process of network connection,
that will receive input information from the user and send it
to the network(such as AP and so on).

The API is based on connman D-Bus APIs, it provides register agent,
input login information and so on netctl agent API.

Signed-off-by: Wu Zheng wu.zheng@intel.com

PR #2326, opened 2016-10-08
comms: Add sol-netctl agent API V7

wzhen12:


wzhen12:
The Version 7 of Add sol-netctl agent API.
1. Modify the version 6 errors. (#2319)

@ceolin and @barbieri Please check it.

wzhen12:
@ceolin
I will submit the new patch to fix the api_version issue after the serial is merged.

ceolin:
Merged, thank you !!!

PR #2328, opened 2016-10-11
iio:cpp: Fix build when compiling with g++

ceolin:
NAME_MAX is defined in linux/limits.h, as sol-iio is linux specific,
there is no problem in including a linux header.

### DIFF
--- a/src/lib/io/include/sol-iio.h
+++ b/src/lib/io/include/sol-iio.h
@@ -21,6 +21,8 @@
 #include <sol-buffer.h>
 #include <sol-common-buildopts.h>
 
+#include <linux/limits.h>
+
 #ifdef __cplusplus
 extern "C" {
 #endif


cabelitos:
@edersondisouza I don't think so, since IIO is only supossed to run on linux. +1

ceolin:
@edersondisouza `sol-missing.h` is not a installed. Also, it'll need to include this header right ? What will happen in the other platforms ?

cabelitos:
@ceolin what you mean? IIO is linux only

ceolin:
@cabelitos i was talking about put this header (`linux/limits.h`) in `sol-missing.h`. This would cause problems in other platforms.

cabelitos:
Indeed.  Anyway +1.

edersondisouza:
+1

ceolin:
merged, thank you guys

PR #2333, opened 2016-10-18
check-api: Change regex to not match LIBSOLETTA_X on extends

edersondisouza:
Current regex was matching everything ended in ';'. Changed to do so
only if not preceeded by '}'.


barbieri:
+1

edersondisouza:
Integrated, tks!

PR #2340, opened 2016-10-21
iio: Add new nodes for light intensity sensors

yongli3:
To support intensity_both/intensity_ir data reading in some light sensors

### DIFF
--- a/src/lib/io/sol-iio.c
+++ b/src/lib/io/sol-iio.c
@@ -335,7 +335,7 @@ create_hrtimer_trigger(struct sol_iio_device *device, const char *trigger_name)
 {
     int success, id, len;
     bool r = false;
-    char *name = NULL, *replace;
+    char *name = NULL;
     char path[PATH_MAX];
 
     if (check_file_existence(CONFIGFS_IIO_HRTIMER_TRIGGERS_PATH)) {
@@ -565,6 +565,7 @@ setup_device_reader(struct sol_iio_device *device)
 
 /* Some channels are named on the form  <type>[_x|_y|_z]
  * This function shall return the name without <x|y|z> components.
+ * And remove both/ir/uv suffix frome some light internsity sensors.
  * The form <type>[Y][_modifier] is also common (Y is a number), and this function
  * remove numbers too, in an attempt to get the 'pure' name.
  * TODO there are other esoteric combinations - for them, if we care about,
@@ -572,19 +573,35 @@ setup_device_reader(struct sol_iio_device *device)
 static char *
 channel_get_pure_name(struct sol_iio_channel *channel)
 {
+    char channel_name[NAME_MAX];
     size_t channel_name_len;
     char *channel_pure_name;
     bool modified = false;
 
-    channel_name_len = strlen(channel->name);
+    strncpy(channel_name, channel->name, sizeof(channel_name) - 1);
+    channel_name[sizeof(channel_name) - 1] = '\0';
+
+    channel_name_len = strlen(channel_name);
+    if (strendswith(channel_name, "_both")) {
+        channel_name[channel_name_len - 5] = '\0';
+        modified = true;
+    }
+
+    if (strendswith(channel_name, "_ir") ||
+        strendswith(channel_name, "_uv")) {
+        channel_name[channel_name_len - 3] = '\0';
+        modified = true;
+    }
+
+    channel_name_len = strlen(channel_name);
     if (channel_name_len > 2) {
         char *channel_name_suffix;
 
-        channel_name_suffix = channel->name + channel_name_len - 2;
+        channel_name_suffix = channel_name + channel_name_len - 2;
         if (streq(channel_name_suffix,  "_x") ||
             streq(channel_name_suffix,  "_y") || streq(channel_name_suffix,  "_z")) {
 
-            channel_pure_name = strndup(channel->name, channel_name_len - 2);
+            channel_pure_name = strndup(channel_name, channel_name_len - 2);
             return channel_pure_name;
         } else {
             /* Recreate channel name without Y_ components (Y is a number).
@@ -595,9 +612,10 @@ channel_get_pure_name(struct sol_iio_channel *channel)
             channel_pure_name = calloc(1, channel_name_len + 1);
             original_channel_pure_name = channel_pure_name;
             for (i = 0; i < channel_name_len; i++) {
-                if (isalpha((uint8_t)channel->name[i]) || channel->name[i] == '-' || channel->name[i] == '_')
-                    *channel_pure_name++ = channel->name[i];
-                else if (isdigit((uint8_t)channel->name[i])) {
+                if (isalpha((uint8_t)channel_name[i]) || channel_name[i] == '-'
+                    || channel_name[i] == '_')
+                    *channel_pure_name++ = channel_name[i];
+                else if (isdigit((uint8_t)channel_name[i])) {
                     modified = true;
                     continue;
                 }
@@ -608,6 +626,11 @@ channel_get_pure_name(struct sol_iio_channel *channel)
             free(original_channel_pure_name);
             return NULL;
         }
+    }
+
+    if (modified) {
+        channel_pure_name = strndup(channel_name, channel_name_len);
+        return channel_pure_name;
     }
 
     return NULL;
@@ -664,7 +687,7 @@ iio_set_channel_scale(struct sol_iio_channel *channel, double scale)
     if (craft_filename_path(path, sizeof(path), CHANNEL_SCALE_PATH,
         channel->device->device_id, channel->name)) {
 
-        if (sol_util_write_file(path, "%lf", scale) > 0) {
+        if (sol_util_write_file(path, "%.9lf", scale) > 0) {
             channel->scale = scale;
             return true;
         }
@@ -673,15 +696,15 @@ iio_set_channel_scale(struct sol_iio_channel *channel, double scale)
     /* If failed, try channel pure name */
     pure_name = channel_get_pure_name(channel);
     if (pure_name && craft_filename_path(path, sizeof(path), CHANNEL_SCALE_PATH,
-        channel->device->device_id, channel->name)) {
+        channel->device->device_id, pure_name)) {
 
-        result = (sol_util_write_file(path, "%lf", scale) > 0);
+        result = (sol_util_write_file(path, "%.9lf", scale) > 0);
         if (result)
             channel->scale = scale;
     }
 
     if (!result)
-        SOL_WRN("Could not set scale to %lf on channel [%s] of device%d",
+        SOL_WRN("Could not set scale to %.9lf on channel [%s] of device%d",
             scale, channel->name, channel->device->device_id);
 
     free(pure_name);
@@ -709,7 +732,7 @@ iio_set_channel_offset(struct sol_iio_channel *channel, int offset)
     /* If failed, try channel pure name */
     pure_name = channel_get_pure_name(channel);
     if (pure_name && craft_filename_path(path, sizeof(path), CHANNEL_OFFSET_PATH,
-        channel->device->device_id, channel->name)) {
+        channel->device->device_id, pure_name)) {
 
         result = sol_util_write_file(path, "%d", offset) > 0;
         if (result)
@@ -1262,7 +1285,7 @@ sol_iio_add_channel(struct sol_iio_device *device, const char *name, const struc
         SOL_WRN("Could not activate device channel [%s] in device%d",
             name, device->device_id);
 
-    SOL_DBG("channel [%s] added. scale: %lf - offset: %d - storagebits: %d"
+    SOL_DBG("channel [%s] added. scale: %.9lf - offset: %d - storagebits: %d"
         " - bits: %d - mask: %" PRIu64, channel->name, channel->scale,
         channel->offset, channel->storagebits, channel->bits, channel->mask);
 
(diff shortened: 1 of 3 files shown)


yongli3:
@edersondisouza @fulong82 @laykuanloon please help to review, this PR is used for issue #2244

src/modules/flow/iio/nodes.c
```diff
@@ -943,6 +943,154 @@ light_open(struct sol_flow_node *node, void *data, const struct sol_flow_node_op
 }
 
 static bool
+intensity_both_create_channels(struct iio_double_data *mdata, int device_id)
+{
+    mdata->iio_base.device = sol_iio_open(device_id, &mdata->iio_base.config);
+    SOL_NULL_CHECK(mdata->iio_base.device, false);
+
+    mdata->channel_val = iio_add_channel(mdata->scale, mdata->offset, "in_intensity_both", &mdata->iio_base);
```
yongli3: @fulong82 Thanks for your test. I am using the TSL2563 as below. Please let me know which the sensor name you are testing, regarding other sensors RPR0521/LTR501/APDS9300/HID ALS, do you know if they are using the in_intensity_offset or in_intensity_both_offset ?
 cat name
tsl2563
root@intel-quark:/sys/bus/iio/devices/iio:device1# ls -l
-r--r--r--    1 root     root          4096 Sep 27 15:27 dev
drwxr-xr-x    2 root     root             0 Sep 27 15:27 events
-rw-r--r--    1 root     root          4096 Sep 27 15:27 in_illuminance0_input
-rw-r--r--    1 root     root          4096 Sep 27 15:27 in_intensity_both_calibscale
-rw-r--r--    1 root     root          4096 Sep 27 15:27 in_intensity_both_raw
-rw-r--r--    1 root     root          4096 Sep 27 15:27 in_intensity_ir_calibscale
-rw-r--r--    1 root     root          4096 Sep 27 15:27 in_intensity_ir_raw
-r--r--r--    1 root     root          4096 Sep 27 15:27 name
drwxr-xr-x    2 root     root             0 Sep 27 15:27 power
lrwxrwxrwx    1 root     root             0 Sep 27 15:27 subsystem -> ../../../../../../../bus/iio
-rw-r--r--    1 root     root          4096 Sep 27 15:27 uevent

yongli3: @fulong82 I updated the PR, please help to test if the scale/offset is correct

fulong82: @yongli3
Tested. It is working now. Thx.

src/lib/io/sol-iio.c
```diff
@@ -544,12 +544,27 @@ channel_get_pure_name(struct sol_iio_channel *channel)
...
     size_t channel_name_len;
     char *channel_pure_name;
+    char *channel_name_suffix;
     bool modified = false;
 
     channel_name_len = strlen(channel->name);
-    if (channel_name_len > 2) {
-        char *channel_name_suffix;
+    if (channel_name_len > 5) {
+        channel_name_suffix = channel->name + channel_name_len - 5;
```
edersondisouza: According to kernel doc, `Documentation/ABI/testing/sysfs-bus-iio`, it's possible to have a channel like `in_intensityY_both_raw`, where Y is an index number. So I'd say that you have to take this into consideration. An idea would be doing what you do now, stripping suffix 'ir' and 'both' (and maybe 'uv') then letting code below process the Y part.
Note that Soletta also has a `strendswith` macro, that may help you to check a string suffix.

src/lib/io/sol-iio.c
```diff
@@ -544,12 +544,27 @@ channel_get_pure_name(struct sol_iio_channel *channel)
 {
```
edersondisouza: Please, also update this function commentary, to include this 'both' and 'ir' thing.
That said, shouldn't you also include 'uv' suffix? It appears on kernel doc...

src/modules/flow/iio/iio.json
```diff
@@ -945,6 +945,210 @@
...
+             "description": "Range of output packet. Usually, is the output range of used sensor. It'll be the 'min' and 'max' fields of drange packets sent on OUT port.",
+             "name": "out_range"
+         }
+       ],
+       "version": 1
+      },
+      "out_ports": [
+        {
+         "data_type": "float",
+         "description": "Light data read in lux.",
```
edersondisouza: I had understood that this data is unit-less, isn't it?

src/modules/flow/iio/iio.json
```diff
@@ -945,6 +945,210 @@
...
```
edersondisouza: Ditto.

src/modules/flow/iio/iio.json
yongli3: @edersondisouza some light sensors provide the data in lux, some are raw data. For raw data we will use the scale factor to process them. For "unit-less", do you mean we need to change it to  
"Light data read in lux(unit-lesss)" ?

src/lib/io/sol-iio.c
yongli3: Please search the "in_intensity_both" in nodes.c, in order to support the Y(a number) part, we need to add the channel_id(multi channel) logic ?

src/modules/flow/iio/iio.json
edersondisouza: That's good.

src/lib/io/sol-iio.c
edersondisouza: No, it's not multi-channel logic. Look in this function, but some lines below. The job of stripping the Y component is already done. Just change the flow so it goes there after you stripped 'both' or 'ir' part.

src/modules/flow/iio/iio.json

yongli3:
@edersondisouza @fulong82 I updated the PR, please help to review/test it

src/modules/flow/iio/nodes.c
```diff
@@ -959,6 +959,160 @@ light_open(struct sol_flow_node *node, void *data, const struct sol_flow_node_op
 }
 
 static bool
+intensity_both_create_channels(struct iio_double_data *mdata, int device_id)
+{
+    mdata->iio_base.device = sol_iio_open(device_id, &mdata->iio_base.config);
+    SOL_NULL_CHECK(mdata->iio_base.device, false);
+
+    mdata->channel_val = iio_add_channel(mdata->scale, mdata->offset, "in_intensity0_both", &mdata->iio_base);
```
edersondisouza: I think that as multiple channels support is already in, you can rebase this patch on top of master and make use of that, instead of this `in_intensity0_both` and `in_intensity_both` trick.

src/modules/flow/iio/nodes.c
```diff
@@ -959,6 +959,160 @@ light_open(struct sol_flow_node *node, void *data, const struct sol_flow_node_op
...
+
+}
+
+static bool
+intensity_ir_create_channels(struct iio_double_data *mdata, int device_id)
+{
+    mdata->iio_base.device = sol_iio_open(device_id, &mdata->iio_base.config);
+    SOL_NULL_CHECK(mdata->iio_base.device, false);
+
+    mdata->channel_val = iio_add_channel(mdata->scale, mdata->offset, "in_intensity0_ir", &mdata->iio_base);
```
edersondisouza: Ditto.

src/lib/io/sol-iio.c
```diff
@@ -673,15 +696,15 @@ iio_set_channel_scale(struct sol_iio_channel *channel, double scale)
     /* If failed, try channel pure name */
     pure_name = channel_get_pure_name(channel);
     if (pure_name && craft_filename_path(path, sizeof(path), CHANNEL_SCALE_PATH,
-        channel->device->device_id, channel->name)) {
+        channel->device->device_id, pure_name)) {
 
-        result = (sol_util_write_file(path, "%lf", scale) > 0);
+        result = (sol_util_write_file(path, "%.9lf", scale) > 0);
```
edersondisouza: Why this change?

src/lib/io/sol-iio.c
```diff
@@ -595,22 +615,25 @@ channel_get_pure_name(struct sol_iio_channel *channel)
...
+            }
 
             free(original_channel_pure_name);
-            return NULL;
+            return internsity_channel_name;
         }
     }
 
-    return NULL;
+    return internsity_channel_name;
```
edersondisouza: I guess that this is over complicated...
A simple approach to use would be (highlevel pseudocode):
  - copy `channel->name` to a `channel_name[NAME_MAX]` buffer
  - `strendswith(channel_name, "_both") || strendswith(channel_name, "_ir")` ->strip from channel_name, doing `channel_name[strlen(channel_name) - 5 (or 3)] = '\0'`
  - let current code act, but on `channel_name` instead of `channel->name`
On last step, you could actually do the same for `_x`, `_y` and `_z` stripping, so code would be uniform, but it isn't a must for now.
Any thoughts?

src/lib/io/sol-iio.c
```diff
@@ -709,7 +732,7 @@ iio_set_channel_offset(struct sol_iio_channel *channel, int offset)
     /* If failed, try channel pure name */
     pure_name = channel_get_pure_name(channel);
     if (pure_name && craft_filename_path(path, sizeof(path), CHANNEL_OFFSET_PATH,
-        channel->device->device_id, channel->name)) {
+        channel->device->device_id, pure_name)) {
```
edersondisouza: Nice catch... guess this was wrong for a long time :cold_sweat:

src/lib/io/sol-iio.c
yongli3: I prefer to keep the current complicated implement, in order to resolve the issue #2244. I will refactor this function in the next PR

edersondisouza: But alternative is easier, actually.

src/lib/io/sol-iio.c
edersondisouza: How about `%g`?

edersondisouza: Ah, true. I was thinking about log - it would be easier to see.

src/modules/flow/iio/nodes.c
yongli3: @edersondisouza The current GEN_CHANNEL_NAME macro is used to append a "%d" at the end of the string. But the two in_intensity0_both and in_intensity0_ir have a different format(%d is in hte middle of the string), any thoughts/suggestions about the change?

edersondisouza: Don't use the macro, treat them differently - they are the exception, macro is for generic case.

src/lib/io/sol-iio.c
yongli3: @edersondisouza I changed this function, please help to review

src/lib/io/sol-iio.c
yongli3: Done

src/modules/flow/iio/nodes.c
```diff
@@ -959,6 +959,178 @@ light_open(struct sol_flow_node *node, void *data, const struct sol_flow_node_op
...
+
+    mdata->iio_base.device = sol_iio_open(device_id, &mdata->iio_base.config);
+    SOL_NULL_CHECK(mdata->iio_base.device, false);
+
+    if (channel_id >= 0) {
+        ret = snprintf(channel_name, sizeof(channel_name), "in_intensity%d_both", channel_id);
+        SOL_INT_CHECK_GOTO(ret, >= (int)sizeof(channel_name), error);
+        SOL_INT_CHECK_GOTO(ret, < 0, error);
+    } else {
+        strncpy(channel_name, "in_intensity_both", sizeof(channel_name) - 1);
```
edersondisouza: Hint: as you are copying a constant into buffer, no strong need for `strncpy` and '\0' worries. But it doesn't hurt, either.

src/lib/io/sol-iio.c
```diff
@@ -610,6 +628,11 @@ channel_get_pure_name(struct sol_iio_channel *channel)
         }
     }
 
+    if (modified) {
+        channel_pure_name = strndup(channel_name, channel_name_len);
```
edersondisouza: `strndup` may fail. I know that there's a non-verified call above. It should be fixed too. It maybe on another patch, though.

edersondisouza:
Integrated, tks!

yongli3:
Thanks @edersondisouza I have tested it on my boards using TSL2563, HMC5843 and L3G4200D. I will submit another PR for the strndup checking. @fulong82 please help to re-test it using your HID sensors

fulong82:
@yongli3,

Tested working. Thanks. 👍

fulong82:
@laykuanloon & @yongli3 

Btw, i saw this warning message when using the latest code base. Do we need to suppress it as not all iio device will have mount_matrix especially  ambient light sensor.

```
# ../runfbsh als.fbp sol-flow-hid.json
WRN: ./src/lib/io/sol-iio.c:770 get_mount_matrix() Could not find mount_matrix for device2
illuminance = 103.583000 (float range) Lux
illuminance = 104.473000 (float range) Lux
illuminance = 103.880000 (float range) Lux
```

yongli3:
@fulong82 Thanks for your quick response! Regarding this WRN message, I think we can change it to DBG, any comments/suggestions?

fulong82:
@yongli3 I am fine for SOL_DBG. :-)

@laykuanloon What do you think? Since you are the requester for it.

laykuanloon:
@fulong82 @yongli3 
I am ok.

PR #2342, opened 2016-10-25
Check availability of RB_SW_SUSPEND

ffontaine:
RB_SW_SUSPEND is not always available (for example on uclibc) so check its
availability before using it

Signed-off-by: Fabrice Fontaine <fabrice.fontaine@orange.com>

### DIFF
--- a/src/lib/common/sol-missing.h
+++ b/src/lib/common/sol-missing.h
@@ -22,6 +22,8 @@
 extern "C" {
 #endif
 
+#include "sol-common-buildopts.h"
+
 #ifndef HAVE_DECL_STRNDUPA
 #include <alloca.h>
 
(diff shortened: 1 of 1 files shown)


ffontaine:
I found this issue with the latest version of uclibc-ng (1.0.19). As you can see here: http://www.uclibc-ng.org/browser/uclibc-ng/include/sys/reboot.h, RB_SW_SUSPEND is not defined. But there is the same issue with the "old" uclibc: https://git.uclibc.org/uClibc/tree/include/sys/reboot.h

glima:
Ok, that entry seems to be coming from Linux headers. +1

ceolin:
@ffontaine @glima I think it's worth to define it in `sol-missing.h` when it is missing. This way we can have the same behaviour not matter which libc is using. What do you think guys ?

ffontaine:
You're right, however the value of RB_SW_SUSPEND seems to have changed over time. In the man page, it is defined as 0xd000fce1 however in the latest kernel version, it is defined as 0xd000fce**2**.
So, if RB_SW_SUSPEND is undefined, I would suggest to update sol-missing.h to include linux/reboot.h and define RB_SW_SUSPEND as LINUX_REBOOT_CMD_SW_SUSPEND. Do you agree with this proposal?

ceolin:
@ffontaine i agree with you. Please, just keep in mind that this header can be included by other platforms !linux, so before include `linux/reboot.h` check if the platform is linux.

```
#ifdef SOL_PLATFORM_LINUX
#include <linux/reboot.h>
 ...
#endif
```

ceolin:
merged, thanks a lot !

PR #2344, opened 2016-10-27
iio: Add strndup return checking

yongli3:
If there is an error, will output a warning

Signed-off-by: Yong Li yong.b.li@intel.com

### DIFF
--- a/src/lib/io/sol-iio.c
+++ b/src/lib/io/sol-iio.c
@@ -578,6 +578,7 @@ channel_get_pure_name(struct sol_iio_channel *channel)
     char *channel_pure_name;
     bool modified = false;
 
+    SOL_NULL_CHECK(channel->name, NULL);
     strncpy(channel_name, channel->name, sizeof(channel_name) - 1);
     channel_name[sizeof(channel_name) - 1] = '\0';
 
@@ -602,6 +603,7 @@ channel_get_pure_name(struct sol_iio_channel *channel)
             streq(channel_name_suffix,  "_y") || streq(channel_name_suffix,  "_z")) {
 
             channel_pure_name = strndup(channel_name, channel_name_len - 2);
+            SOL_NULL_CHECK(channel_pure_name, NULL);
             return channel_pure_name;
         } else {
             /* Recreate channel name without Y_ components (Y is a number).
@@ -610,6 +612,7 @@ channel_get_pure_name(struct sol_iio_channel *channel)
             char *original_channel_pure_name;
 
             channel_pure_name = calloc(1, channel_name_len + 1);
+            SOL_NULL_CHECK(channel_pure_name, NULL);
             original_channel_pure_name = channel_pure_name;
             for (i = 0; i < channel_name_len; i++) {
                 if (isalpha((uint8_t)channel_name[i]) || channel_name[i] == '-'
@@ -630,6 +633,7 @@ channel_get_pure_name(struct sol_iio_channel *channel)
 
     if (modified) {
         channel_pure_name = strndup(channel_name, channel_name_len);
+        SOL_NULL_CHECK(channel_pure_name, NULL);
         return channel_pure_name;
     }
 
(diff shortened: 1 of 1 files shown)


yongli3:
@ceolin @edersondisouza This is a follow-up to PR #2340, please help to review

src/lib/io/sol-iio.c
```diff
@@ -1107,10 +1111,12 @@ channel_get_offset(struct sol_iio_channel *channel)
 
 #define GET_OFFSET(_name) \
     do { \
-        r = craft_filename_path(path, sizeof(path), CHANNEL_OFFSET_PATH, \
-            device->device_id, _name); \
-        if (r && sol_util_read_file(path, "%d", &channel->offset) > 0) { \
-            goto end; \
+        if (_name) { \
```
ceolin: why are you doing this check here ? Shouldn't you handle properly when `channel->name` is `NULL` ? This check will possibly avoid a crash here but the next instruction is `channel_get_pure_name` that crashes when `channel->name` is `NULL`

src/lib/io/sol-iio.c
```diff
@@ -602,6 +602,7 @@ channel_get_pure_name(struct sol_iio_channel *channel)
             streq(channel_name_suffix,  "_y") || streq(channel_name_suffix,  "_z")) {
 
             channel_pure_name = strndup(channel_name, channel_name_len - 2);
+            SOL_NULL_CHECK(channel_pure_name, NULL);
```
ceolin: :+1:  for that. Please since you are touching this function, there is a `calloc` here that should be checked too.

src/lib/io/sol-iio.c
yongli3: @ceolin This checking is used for the second GET_OFFSET call:
    channel_pure_name = channel_get_pure_name(channel);
    GET_OFFSET(channel_pure_name);

I found the below log during testing, since the channel_get_pure_name returns NULL
DBG: ./src/lib/io/sol-iio.c:165 craft_filename_path() available=4096, used=46, path='/sys/bus/iio/devices/iio:device1/(null)_offset'

src/lib/io/sol-iio.c
yongli3: Done, thanks!

src/lib/io/sol-iio.c
```diff
@@ -1107,10 +1112,12 @@ channel_get_offset(struct sol_iio_channel *channel)
...
-        r = craft_filename_path(path, sizeof(path), CHANNEL_OFFSET_PATH, \
-            device->device_id, _name); \
-        if (r && sol_util_read_file(path, "%d", &channel->offset) > 0) { \
-            goto end; \
+        if (_name) { \
+            r = craft_filename_path(path, sizeof(path), CHANNEL_OFFSET_PATH, \
+                device->device_id, _name); \
+            if (r && sol_util_read_file(path, "%d", &channel->offset) > 0) { \
+                goto end; \
+            } \
```
ceolin: my point is, if `_name` is NULL this macro will do nothing, then the next instruction (line 1126) will be `channel_get_pure_name()` with the parameter `channel` (please remember that `channel->name` is NULL previous check). Now look the function `channel_get_pure_name()`:
 ```
static char *                                                                                                  
channel_get_pure_name(struct sol_iio_channel *channel)                                                         
{                                                                                                              
    char channel_name[NAME_MAX];                                                                               
    size_t channel_name_len;                                                                                   
    char *channel_pure_name;                                                                                   
    bool modified = false;                                                                                     
                                                                                                               
    strncpy(channel_name, channel->name, sizeof(channel_name) - 1);
    ....
```
As you can see, it will crash on `strncpy()`

So, it seems that you can not deal with `channel->name == NULL` in any case. Probably the check should be in public entry point.

yongli3: @ceolin Thanks for your comment. The channel_get_offset was called by sol_iio_add_channel.  The channle->name should be !NULL in all cases.  But adding the checking is always safe. Let me add the NULL checking before GET_OFFSET call

ceolin:
merged, thanks !!

Issue #2117, opened 2016-06-02
New IIO C-API to read configuration attributes


lblim:
#### Task Description

The existing IIO C API does not allow the users to read out the existing configuration. The API only support reading out the sensor raw data or processed data. 

Need new API be introduced to allow users to read and modify the attributes of IIO devices.

lblim:
May I know when is the target date for v2?

barbieri:
no dates set in stone, but we're looking into 2-3 months maximum.

yongli3:
@laykuanloon @fulong82  could you share more information about which configuration items are needed for the reading/modifying?

fulong82:
Hi @yongli3,

Sorry for the late response. Here are the list that we would like to have a read API that can be used in sensor application:
- sampling frequency
- hysteresis
- scale
- offset

@laykuanloon do you have any interfaces want to add on?


Thanks & Regards,
Wilson

yongli3:
@fulong82 which sensor has the "hysteresis" ?

fulong82:
Hi @yongli3,

All of the IIO-HID sensors would have "hysteresis" attributes:
http://lxr.free-electrons.com/source/drivers/iio/common/hid-sensors/hid-sensor-attributes.c#L224
For example:
- HID accelerometer: http://lxr.free-electrons.com/source/drivers/iio/accel/hid-sensor-accel-3d.c#L150
- HID gyro: http://lxr.free-electrons.com/source/drivers/iio/gyro/hid-sensor-gyro-3d.c#L150
- HID magneto: http://lxr.free-electrons.com/source/drivers/iio/magnetometer/hid-sensor-magn-3d.c#L197
- HID ambient light: http://lxr.free-electrons.com/source/drivers/iio/light/hid-sensor-als.c#L127
- HID pressure: http://lxr.free-electrons.com/source/drivers/iio/pressure/hid-sensor-press.c#L124
- HID inclinometer: http://lxr.free-electrons.com/source/drivers/iio/orientation/hid-sensor-incl-3d.c#L150


Regards,
Wilson

yongli3:
Just like the sol_iio_mount_calibration API, we can add a new one as below:
Int sol_iio_read_device_config(struct sol_iio_device *device, struct iio_device_config *config);

The API will query the device configuration, and save them into config

 struct iio_device_config {
	double hysteresis;
	double offset;
	double sampling_frequency;
	double scale;
}

[...]

PR #2346, opened 2016-10-28
New IIO C-API to read configuration attributes

guchaojie:
Add 3 C-API to read scale, offset and sampling_frequency attribute.

Signed-off-by: guchaojie chao.jie.gu@intel.com


guchaojie:
@ceolin @edersondisouza @fulong82 @laykuanloon  This PR is for issue #2117. All thie C-API is to get configuration attribute from device level.  Please have  a review, any comment is welcome.

src/lib/io/include/sol-iio.h
```diff
@@ -249,6 +249,58 @@ int
...
+char *sol_iio_channel_get_name(struct sol_iio_channel *channel);
+
+
+/**
+ * @brief Gets the configuration scale attribute from one device
+ *
+ * This function gets configuration scale attribute from one device,
+ * after the device sol_iio_device structure has been desired.
+ *
+ * @param device The sol_iio_device structure which the device is desired
```
ceolin: which the scale is desired ...

src/lib/io/include/sol-iio.h
```diff
@@ -249,6 +249,58 @@ int
...
+ *
+ */
+char *sol_iio_channel_get_name(struct sol_iio_channel *channel);
+
+
+/**
+ * @brief Gets the configuration scale attribute from one device
+ *
+ * This function gets configuration scale attribute from one device,
+ * after the device sol_iio_device structure has been desired.
```
ceolin: this paragraph is adding nothing, i think you can remove it.

src/lib/io/include/sol-iio.h
```diff
@@ -249,6 +249,58 @@ int
...
+ *
+ * @return 0 on success, -errno on failure.
+ */
+int sol_iio_device_get_scale(struct sol_iio_device *device, const char *prefix_name, double *scale);
+
+/**
+ * @brief Gets the configuration offset attribute from one device
+ *
+ * This function gets configuration offset attribute from one device,
+ * after the device sol_iio_device structure has been desired.
```
ceolin: ditto

src/lib/io/sol-iio.c
```diff
@@ -571,14 +571,16 @@ setup_device_reader(struct sol_iio_device *device)
...
-channel_get_pure_name(struct sol_iio_channel *channel)
+channel_get_pure_name(char *name)
 {
     char channel_name[NAME_MAX];
     size_t channel_name_len;
     char *channel_pure_name;
     bool modified = false;
 
-    strncpy(channel_name, channel->name, sizeof(channel_name) - 1);
+    SOL_NULL_CHECK(name, NULL);
```
ceolin: I've spotted this problem in another commit :)

src/lib/io/sol-iio.c
```diff
@@ -1762,3 +1764,108 @@ sol_iio_mount_calibration(struct sol_iio_device *device, sol_direction_vector *v
...
+            device->device_id, _name)) { \
+            r = sol_util_read_file(path, "%lf", scale); \
+            goto end; \
+        } \
+    } while (0)
+
+    GET_SCALE(prefix_name);
+    pure_name = channel_get_pure_name(prefix_name);
+    GET_SCALE(pure_name);
+
```
ceolin: this code is almost the same of `channel_get_scale()`. Wouldn't be better both use same code ? You can put this code in another function and both call it. 

One more thing, the scale is not associated with a channel ? if yes, is it possible get it directly from the desired channel looking the channel in the channels vector ?

src/modules/flow/iio/iio.json
```diff
@@ -308,6 +368,21 @@
...
+        },
+        {
+         "data_type": "float",
+         "description": "Offset to be added to device raw readings",
+        "name": "OFFSET"
+        },
+        {
+         "data_type": "int",
+         "description": "Sampling frequency of the sensor. If -1, use device default",
+         "name": "SAMPLING_FREQ"
```
ceolin: since we are using frequency everywhere, I guess it would be better `SAMPLIGING_FREQUENCY` don't you think ?

ceolin:
patch looks good, just check whether my comments make sense or not :)

src/lib/io/sol-iio.c
guchaojie: the scale in base function is not assiociated with a channel.

PR #2347, opened 2016-10-28
Fix tinycbor API calls

ffontaine:
Since tinycbor 0.3.2, ptr and bytes_needed of CborEncoder structure are
accessible through a named union called data instead of an unnamed union
so add the data structure when needed

Signed-off-by: Fabrice Fontaine fabrice.fontaine@orange.com

### DIFF
--- a/src/lib/comms/sol-oic-cbor.c
+++ b/src/lib/comms/sol-oic-cbor.c
@@ -63,7 +63,7 @@ initialize_cbor_payload(struct sol_oic_map_writer *encoder)
 
     cbor_encoder_init(&encoder->encoder, encoder->payload,
         buf->capacity - offset, 0);
-    old_ptr = encoder->encoder.ptr;
+    old_ptr = encoder->encoder.data.ptr;
 
     encoder->type = SOL_OIC_MAP_CONTENT;
 
@@ -72,7 +72,7 @@ initialize_cbor_payload(struct sol_oic_map_writer *encoder)
     err = cbor_encoder_create_map(&encoder->encoder, &encoder->rep_map,
         CborIndefiniteLength);
     if (err == CborNoError) {
-        new_ptr = encoder->rep_map.ptr;
+        new_ptr = encoder->rep_map.data.ptr;
         r = buffer_used_bump(encoder, new_ptr - old_ptr);
         if (r < 0)
             err = CborErrorUnknownType;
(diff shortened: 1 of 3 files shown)


src/lib/comms/sol-oic-cbor.c
```diff
@@ -63,7 +63,7 @@ initialize_cbor_payload(struct sol_oic_map_writer *encoder)
 
     cbor_encoder_init(&encoder->encoder, encoder->payload,
         buf->capacity - offset, 0);
-    old_ptr = encoder->encoder.ptr;
+    old_ptr = encoder->encoder.data.ptr;
```
ffontaine: You're right, sorry. This change has been applied on master, it is not a part of v0.3.2. Will you agree if I update the reference and the commit message to reflect that this change is needed to be compatible with the master branch of tinycbor? If you prefer, You can also close this pull request and wait for the next tinycbor release.

ceolin: I prefer work with releases, unless someone has something that depends on this new version. Is it your case ?

thiagomacieira: Please don't use the structure members. Use the new API isntead.

thiagomacieira: You're supposed to use cbor_encoder_get_used_bytes and add to the original pointer.

ffontaine: I suppose that you mean cbor_encoder_get_extra_bytes_needed?
However, I think I will wait @glima for its expertise. The buffer_used_bump function and all its relative code seems to have been added by the commit 487af3523e230fa78dfe4a8dc00a3772adeda76e. It seems that soletta is "abusing" tinycbor API for a purpose that I do not master.

glima: I recall well (it's been a while now), the memory region we pass to tinycbor to encode is a Soletta buffer, that is memory that can be realloc-ed if it grows too much. That hack is to hard-wire into an existing cbor encoder the new memory region with its size and let the encoding proceed without noticing that. At the time it was written, there was no API supporting that mem region changing on the fly. Not sure if there is one already on this last release. From what I checked as well, this data field does not exist indeed (only checked now). So not sure what you wanted.

ffontaine: I updated the PR using the tag on tinycbor v0.4. Can you check if it's ok?

glima: I'm a bit confused: doesn't the v0.4 tag belong in the master branch there?

ffontaine: This is the first time I update a git submodule but it seems ok in the patch, a088996aa5f59b4f27f20fadad053d88bee357d4 is the commit id of tinycbor v0.4: 
```
diff --git a/src/thirdparty/tinycbor b/src/thirdparty/tinycbor
index 629d5b7..a088996 160000
--- a/src/thirdparty/tinycbor
+++ b/src/thirdparty/tinycbor
@@ -1 +1 @@
-Subproject commit 629d5b71823ad91faab60f3f51597f010ad98c0c
+Subproject commit a088996aa5f59b4f27f20fadad053d88bee357d4
```
It seems git will not explicitly reference/show the v0.4 tag.

glima:
Yes, time for a new release. I'll definitely address this patch when doing so, soon.

glima:
Local tests made (samples ran, all doing fine). The commit is now sound, with the upstream tag for tinycbor fixed. Merging, thanks a lot.

PR #2350, opened 2016-11-04
New IIO C-API to read configuration attributes v3

guchaojie:
including 3 C-API to read scale, offset and sampling_frequency attribute


guchaojie:
@ceolin @edersondisouza @fulong82 @laykuanloon This PR is for issue #2117.

ceolin:
merged, thanks a lot !

src/modules/flow/iio/iio.json
```diff
@@ -18,6 +18,14 @@
           "methods": {
             "process": "iio_common_tick"
           }
+        },
+        {
+          "data_type": "any",
+          "description": "Packets sent in here will trigger a gyroscope configuration attribute reading.",
+          "name": "TRIGGER",
```
edersondisouza: I know that this is already in, but I think that this name `TRIGGER` can be a bit confusing, don't you think? I mean, one can think that it has anything to do with IIO triggers, while it has nothing to do. Maybe something like 'CONFIG' or 'INFO'?

Issue #2154, opened 2016-06-10
Proposal: add wildcard characters support to IIO absolute path json input


fulong82:
#### Task Description

Hi @barbieri, @edersondisouza and @yongli3,

I would like to make a suggestion here to **add support of wildcard characters in IIO absolute path input**. Below are my points of why this is useful to users when they choose to use absolute path:
- iio device id (iio:deviceX) is dynamically assigned base on device load sequence: This may cause the absolute path `/sys/bus/i2c/devices/6-0077/iio:device0/` changed on the next boot if it is not being loaded 1st. So if wildcard character is supported, then there won't be an issue if the absolute path wrote in `/sys/bus/i2c/devices/6-0077/iio:device*/`. Understand that you may suggest using `i2c/X-YYYY` form instead, below bullet shows another possibility.
-  i2c bus id (i2c-X) may be changed: When user choose to enable more i2c buses (e.g. smbus, graphic port or usb-to-i2c device), the i2c bus id may be changed. So the `i2c/X-YYYY` form could not help whereas the absolute path input `/sys/bus/i2c/devices/*-0077/iio:device*/` will do.

Besides letting user changing the sol-flow.json file, a wildcard characters support could eliminate the troublesome work by just adding a `glob()` function call before the `realpath()` function call in the `sol-iio.c` file `resolve_absolute_path()` function code.
#### Dependencies

N/A
#### Example Code

```
#include <limits.h>
#include <stdlib.h>
#include <glob.h>
#include <stdio.h>

void main (void)
{
        char path[PATH_MAX] = "/sys/bus/i2c/devices/*-0077/iio:device*/";
        char resolved_path[PATH_MAX];
        glob_t resolved_glob;
        int flags;

        flags = GLOB_MARK | GLOB_TILDE;
        glob(path, flags, NULL, &resolved_glob);
        printf("glob return path = %s\n", resolved_glob.gl_pathv[0]);

        realpath(resolved_glob.gl_pathv[0], resolved_path);
        printf("realpath resolved path = %s\n\n", resolved_path);

        globfree(&resolved_glob);
}
```

edersondisouza:
On 10-06-2016 08:54, Wilson Kweh wrote:
>
>         Task Description
>
> Hi @barbieri <https://github.com/barbieri>, @edersondisouza
> <https://github.com/edersondisouza> and @yongli3
> <https://github.com/yongli3>,
>
> I would like to make a suggestion here to *add support of wildcard
[...]

PR #2353, opened 2016-11-14
iio: Support wildcard in IIO path input

ceolin:
It allows using wildcard characters to select an IIO path. It's
important to know that only the first result is used.

It internally uses glob().

### DIFF
--- a/src/lib/io/sol-iio.c
+++ b/src/lib/io/sol-iio.c
@@ -20,6 +20,7 @@
 #include <dirent.h>
 #include <errno.h>
 #include <fcntl.h>
+#include <glob.h>
 #include <limits.h>
 #include <stddef.h>
 #include <stdio.h>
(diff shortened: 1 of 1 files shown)


ceolin:
@fulong82 is it that you have in mind on https://github.com/solettaproject/soletta/issues/2154 ?

I've scratched it but I was not able to test yet. I'm creating this pullreq to see whether I understood correctly or not .

fulong82:
@ceolin, 

Yes. So base on your pull request implementation, you are taking the 1st matching entry to proceed.
Thanks.

src/lib/io/sol-iio.c
```diff
@@ -1572,6 +1573,33 @@ resolve_absolute_path_cb(void *data, const char *dir_path, struct dirent *ent)
...
+                errno = ENOMEM;
+                return NULL;
+            case GLOB_ABORTED:
+            default:
+                errno = EINVAL;
+                return NULL;
+        }
+    }
+
+    resolved_path = realpath(result.gl_pathv[0], resolved_path);
```
fulong82: Isn't this should be:
`resolved_path = realpath(result.gl_pathv[0], *resolved_path);`

ceolin: nope,  realpath() receives a pointer to an allocated memory and change this memory. But there is a problem in this code, sending a new pr. @fulong82 are you able to test it ?

fulong82:
@ceolin 

Okay, tested your patch. It is working. Thx. 👍

ceolin:
@fulong82 thanks for testing. Is it address the issue https://github.com/solettaproject/soletta/issues/2154 ?

fulong82:
@ceolin 
Yes and Thanks.

PR #2356, opened 2017-03-01
Fixes

glima:
Preparing for v2


glima:
The last (style) commit might cause trolls, I already got my 🍿 and am prepared to drop it, if need be.

src/lib/io/sol-iio.c
```diff
@@ -470,14 +470,14 @@ get_buffer_value(struct sol_iio_device *device, int *value)
...
-    if (!r)
-        return false;
+        if (!craft_filename_path(path, sizeof(path), CHANNEL_ENABLE_PATH,
+            device->device_id, channel_name))
+            return false;
 
-    if (check_file_existence(path)) {
         if (sol_util_write_file(path, "%d", enabled) < 0)
             return false;
     }
```
ceolin: both implementation are calling snprintf twice for the same path :( but anyway, the code is cleaner now
+1

src/modules/flow/iio/nodes.c
```diff
@@ -126,7 +126,7 @@ iio_common_close(struct sol_flow_node *node, void *data)
         goto end;
 
     for (iter = mdata->config.oversampling_ratio_table; iter->key; iter++)
-        free(iter->key);
+        free((void *)iter->key);
```
ceolin: indeed, it's not beauty :(
+1

src/lib/io/sol-iio.c
glima: Yeah, but then the first call was not meant to be called without format args. Not a hot path anyway, it seems.

ceolin: it's not a problem  call with the full path, ideally with full path only the last line of that function should be used :)

ceolin:
patches are great, only the last one, please remove it, we have agreed in not doing these commits changing only code-style

glima:
OK, will merge those with the last bumping uncrustify`s version, only

PR #2357, opened 2017-03-01
samples: make them more easily ditched from build

glima:
Also make the arctangent node, which is an out-of-tree node contribution
node+sample example, depend on the sample being built.

The Fedora packaging build won't have samples built anymore, thus this
cleanup.


edersondisouza:
lgtm

glima:
thanks, merging

PR #2358, opened 2017-03-01
Remove nodejs binding

ceolin:
Nodejs binding was outdated and incomplete, as there is nobody
maintaining it we are removing it. This will make the build faster
and simplify the CI.
Whether somebody get interest in it again we can revert it.


glima:
Bunch of builds here and it seems we're fine. There's an unrelated issue, though, I want to sort out first: clean cashes/repo and make `alldefconfig; make -j check-fbp-bin` is screwing up ­— the module for js flow metatype is not being found prior to compiling the sample that has `MyScript` entry :(

cabelitos:
Wow, the JS guys are not maintaining it anymore?

Is this the correct approach?

If we remove, we should remenber to change de CI rules so we don't try to execute the binding stuff.

cabelitos:
@ceolin @glima ^

glima:
Nope, no sign of them for a great deal of time, now. Well, we do not seem to have users of this part as well, and as you remembered, there's complexity on the CI as well to take care of. We should clean on that side as well, we are aware.

ceolin:
@cabelitos agree about the CI, but  as the buildsystem itself does not have anything about js bindings anymore, even don't changing it now will not be a problem

cabelitos:
ok then, +1

cabelitos:
@ceolin I guess you forgot to close this PR.

glima:
True, closing. I still want to figure out all build issues before proceeding with the release, doing that soon.

PR #2361, opened 2017-04-07
Random bs fixes

dorileo:
A few build issues were brought to my attention by @glima.

1. a parallel breakage with check-fbp-bin;
2. a parallel breakage with samples;

And I got a bonus fix with the gen-alias target - now converted to an actual rule and properly set into dep chain.

Regards....

### DIFF
--- a/tools/build/Makefile.targets
+++ b/tools/build/Makefile.targets
@@ -14,7 +14,7 @@ check-cpp: $(SOL_LIB_OUTPUT) $(HEADERS_CC)
 PHONY += check-cpp
 
 ifeq (y,$(FBP_RUNNER))
-check-fbp: $(SOL_LIB_OUTPUT) $(SOL_FBP_RUNNER_BIN) $(modules-out) $(PRE_GEN) gen-aliases
+check-fbp: $(SOL_LIB_OUTPUT) $(SOL_FBP_RUNNER_BIN) $(modules-out) $(PRE_GEN) $(DEFAULT_ALIAS_FILE)
 	$(Q)$(PYTHON) $(TEST_FBP_SCRIPT) --runner="$(abspath $(SOL_FBP_RUNNER_BIN))"
 else
 check-fbp:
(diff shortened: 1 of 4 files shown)


tools/build/Makefile.targets
```diff
@@ -202,11 +202,44 @@ endif
...
 
-DEFAULT_ALIAS_FILE := $(build_aliasesdir)/50-default.json
-
-ifeq (y,$(USE_FLOW))
-gen-aliases: $(FLOW_NODE_TYPE_ALIAS)
-	$(Q)$(MKDIR) -p $(build_aliasesdir)
-	$(Q)echo "     "GEN"   "$(DEFAULT_ALIAS_FILE)
-	$(Q)$(FLOW_NODE_TYPE_ALIAS) --output=$(DEFAULT_ALIAS_FILE) $(all-gens)
-endif
+ifneq ($(USE_NODEJS),)
```
glima: From here on it seems you got your rebase wrong (remove the + lines, please). Testing the rest now, they look good from a 1st view

tools/check-style.py
```diff
@@ -29,7 +29,7 @@
 DIFF_REF_COLOR = "\033[36m"
 END_COLOR = '\033[0m'
 
-UNCRUSTIFY_VERSION = "0.61"
+UNCRUSTIFY_VERSION = "0.60"
```
glima: Long story short, let's try:

- checking style on CI *always* went through whole files, not *diffs*
- false positives, due to that, have been happening since ages now (it won't change with version being .60 solely)
- IMO a commit fixing style should take place to fix that for good, since having diff checks is much harder with the tools we got
- Some people dislike a patching fixing style, so I did just bump the tool's version. That way, we at least have saner approaches to style fixes when people *do* use the tool (lots of bugs from .60 were corrected)
- I second that re-packaging .61 for Ubuntu now is a PITA, but not convinced that reverting this will bring any real good

:-1:

dorileo:
new version fixes rebase issues and removes the revert patch, lets try to fix semaphore.

PR #2364, opened 2017-05-11
Fixes (one more round towards v2)

glima:
Please help validate those (lack of) breaks were bugs.


src/modules/flow/json/json.c
```diff
@@ -992,6 +993,7 @@ json_path_add_new_element(struct sol_flow_node *node, struct json_element *base_
             SOL_INT_CHECK(r, == -ENOMEM, r);
             SOL_INT_CHECK_GOTO(r, < 0, error_parse);
 
+            break;
```
edersondisouza: Maybe you should try to ping original authors of these codes - maybe they can remember some peculiarity. That said, I guess that _not_ having breaks is correct for json patch: note that both handle 'YXZ_BLOB' then falls through 'XYZ' case: it appears to prepare data for next case.

glima: Mm, maybe. @otaviobp ?

otaviobp: The reason to not having that break is the reason @edersondisouza commented. It is not an error. it is intended. I remember having tests that explores that, and they are suppose to fail after your changes. Have you tested that?

glima:
* Otavio Pontes <notifications@github.com> [2017-05-11 14:11:54 -0700]:

> otaviobp commented on this pull request.
> 
> 
> 
> > @@ -992,6 +993,7 @@ json_path_add_new_element(struct sol_flow_node *node, struct json_element *base_
>              SOL_INT_CHECK(r, == -ENOMEM, r);
>              SOL_INT_CHECK_GOTO(r, < 0, error_parse);
>  
> +            break;
> 
> The reason to not having that break is the reason @edersondisouza commented. It is not an error. it is intended. I remember having tests that explores that, and they are suppose to fail after your changes. Have you tested that?

Maybe I missed a test suite. OK, then, I'll rewrite it commenting the
explicit pass through, then. Thanks, guys.

> 
> -- 
> https://github.com/solettaproject/soletta/pull/2364#discussion_r116103512

-- 
Gustavo Lima Chaves

glima:
The tests were passing, just checked. Anyway, breaks that should not be there removed.

glima:
Merging it at the end of the day

src/lib/comms/sol-http-server-impl-microhttpd.c
```diff
@@ -1034,6 +1034,7 @@ free_request(struct sol_http_request *request)
         switch (param->type) {
         case SOL_HTTP_PARAM_POST_DATA:
             free((char *)param->value.data.filename.data);
+            break;
```
edersondisouza: I guess that @ceolin is the one who can validate this - but I guess that current code is correct as well: code after `FOREACH` does basically the same for `request->param` - frees unconditionally `param.value.key_value.value.data` and `param.value.key_value.key.data`, and if `SOL_HTTP_PARAM_POST_DATA`, frees `param.value.data.filename.data`.

glima: After the loop I see no problem, just wondering if inside it we have a double-free scenario.

glima: @ceolin ping

src/lib/flow/sol-flow-node-options.c
```diff
@@ -781,7 +781,7 @@ split_option(const char *input, const char **key, unsigned int *key_len, const c
 {
     const char *equal = strchr(input, '=');
 
-    if (!equal || equal == input || equal + 1 == '\0')
+    if (!equal || equal == input || *equal + 1 == '\0')
```
ceolin: shouldn't be: `*(equal + 1) == '\0'`

src/lib/comms/sol-http-server-impl-microhttpd.c
ceolin: Don't need this break. It'll cause memory leak.
If you really want to change it you will need to do:

```
    SOL_HTTP_PARAMS_FOREACH_IDX (&request->params, param, idx) {                                               
        switch (param->type) {                                                                                 
        case SOL_HTTP_PARAM_POST_DATA:
            free((char*)param->value.data.value.data);                                                   
            free((char *)param->value.data.key.data);          
            free((char *)param->value.data.filename.data);                                                     
            break;                                                                                             
        case SOL_HTTP_PARAM_POST_FIELD:                                                                        
            free((char *)param->value.key_value.value.data);                                                   
            free((char *)param->value.key_value.key.data);                                                     
            break;                                                                                             
        default:                                                                                               
            break;                                                                                             
        }                                                                                                      
    }    
```

To my mind, just let it as it is.

ceolin:
Please, remove the commit with the break from the serie and check the comment about the sol-flow-node-options.c

edersondisouza:
On Thu, 2017-06-08 at 13:32 -0700, Flávio Vinicius Alvares Ceolin
wrote:
> @ceolin commented on this pull request.
> In src/lib/comms/sol-http-server-impl-microhttpd.c:
> > @@ -1034,6 +1034,7 @@ free_request(struct sol_http_request
> *request)
>          switch (param->type) {
>          case SOL_HTTP_PARAM_POST_DATA:
>              free((char *)param->value.data.filename.data);
> +            break;
> Don't need this break. It'll cause memory leak.
> If you really want to change it you will need to do:
>     SOL_HTTP_PARAMS_FOREACH_IDX (&request->params, param, idx)
> {                                               
>         switch (param->type)
> {                                                                    
>              
>         case SOL_HTTP_PARAM_POST_DATA:
>             free((char*)param-
> >value.data.value.data);                                             
>       
>             free((char *)param->value.data.key.data);          
>             free((char *)param-
> >value.data.filename.data);                                          
>            
>             break;                                                   
>                                           
>         case
> SOL_HTTP_PARAM_POST_FIELD:                                           
>                              
>             free((char *)param-
> >value.key_value.value.data);                                        
>            
>             free((char *)param-
> >value.key_value.key.data);                                          
>            
>             break;                                                   
>                                           
>         default:                                                     
>                                           
>             break;                                                   
>                                           
>         }                                                            
>                                           
>     }    
> To my mind, just let it as it is.

To prevent this confusion in future, better put a comment making sure
fallthrough is desired.

src/lib/flow/sol-flow-node-options.c
glima: Damn, thanks for that.

glima:
Yeah, I'll just commit the fall-through comment and ship it. Thanks, guys.

glima:
Merged. We're failing make samples, damn. Will look more closely into that later.

src/lib/flow/sol-flow-node-options.c
barbieri: `equal[1]` is less ugly than `*(equal + 1)`

PR #2365, opened 2017-11-06
Openssl

glima:
Bring make check back to life on current distros. make check-valgrind is b0rked for test-mainloop-linux, gotta figure out why now.

### DIFF
--- a/src/lib/comms/sol-gatt-impl-bluez.c
+++ b/src/lib/comms/sol-gatt-impl-bluez.c
@@ -229,6 +229,7 @@ attr_method(enum pending_type type, sd_bus_message *m, void *userdata, sd_bus_er
 
         r = attr->write(pending, &buf, 0);
         SOL_INT_CHECK_GOTO(r, < 0, error_op);
+    /* fall through */
     default:
         r = -EINVAL;
         goto error_op;
--- a/src/lib/comms/sol-network-impl-linux.c
+++ b/src/lib/comms/sol-network-impl-linux.c
@@ -282,6 +282,7 @@ _on_event(void *data, int nl_socket, uint32_t cond)
             switch (h->nlmsg_type) {
             case NLMSG_ERROR:
                 SOL_WRN("read_netlink: Message is an error");
+            /* fall through */
             case NLMSG_DONE:
                 return true;
             case RTM_NEWADDR:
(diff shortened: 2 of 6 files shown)


glima:
@sergiomb2 is it you that pinged about this and the spec bump?

sergiomb2:
I pinged for update solleta in Fedora https://bugzilla.redhat.com/show_bug.cgi?id=1424460 , as I wrote Fedora 26  have openssl-1.0 as compat package and openssl-1.1 as the main package , so migrate to openssl 1.1 is not required

sergiomb2:
Add this patch [2] fixes build for F26 
F27 fails to build on new curl according my build in [1]

[1] 
https://copr.fedorainfracloud.org/coprs/sergiomb/builds_for_Stable_Releases/build/658499/
[2]
[soletta.spec.patch.txt](https://github.com/solettaproject/soletta/files/1447147/soletta.spec.patch.txt)

glima:
Thanks, @sergiomb2 . Will take a look at the curl issue as soon as I can (and test/apply the other patch).

sergiomb2:
Anyway I don't see any openssl package in Fedora

sergiomb2:
[soletta.spec.patch.v2.txt](https://github.com/solettaproject/soletta/files/1447447/soletta.spec.patch.v2.txt)
Soletta 2 builds on F25 and F26 but stil fails on F27 with curl

glima:
@sergiomb2 thanks for the python2 requires thing. I'm not sure yet when we need it, but will test/incur with it soon. I know the builds may pass (why do send patches against an old spec, BTW? I have a more up to date in this very same repo), but I'm not talking x86 builds, but the alien architectures that fail. We *have* to address those too, plus the curl thing.

sergiomb2:
I'm using the latest spec (in Fedora) [1] 
[1]
https://src.fedoraproject.org/rpms/soletta/blob/master/f/soletta.spec
https://src.fedoraproject.org/cgit/rpms/soletta.git/tree/soletta.spec

glima:
make: *** [tools/build/Makefile.rules:377: build/stage/lib/comms/sol-http-client-impl-curl.o] Error 1

was seen on logs now, indeed. Will attack this as soon as I got time.

BTW, @ceolin care to look at the patch, when you can, please?

sergiomb2:
F27 have one  curl update [1] which breaks the build 

[1] 
 https://apps.fedoraproject.org/packages/libcurl

glima:
Timeout. Merging this.

PR #2366, opened 2018-07-10
Fixes: starting effort to make the package live again on Fedora

glima:


edersondisouza:
`make check-valgrind` fails on `test-mainloop-linux` here. Is it expected? 
Did you check semaphore complaints?

glima:
Yes, this is new. It seems to me like a false positive, what do you think? I'm not too much alarmed with that because check-valgrind is not run on Fedora specs.

glima:
@edersondisouza ping?

edersondisouza:
:+1:

